PR #7, opened 2016-07-25
Improve error handling for no-such-object type processing

robbrockbank:
This PR fixes up the following:

-  List methods return empty slices of objects when the enumeration parent does not exist.
-  Improved error messages for gets/creates etc. when the request is missing some mandatory identifiers (it now includes the name of the identifier)


tomdee:
How would you feel about splitting this into different PRs for the different things you're adding. It's tricky to review as it stands.

robbrockbank:
Closing this PR in preference to another that i'm about to open.

PR #8, opened 2016-07-25
Add conflict error type

robbrockbank:
Add error type indicating resource conflict.

Casey - I think you'll need this for the IPAM work.  It looks like the current IPAM changes defines a cas error type but doesn't use it.

As discussed - I/we also need to fix up the backend API so that the delete processing is also atomic -by ensuing we pass in the revision information as well.  This PR is just to handle the compare error.


caseydavenport:
LGTM.

Looking forward to the subsequent PRs as well:
- Better defined error interface for libcalico.
- Support compare on the delete operation.

PR #9, opened 2016-07-25
Start adding docs for frontend and backend client

robbrockbank:
Tom - this is WIP - but will hopefully provide enough context for us to have the chat about libcalico.

### DIFF
--- a/lib/api/doc.go
+++ b/lib/api/doc.go
@@ -13,6 +13,12 @@
 // limitations under the License.
 
 /*
-Package api implements the struct definitions used by the northbound client API.
+Package api implements the struct definitions used on the northbound client API.
+
+In particular this includes:
+-  The various resource types as exposed on the Northbound API
+-  The client configuration
+
+The resource type structures include the YAML tags for the public facing interface.
 */
 package api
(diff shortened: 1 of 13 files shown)


lib/api/doc.go
```diff
@@ -13,6 +13,12 @@
 // limitations under the License.
 
 /*
-Package api implements the struct definitions used by the northbound client API.
+Package api implements the struct definitions used on the northbound client API.
+
+In particular this includes:
```
tomdee: This doesn't format like you think it does

tomdee:
```
/usr/lib/go/bin/godoc github.com/tigera/libcalico-go/lib/api/
use 'godoc cmd/github.com/tigera/libcalico-go/lib/api/' for documentation on the github.com/tigera/libcalico-go/lib/api/ command 

PACKAGE DOCUMENTATION

package api
    import "github.com/tigera/libcalico-go/lib/api"

    Package api implements the struct definitions used on the northbound
    client API.

    In particular this includes: - The various resource types as exposed on
    the Northbound API - The client configuration

    The resource type structures include the YAML tags for the public facing
    interface.
```

PR #10, opened 2016-07-25
Re-implement ReleasePoolAffinities

caseydavenport:


lib/client/ipam.go
```diff
@@ -577,28 +577,26 @@ func (c ipams) RemoveIPAMHost(host *string) error {
...
+	if err != nil {
+		glog.Errorf("Error querying block affinities: %s", err)
+		return nil, err
+	}
+
+	// Iterate through each block affinity and build up a mapping
+	// of blockCidr -> host.
+	glog.V(4).Infof("Getting block -> host mappings")
+	for _, o := range objs {
+		host := o.Key.(backend.BlockAffinityKey).Host
```
robbrockbank: better to just cast into a local variable rather than casting twice

lib/client/ipam.go
```diff
@@ -577,28 +577,26 @@ func (c ipams) RemoveIPAMHost(host *string) error {
 	return nil
 }
 
-func (c ipams) hostBlockPairs(pool common.IPNet) (map[string]string, error) {
```
robbrockbank: Should these be common.IPNet rather than pointers?  I thought go didn't copy structs if they weren't modified. 

I'm not really fussed either way - just trying to get more opinions on pointers vs object... :-)

caseydavenport: I made it a pointer since `common.ParseCIDR` returns pointers, but it's kind of a silly reason.

lib/client/ipam.go
caseydavenport: yep, fixed.

lib/backend/block_affinity.go
```diff
@@ -52,19 +52,21 @@ func (key BlockAffinityKey) valueType() reflect.Type {
...
+		k = k + options.Host
+	}
 	return k
 }
 
 func (options BlockAffinityListOptions) keyFromEtcdResult(ekey string) KeyInterface {
 	glog.V(2).Infof("Get Block affinity key from %s", ekey)
-	r := matchBlock.FindAllStringSubmatch(ekey, -1)
-	if len(r) != 2 {
+	r := matchBlockAff.FindAllStringSubmatch(ekey, -1)
```
robbrockbank: Any reason we don't just use the full word Affinity rather than Aff?

robbrockbank: The match string:

`"^/?/ ... "`  should not have that second /

caseydavenport:
Ok - I think that's all.  Could you take another look? @robbrockbank

lib/client/ipam.go
```diff
@@ -513,7 +513,7 @@ func (c ipams) ReleaseHostAffinities(host *string) error {
 
 // ReleasePoolAffinities releases affinity for all blocks within
 // the specified pool across all hosts.
-func (c ipams) ReleasePoolAffinities(pool common.IPNet) error {
+func (c ipams) ReleasePoolAffinities(pool *common.IPNet) error {
```
robbrockbank: ReleasePoolAffinities() is part of the northbound API - so we need to decide whether a pointer to an IPNet is the right thing or just an IPNet.

caseydavenport: I think the right thing is to pass a struct, not a pointer.

Don't want the value to accidentally be changed in ReleasePoolAffinities.  I will update.

caseydavenport:
Commits now included in #18.

PR #11, opened 2016-07-25
Add omitempty directives to slice fields in backend workload endpoint

robbrockbank:
Tom - I believe this should fix the issue you were having with the workload endpoint data being invalid.


tomdee:
LGTM - merging

PR #12, opened 2016-07-25
Add conflict error handling for Delete backend

robbrockbank:
Casey - this is the other part of the conflict error handling - adding support to the backend Delete method by passing in a DatastoreObject rather than just a Key.

### DIFF
--- a/lib/backend/etcd.go
+++ b/lib/backend/etcd.go
@@ -34,7 +34,6 @@ import (
 var (
 	etcdApplyOpts  = &etcd.SetOptions{PrevExist: etcd.PrevIgnore}
 	etcdCreateOpts = &etcd.SetOptions{PrevExist: etcd.PrevNoExist}
-	etcdDeleteOpts = &etcd.DeleteOptions{Recursive: true}
 	etcdGetOpts    = &etcd.GetOptions{Quorum: true}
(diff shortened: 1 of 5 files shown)


caseydavenport:
LGTM.

How will we support this on the frontend? Do we have a way of exposing object versioning in a datastore-agnostic way?

lib/backend/etcd.go
```diff
@@ -130,11 +129,15 @@ func (c *EtcdClient) Apply(d *DatastoreObject) (*DatastoreObject, error) {
 }
 
 // Delete an entry in the datastore.  This errors if the entry does not exists.
```
caseydavenport: typo: exists.

robbrockbank:
At the moment this is not exposed on the northbound API - but I'm imagining we'd expose an interface{} field on the Metadata that would not get written into the YAML files - so this is more for use with the programmatic API - so the actual thing in the revision info could be anything.

If we do need to expose it in the YAML then we'd have to sanitize it in so much as to make it a single value that could be written and read (for example a string or an int).  I have thought a little about how we'd do that, for example if our backend only supported atomic operations by passing in the whole of the original set of data then we might have to, for example, take a checksum of the data and pull and re-push to perform an update ... but for now, etcd returns an integer index so it's a non-issue if we want to write it to file.

PR #13, opened 2016-07-26
Return empty slice when List enumeration root does not exist

robbrockbank:
Fix List processing:
-  If enumeration root does not exist, return empty slice rather than a "no such resource" exception.
-  Make sure we convert etcd exceptions so not exposing etcd errors directly.


tomdee:
LGTM

PR #14, opened 2016-07-26
Improve error handling for no-such-object type processing

robbrockbank:
Casey - you may care about this one in relation to your IPAM stuff - that said, the types of error this affects should not trickle their way up out of IPAM so perhaps it doesn't matter.

This PR improves resource identification in error messages.  The changes consist of:
-  Adding a String() method to KeyInterfaces for resources that are exposed directly on the northbound API.  This provides a nice description of the type of object and it's main identifiers.
-  Adding the KeyInterface into certain types of error so that the message can include the above friendly resource description.
-  Adding details about which identifiers are missing from a create/get/update etc. request.  This involved adding a new field to an exception and updating the KeyInterfaces to specify the missing fields.

### DIFF
--- a/lib/backend/pool.go
+++ b/lib/backend/pool.go
@@ -36,7 +36,7 @@ type PoolKey struct {
 
 func (key PoolKey) asEtcdKey() (string, error) {
 	if key.CIDR.IP == nil {
-		return "", common.ErrorInsufficientIdentifiers{}
+		return "", common.ErrorInsufficientIdentifiers{Name: "cidr"}
 	}
 	c := strings.Replace(key.CIDR.String(), "/", "-", 1)
 	e := fmt.Sprintf("/calico/v1/ipam/v%d/pool/%s", key.CIDR.Version(), c)
@@ -49,6 +49,10 @@ func (key PoolKey) asEtcdDeleteKey() (string, error) {
 
 func (key PoolKey) valueType() reflect.Type {
 	return typePool
+}
+
+func (key PoolKey) String() string {
+	return fmt.Sprintf("Policy(cidr=%s)", key.CIDR)
 }
 
 type PoolListOptions struct {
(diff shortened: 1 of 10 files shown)


lib/backend/pool.go
```diff
@@ -51,6 +51,10 @@ func (key PoolKey) valueType() reflect.Type {
 	return typePool
 }
 
+func (key PoolKey) String() string {
+	return fmt.Sprintf("Policy(cidr=%s)", key.CIDR)
```
caseydavenport: I think Policy -> Pool?

lib/backend/workloadendpoint.go
```diff
@@ -52,6 +61,11 @@ func (key WorkloadEndpointKey) valueType() reflect.Type {
 	return reflect.TypeOf(WorkloadEndpoint{})
 }
 
+func (key WorkloadEndpointKey) String() string {
+	return fmt.Sprintf("WorkloadEndpoint(hostname=%s,orchestrator=%s,workload=%s,name=%s)",
```
caseydavenport: Should have spaces after commas.

caseydavenport:
So all of these "InsufficientIdentifier" errors are in the backend, and thus probably shouldn't be propagated through to the frontend (where field names, etc might be different).

Is the plan to handle that in a separate PR by adding some corresponding function to the frontend?

robbrockbank:
@casey regarding your point about field names being different.  Yup that is a concern - so we'd need to propagate the error, catch it and then convert the field name.  At the moment, I've just named the fields in the backend as per the front-end (So I've effectively postponed this until a later date).  Not ideal I know - but I think we'd need to introduce some generic field mapping machinery which we don't have in place just yet.

caseydavenport:
@robbrockbank Yeah, I definitely don't think we should do it now.  Just wanted to make sure my understanding was right.

The plan for another PR certainly could be "yeah we'll do that in a month or two when we have a real need"

caseydavenport:
LGTM

PR #17, opened 2016-07-26
Cleanup pool resource

caseydavenport:
Updates the api.Pool resource and GoDocs.

Based off #16

### DIFF
--- a/lib/api/pool.go
+++ b/lib/api/pool.go
@@ -26,16 +26,36 @@ type PoolMetadata struct {
 }
 
 type PoolSpec struct {
-	IPIP        bool `json:"ipip,omitempty"`
-	NATOutgoing bool `json:"nat-outgoing"`
-	Ipam        bool `json:"ipam"`
-	Disabled    bool `json:"disabled"`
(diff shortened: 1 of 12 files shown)


lib/api/pool.go
```diff
@@ -26,16 +26,36 @@ type PoolMetadata struct {
...
+	// containers in this pool to destinations outside of this pool
+	// will be masqueraded.
+	NATOutgoing bool `json:"nat-outgoing,omitempty"`
+
+	// When disabled is true, Calico IPAM will not assign
+	// addreses from this pool.
+	Disabled bool `json:"disabled,omitempty"`
+}
+
+type IPIPConfiguration struct {
```
robbrockbank: What's the motivation for having this as a sub-struct?

caseydavenport: In case we need additional configuration for `ipip` in the future.

For example if we want to expose the interface name.  For now that defaults to `tunl0`, but maybe one day we'll want to make it configurable on this API.

caseydavenport:
Commits in #18

PR #18, opened 2016-07-26
IPAM Features / error handling

caseydavenport:
- Fix a bunch of error handling cases to actually handle different errors.

Includes:
- #10 #15 #16 #17

### DIFF
--- a/lib/api/pool.go
+++ b/lib/api/pool.go
@@ -26,16 +26,36 @@ type PoolMetadata struct {
 }
 
 type PoolSpec struct {
-	IPIP        bool `json:"ipip,omitempty"`
-	NATOutgoing bool `json:"nat-outgoing"`
-	Ipam        bool `json:"ipam"`
-	Disabled    bool `json:"disabled"`
+	// Contains configuration for ipip tunneling
+	// for this pool. If not specified, then ipip
+	// tunneling is disabled for this pool.
+	IPIP *IPIPConfiguration `json:"ipip,omitempty"`
+
+	// When nat-outgoing is true, packets sent from Calico networked
+	// containers in this pool to destinations outside of this pool
+	// will be masqueraded.
+	NATOutgoing bool `json:"nat-outgoing,omitempty"`
+
+	// When disabled is true, Calico IPAM will not assign
+	// addreses from this pool.
+	Disabled bool `json:"disabled,omitempty"`
+}
+
+type IPIPConfiguration struct {
+	// When enabled is true, ipip tunneling will be
+	// used to deliver packets to destinations within this
+	// pool.
+	Enabled bool `json:"enabled,omitempty"`
 }
 
 type Pool struct {
 	TypeMetadata
+
+	// Metadata for a Pool.
 	Metadata PoolMetadata `json:"metadata,omitempty"`
-	Spec     PoolSpec     `json:"spec,omitempty"`
+
+	// Specification for a Pool.
+	Spec PoolSpec `json:"spec,omitempty"`
 }
 
 func NewPool() *Pool {
--- a/lib/backend/block_affinity.go
+++ b/lib/backend/block_affinity.go
@@ -25,8 +25,8 @@ import (
 )
 
 var (
-	matchBlockAff = regexp.MustCompile("^/?/calico/ipam/v2/host/([^/]+)/ipv./block/([^/]+)$")
-	typeBlockAff  = reflect.TypeOf(BlockAffinity{})
+	matchBlockAffinity = regexp.MustCompile("^/?calico/ipam/v2/host/([^/]+)/ipv./block/([^/]+)$")
+	typeBlockAff       = reflect.TypeOf(BlockAffinity{})
 )
 
 type BlockAffinityKey struct {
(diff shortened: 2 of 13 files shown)


lib/backend/block_affinity.go
```diff
@@ -25,8 +25,8 @@ import (
 )
 
 var (
-	matchBlockAff = regexp.MustCompile("^/?/calico/ipam/v2/host/([^/]+)/ipv./block/([^/]+)$")
-	typeBlockAff  = reflect.TypeOf(BlockAffinity{})
+	matchBlockAffinity = regexp.MustCompile("^/?calico/ipam/v2/host/([^/]+)/ipv./block/([^/]+)$")
+	typeBlockAff       = reflect.TypeOf(BlockAffinity{})
```
robbrockbank: Aff -> Affinity

lib/client/ipam.go
```diff
@@ -83,7 +79,7 @@ type IPAMInterface interface {
 
 	// ReleasePoolAffinities releases affinity for all blocks within
 	// the specified pool across all hosts.
-	ReleasePoolAffinities(pool common.IPNet) error
+	ReleasePoolAffinities(pool *common.IPNet) error
```
robbrockbank: i feel this shouldn't be a pointer

lib/client/ipam.go
```diff
@@ -419,47 +423,58 @@ func (c ipams) assignFromExistingBlock(
...
 	}
 
 	// Determine the hostname to use.
 	hostname := decideHostname(host)
+	failed := []common.IPNet{}
+	claimed := []common.IPNet{}
 
 	// Verify the requested CIDR falls within a configured pool.
 	if !c.blockReaderWriter.withinConfiguredPools(common.IP{cidr.IP}) {
 		estr := fmt.Sprintf("The requested CIDR (%s) is not within any configured pools.", cidr.String())
```
robbrockbank: Do these errors bubble up to the top?  Reason I ask is that i'm not a big fan of an error where the only details about the error are in a description string.  Suppose (much further down the line) we wanted to provide multi-language support, this doesn't lend itself well to that.  Not suggesting you fix this as I'm still not 100% sure what our error handling plan is.

lib/client/ipam.go
```diff
@@ -513,7 +532,7 @@ func (c ipams) ReleaseHostAffinities(host *string) error {
 
 // ReleasePoolAffinities releases affinity for all blocks within
 // the specified pool across all hosts.
-func (c ipams) ReleasePoolAffinities(pool common.IPNet) error {
+func (c ipams) ReleasePoolAffinities(pool *common.IPNet) error {
```
robbrockbank: don't think this should be a *

lib/client/ipam_block_reader_writer.go
```diff
@@ -224,7 +239,8 @@ func (rw blockReaderWriter) releaseBlockAffinity(host string, blockCIDR common.I
 func (rw blockReaderWriter) withinConfiguredPools(ip common.IP) bool {
 	allPools, _ := rw.client.Pools().List(api.PoolMetadata{})
 	for _, p := range allPools.Items {
-		if p.Metadata.CIDR.IPNet.Contains(ip.IP) {
+		// Compare any enabled pools.
+		if !p.Spec.Disabled && p.Metadata.CIDR.IPNet.Contains(ip.IP) {
```
robbrockbank: Can you not just do Metadata.CIDR.Contains()

lib/backend/block_affinity.go
caseydavenport: Fixed.

lib/client/ipam.go
caseydavenport: Fixed.

lib/client/ipam.go
caseydavenport: Yeah, it does.

At the moment I'm not exposing different error types over the public API for IPAM.  That will be a part of a more comprehensive discussion on how we do error handling in libcalico (which we should have soon-ish).

Until then the errors are just treated opaquely.

lib/client/ipam.go
caseydavenport: Fixed.

lib/client/ipam_block_reader_writer.go
caseydavenport: Hmm.. maybe.  I'll see.

caseydavenport: Yep, can do.

lib/api/pool.go
```diff
@@ -26,16 +26,36 @@ type PoolMetadata struct {
...
 
 type PoolSpec struct {
-	IPIP        bool `json:"ipip,omitempty"`
-	NATOutgoing bool `json:"nat-outgoing"`
-	Ipam        bool `json:"ipam"`
-	Disabled    bool `json:"disabled"`
+	// Contains configuration for ipip tunneling
+	// for this pool. If not specified, then ipip
+	// tunneling is disabled for this pool.
+	IPIP *IPIPConfiguration `json:"ipip,omitempty"`
```
robbrockbank: I don't think this needs to be a pointer.  The zero value for that struct will have IPIP set to false.

caseydavenport: Making it a pointer is more about making it an optional field in the json.

If this isn't a pointer and it's not specified, your json looks like this:

```
spec:
  ipip: {}
```

If it is a pointer and not specified, you get this:
```
spec:
```

This is also the Kubernetes way of doing "optional" fields.  Any non-mandatory fields are pointers and have `omitempty` on them.

tomdee:
LGTM

robbrockbank:
LTGMT

PR #19, opened 2016-07-27
Endpoints

tomdee:


tomdee:
Temporary code until we get a proper fix in place.

PR #20, opened 2016-07-27
Finish off IPAM TODOs

caseydavenport:
Includes a few fixes each in its own commit:
- [x] RemoveIPAMHost implementation
- [x] Minimize datastore writes when releasing IP address
- [x] Implement random block hunt when out of free blocks

### DIFF
--- a/lib/client/ipam.go
+++ b/lib/client/ipam.go
@@ -401,7 +401,7 @@ func (c ipams) assignFromExistingBlock(
 		// Pull out the block.
 		b := allocationBlock{obj.Value.(model.AllocationBlock)}
 
-		glog.V(4).Infof("Got block: %v", b)
+		glog.V(4).Infof("Got block: %+v", b)
 		ips, err = b.autoAssign(num, handleID, host, attrs, true)
 		if err != nil {
 			glog.Errorf("Error in auto assign: %s", err)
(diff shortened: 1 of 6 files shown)


lib/client/ipam.go
```diff
@@ -589,22 +655,18 @@ func (c ipams) RemoveIPAMHost(host *string) error {
...
-	// 		return err
-	// 	}
-	// }
+	// Remove the host tree from the datastore.
+	err := c.client.backend.Delete(&model.KVPair{
+		Key: model.IPAMHostKey{Host: hostname},
+	})
+	if err != nil {
+		if _, ok := err.(common.ErrorResourceDoesNotExist); ok {
+			// Already deleted - carry on.
```
robbrockbank: contract to a single branch?   !ok

caseydavenport: d'oh, yeah...

lib/client/ipam.go
```diff
@@ -589,22 +655,17 @@ func (c ipams) RemoveIPAMHost(host *string) error {
 	// Determine the hostname to use.
 	hostname := decideHostname(host)
 
-	// Release host affinities.
```
robbrockbank: Did you mean to remove this as well?

caseydavenport: Well, I could leave it.  However, removing the IPAM host actually removes all the affinities anyway so I took it out.

I'm not a super big fan of it, but I suppose it's sort of like removing a tier and thus it removes all the policies within the tier with it.

robbrockbank:
One comment - looks like you removed some code that I don't think should have been removed.

lib/client/ipam.go
robbrockbank: I thought it did more than just that - I thought it reconfigured the allocation block to remove the host affinity.

caseydavenport: Right you are.

caseydavenport:
OK - comment addressed!

Good to go?

robbrockbank:
LGTM

PR #23, opened 2016-07-28
Use a generator to iterate through blocks for perf

caseydavenport:


tomdee:
I'm not sure about using a generator. Could you not just make the .getNext() function stateless?

caseydavenport:
@tomdee Let's chat this afternoon.

tomdee:
It's good enough to merge and unblocks broken IPv6 so merging.

PR #25, opened 2016-07-28
Refactor backend API

fasaxc:
- Move etcd client to its own sub-package to make way for making other backends.
- Move datamodel keys and values into a `model` sub-package.
- Remove etcd-ness from model classes.  Generally changed the `asEtcd(Delete)Key()` methods to be `Default(Delete)Path()`; and commented that they're paths suitable for most datastores and used on the Felix API.
- Remove Client struct and use an interface instead.
- Rename DatastoreObject to KVPair
- Rename some other types to shorten.  E.g. `model.KeyInterface` -> `model.Key`.
- ~~Remove special-case in etcd driver for the `Profile` objects and replace with `OverrideGet|Apply|Update` interfaces.  I think we still need to rejig this based on the V1/V2 discussions we had yesterday.~~
- Introduce an adapter object to do the composite key mapping.
- Add config and ready flag to model.


lib/backend/etcd/etcd.go
```diff
@@ -92,21 +103,18 @@ func ConnectEtcdClient(config *api.ClientConfig) (DatastoreReadWriteInterface, e
 }
 
 // Create an entry in the datastore.  This errors if the entry already exists.
-func (c *EtcdClient) Create(d *DatastoreObject) (*DatastoreObject, error) {
-	// Special case the profile
-	switch t := d.Key.(type) {
-	case ProfileKey:
-		return t.Create(c, d)
+func (c *EtcdClient) Create(d *KVPair) (*KVPair, error) {
+	if co, ok := d.Key.(CreateOverrider); ok {
```
robbrockbank: Hmm, this was how i had originally thought of doing it, but I was worried that this would happen:

-  Create is called with a Profile which calls the override (because Profile key implements CreateOverrider)
-  Overridden create calls back into the client with a ProfilesTag ... but I think the ProfileTag also implements the CreateOverrider because the key has the profile key embedded in it.

Is that not the case?

lib/client/tier.go
```diff
@@ -77,34 +77,34 @@ func (h *tiers) List(metadata api.TierMetadata) (*api.TierList, error) {
...
-	l := backend.TierListOptions{
+	l := model.TierListOptions{
 		Name: hm.Name,
 	}
 	return l, nil
 }
 
 // Convert a TierMetadata to a TierKeyInterface
-func (h *tiers) convertMetadataToKeyInterface(m interface{}) (backend.KeyInterface, error) {
+func (h *tiers) convertMetadataToKeyInterface(m interface{}) (model.Key, error) {
```
robbrockbank: Probably should rename these `convertMetadataToKey`

lib/backend/etcd/etcd.go
robbrockbank: I've just tested.. this is the case :-(

robbrockbank: Perhaps we should add the override methods to the Profile object instead of the Key... although I had a sneaky feeling there was a reason I didn't like that - but I can't think of it now.

lib/backend/compat/compat.go
```diff
@@ -0,0 +1,151 @@
...
+			d.Revision = t.Revision
+			return d, nil
+		}
+	}
+	return c.client.Apply(d)
+}
+
+// Delete an entry in the datastore.  This errors if the entry does not exists.
+func (c *ModelAdaptor) Delete(d *KVPair) error {
+	if _, ok := d.Key.(ProfileKey); ok {
```
robbrockbank: don't need this branch do we?

robbrockbank:
LGTM apart from the circle CI failure :-)

fasaxc:
@robbrockbank CI failure is because there's no CI config on this branch. Circle CI branch needs to be merged first

PR #26, opened 2016-07-28
Parse may return a pointer - dereference if it does

robbrockbank:
I think this should fix the problem you were hitting invoking Get() - I had a similar issue with the List() command which I had already fixed.

I've shot Shaun a message asking why the ParseValue() only dereferences non-structs, as my preferred approach would be to fix up that - but I'd like his input on that before I change it that way.


tomdee:
This code looks kind of complicated but if it fixes the problem, I'm going to merge it.

PR #28, opened 2016-07-29
Move files from common into separate specific sub packages

robbrockbank:
Whilst this looks like a big change, it is basically just a renaming PR.

The old common sub package has been deleted, and things have been moved from there to:

```
lib/errors
lib/net
lib/numorstring
lib/validator
```

And the resource manager which was in the API package has been moved into calicoctl since that is the only thing that makes use of it.

The only thing that falls outside those changes (it got caught in the cross fire) was the handling of the default Tier ... basically to make it so that on a create a blank tier is assumed to be "default", otherwise you can play around with the default tier as if it were any other tier.... which is some feedback you gave me ages ago.

### DIFF
--- a/lib/resourcemgr/resourcemgr.go
+++ b/lib/resourcemgr/resourcemgr.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package api
+package resourcemgr
 
 import (
 	"errors"
@@ -24,11 +24,12 @@ import (
 	"io/ioutil"
 	"os"
 
+	"github.com/tigera/libcalico-go/lib/api"
 	. "github.com/tigera/libcalico-go/lib/api/unversioned"
 
 	"github.com/ghodss/yaml"
 	"github.com/golang/glog"
-	"github.com/tigera/libcalico-go/lib/common"
+	"github.com/tigera/libcalico-go/lib/validator"
 )
 
 var helpers map[TypeMetadata]resourceHelper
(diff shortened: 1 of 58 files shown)


lib/types/ip.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package common
+package types
```
caseydavenport: I think we should be looking to break up the "types" package (as it is it isn't much better than "common").

For example, all the IP stuff should probably in a package called `net` or `network`.

lib/types/protocol.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package common
+package types
 
 type Protocol struct {
```
caseydavenport: Does this (and Port, Int32OrString, etc) belong in the `api` package?

lib/resourcemgr/resourcemgr.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package api
+package resourcemgr
```
caseydavenport: Does this actually need its own package?  Where is all of this code called from?

From what I can tell, it has a bunch methods declared as public, but only one of them is ever called from outside of this file in calicoctl/commands/util.go.

caseydavenport:
Generally looking good.

My summary of thoughts would be:
- The `types` package should be dissolved and turned into packages that describe individual sets of function (e.g "net")
- The resourcemgr package doesn't seem to be used anywhere except in calicoctl as far as I can tell.  Do we think other pieces of code will want this someday? Or is it really a part of calicoctl, not libcalico?

lib/types/protocol.go
robbrockbank: they are also fields used in the backend models - so living in api doesn't feel right.

lib/types/ip.go
robbrockbank: maybe, but i don't feel these type definitions are actually adding any additional function - as it happens I'm currently looking into whether we can do something about handling our own marshaling etc. and different ways to validate ... upshot of which is that we could probably do away with all of these additional types - so a little loathe to keep splitting up at the moment.

robbrockbank:
yup resource manager is only used by calicoctl - but arguably it could be used outside if, for example someone wants to write their own YAML based provisioning system or similar.  (seems unlikely... but doesn't see unreasonable having in libcalico.)

lib/types/protocol.go
caseydavenport: Do they belong in the "net" package then, along with `IP` and `IPNet`?

Or, should there be specific "api" and "backend" implementations of all of these?  If we're sharing structs between the `api` and `backend` packages aren't we breaking the isolation to an extent? Sort of plays into what we want to do about the IP and IPNet structs we've defined (I think Kubernetes treats these as strings on the API, so maybe we should do the same and just convert to back as needed?)

caseydavenport:
re: resourcemgr - I think my preference would be to keep it private / part of calicoctl until we think there is a solid need for other clients to use this function, in an effort to keep the public API of libcalico as small as possible.

lib/types/protocol.go
robbrockbank: I don't see that sharing structs between the back end and front end is an issue.  If we were using net.IP as is in both the front end and back end I'm sure we wouldn't question that.

lib/resourcemgr/resourcemgr.go
robbrockbank: I can move into calicoctl.

lib/types/protocol.go
caseydavenport: Yeah, good point.

Though in that case I think I would still question the use of net.IP in the `api` structs at all (but that is a separate issue from this).

My original concern was to make sure package names were based more around specific sets of function.  A package named `types` has strong potential to grow into something that's big and ugly, and not well defined, whereas a package named `net` has much more clearly defined boundaries. This also matches the go package naming conventions.

i.e keeping a `types` package sort of defeats a lot of the purpose of splitting up the old `common` package.

robbrockbank:
I'll move resource manager as part of the next PR - which I've partly implemented and has already mucked around a bit with that module....

robbrockbank:
As discussed, I've split up the types and moved into net and numorstring.  I've also moved the resourcemgr into calicoctl.

PR #29, opened 2016-07-29
Fix host affinity for blocks

caseydavenport:
We weren't limiting the returned blocks based on their affinity, so IP addresses were being assigned from the same block on multiple hosts.

### DIFF
--- a/lib/backend/model/block_affinity.go
+++ b/lib/backend/model/block_affinity.go
@@ -52,13 +52,18 @@ func (key BlockAffinityKey) valueType() reflect.Type {
 }
 
 type BlockAffinityListOptions struct {
-	Host string
+	Host      string
+	IPVersion int
 }
 
 func (options BlockAffinityListOptions) DefaultPathRoot() string {
 	k := "/calico/ipam/v2/host/"
 	if options.Host != "" {
 		k = k + options.Host
+
+		if options.IPVersion != 0 {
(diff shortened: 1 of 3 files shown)


lib/backend/model/block_affinity.go
```diff
@@ -52,13 +52,24 @@ func (key BlockAffinityKey) valueType() reflect.Type {
...
 	if options.Host != "" {
 		k = k + options.Host
+
+		if options.IPVersion != 0 {
+			k = k + fmt.Sprintf("/ipv%d/block", options.IPVersion)
+		}
+	} else {
+		// If no host is specified, the IP version must also
+		// be unspecified.
+		if options.IPVersion != 0 {
```
robbrockbank: Is this necessary?  Why not construct this as per the other backend types?

caseydavenport: Yeah I guess I could do that, but I didn't need that function so I didn't do it and just added this so that no-one was surprised when they tried to get all IPv4 block affinities and it didn't work.

caseydavenport:
@robbrockbank - ready for another look!

robbrockbank:
LGTM (does this require the other PR? ... wasn't sure so didn't merge)

caseydavenport:
Nope, this should be good on its own.

PR #31, opened 2016-08-02
Add log action and prefix

robbrockbank:


lib/backend/model/rule.go
```diff
@@ -20,7 +20,8 @@ import (
 )
 
 type Rule struct {
-	Action string `json:"action,omitempty" validate:"backendaction"`
+	Action    string  `json:"action,omitempty" validate:"backendaction"`
+	LogPrefix *string `json:"log_prefix,omitempty" validate:"omitempty"`
```
caseydavenport: Is there a mechanism in the code to return an error if LogPrefix is specified but the action is not "log"?

lib/api/rule.go
```diff
@@ -24,7 +24,8 @@ import (
 )
 
 type Rule struct {
-	Action string `json:"action" validate:"action"`
+	Action    string  `json:"action" validate:"action"`
+	LogPrefix *string `json:"logPrefix,omitempty" validate:"omitempty"`
```
caseydavenport: If the action is "log", but the LogPrefix is nil, is there a default value, or just no prefix (null in etcd)?

caseydavenport:
A couple of questions but otherwise LGTM.

I think we'll want that validation on the frontend API.  We shouldn't let people write config on that API that doesn't make sense.

lib/backend/model/rule.go
robbrockbank: The action does not need to be log.

caseydavenport:
@robbrockbank - which milestone should this be placed in?

PR #33, opened 2016-08-02
Use ResourceMetadata interface

robbrockbank:
Add a ResourceMetadata interface and update internal interfaces to use ResourceMetadata instead of interface{}

Whilst ObjectMetadata doesn't yet contain any fields - it will do - namely revision information, and possibly other stuff in the future.   Adding the interface now allows the internal interfaces to look a little nicer.


caseydavenport:
LGTM.

PR #38, opened 2016-08-02
Add ID to profile rules and ensure null is to written to tags and labels

robbrockbank:
Add the profile ID to the profile rules object.  This is not part of the actual etcd felix spec, but is used by calioctl for some unknown reason.

Also, fix up processing to ensure null values are not written into empty profile tags or empty profile labels.

### DIFF
--- a/lib/backend/compat/compat.go
+++ b/lib/backend/compat/compat.go
@@ -15,6 +15,7 @@
 package compat
 
 import (
+	"github.com/golang/glog"
 	"github.com/tigera/libcalico-go/lib/backend/api"
 	. "github.com/tigera/libcalico-go/lib/backend/model"
 )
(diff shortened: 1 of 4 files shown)


lib/backend/compat/compat.go
```diff
@@ -144,5 +144,13 @@ func toTagsLabelsRules(d *KVPair) (t, l, r *KVPair) {
 		Value: p.Rules,
 	}
 
+	// Fix up tags and labels so to be empty values rather than nil.
```
caseydavenport: Is this because Felix can't handle null values? Or calicoctl? Or both?

robbrockbank: This is purely for Felix - we shouldn't be writing null where Felix would be expecting a dictionary or a list.

robbrockbank:
@caseydavenport :  I added a couple of logs and a little more commenting.

caseydavenport:
LGTM

PR #50, opened 2016-08-04
produce artifacts.

djosborne:


circle.yml
```diff
@@ -25,3 +25,8 @@ checkout:
 test:
   override:
     - cd ${HOME}/.go_project/src/github.com/tigera/libcalico-go && ./run-uts
+    - cd ${HOME}/.go_project/src/github.com/tigera/libcalico-go && make bin/calicoctl
```
robbrockbank: should this be part of test?

djosborne: I actually think it should. Doing it anywhere else doesn't really fit. Options include:
```
machine: adjusting the VM to your preferences and requirements
checkout: checking out and cloning your git repo
dependencies: setting up your project’s language-specific dependencies
database: preparing the databases for your tests
test: running your tests
deployment: deploying your code to your web servers
```

I can't see a better fit that doesn't upset the current testing pipeline.

robbrockbank:
ok.  LGTM

PR #53, opened 2016-08-05
Don't omit empty endpoint data in backend

caseydavenport:
Python `calicoctl`, for one, seems to expect fields to be there even if empty.

e.g
```
core@k8s-master ~ $ calicoctl endpoint show
Traceback (most recent call last):
  File "<string>", line 145, in <module>
  File "calico_ctl/endpoint.py", line 145, in endpoint
  File "calico_ctl/endpoint.py", line 165, in endpoint_show
  File "site-packages/pycalico/datastore.py", line 128, in wrapped
  File "site-packages/pycalico/datastore.py", line 1184, in get_endpoints
  File "site-packages/pycalico/datastore_datatypes.py", line 292, in from_json
KeyError: 'ipv6_nets'
```

### DIFF
--- a/lib/backend/model/workloadendpoint.go
+++ b/lib/backend/model/workloadendpoint.go
@@ -144,8 +144,8 @@ type WorkloadEndpoint struct {
 	State      string            `json:"state"`
 	Name       string            `json:"name"`
 	Mac        net.MAC           `json:"mac"`
-	ProfileIDs []string          `json:"profile_ids,omitempty"`
-	IPv4Nets   []net.IPNet       `json:"ipv4_nets,omitempty"`
-	IPv6Nets   []net.IPNet       `json:"ipv6_nets,omitempty"`
-	Labels     map[string]string `json:"labels,omitempty"`
+	ProfileIDs []string          `json:"profile_ids"`
+	IPv4Nets   []net.IPNet       `json:"ipv4_nets"`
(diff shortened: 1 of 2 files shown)


lib/backend/model/workloadendpoint.go
```diff
@@ -144,8 +144,8 @@ type WorkloadEndpoint struct {
 	State      string            `json:"state"`
 	Name       string            `json:"name"`
 	Mac        net.MAC           `json:"mac"`
-	ProfileIDs []string          `json:"profile_ids,omitempty"`
-	IPv4Nets   []net.IPNet       `json:"ipv4_nets,omitempty"`
-	IPv6Nets   []net.IPNet       `json:"ipv6_nets,omitempty"`
-	Labels     map[string]string `json:"labels,omitempty"`
+	ProfileIDs []string          `json:"profile_ids"`
+	IPv4Nets   []net.IPNet       `json:"ipv4_nets"`
```
robbrockbank: You might need to initialize the slices and maputo be empty if they aren't specified in the API - otherwise I think they'll be null in the JSON

caseydavenport: Any suggestions for where this should be done?

caseydavenport: See below.

caseydavenport:
@robbrockbank - Care to take another look?

PR #55, opened 2016-08-08
Add String() for Rules and Policies.

fasaxc:


lib/backend/model/rule.go
```diff
@@ -46,3 +49,103 @@ type Rule struct {
...
+		parts = append(parts, "code", strconv.Itoa(*r.ICMPCode))
+	}
+	if r.NotICMPType != nil {
+		parts = append(parts, "!type", strconv.Itoa(*r.NotICMPType))
+	}
+	if r.NotICMPCode != nil {
+		parts = append(parts, "!code", strconv.Itoa(*r.NotICMPCode))
+	}
+
+	if r.SrcTag != "" || len(r.SrcPorts) > 0 || r.SrcNet != nil || r.SrcSelector != "" ||
```
robbrockbank: More maintainable if you keep separate 'from' and 'to' parts lists then at the end you can include the word "from" if there are any from parts, sim. for to.   WDYT?

lib/backend/model/rule.go
```diff
@@ -46,3 +49,103 @@ type Rule struct {
...
 }
+
+func (r Rule) String() string {
+	parts := make([]string, 0)
+	if r.Action != "" {
+		parts = append(parts, r.Action)
+	} else {
+		parts = append(parts, "allow")
+	}
+	if r.Protocol != nil {
```
robbrockbank: Missing !Protocol

lib/backend/model/rule.go
fasaxc: Sounds good.

robbrockbank:
LGTM - but branch has conflicts.

PR #58, opened 2016-08-12
Remove profile ID from backend rules

robbrockbank:
To be merged after calicoctl release that handles profiles without the id field in the backend,


tomdee:
OK, LGTM

PR #59, opened 2016-08-15
Remove utils file and don't allow workload endpoints to be created

robbrockbank:
Split the utils file into:
-  client.go   [handles creating a new client]
-  resources.go  [handles generic resource command processing]

A few additional code comments.

Don't allow workload endpoints to be created.  They can be viewed though using get.


tomdee:
Looks fine to me - as per our discussion you'll think some more and open another PR to handle the get part of WorkloadEndpoints

PR #61, opened 2016-08-16
Added BGP peer resource management commands

robbrockbank:


lib/backend/model/bgppeer.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+type BGPPeerListOptions struct {
+	Hostname string
+	PeerIP   net.IP
+}
+
+func (options BGPPeerListOptions) DefaultPathRoot() string {
+	if options.Hostname == "" {
+		return "/calico/bgp/v1"
```
gunjan5: the string "/calico/bgp/v1" is being repeated multiple times, might be a good idea to make it a constant or a global variable, it would be easier to modify if it ever changes

lib/client/bgppeer.go
```diff
@@ -0,0 +1,130 @@
...
+func newBGPPeers(c *Client) *peers {
+	return &peers{c}
+}
+
+// Create creates a new pool.
+func (h *peers) Create(a *api.BGPPeer) (*api.BGPPeer, error) {
+	return a, h.c.create(*a, h)
+}
+
+// Create creates a new pool.
```
gunjan5: comment should be for Update :)

lib/client/bgppeer.go
```diff
@@ -0,0 +1,130 @@
...
+func (h *peers) Update(a *api.BGPPeer) (*api.BGPPeer, error) {
+	return a, h.c.update(*a, h)
+}
+
+// Create creates a new pool.
```
gunjan5: should be a comment for `Apply`

lib/client/bgppeer.go
```diff
@@ -0,0 +1,130 @@
...
+
+// List takes a Metadata, and returns the list of peers that match that Metadata
+// (wildcarding missing fields)
+func (h *peers) List(metadata api.BGPPeerMetadata) (*api.BGPPeerList, error) {
+	l := api.NewBGPPeerList()
+	err := h.c.list(metadata, h, l)
+	return l, err
+}
+
+// Convert a BGPPeerMetadata to a BGPPeerListInterface
```
gunjan5: this is minor, but GoDoc can pick up the comment for documents if it begins with `convertMetadataToListInterface`. So like most of the other comments in this, it would be nice if it's `// convertMetadataToListInterface converts a BGPPeerMetadata to a BGPPeerListInterface`

lib/client/bgppeer.go
```diff
@@ -0,0 +1,130 @@
...
+func (h *peers) convertMetadataToListInterface(m unversioned.ResourceMetadata) (model.ListInterface, error) {
+	pm := m.(api.BGPPeerMetadata)
+	l := model.BGPPeerListOptions{
+		PeerIP:   pm.PeerIP,
+		Hostname: pm.Hostname,
+	}
+	return l, nil
+}
+
+// Convert a BGPPeerMetadata to a BGPPeerKeyInterface
```
gunjan5: Same as previous comment

lib/client/bgppeer.go
```diff
@@ -0,0 +1,130 @@
...
+func (h *peers) convertMetadataToKey(m unversioned.ResourceMetadata) (model.Key, error) {
+	pm := m.(api.BGPPeerMetadata)
+	k := model.BGPPeerKey{
+		PeerIP:   pm.PeerIP,
+		Hostname: pm.Hostname,
+	}
+	return k, nil
+}
+
+// Convert an API BGPPeer structure to a Backend BGPPeer structure
```
gunjan5: Same as previous comment

lib/client/bgppeer.go
```diff
@@ -0,0 +1,130 @@
...
+		Value: model.BGPPeer{
+			PeerIP: ap.Metadata.PeerIP,
+			ASNum:  ap.Spec.ASNum,
+		},
+	}
+
+	return &d, nil
+}
+
+// Convert a Backend BGPPeer structure to an API BGPPeer structure
```
gunjan5: same here :)

lib/client/client.go
```diff
@@ -78,6 +78,11 @@ func (c *Client) WorkloadEndpoints() WorkloadEndpointInterface {
 	return newWorkloadEndpoints(c)
 }
 
+// Profiles returns an interface for managing BGP peer resources.
```
gunjan5: comment seems to be for another method

lib/backend/model/bgppeer.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/tigera/libcalico-go/lib/net"
+)
+
+var (
+	matchGlobalBGPPeer = regexp.MustCompile("^/?calico/bgp/v1/global/peer_v./([^/]+)$")
+	matchHostBGPPeer   = regexp.MustCompile("^/?calico/bgp/v1/host/([^/]+)/peer_v./([^/]+)$")
+	typeBGPPeer        = reflect.TypeOf(BGPPeer{})
+)
+
+//TODO: Can we get rid of the IP in the IDs and replace with an arbitrary name.
```
gunjan5: might be a good idea to remove `TODO` (unless it's necessary)

lib/client/bgppeer.go
robbrockbank: Yup good spot - this was copied from code that I wrote before I knew about GoDoc - so still getting used to the different conventions.

lib/backend/model/bgppeer.go
robbrockbank: These string values cannot change without breaking things so I'm more inclined to leave this as is so that it's more readable what the key actually is.

calicoctl/commands/resources.go
```diff
@@ -124,6 +124,16 @@ func getResourceFromArguments(args map[string]interface{}) (unversioned.Resource
 			p.Metadata.CIDR = *cidr
 		}
 		return *p, nil
+	case "bgpPeer":
```
caseydavenport: A couple quick questions:

1) As it is now I think we're pretty strict about wording.  kubectl lets you do things like:
```
kubectl get networkpolicy
kubectl get networkpolicies
```
both of which return the same stuff, which is pretty nice in terms of UX.  Definitely _not_ something to fix in this PR, but might be worth an issue in the GA milestone.

lib/backend/model/bgppeer.go
robbrockbank: Yeah I'll just raise an issue.

calicoctl/commands/resources.go
robbrockbank: I was just about to raise an issue about that.

robbrockbank: issue raised - this also includes the plural thing too

calicoctl/commands/resources.go
```diff
@@ -124,6 +124,16 @@ func getResourceFromArguments(args map[string]interface{}) (unversioned.Resource
...
+		p := api.NewBGPPeer()
+		if name != "" {
+			err := p.Metadata.PeerIP.UnmarshalText([]byte(name))
+			if err != nil {
+				return nil, err
+			}
+		}
+		p.Metadata.Hostname = hostname
```
caseydavenport: Does this need to set `p.Metadata.Name` here as well?  Doesn't look like `name` is being used.

Also, so I understand, why is the defaulting of `Name` occurring here and not in libcalico?

lib/api/bgppeer.go
```diff
@@ -0,0 +1,60 @@
...
+	. "github.com/tigera/libcalico-go/lib/net"
+)
+
+
+type BGPPeerMetadata struct {
+	ObjectMetadata
+	// The hostname of the node that is peering with this peer.  If left blank,
+	// Calico assumes this is a "global" BGP peer - i.e. it peers with every Calico
+	// node in the deployment.
+	Hostname string `json:"hostname,omitempty" validate:"omitempty,name"`
```
caseydavenport: Stick a newline before the comments.

lib/api/bgppeer.go
```diff
@@ -0,0 +1,60 @@
...
+	// The IP address of the peer.
+	PeerIP   IP     `json:"peerIP" validate:"omitempty,ip"`
+}
+
+type BGPPeerSpec struct {
+	// The AS Number of the peer
```
caseydavenport: Be consistent with punctuation!

lib/api/bgppeer.go
```diff
@@ -0,0 +1,60 @@
...
+	ASNum int `json:"asNum" validate:"omitempty,asn"`
```
caseydavenport: I'd prefer `asNumber` in the json to avoid excessive abbreviation.

lib/api/bgppeer.go
```diff
@@ -0,0 +1,60 @@
...
+
+
+type BGPPeerMetadata struct {
+	ObjectMetadata
+	// The hostname of the node that is peering with this peer.  If left blank,
+	// Calico assumes this is a "global" BGP peer - i.e. it peers with every Calico
+	// node in the deployment.
+	Hostname string `json:"hostname,omitempty" validate:"omitempty,name"`
+	// The IP address of the peer.
+	PeerIP   IP     `json:"peerIP" validate:"omitempty,ip"`
```
caseydavenport: What does the `validate:"omitempty"` part do here (and above?).  This field isn't optional, yeah?

lib/backend/model/bgppeer.go
```diff
@@ -0,0 +1,117 @@
...
+	PeerIP   net.IP `json:"-" validate:"required"`
+}
+
+func (key BGPPeerKey) DefaultPath() (string, error) {
+	if key.PeerIP.IP == nil {
+		return "", errors.ErrorInsufficientIdentifiers{Name: "peerIP"}
+	}
+	if key.Hostname == "" {
+		e := fmt.Sprintf("/calico/bgp/v1/global/peer_v%d/%s",
+			key.PeerIP.Version(), key.PeerIP)
```
caseydavenport: Doesn't this need to be `key.PeerIP.String()`?

caseydavenport: Nevermind, Go detects the `String()` method and does it automatically, cool!

robbrockbank: Not with the %s format - that will look for a Stringer interface.

lib/api/bgppeer.go
robbrockbank: Yup - good point - the Northbound API needs to have less abbreviations.  Consider that noted for future.

lib/backend/model/bgppeer.go
```diff
@@ -0,0 +1,117 @@
...
+type BGPPeerListOptions struct {
+	Hostname string
+	PeerIP   net.IP
+}
+
+func (options BGPPeerListOptions) DefaultPathRoot() string {
+	if options.Hostname == "" {
+		return "/calico/bgp/v1"
+	} else if options.PeerIP.IP == nil {
+		return fmt.Sprintf("/calico/bgp/v1/host/%s",
```
caseydavenport: This doesn't need to be split on two lines.

calicoctl/commands/resources.go
robbrockbank: BGPPeer does not have a Name field.  It's primary identifier is the IP address.

Now, the get command syntax currently assumes that every object is identified by a name field ... which it isn't.  :-(   So I think some work is actually required to reconsider what the shape of a generic get command should be.  I'll raise this as an issue to make sure I don't forget about it - but I *think* this will be a post-alpha thing.  Maybe we could have a quick brain storm about that.

lib/backend/model/bgppeer.go
```diff
@@ -0,0 +1,117 @@
...
+	} else {
+		glog.V(2).Infof("%s didn't match regex", ekey)
+		return nil
+	}
+
+	if options.PeerIP.IP != nil && !options.PeerIP.Equal(peerIP.IP) {
+		glog.V(2).Infof("Didn't match peerIP %s != %s", options.PeerIP.String(), peerIP.String())
+		return nil
+	}
+	if options.Hostname != "" && hostname != options.Hostname {
```
caseydavenport: If I understand correctly, this means there is no way to get just global peers, or just non-global peers?

i.e If `options.Hostname == ""` then _all_ BGP peers will be returned no matter what the hostname on the peer is, since we only do the hostname check when `options.Hostname != ""`.  Don't we want a way to differentiate between "get me all peers across all hosts" and "get me all global peers"?

lib/api/bgppeer.go
robbrockbank: Oooh - I'm glad you spotted that!   My original thought process was we need to be able to handle the Spec fields not being assigned for a get or a delete!  But that just doesn't make sense since we don't validate the Spec on a get or a delete.  I will remove!  That covers the ASNumber field.

Now the IP address field is interesting.  That is part of the Metadata - and arguably we should validate the Metadata - but a get needs to allow fields being missing from the Metadata - so it'll list rather than get a single value.  But in hindsight again - I think we just don't want to validate this on a get at all - therefore I'll remove for the IP field as well.  I'll check the other resource types for this too.

robbrockbank: Thinking about it we could have a separate validate-list tag that validates metadata fields for enumerating

calicoctl/commands/resources.go
caseydavenport: RIght, so there is no Name field right now, but there should be!

caseydavenport:
@robbrockbank - First pass is done, back to you!

calicoctl/commands/resources.go
robbrockbank: Hmmm, what about IP Pools?  ;-)

lib/backend/model/bgppeer.go
robbrockbank: Grrr - I knew there was something I wanted to raise yesterday in the demo and that was it.  Ugh.

Yes that is completely true.   Let's have a quick chat about that.

calicoctl/commands/resources.go
robbrockbank: Yes as per my TODO.

However, it is not as simple as that.  The IP index is used to construct an IP-version specific datastore key - therefore with the current data model it is necessary to have the IP address as an index field (i.e. have it in the Metadata).   I assume that was the reason you kept the CIDR as an ID field for the IP Pool rather than having an arbitrary name field?

lib/backend/model/bgppeer.go
robbrockbank: We could get into the realm of having the hostname as a pointer in the struct so it has a "" and a nil value - but I think that interface kinda sucks.

calicoctl/commands/resources.go
caseydavenport: Yeah... the IP version is the bit that messed this up for pools as well.

Fine if this doesn't happen in the alpha timeframe, but I think we want to do this for GA.  It probably means that the way we handle identifying information isn't quite right.

lib/scope/globalornode.go
```diff
@@ -0,0 +1,84 @@
...
+func (f GlobalOrNode) MarshalJSON() ([]byte, error) {
+	switch f {
+	case Undefined:
+		return json.Marshal(UndefinedStr)
+	case Global:
+		return json.Marshal(GlobalStr)
+	case Node:
+		return json.Marshal(NodeStr)
+	default:
+		return nil, fmt.Errorf("Unrecognised scope value `%d`", f)
```
caseydavenport: "Unrecognized" ;)

lib/api/bgppeer.go
```diff
@@ -0,0 +1,66 @@
...
+
+type BGPPeerMetadata struct {
+	ObjectMetadata
+
+	// The scope of the peer.  This may be global or node.  A global peer is a
+	// BGP device that peers with all Calico nodes.  A node peer is a BGP device that
+	// peers with the specified Calico node (specified by the node hostname).
+	Scope scope.GlobalOrNode `json:"scope" validate:"omitempty,scopeglobalornode"`
+
+	// The hostname of the node that is peering with this peer.  This is only valid
```
caseydavenport: "only valid" is a bit too weak for what we want, I think.  This field is `mandatory` for host scoped peers, and must not be defined for global scoped peers.

lib/api/bgppeer.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/tigera/libcalico-go/lib/scope"
+)
+
+type BGPPeerMetadata struct {
+	ObjectMetadata
+
+	// The scope of the peer.  This may be global or node.  A global peer is a
+	// BGP device that peers with all Calico nodes.  A node peer is a BGP device that
+	// peers with the specified Calico node (specified by the node hostname).
+	Scope scope.GlobalOrNode `json:"scope" validate:"omitempty,scopeglobalornode"`
```
caseydavenport: I think a `GlobalOrNode` field is a needlessly restrictive concept in the API. 

Why not just make it of type `Scope`, which is essentially a string and then define a few const scopes.  That leaves us expansion for things like `Rack` scope, `Cluster` scope, etc as we add these concepts to the API.

lib/backend/model/bgppeer.go
```diff
@@ -0,0 +1,149 @@
...
+		return e, nil
+	case scope.Node:
+		if key.Hostname == "" {
+			return "", errors.ErrorInsufficientIdentifiers{Name: "hostname"}
+		}
+		e := fmt.Sprintf("/calico/bgp/v1/host/%s/peer_v%d/%s",
+			key.Hostname, key.PeerIP.Version(), key.PeerIP)
+		return e, nil
+	}
+	panic(fmt.Errorf("Unexpected scope value: %d", key.Scope))
```
caseydavenport: Is `panic` right here?  Shouldn't this just be a validation error?

lib/backend/model/bgppeer.go
```diff
@@ -0,0 +1,149 @@
...
+func (key BGPPeerKey) DefaultDeletePath() (string, error) {
+	return key.DefaultPath()
+}
+
+func (key BGPPeerKey) valueType() reflect.Type {
+	return typeBGPPeer
+}
+
+func (key BGPPeerKey) String() string {
+	if key.Hostname == "" {
```
caseydavenport: This should use key.Scope

lib/backend/model/bgppeer.go
robbrockbank: possibly, but the only way we can get a bad value here is if the a user of the client API passes in a bad value.   This isn't possible through calicoctl - since the json marshaling won't allow it.  

Odd though - for some reason I thought this didn't return an error which is why i went for the pragmatic approach here.  I'll fix up to return an error.  If we do define a more generic scope field then we'll need to handle that.

lib/api/bgppeer.go
caseydavenport: Also, I don't see anywhere that we're actually validating the second bit to be the case. Maybe we can do something quick in this PR that prevents people from specifying Hostname for Global peers?

lib/backend/model/bgppeer.go
caseydavenport: I think that is right.  It might not be possible through calicoctl, but that doesn't mean that it won't be possible through all possible future users of this interface.

lib/scope/globalornode.go
```diff
@@ -79,6 +79,6 @@ func (f GlobalOrNode) MarshalJSON() ([]byte, error) {
 	case Node:
 		return json.Marshal(NodeStr)
 	default:
-		return nil, fmt.Errorf("Unrecognised scope value `%d`", f)
+		return nil, fmt.Errorf("unknown scope value `%d`", f)
```
caseydavenport: I'll accept it... this time.

robbrockbank: buah ha ha.... stoopid US spellings

caseydavenport:
LGTM - give it a squash!

Issue #75, opened 2016-08-19
Order table output row


robbrockbank:
The get command does not order the table output for the results.  We should consider adding a sort column or columns.

May not be easy to do with the template engine - so may need to reconsider how we output the tables.

robbrockbank:
I would probably just re-implement the tab writer so that it can handle ordering based on the contents of the columns (in column order)

caseydavenport:
This is actually a calicoctl issue.

PR #77, opened 2016-08-21
Build out calicoctl get command to display PS style table outputs (and group resource specific processing)

robbrockbank:
This PR adds a number of different output options to the `calicoctl get` command.  This includes:
-  PS style output (now the default)
-  Wide PS style output (i.e. more columns)
-  Custom headings PS style output (user can choose from any column that has been defined)
-  Custom golang template
-  Custom golang template file
-  JSON

As part of this work, I've added more functionality to the resource manager to tidy up the resource-specific code that was littered around calicoctl.  Now the resource manager has all of the resource-specific mappings to provide all resource management actions (apply, create etc.).

### DIFF
--- a/calicoctl/commands/get.go
+++ b/calicoctl/commands/get.go
@@ -16,23 +16,18 @@ package commands
 
 import (
 	"github.com/docopt/docopt-go"
-	"github.com/tigera/libcalico-go/lib/api"
-	"github.com/tigera/libcalico-go/lib/client"
 
 	"fmt"
+	"strings"
 
-	"github.com/ghodss/yaml"
 	"github.com/golang/glog"
-	"github.com/tigera/libcalico-go/lib/api/unversioned"
 )
 
 func Get(args []string) error {
 	doc := EtcdIntro + `Display one or many resources identified by file, stdin or resource type and name.
 
-Possible resource types include: policy
-
-By specifying the output as 'template' and providing a Go template as the value
-of the --template flag, you can filter the attributes of the fetched resource(s).
+By specifying the output as 'go-template' and providing a Go template as the value
+of the --go-template flag, you can filter the attributes of the fetched resource(s).
 
 Usage:
   calicoctl get ([--tier=<TIER>] [--hostname=<HOSTNAME>] [--scope=<SCOPE>] (<KIND> [<NAME>]) |
@@ -49,7 +44,8 @@ Examples:
 
 Options:
   -f --filename=<FILENAME>     Filename to use to get the resource.  If set to "-" loads from stdin.
-  -o --output=<OUTPUT FORMAT>  Output format.  One of: yaml, json.  [Default: yaml]
+  -o --output=<OUTPUT FORMAT>  Output format.  One of: ps, wide, custom-columns=..., yaml, json,
+                               go-template=..., go-template-file=...   [Default: ps]
   -t --tier=<TIER>             The policy tier.
   -n --hostname=<HOSTNAME>     The hostname.
   -c --config=<CONFIG>         Filename containing connection configuration in YAML or JSON format.
@@ -66,8 +62,54 @@ Options:
 		return nil
 	}
 
-	cmd := get{}
-	results := executeConfigCommand(parsedArgs, cmd)
+	var rp resourcePrinter
+	output := parsedArgs["--output"].(string)
+	switch output {
+	case "yaml":
+		rp = resourcePrinterYAML{}
+	case "json":
+		rp = resourcePrinterJSON{}
+	case "ps":
+		rp = resourcePrinterTable{wide: false}
+	case "wide":
+		rp = resourcePrinterTable{wide: true}
+	default:
+		// Output format may be a key=value pair, so split on "=" to find out.  Pull
+		// out the key and value, and split the value by "," as some options allow
+		// a multiple-valued value.
+		outputParms := strings.SplitN(output, "=", 2)
+		outputKey := outputParms[0]
+		outputValue := ""
+		outputValues := []string{}
+		if len(outputParms) == 2 {
+			outputValue = outputParms[1]
+			outputValues = strings.Split(outputValue, ",")
+		}
+
+		switch outputKey {
+		case "go-template":
+			if outputValue == "" {
+				return fmt.Errorf("need to specify a template")
+			}
+			rp = resourcePrinterTemplate{template: outputValue}
+		case "go-template-file":
+			if outputValue == "" {
+				return fmt.Errorf("need to specify a template file")
+			}
+			rp = resourcePrinterTemplateFile{templateFile: outputValue}
+		case "custom-columns":
+			if outputValue == "" {
+				return fmt.Errorf("need to specify at least one column")
+			}
+			rp = resourcePrinterTable{headings: outputValues}
+		}
+	}
+
+	if rp == nil {
+		return fmt.Errorf("unrecognized output format '%s'", output)
+	}
+
+	results := executeConfigCommand(parsedArgs, actionList)
 	glog.V(2).Infof("results: %+v", results)
 
 	if results.err != nil {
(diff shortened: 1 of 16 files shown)


calicoctl/commands/get.go
```diff
@@ -66,53 +64,50 @@ Options:
...
+	default:
+		// Output format may be a key=value pair, so split on "=" to find out.  Pull
+		// out the key and value, and split the value by "," as some options allow
+		// a multiple-valued value.
+		outputParms := strings.SplitN(output, "=", 2)
+		if len(outputParms) == 2 {
+			outputKey := outputParms[0]
+			outputValue := outputParms[1]
+			outputValues := strings.Split(outputValue, ",")
+			if len(outputParms) == 2 {
```
caseydavenport: Is this right?  We're already within a `len(outputParms) == 2` if statement.

calicoctl/commands/get.go
```diff
@@ -66,53 +64,50 @@ Options:
...
+				switch outputKey {
+				case "go-template":
+					rp = resourcePrinterTemplate{template: outputValue}
```
caseydavenport: Should these all be using outputValues? Split will return the original if the split character isn't present.

robbrockbank: I don't think we should - only certain options expect a comma separated value - so for those it's better to treat the value as a single thing rather than break it up and ignore anything after the first comma.

calicoctl/commands/get.go
robbrockbank: Urm no - that would be surplus to requirements.

calicoctl/commands/get.go
caseydavenport: Shouldn't we be error-ing if a comma separated value is passed to them then?  At the moment the extra stuff will just be ignored.

calicoctl/commands/get.go
```diff
@@ -66,53 +64,60 @@ Options:
...
+	case "ps":
+		rp = resourcePrinterTable{wide: false}
+	case "wide":
+		rp = resourcePrinterTable{wide: true}
+	default:
+		// Output format may be a key=value pair, so split on "=" to find out.  Pull
+		// out the key and value, and split the value by "," as some options allow
+		// a multiple-valued value.
+		outputParms := strings.SplitN(output, "=", 2)
+		outputKey := outputParms[0]
```
robbrockbank: Casey - I reworked this to provide better error messages when the value is not supplied for the template, template file or custom columns.

calicoctl/commands/get.go
robbrockbank: Extra stuff will not be ignored!  For the options where we only expect a single value, then the single value is passed in, commas and all.

For the template-filename we will error if there is a comma in the name - because that won't be valid.

In the case of a template string, that may validly have a comma in it.  At the moment only the custom-columns uses the outputValues so I could move it into there if it's bugging you, but I kept it outside because I think it looked a little neater and we might end up adding other options that do have a comma separated list of things.

calicoctl/commands/get.go
```diff
@@ -66,53 +62,60 @@ Options:
...
+		outputValue := ""
+		outputValues := []string{}
+		if len(outputParms) == 2 {
+			outputValue = outputParms[1]
+			outputValues = strings.Split(outputValue, ",")
+		}
+
+		switch outputKey {
+		case "go-template":
+			if outputValue == "" {
```
caseydavenport: This doesn't validate against the case that someone puts an `outputValue` that has commas in it (meaning multiple output values).  Not sure if we care so much... but you could probably do this instead:

```
if len(outputValues) != 1 {
  ...
}
```

calicoctl/commands/printer.go
```diff
@@ -0,0 +1,194 @@
...
+	print(resources []unversioned.Resource) error
+}
+
+// JSON output formatter
+type resourcePrinterJSON struct{}
+
+func (r resourcePrinterJSON) print(resources []unversioned.Resource) error {
+	// For simplicity convert the slice of resources (which may contain resource lists)
+	// to expand any lists - so we just work with a slice of real resource types.
+	resources = convertToSliceOfResources(resources)
```
caseydavenport: I'm not sure I really understand this...

calicoctl/commands/get.go
caseydavenport: d'oh, you're right...

calicoctl/commands/printer.go
```diff
@@ -0,0 +1,194 @@
...
+	if output, err := json.MarshalIndent(resources, "", "  "); err != nil {
+		return err
+	} else {
+		fmt.Printf("%s\n", string(output))
+	}
+	return nil
+}
+
+//YAML output formatter
```
caseydavenport: Add a space after those `//`

caseydavenport: Also, GoDoc-ify it yeah?  So, `resourcePrinterYAML is ...`

calicoctl/commands/printer.go
```diff
@@ -0,0 +1,194 @@
...
+
+func (r resourcePrinterTable) print(resources []unversioned.Resource) error {
+	glog.V(2).Infof("Output in table format (wide=%v)", r.wide)
+	for idx, resource := range resources {
+		// Get the resource manager for the resource type.
+		rm := resourcemgr.GetResourceManager(resource)
+
+		// If there are multiple resources then make sure we leave a gap
+		// between each table.
+		if idx > 0 {
```
caseydavenport: Don't we want to do this all the time?  Just do it _after_ the resource is printed.

This way we get a nice gap between the printed table and the next shell prompt, even when only a single resource is being printed.

calicoctl/commands/get.go
robbrockbank: no you can't - at least not very easily  - the template could have commas in...

calicoctl/commands/printer.go
robbrockbank: Yeah - it's a brain reprogramming thing! :-)

calicoctl/commands/printer.go
robbrockbank: Yeah it probably makes sense.

calicoctl/commands/printer.go
robbrockbank: I'll try and clarify the text a little - I have an issue to get rid of this code at some point.

calicoctl/resourcemgr/tier.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/tigera/libcalico-go/lib/client"
+)
+
+func init() {
+	registerResource(
+		api.NewTier(),
+		api.NewTierList(),
+		[]string{"NAME"},
+		[]string{"NAME", "ORDER"},
+		map[string]string{
```
caseydavenport: I don't see why we shouldn't have the order shown by default.  It's only two columns, and there is no reason the wide output can't be the same as the non-wide output for some resources.

calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -19,63 +19,100 @@ import (
...
 // ResourceHelper encapsulates details about a specific version of a specific resource:
 //
 // 	-  The type of resource (Kind and Version).  This includes the list types (even
 //	   though they are not strictly resources themselves).
 // 	-  The concrete resource struct for this version
+//	-  Template strings used to format output for each resource type.
+//	-  Functions to handle resource management actions (apply, create, update, delete, list).
+//         These functions are an untyped interface (generic Resource interfaces) that map through
+//         to the Calicc clients typed interface.
 type resourceHelper struct {
```
caseydavenport: Not sure if this is the right place.. but we need a way of specifying sorting behavior when listing resources.

For example, tiers are currently listed alphabetically, but it would be really nice if they were sorted by their order instead.

robbrockbank: Agreed - this was already being tracked by #75

calicoctl/resourcemgr/tier.go
robbrockbank: I'm happy to include - just that up until now the ps format just included the metadata and the wide added in some spec fields.  It seemed like a good distinction, but it's all pretty arbitrary

caseydavenport: Yeah, I'm not too worried about spec vs metadata fields corresponding to wide or not.  I think we just need to make the default as useful as possible without cluttering the screen.

Wide can clutter the screen.

caseydavenport:
It's hard to review this effectively by _just_ reading the code - I've tried playing around with it a bit and it looks like a good start - I think we'll have to tweak and improve based on UX feedback as we use it a bit more, but I think this should go in so we can raise issues against the UX. 

@robbrockbank thoughts?

caseydavenport:
@robbrockbank How is the go-template stuff supposed to work?

I've tried this:

```
macncheese:libcalico-go casey$ ./bin/calicoctl get bgppeer -o go-template-file=temp.temp
Error executing command: template: get:2:12: executing "get" at <$value.Spec>: Spec is not a field of struct type unversioned.Resource
```

But since the returned value is just a Resource, I can't seem to access Spec fields.

robbrockbank:
I'm all for merging and raising feedback issues - I think people need to play with the interface - but of course I'll need to document things a little more before they can.

caseydavenport:
LGTM

PR #90, opened 2016-08-22
Allow plural and case-insensitive resource names on CLI

### LINKED ISSUE #43: Support singular and plural names in calicoctl and caps insensitive
e.g.

calicoctl get pools
calicoctl get pool
calicoctl get hostendpoints
calicoctl get hostEndpoints

robbrockbank:
PR to allow case-insensitive and plural versions of resource names in the CLI.  This does not change the format required in the YAML or JSON files - these are still case sensitive and only accept the singular versions.

Fixes #43


caseydavenport:
LGTM.

PR #93, opened 2016-08-22
Bundle ICMP code and type

gunjan5:
for issue [66](https://github.com/tigera/libcalico-go/issues/66)

### DIFF
--- a/lib/api/rule.go
+++ b/lib/api/rule.go
@@ -26,16 +26,20 @@ import (
 type Rule struct {
 	Action string `json:"action" validate:"action"`
 
-	Protocol *Protocol `json:"protocol,omitempty" validate:"omitempty"`
-	ICMPType *int      `json:"icmpType,omitempty" validate:"omitempty,gte=0,lte=255"`
-	ICMPCode *int      `json:"icmpCode,omitempty" validate:"omitempty,gte=0,lte=255"`
+	Protocol *Protocol   `json:"protocol,omitempty" validate:"omitempty"`
+	ICMP     *ICMPFields `json:"icmp,omitempty" validate:"omitempty"`
 
-	NotProtocol *Protocol `json:"!protocol,omitempty" validate:"omitempty"`
-	NotICMPType *int      `json:"!icmpType,omitempty" validate:"omitempty,gte=0,lte=255"`
-	NotICMPCode *int      `json:"!icmpCode,omitempty" validate:"omitempty,gte=0,lte=255"`
+	NotProtocol *Protocol   `json:"!protocol,omitempty" validate:"omitempty"`
+	NotICMP     *ICMPFields `json:"!icmp,omitempty" validate:"omitempty"`
 
 	Source      EntityRule `json:"source,omitempty" validate:"omitempty"`
 	Destination EntityRule `json:"destination,omitempty" validate:"omitempty"`
+}
+
+// ICMPFields defines structure for ICMP and NotICMP sub-struct for ICMP code and type
+type ICMPFields struct {
+	Type *int `json:"type,omitempty" validate:"omitempty,gte=0,lte=255"`
+	Code *int `json:"code,omitempty" validate:"omitempty,gte=0,lte=255"`
 }
 
 type EntityRule struct {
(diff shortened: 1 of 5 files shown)


lib/api/rule.go
```diff
@@ -27,17 +27,21 @@ type Rule struct {
...
-	NotICMPType *int      `json:"!icmpType,omitempty" validate:"omitempty,gte=0,lte=255"`
-	NotICMPCode *int      `json:"!icmpCode,omitempty" validate:"omitempty,gte=0,lte=255"`
+	NotICMP     *CodeType `json:"!icmp,omitempty" validate:"omitempty"`
 
 	Source      EntityRule `json:"source,omitempty" validate:"omitempty"`
 	Destination EntityRule `json:"destination,omitempty" validate:"omitempty"`
 }
 
+// CodeType defines structure for ICMP and NotICMP sub-struct for ICMP code and type
+type CodeType struct {
```
caseydavenport: I think a better name for this might be something like `ICMPFields` - I think it captures the intention of the struct a bit better.

@gunjan5 thoughts?

lib/client/rule.go
```diff
@@ -39,14 +39,25 @@ func ruleActionBackendToAPI(action string) string {
 
 // ruleAPIToBackend converts an API Rule structure to a Backend Rule structure.
 func ruleAPIToBackend(ar api.Rule) model.Rule {
+	var icmpCode, icmpType, notIcmpCode, notIcmpType *int
```
caseydavenport: I think the naming conventions say `notIcmpCode` should be `notICMPCode`, and the same for type.

caseydavenport:
Are there no test changes associated with this?

lib/api/rule.go
gunjan5: Yes, agree. Couldn't think of anything better, but `ICMPFields` sounds good! I'll change it

lib/client/rule.go
gunjan5: Yup, I knew someone was gonna catch this ;)

lib/api/rule.go
```diff
@@ -58,8 +62,12 @@ func init() {
...
 func validateRule(v *validator.Validate, structLevel *validator.StructLevel) {
 	rule := structLevel.CurrentStruct.Interface().(Rule)
-	if rule.ICMPCode != nil && rule.ICMPType == nil {
-		structLevel.ReportError(reflect.ValueOf(rule.ICMPCode), "ICMPCode", "icmpCode", "icmpCodeWithoutType")
+	if rule.ICMP != nil && rule.ICMP.Code != nil && rule.ICMP.Type == nil {
+		structLevel.ReportError(reflect.ValueOf(rule.ICMP.Code), "Code", "code", "icmpCodeWithoutType")
+	}
+
+	if rule.NotICMP != nil && rule.NotICMP.Code != nil && rule.NotICMP.Type == nil {
+		structLevel.ReportError(reflect.ValueOf(rule.NotICMP.Code), "Code", "code", "notIcmpCodeWithoutType")
```
caseydavenport: I think `notIcmpCodeWithoutType` should just be `icmpCodeWithoutType`.

We can distinguish the two cases by doing `Code` -> `NotICMP.Code`?

gunjan5:
I think there is one test for rules, but it's between `model` and `backend`, and this PR doesn't change `model` or `backend` code.
https://github.com/tigera/libcalico-go/blob/master/lib/backend/model/rule_test.go

I couldn't find any existing tests for `api` and `client` rules

caseydavenport:
You probably need to update the following as well (they all mention icmpCode!)

```
test/all-no-profile.yaml
test/all.yaml
test/policy.yaml
```

lib/api/rule.go
caseydavenport: Looks to me like the fieldName / customName are used in the error reporting (but calicoctl isn't using them in the message it outputs).  I'd still update the tag as suggested above, and we can look at the other fields when we address #94.

lib/api/rule.go
```diff
@@ -26,18 +26,22 @@ import (
...
+	NotProtocol *Protocol   `json:"!protocol,omitempty" validate:"omitempty"`
+	NotICMP     *ICMPFields `json:"!icmp,omitempty" validate:"omitempty"`
 
 	Source      EntityRule `json:"source,omitempty" validate:"omitempty"`
 	Destination EntityRule `json:"destination,omitempty" validate:"omitempty"`
 }
 
+// ICMPFields defines structure for ICMP and NotICMP sub-struct for ICMP code and type
+type ICMPFields struct {
+	Type *int `json:"type,omitempty" validate:"omitempty,gte=0,lte=255"`
```
robbrockbank: The Type and Code should not be pointers anymore.

gunjan5: But then wouldn't they get 0 as their Zero Value by default, if not specified (which could be a valid Code and Type to the validator)?  
validate:"omitempty,**gte=0**,lte=255"

caseydavenport: @robbrockbank You may still want to specify a type without a code (though not the other way around).

Since `0` is a valid ICMP code, you want it to default to `nil` so that we can distinguish "I don't care about ICMP code" from "ICMP code 0".

Type should always be present so could be made not a pointer, but changing from pointer -> int means that a default value of `0` will be picked, which is kinda unintuitive.

caseydavenport:
LGTM - @gunjan5 could you squash the commits before merging?

gunjan5:
Fixes #93

PR #98, opened 2016-08-23
Wrong error message for missing resource type

gunjan5:
This fixes issue [97](https://github.com/tigera/libcalico-go/issues/97)

**Before**: 
```bash
./bin/calicoctl get foo
Error getting resources: Resource type 'foo' is not unsupported
```

**After**:
```bash
./bin/calicoctl get foo
Error getting resources: Resource type 'foo' is not supported
```


caseydavenport:
LGTM

PR #102, opened 2016-08-24
Config options rebased

robbrockbank:
This PR allows user to configure various system config options:
-  IPIP
-  default node AS number
-  node-to-node mesh
-  BGP and Felix logging levels (node and global)

It does not yet allow the user to specify arbitrary config options.

This was knocked together, but I think the implementation needs some tweaks (although since this works I'd probably keep this as is for alpha).  My thoughts as follows:
-  I think the programmatic API should always expect the scope and component to be specified.  Therefore, it's just calicoctl that defaults them when the name provides uniqueness.  Not 100% convinced about this though.
-  Where we do the mapping to the backend values needs moving - the values could quite possibly change based on version of the backend data.  However, I don't think it is worth designing this at the moment because that is potentially a big job - for example we might want to move some of the config into a big config structure rather than having single key/value pairs in the datastore.


matthewdupre:
Does this need to be picked back up?

fasaxc:
TODO:

PR #103, opened 2016-08-25
Add Syncer API and etcdv2 Syncer.

fasaxc:


robbrockbank:
This is orthogonal to the main releases that we are doing, so in the interest of preventing things becoming stale let's get this in.  We need to have a full review of the sync code in the future.

PR #104, opened 2016-08-25
Backend model updates

fasaxc:
@robbrockbank this makes some minor fixes, adds TTL support and the endpoint status reporting keys.  Those are used in golang felix.


robbrockbank:
This was merged in #109

PR #105, opened 2016-08-25
Try to improve backend API

fasaxc:
To avoid conflicts, this builds on the PR to add the Syncer.  That should be merged first, or only the newer two commits reviewed.

This PR:

- Improves the backend docs.
- Replace DefaultPath/DefaultDeletePath with conversion functions.

The latter change 

- hides the fact that we use methods to implement the default path/delete path function behind a function-based interface
- prevents our Keys from "leaking" the path information out 
- gives a central choke point to impose new path guarantees and document them (such as adding a '/' onto the keys for etcdv3 compatibility)
- gives a place to document the paths more thoroughly, and it paves the way for adding more types of conversion function later.

One last wrinkle is that the ListInterface still declares a public method to parse the key, but only if it's of the correct type.  I propose that we just drop that method and use the generic `KeyFromDefaultPath` function, which is required for the `Syncer` and duplicates the functionality of the individual methods.  We'd still need to be able to filter the resulting keys but we could do that by, say, exposing the key type on the `ListInterface` or adding an `IsMatchingKey` method that checks the type of a given key.

### DIFF
--- a/lib/backend/model/keys.go
+++ b/lib/backend/model/keys.go
@@ -43,6 +43,7 @@ type ListInterface interface {
 	// to the directory containing all the keys to be listed.
 	defaultPathRoot() string
 
+	// BUG(smc) I think we should remove this and use the package KeyFromDefaultPath function.
 	// KeyFromDefaultPath parses the default path representation of the
 	// Key type for this list.  It returns nil if passed a different kind
 	// of path.
--- a/lib/backend/model/keys.go
+++ b/lib/backend/model/keys.go
@@ -16,7 +16,6 @@ package model
 
 import (
 	"encoding/json"
-	"errors"
 	"reflect"
 	"strings"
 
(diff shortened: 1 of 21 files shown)


lib/backend/model/keys.go
```diff
@@ -33,57 +32,130 @@ var rawBoolType = reflect.TypeOf(rawBool(true))
...
+	// defaultPathRoot() returns a default stringified root path, i.e. path
 	// to the directory containing all the keys to be listed.
-	DefaultPathRoot() string
-	ParseDefaultKey(key string) Key
+	defaultPathRoot() string
+
+	// KeyFromDefaultPath parses the default path representation of the
+	// Key type for this list.  It returns nil if passed a different kind
+	// of path.
+	KeyFromDefaultPath(key string) Key
```
lxpollitt: Presumably the key parameter should be called path?  And perhaps this method should be private, similar to the other path related methods  wrapped by public helper functions?

lib/backend/model/profile.go
```diff
@@ -113,7 +113,7 @@ func (options ProfileListOptions) DefaultPathRoot() string {
 	return k
 }
 
-func (options ProfileListOptions) ParseDefaultKey(ekey string) Key {
+func (options ProfileListOptions) KeyFromDefaultPath(ekey string) Key {
```
lxpollitt: Should the ekey parameter be named path?

robbrockbank:
This looks like a good start to me.  I'll merge.  We can sort out the ListInterface later.

PR #107, opened 2016-08-25
Various docs tweaks and fix of config load

robbrockbank:


calicoctl/commands/delete.go
```diff
@@ -25,6 +25,9 @@ import (
 func Delete(args []string) error {
 	doc := EtcdIntro + `Delete a resource identified by file, stdin or resource type and name.
 
+Valid resource kinds are bgpPeer, hostEndpoint, policy, pool, profile and tier.  The <KIND>
```
djosborne: should these be camel case in the UI?

djosborne: Looks like they do all lowercase mushed together. I guess making them camel-case but case-insensitive means its just a tad easier to read, so I'm fine with it.

robbrockbank: The reason I added them as camel-case here was because the YAML/JSON is case sensitive - so this makes it less confusing if you are using those.

djosborne:
LGTM

PR #115, opened 2016-08-30
Calicoctl node status cmd

gunjan5:


gunjan5:
The failing test in CircleCI is because I'm using Golang version 1.7 (where `"context"` is part of the core library) vs pre-1.7 versions had it at `"golang.org/x/net/context"` 
```bash
calicoctl/commands/status.go:18:2: cannot find package "context" in any of:
	/home/ubuntu/.go_project/src/github.com/tigera/libcalico-go/vendor/context (vendor tree)
	/usr/local/go/src/context (from $GOROOT)
	/home/ubuntu/.go_project/src/context (from $GOPATH)
make: *** [bin/calicoctl] Error 1
```

Should we be updating our testing environment in CircleCI to Golang version 1.7 or should I change my imports to support 1.6? @robbrockbank @caseydavenport

calicoctl/commands/node.go
```diff
@@ -0,0 +1,55 @@
...
+
+	"github.com/docopt/docopt-go"
+)
+
+// Node function is a switch to node related sub-commands
+func Node(args []string) error {
+	var err error
+	doc := `Usage: calicoctl node <command> [<args>...]
+
+    ststus          Shows the status of the node.
```
caseydavenport: typo: ststus

calicoctl/commands/status.go
```diff
@@ -0,0 +1,206 @@
+// Copyright (c) 2016 Tigera, Inc. All rights reserved.
```
caseydavenport: I _think_ we probably want this file to live in a subdirectory specific for the `node` command.

e.g `calicoctl/commands/node/status.go`

Keeps all the node options separated so they're easier to find, and makes it easier for us to add, for example, a `calicoctl status ...` command down the road if the need arises.

caseydavenport:
@gunjan5 - I say let's update to use Go 1.7.

I think this is what we want to do across all our Go repos anyway.

calicoctl/commands/node.go
```diff
@@ -0,0 +1,55 @@
...
+	doc := `Usage: calicoctl node <command> [<args>...]
+
+    ststus          Shows the status of the node. 
+
+Description:
+  Node specific commands for calicoctl
+  
+  See 'calicoctl node --help' to read about a specific subcommand.`
+
+	arguments, _ := docopt.Parse(doc, args, true, "calicoctl", false, false)
```
caseydavenport: any reason we're throwing away the error here?

calicoctl/commands/node.go
```diff
@@ -0,0 +1,55 @@
...
+
+	if arguments["<command>"] != nil {
```
caseydavenport: I might be missing something here (I don't know docopt super well!) but isn't `arguments["<command>"] == nil` essentially an error case?  When would that be true?

Might the logic look better like this?

```
if arguments[command] == nil {
  // Return an error.
}

// Otherwise, we parsed successfully
args := ...

// Check the command
switch command {
 ...
}
```

calicoctl/commands/status.go
```diff
@@ -0,0 +1,206 @@
...
+	"strings"
+
+	"github.com/docker/engine-api/client"
+	"github.com/docker/engine-api/types"
+	"github.com/docopt/docopt-go"
+
+	"github.com/olekukonko/tablewriter"
+)
+
+var ipv4 = true
```
caseydavenport: As far as I can tell, this is never assigned anywhere, and is only used in an if statement below to determine the separator for an IP address.  I don't think we need this guy.

calicoctl/commands/status.go
```diff
@@ -0,0 +1,206 @@
...
+
+	drawTable(resp)
+
+}
+
+func drawTable(in string) {
+	data := [][]string{}
+
+	ipSep := "."
+	if !ipv4 {
```
caseydavenport: `ipv4` is _always_ true.

calicoctl/commands/status.go
```diff
@@ -0,0 +1,206 @@
...
+
+var ipv4 = true
+
+// Status prings status of the node and returns error (if any)
+func Status(args []string) error {
+	doc := `Usage:
+calicoctl node status
+
+Description:
+  Display the status of the calico node`
```
caseydavenport: `calico` should be capitalized.

calicoctl/commands/status.go
```diff
@@ -0,0 +1,206 @@
...
+
+	_, _ = docopt.Parse(doc, args, true, "calicoctl", false, false)
```
caseydavenport: What does this line do? We're not using the output, and I think that since we're passing `false` as the last argument, `Parse` won't exit on parsing failures.

calicoctl/commands/status.go
```diff
@@ -0,0 +1,206 @@
...
+
+	ctx := context.Background()
+
+	defaultHeaders := map[string]string{"User-Agent": "engine-api-cli-1.0"}
+	cli, err := client.NewClient("unix:///var/run/docker.sock", "v1.22", nil, defaultHeaders)
```
caseydavenport: Why not try this on for size: https://godoc.org/github.com/docker/engine-api/client#NewEnvClient

I think it makes this code simpler, and will also allow people to configure our Docker interactions through environment variables.  The defaults look sensible enough to me (and if they're not, we can override them via env vars)

caseydavenport: Hm, it doesn't let you pass the defaultHeaders field.  Care to enlighten me? What does `defaultHeaders := map[string]string{"User-Agent": "engine-api-cli-1.0"}` do?

calicoctl/commands/status.go
```diff
@@ -0,0 +1,206 @@
...
+	if err != nil {
+		panic(err)
```
caseydavenport: Not sure a `panic` is the right thing to do here - we probably want to return an error to the user rather than crash.

calicoctl/commands/status.go
```diff
@@ -0,0 +1,206 @@
...
+func Status(args []string) error {
+	doc := `Usage:
+calicoctl node status
+
+Description:
+  Display the status of the calico node`
+
+	_, _ = docopt.Parse(doc, args, true, "calicoctl", false, false)
+
+	ctx := context.Background()
```
caseydavenport: `ctx` is defined way up here, but only used on line 56.  You should try to keep the variable declarations as close to where they are used as possible - especially if they're only used once!

calicoctl/commands/status.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+
+	options := types.ContainerListOptions{All: true}
+	containers, err := cli.ContainerList(ctx, options)
+	if err != nil {
+		panic(err)
+	}
+
+	for _, c := range containers {
+		if c.Names[0] == "/calico-node" && c.State == "running" {
```
caseydavenport: See this relevant issue: https://github.com/projectcalico/calico-containers/issues/1086

calicoctl/commands/status.go
```diff
@@ -0,0 +1,206 @@
...
+
+			fmt.Printf("calico-node container is running. Status: %s\n", c.Status)
```
caseydavenport: Any reason we don't want to make this something like this instead?
```
fmt.Printf("calico-node container is %s. Status: %s\n", c.State, c.Status)
```

Not critical for getting this merged, but just a thought.  Provides more information to the user this way.

calicoctl/commands/status.go
```diff
@@ -0,0 +1,206 @@
...
+
+			// if len(parts) == 2 {
+			// 	felixVersion := parts[1][1 : len(parts[1])-1]
+
+			// 	fmt.Printf("Running felix version %v\n", felixVersion)
+			// }
+
+			//
+
+			//if calico-node container is running, then connect to the bird socket and get the data
```
caseydavenport: Comments should have proper punctuation and capitalization, and should have a space after the `//`.

calicoctl/commands/status.go
```diff
@@ -0,0 +1,206 @@
...
+			c, err := net.Dial("unix", "/var/run/calico/bird.ctl")
+			if err != nil {
+				panic(err)
```
caseydavenport: Again, not sure about the panic here.  What is the UX when this happens? Don't they get a big stack trace printed to screen?

calicoctl/commands/status.go
```diff
@@ -0,0 +1,206 @@
...
+			if err != nil {
+				log.Fatal("Error writing to BIRD socket:", err)
+			}
+			reader(c)
+
+			return nil
+		}
+	}
+
+	// return and print message if calio-node is not running
```
caseydavenport: Capitalization, and typo: `calio-node`

calicoctl/commands/status.go
```diff
@@ -0,0 +1,206 @@
...
+	fmt.Println("calico-node container not running")
+	return nil
+
+}
+
+func reader(r io.Reader) {
```
caseydavenport: I think this could use a better name.  It has a pretty specific purpose - to read from the bird socket and print out the data in a table format.

perhaps it should be `printBIRDResponse` or similar?

calicoctl/commands/status.go
```diff
@@ -0,0 +1,206 @@
...
+	buf := make([]byte, 1024)
```
caseydavenport: How do we know that `1024

calicoctl/commands/status.go
```diff
@@ -0,0 +1,206 @@
...
+		return
+	}
+
+	resp := string(buf[:n])
+
+	drawTable(resp)
+
+}
+
+func drawTable(in string) {
```
caseydavenport: `in` has a specific format that's associated with it (it's not just going to be an arbitrary string) - I'd recommend naming it something like `birdOut` for readability.

calicoctl/commands/status.go
```diff
@@ -0,0 +1,206 @@
...
+				col = append(col, "node specific")
+				col = append(col, fields...)
+			} else if strings.HasPrefix(s, "Global_") {
+				s = s[7:]
+				s = strings.Replace(s, "_", ipSep, -1)
+				col = append(col, s)
+				col = append(col, "global")
+				col = append(col, fields...)
+			} else {
+				// did not match any of the predefined options for BIRD
```
caseydavenport: Capitalization / punctuation in comments ;)

calicoctl/commands/status.go
```diff
@@ -0,0 +1,206 @@
...
+			col := []string{}
+			// split columns
+			fields := strings.Fields(line)[3:6]
+			if strings.HasPrefix(s, "Mesh_") {
+				s = s[5:]
+				s = strings.Replace(s, "_", ipSep, -1)
+				col = append(col, s)
+				col = append(col, "node-to-node mesh")
+				col = append(col, fields...)
+
```
caseydavenport: extra newline here

calicoctl/commands/status.go
```diff
@@ -0,0 +1,206 @@
...
+				col = append(col, s)
+				col = append(col, "global")
+				col = append(col, fields...)
+			} else {
+				// did not match any of the predefined options for BIRD
+				fmt.Println(errors.New("Error: Did not match any of the predefined options for BIRD"))
+				break
+			}
+			data = append(data, col)
+
```
caseydavenport: extra newline here

calicoctl/commands/status.go
```diff
@@ -0,0 +1,206 @@
...
+	n, err := r.Read(buf[:])
+	if err != nil {
+		return
+	}
+
+	resp := string(buf[:n])
+
+	drawTable(resp)
+
```
caseydavenport: extra newline

calicoctl/commands/status.go
```diff
@@ -0,0 +1,206 @@
...
+	table.SetHeader([]string{"Peer address", "Peer type", "State", "Since", "Info"})
+
+	r, _ := regexp.Compile(`\w+_\d+_\d+_\d+_\d+`)
+	for _, line := range strings.Split(in, "\n") {
+
+		s := r.FindString(line)
+
+		if s != "" {
+			col := []string{}
+			// split columns
```
caseydavenport: Not a super useful comment :)

calicoctl/commands/status.go
caseydavenport: I think you're going to want to encapsulate all of this functionality - it needs to be done twice (once for ipv4, once for ipv6).

caseydavenport: Ideally, I think we also separate this logic from whether or not `calico-node` is running.  That check is inherently brittle, whereas the BIRD socket can exist even if the container isn't running (e.g BIRD is running on its own).

The output can print an error if one of the checks fails, but it doesn't need to gate one on the other.

That said, I think that's more of a GA feature, so happy to let that go for this PR.

calicoctl/commands/status.go
```diff
@@ -0,0 +1,206 @@
...
+
+	ipSep := "."
+	if !ipv4 {
+		ipSep = ":"
+	}
+
+	table := tablewriter.NewWriter(os.Stdout)
+	table.SetHeader([]string{"Peer address", "Peer type", "State", "Since", "Info"})
+
+	r, _ := regexp.Compile(`\w+_\d+_\d+_\d+_\d+`)
```
caseydavenport: Worth adding a comment describing what each of the things this regex pulls out are.

Also, I think it is generally good to stick static regexes like this at the top of the file.  They only need to be compiled once since they never change.  Give it a good name (like `bgpPeerRegex`) and include a detailed comment as to what the expected input / output is.

calicoctl/commands/node.go
gunjan5: sorry, copied it from another file and forgot to add the error check. 
should probably do it for other files too.
I'll fix this one for now

calicoctl/commands/node.go
gunjan5: I'm not a big fan of this library! :) 
you don't have to check `arguments["<command>"] == nil` because when you describe the command with `Usage: calicoctl node <command> [<args>...]` this library makes sure that is fulfilled, if not it returns an error and prints proper usage message itself.

I hope that's what you meant, lmk if I'm missing the point.

calicoctl/commands/status.go
gunjan5: I added this for future when we get information from etcd. Python version switches `ipSep` based on the IP version info it gets from etcd.

It's not super useful right now, but I wanted to leave it for future.

Let me know if you think we should get rid of this.

calicoctl/commands/status.go
gunjan5: added the error check, but not left the last argument for convention (that's what it's set to for other commands). From the docs, it seems like setting it to `true` might be better, I'll ask @robbrockbank if he has a preference

calicoctl/commands/status.go
gunjan5: yea so if you use `NewEnvClient` and if Docker API version is not specified, it will use the latest by default, which can break things. setting `defaultHeaders` is optional, and if left empty it won't enforce any version, which is also bad, so setting it to a stable version which is tested and works.

lmk, if you still prefer `NewEnvClient`

calicoctl/commands/status.go
gunjan5: `ctx` is supposed to be used at 2 more places in code, but that's the code which isn't working as expected and hence commented out for now.

calicoctl/commands/status.go
gunjan5: ok fixed this to check if the name contains "calico-node" and tested it with a nasty name like `k8s_calico-node.215a1ad0_calico-n7l79_default_6e29a86d-5868-11e6-9250-080027397496_d3d6d6b6`

calicoctl/commands/status.go
gunjan5: yeah, I think it would make sense to refactor the code/encapsulate when we do the etcd part. i.e. in GA

calicoctl/commands/status.go
gunjan5: 1024 is the initial length of the slice. It will resize automatically if needed (won't overflow).
I picked 1024 because Go runtime doubles the slice size until 1024 then it's linear afterwards

gunjan5:
@caseydavenport Re: upgrade to 1.7, I just made the import compatible for pre-1.7 for now. I think the upgrade to 1.7 should be in a different PR since it will require many other files.

calicoctl/commands/status.go
caseydavenport: OK cool - good to know :)

calicoctl/commands/node/status.go
```diff
@@ -28,19 +27,26 @@ import (
...
 	"github.com/docker/engine-api/types"
 	"github.com/docopt/docopt-go"
+	"golang.org/x/net/context"
 
 	"github.com/olekukonko/tablewriter"
 )
 
-var ipv4 = true
+var (
+	ipv4 = true
```
caseydavenport: I think this can just be removed - it's not used anywhere!

calicoctl/commands/status.go
caseydavenport: Yeah, I think we should get rid of it.  It's easy to add back later, and confusing to keep things around that aren't being used.

I think when we add this in the future, we don't want it to be a file-scoped variable anyway - probably just an argument passed to the BGP print function.

calicoctl/commands/status.go
caseydavenport: OK - that is confusing.  Could you stick a comment in there to explain that this is needed so that the help message will be printed, and that actual syntax checking is done at the level above?

calicoctl/commands/node/status.go
```diff
@@ -109,12 +114,10 @@ Description:
...
 			// }
 
-			//
-
-			//if calico-node container is running, then connect to the bird socket and get the data
+			// Connect to the bird socket and get the data if calico-node container is running
 			c, err := net.Dial("unix", "/var/run/calico/bird.ctl")
 			if err != nil {
-				panic(err)
+				log.Fatalf("Error connecting to the BIRD socker: %v", err)
```
caseydavenport: typo: socker

calicoctl/commands/status.go
caseydavenport: It doesn't look like we do this in the Python code.  Looks like we essentially do the equivalent of `NewEnvClient` (though we hardcode the version to 1.16)

https://github.com/projectcalico/calico-containers/blob/master/calicoctl/calico_ctl/connectors.py#L34

https://github.com/projectcalico/calico-containers/blob/master/calicoctl/calico_ctl/utils.py#L26

Mostly worried about people who are using the existing DOCKER_HOST var, which won't work when they switch to this calicoctl.

calicoctl/commands/node.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+	if arguments["<command>"] != nil {
+		command := arguments["<command>"].(string)
+		args := append([]string{command}, arguments["<args>"].([]string)...)
+
+		switch command {
+		case "status":
+			err = node.Status(args)
+		default:
```
caseydavenport: I think that we should print another line before the doc. Something like:

`(Invalid option '%s', command)`

This let's them know _why_ they command they typed in isn't working.

calicoctl/commands/node/status.go
```diff
@@ -0,0 +1,203 @@
...
+	options := types.ContainerListOptions{All: true}
+	containers, err := cli.ContainerList(ctx, options)
+	if err != nil {
+		panic(err)
+	}
+
+	for _, c := range containers {
+		if strings.Contains(c.Names[0], "calico-node") && c.State == "running" {
+
+			fmt.Printf("calico-node container is %s. Status: %s\n", c.State, c.Status)
```
caseydavenport: Hmm.. yeah I guess this was a bit silly (since c.State will _always_ be running).

I had originally though we could show non-running state as well, but there might be a bunch of exited calico-node containers that we don't care about (so long as one is running), so the way you had it is probably best... sorry!

calicoctl/commands/node/status.go
```diff
@@ -0,0 +1,203 @@
...
+			printBIRDResponse(c)
+
+			return nil
+		}
+	}
+
+	// Return and print message if calico-node is not running
+	fmt.Println("calico-node container not running")
+	return nil
+
```
caseydavenport: extra newline here.

calicoctl/commands/node/status.go
```diff
@@ -0,0 +1,203 @@
...
+	if err != nil {
+		panic(err)
+	}
+
+	for _, c := range containers {
+		if strings.Contains(c.Names[0], "calico-node") && c.State == "running" {
+
+			fmt.Printf("calico-node container is %s. Status: %s\n", c.State, c.Status)
+
+			// config := types.ExecConfig{
```
caseydavenport: As discussed, pull out all of the commented code from this PR and we'll address in another one.

calicoctl/commands/node/status.go
```diff
@@ -0,0 +1,203 @@
...
+func drawTable(birdOut string) {
+	data := [][]string{}
+
+	ipSep := "."
+	if !ipv4 {
+		ipSep = ":"
+	}
+
+	table := tablewriter.NewWriter(os.Stdout)
+	table.SetHeader([]string{"Peer address", "Peer type", "State", "Since", "Info"})
```
caseydavenport: yeah, we're going to need to re-structure this code a bit to make IPv6 work.

It shouldn't be too crazy, but let's chat about it when you have a chance.  Given the alpha release closing in, I say we do that in a separate PR, but let's do it _soon_.

glide.lock
```diff
@@ -40,6 +75,12 @@ imports:
   version: 6acef71eb69611914f7a30939ea9f6e194c78172
   subpackages:
   - context
+  - proxy
+- name: golang.org/x/sys
```
caseydavenport: I don't see this in use else where in this PR.

glide.lock
```diff
@@ -32,6 +63,10 @@ imports:
   version: 1e291572dcb9ba673cdcdb15e0f422702415ebaf
 - name: github.com/satori/go.uuid
   version: 0aa62d5ddceb50dbcb909d790b5345affd3669b6
+- name: github.com/Sirupsen/logrus
```
caseydavenport: I don't see this in use else where in this PR. Was it just not added some other time when it should have been?

glide.lock
```diff
@@ -22,6 +22,37 @@ imports:
...
+  - types/mount
+  - types/blkiodev
+  - types/strslice
+- name: github.com/docker/go-connections
+  version: 0bad1a3951398f88ef4dd75fdb42af374430be89
+  subpackages:
+  - sockets
+  - tlsconfig
+  - nat
+- name: github.com/docker/go-units
```
caseydavenport: I don't see this in use else where in this PR. Was it just not added some other time when it should have been?

glide.lock
```diff
@@ -22,6 +22,37 @@ imports:
...
+  - types/network
+  - types/reference
+  - types/registry
+  - types/swarm
+  - types/time
+  - types/versions
+  - types/mount
+  - types/blkiodev
+  - types/strslice
+- name: github.com/docker/go-connections
```
caseydavenport: I don't see this in use else where in this PR. Was it just not added some other time when it should have been?

glide.lock
```diff
@@ -22,6 +22,37 @@ imports:
   version: fa29b1d70f0beaddd4c7021607cc3c3be8ce94b8
   subpackages:
   - capnslog
+- name: github.com/docker/distribution
```
caseydavenport: I don't see this in use else where in this PR. Was it just not added some other time when it should have been?

glide.lock
gunjan5: `logrus` is used by docker engine-api, second layer of dependancy

calicoctl/commands/node/status.go
```diff
@@ -0,0 +1,203 @@
...
+	"github.com/docker/engine-api/client"
+	"github.com/docker/engine-api/types"
+	"github.com/docopt/docopt-go"
+	"golang.org/x/net/context"
+
+	"github.com/olekukonko/tablewriter"
+)
+
+var (
+	ipv4 = true
```
caseydavenport: I stuck a comment on the comment, but saying it again here - definitely remove things that aren't in use.

It's misleading for readers of the code to see something like this that isn't actually used :)

caseydavenport:
@gunjan5 OK - second pass done.  I think we're getting close with this guy.

As mentioned in the comments, there are some things we'll need to follow up on (like IPv6) shortly after this goes in.

calicoctl/commands/node/status.go
gunjan5: i was confused for a bit about that, but I removed it :)

calicoctl/commands/node.go
gunjan5: ok so I just tested this. if you remove `arguments["<command>"] != nil` check, it panics when you pass `calicoctl node status --help`, so I'm thinking we should just leave it there. (it doesn't panic with `-h` `--help` when `arguments["<command>"] != nil` is checked)

caseydavenport: OK, could you stick a comment right above that if statement that says "if --help is passed, then arguments[<command>] will be nil"?

calicoctl/commands/node/status.go
```diff
@@ -48,12 +44,27 @@ calicoctl node status
...
-	defaultHeaders := map[string]string{"User-Agent": "engine-api-cli-1.0"}
-	cli, err := client.NewClient("unix:///var/run/docker.sock", "v1.22", nil, defaultHeaders)
+	if os.Getenv("DOCKER_HOST") == "" {
+		err := os.Setenv("DOCKER_HOST", "unix:///var/run/docker.sock")
+		if err != nil {
+			log.Fatalf("Error setting DOCKER_HOST: %v", err)
+		}
+	}
+	err := os.Setenv("DOCKER_API_VERSION", "1.16")
+	if err != nil {
```
caseydavenport: You've got nested `err != nil` statements here :)

calicoctl/commands/node/status.go
```diff
@@ -48,12 +44,27 @@ calicoctl node status
...
+		if err != nil {
+			log.Fatalf("Error setting DOCKER_API_VERSION: %v", err)
+		}
+	}
+
+	cli, err := client.NewEnvClient()
```
caseydavenport: `cli` is a confusing name, I think `c` is sufficient (or `dockerClient`)

calicoctl/commands/node/status.go
```diff
@@ -48,12 +44,27 @@ calicoctl node status
...
 
-	defaultHeaders := map[string]string{"User-Agent": "engine-api-cli-1.0"}
-	cli, err := client.NewClient("unix:///var/run/docker.sock", "v1.22", nil, defaultHeaders)
+	if os.Getenv("DOCKER_HOST") == "" {
+		err := os.Setenv("DOCKER_HOST", "unix:///var/run/docker.sock")
+		if err != nil {
+			log.Fatalf("Error setting DOCKER_HOST: %v", err)
+		}
+	}
+	err := os.Setenv("DOCKER_API_VERSION", "1.16")
```
caseydavenport: I don't think this is quite right - This will override any variable that the user passes in!

calicoctl/commands/node/status.go
```diff
@@ -48,12 +44,27 @@ calicoctl node status
...
+	// Note: This call is ignoring the error because error check happens at the level above
+	// i.e at `node.go` before it calls `node.Status`. This call is just so help message gets
+	// printed for this option
 	_, _ = docopt.Parse(doc, args, true, "calicoctl", false, false)
 
 	ctx := context.Background()
 
-	defaultHeaders := map[string]string{"User-Agent": "engine-api-cli-1.0"}
-	cli, err := client.NewClient("unix:///var/run/docker.sock", "v1.22", nil, defaultHeaders)
+	if os.Getenv("DOCKER_HOST") == "" {
```
caseydavenport: `NewEnvClient()` does this automatically, doesn't it?

gunjan5: yes, just looked at the code. Wasn't clear what it does from the godocs. Fixed.

calicoctl/commands/node/status.go
gunjan5: added a check to see if it's set, if it is then don't set it, else set it to 1.16 (like the python version

calicoctl/commands/node/status.go
```diff
@@ -0,0 +1,159 @@
...
+  Display the status of the Calico node`
+
+	// Note: This call is ignoring the error because error check happens at the level above
+	// i.e at `node.go` before it calls `node.Status`. This call is just so help message gets
+	// printed for this option
+	_, _ = docopt.Parse(doc, args, true, "calicoctl", false, false)
+
+	ctx := context.Background()
+
+	if os.Getenv("DOCKER_HOST") == "" {
```
caseydavenport: Not sure I understand this bit.  Why are we checking DOCKER_HOST to determine if we should set DOCKER_API_VERSION?

gunjan5: sorry, my b. It should be DOCKER_API_VERSION. Fixed

caseydavenport:
LGTM!

@gunjan5 - could you squash these up and then merge?

PR #129, opened 2016-09-09
calicoctl diags command

gunjan5:


calicoctl/commands/diags.go
```diff
@@ -0,0 +1,188 @@
...
+		return err
+	}
+
+	runDiags(arguments["--log-dir"].(string))
+
+	return nil
+}
+
+func runDiags(logDir string) {
+	cmds := []struct {
```
robbrockbank: Can we just declare all this as a constant at the top - it's easier to document and see what the list of commands are rather than buried in here.

calicoctl/commands/diags.go
```diff
@@ -0,0 +1,188 @@
...
+	"os/exec"
+	"path/filepath"
+	"regexp"
+	"strings"
+	"time"
+
+	"github.com/docopt/docopt-go"
+)
+
+type regPair struct {
```
robbrockbank: What is this struct? - needs some documentation.

calicoctl/commands/diags.go
```diff
@@ -0,0 +1,188 @@
...
+	if err != nil {
+		return err
+	}
+
+	runDiags(arguments["--log-dir"].(string))
+
+	return nil
+}
+
+func runDiags(logDir string) {
```
robbrockbank: Missing function comments on all the internal functions.

calicoctl/commands/diags.go
```diff
@@ -0,0 +1,188 @@
...
+type regPair struct {
+	reg     string
+	replace string
+}
+
+// Diags function collects diagnostic information and logs
+func Diags(args []string) error {
+	var err error
+	doc := `Usage: 
+            calicoctl diags [--log-dir=<LOG_DIR>]
```
robbrockbank: Had we not decided that this should be a `node` command - since it has to be run on a specific node.

calicoctl/commands/diags.go
```diff
@@ -0,0 +1,188 @@
...
+    such as transfer.sh using curl or similar.  For example:
+
+      curl --upload-file %s https://transfer.sh/%s\n`, tarFilePath, tarFilePath)
+
+}
+
+func writeDiags(info, cmd, dir string) {
+	fname := cmd
+
+	r := []regPair{
```
robbrockbank: I think lines 148-152 would be better as a separate function - this is basically converting the function name to a suitable filename.

In actual fact - do we even need this.  Since you've added a nice structure that links the user description and the command, you could easily add into that a filename - so you don't have to do this munging of the command.

calicoctl/commands/diags.go
```diff
@@ -0,0 +1,188 @@
...
+		}
+	} else {
+		fmt.Printf("No logs found in %s; skipping log copying", logDir)
+	}
+
+	// Get the current time and create a tar.gz file with the timestamp in the name
+	timeNow := time.Now()
+	y, mo, d := timeNow.Date()
+	h, m, s := timeNow.Clock()
+	tarFile := fmt.Sprintf("diags-%d%d%d_%d%d%d.tar.gz", d, mo, y, h, m, s)
```
robbrockbank: Can we just use time.Format here?  We don't have to stick to the exact same format as before - perhaps just use the RFC3339 format or similar?

calicoctl/commands/diags.go
```diff
@@ -0,0 +1,188 @@
...
+	if info != "" {
+		fmt.Println(info)
+	}
+
+	writeToFile(cmd, fname, dir)
+}
+
+func writeToFile(cmd, fname, dir string) {
+
+	parts := strings.Fields(cmd)
```
robbrockbank: Should we be declaring the command as a []string rather than splitting it after.  I guess not an issue provided none of the arguments have a space in them.

calicoctl/commands/diags.go
gunjan5: I thought that was only for `status`, but what you're saying makes sense.
Should I move this over to `calicoctl node diags`?

calicoctl/commands/diags.go
gunjan5: don't think Go lets you have a struct as a constant, but I can make it `var` on top

calicoctl/commands/diags.go
gunjan5: I did this because the python version prints filenames in that format, so we keep format consistent in case if someone is using a custom script for diags

calicoctl/commands/diags.go
gunjan5: on second thought, if it's a `var`, I'd rather keep it's scope tight and not make it accessible to other functions that don't need it, thoughts?

calicoctl/commands/diags.go
gunjan5: yeah, good point, but I'm using the command as a string to make the output filename before, and also having the whole command as a single string makes it more readable, and less hassle of doing variadic ...string

calicoctl/commands/diags.go
gunjan5: using RFC3399 also means our output tar file name format would be different. I did the same format DMY_HMS python version has for the filename, or should I change it to `20160913_131450`?

calicoctl/commands/diags.go
robbrockbank: I think this makes sense - the set of `calicoctl node` commands will include all of the commands that have to be run on a specific node.  Feel free to float past Tom and Casey too - but including the "node" is my preference.

calicoctl/commands/diags.go
robbrockbank: We aren't tied to the old format.  In fact the old format file names won't sort correctly if the day is first - so I think it's better to change it.  The use of RFC3339 is just a suggestion.  In either case it looks like time.Format is what we should be using no?

calicoctl/commands/diags.go
robbrockbank: We don't document these, and most likely we're the only ones that look in the drags bundles - I think we should be fine changing these.

calicoctl/commands/diags.go
robbrockbank: I'm not super fussed either way - but if you end up adding more information to these structures it gets a little cumbersome as an inline).  If you declared it outside the scope of the function then I'd suggest just making this a structure rather than inlining that too.

robbrockbank: var/const ... yeah I'm not sure what you're able to do with a const - but I can believe that you can't define structures there.

calicoctl/commands/diags.go
robbrockbank: It's fine sticking with this, but you should probably document somewhere as a warning to anyone adding more commands that if an argument contains a space this will need to be reworked.

calicoctl/commands/diags.go
gunjan5: I think `time.Format` is a much better solution.
I'm thinking of using this:
```go
fmt.Printf("diags-%s.tar.gz",time.Now().Format("20060102_150405"))
```
`diags-20091110_230000.tar.gz`

how does that look?

calicoctl/commands/diags.go
gunjan5: not sure if I understand it correctly. Even if you have a space in the argument, Fields will split it and pass it to `exec.Command` which should take care of it

robbrockbank: I mean if a single argument contains a space.

For example you could have an argument that is a string `"do this"` which would then get split to `"do` and `"this`

calicoctl/commands/node.go
```diff
@@ -25,30 +25,37 @@ import (
...
 
-		switch command {
-		case "status":
-			err = node.Status(nodeArgs)
-		default:
-			fmt.Printf("Invalid option: %s\n", command)
+		if arguments["status"].(bool) {
+			err = node.Status()
+		} else if arguments["diags"].(bool) {
+			err = node.Diags(logDir)
```
robbrockbank: It feels like we should be consistent here and pass in `args []string` as we do for Node.

calicoctl/commands/node/diags.go
```diff
@@ -0,0 +1,161 @@
...
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/docopt/docopt-go"
+)
+
+// diagCmds is a slice of struct to hold commands, their info and filename to run diagnostic on
+type diagCmds []struct {
```
robbrockbank: Wouldn't this be better as just a single command rather than a []?  That way you can simply pass the structure straight through to the `writeDiags` function rather than pulling out each field from the anonymous inner structure.

calicoctl/commands/node/diags.go
```diff
@@ -0,0 +1,161 @@
...
+	err = os.Chdir(tmpDir)
+	if err != nil {
+		log.Fatalf("Error changing directory to temp directory to dump logs: %v\n", err)
+	}
+
+	os.Mkdir("diagnostics", os.ModeDir)
+	diagsTmpDir := filepath.Join(tmpDir, "diagnostics")
+
+	for _, v := range cmds {
+		writeDiags(v.info, v.cmd, v.filename, diagsTmpDir)
```
robbrockbank: See comment above about just passing in the command structure.

calicoctl/commands/node.go
gunjan5: we are not processing args anymore in node.go because status doesn't need any args and diags needs a --log-dir. Adding <args> and passing it along would be confusing I think

calicoctl/commands/node/diags.go
```diff
@@ -0,0 +1,161 @@
...
+
+	os.Mkdir("diagnostics", os.ModeDir)
+	diagsTmpDir := filepath.Join(tmpDir, "diagnostics")
+
+	for _, v := range cmds {
+		writeDiags(v, diagsTmpDir)
+	}
+
+	fmt.Println("Dumping felix stats")
+	err = exec.Command("pkill", "-SIGUSR1", "felix").Run()
```
robbrockbank: any reason this can't be included in the cmd list?

calicoctl/commands/node/diags.go
```diff
@@ -0,0 +1,161 @@
...
+	if err != nil {
+		log.Printf("Error killing the felix process: %v\n", err)
+	}
+
+	fileInfo, err := os.Stat(logDir)
+
+	// Check if the logDir provided/default exists and is a directory
+	if fileInfo.IsDir() {
+		fmt.Println("Copying Calico logs")
+		cpCmd := exec.Command("cp", "-rf", logDir, diagsTmpDir)
```
robbrockbank: We shouldn't really use `cp` here.  The file operations should really be done using a golang library - it's generally bad form to assume a particular command will be available (of course cp always will be I'm sure).

calicoctl/commands/node/diags.go
```diff
@@ -0,0 +1,161 @@
...
+			log.Fatalf("Error copying log files: %v\n", err)
+		}
+	} else {
+		fmt.Printf("No logs found in %s; skipping log copying", logDir)
+	}
+
+	// Get the current time and create a tar.gz file with the timestamp in the name
+	tarFile := fmt.Sprintf("diags-%s.tar.gz", time.Now().Format("20060102_150405"))
+
+	err = exec.Command("tar", "-zcvf", tarFile, "diagnostics").Run()
```
robbrockbank: Same with `tar` - we should really use the golang archive/tar library or similar to do this.

calicoctl/commands/node/diags.go
```diff
@@ -0,0 +1,161 @@
...
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/docopt/docopt-go"
+)
+
+// diagCmds is a slice of struct to hold commands, their info and filename to run diagnostic on
```
robbrockbank: Comment is no longer correct.  Probably want to remove the "s" too -> diagCmd

calicoctl/commands/node/diags.go
gunjan5: yup, good point! I originally had it outside because we're not saving it's output/it doesn't produce any output to stdout, but I added it to the cmd list now

robbrockbank:
LGTM - squash and commit.

PR #130, opened 2016-09-12
`calicoctl node status` BGP ipv6 peer info

gunjan5:


gunjan5:
output:
```bash
vagrant@calico-01:~$ sudo ./lincalico node status
calico-node container is running. Status: Up 58 seconds

IPv4 BGP status
+--------------+-------------------+-------+----------+-------------+
| PEER ADDRESS |     PEER TYPE     | STATE |  SINCE   |    INFO     |
+--------------+-------------------+-------+----------+-------------+
| 172.17.8.102 | node-to-node mesh | up    | 23:18:20 | Established |
+--------------+-------------------+-------+----------+-------------+

IPv6 BGP status
+------------------------+-------------------+-------+----------+-------------+
|      PEER ADDRESS      |     PEER TYPE     | STATE |  SINCE   |    INFO     |
+------------------------+-------------------+-------+----------+-------------+
| fd80:24e2:f998:72d7::2 | node-to-node mesh | up    | 23:18:20 | Established |
+------------------------+-------------------+-------+----------+-------------+
```

Output with `INFO` not in `Established` state:
```bash
vagrant@calico-01:~$ sudo ./lincalico node status
calico-node container is running. Status: Up 4 hours

IPv4 BGP status
+--------------+-------------------+-------+----------+-------------+
| PEER ADDRESS |     PEER TYPE     | STATE |  SINCE   |    INFO     |
+--------------+-------------------+-------+----------+-------------+
| 172.17.8.102 | node-to-node mesh | up    | 17:56:25 | Established |
+--------------+-------------------+-------+----------+-------------+

IPv6 BGP status
+------------------------+-------------------+-------+----------+--------------------------------+
|      PEER ADDRESS      |     PEER TYPE     | STATE |  SINCE   |              INFO              |
+------------------------+-------------------+-------+----------+--------------------------------+
| fd80:24e2:f998:72d7::2 | node-to-node mesh | start | 17:56:21 | Active Socket: Connection      |
|                        |                   |       |          | closed                         |
+------------------------+-------------------+-------+----------+--------------------------------+
```

calicoctl/commands/node/status.go
```diff
@@ -160,5 +193,9 @@ func drawTable(birdOut string) {
 	for _, v := range data {
 		table.Append(v)
 	}
+	if len(data) == 0 {
+		return errors.New("No peers found")
```
caseydavenport: Instead of returning an error here, why don't we just print out "No IPv%s BGP peers found"?  

This function doesn't really need to return errors.

calicoctl/commands/node/status.go
```diff
@@ -79,21 +79,44 @@ Description:
...
+				if v == "6" {
+					birdSuffix = "6"
+					ipSep = ":"
+				}
+				fmt.Printf("\nIPv%s BGP status", v)
+
+				// Connect to the bird socket and get the data
+				c, err := net.Dial("unix", fmt.Sprintf("/var/run/calico/bird%s.ctl", birdSuffix))
+				if err != nil {
+					log.Printf("Error connecting to BIRD socket: %v", err)
```
caseydavenport: This should include the IP version.

calicoctl/commands/node/status.go
```diff
@@ -79,21 +79,44 @@ Description:
...
+					continue
+				}
+				defer c.Close()
+
+				fmt.Println()
```
caseydavenport: Is this just to add a newline?  Why not just stick a `\n` at the end of the previous printf?

calicoctl/commands/node/status.go
```diff
@@ -79,21 +79,44 @@ Description:
...
+
+				_, _ = c.Write([]byte("show protocols\n"))
```
caseydavenport: You're throwing away the error here.

calicoctl/commands/node/status.go
```diff
@@ -79,21 +79,44 @@ Description:
...
+				if err != nil {
+					log.Fatal("Error writing to BIRD socket:", err)
+				}
+				err = printBIRDResponse(c, ipSep)
```
caseydavenport: I'd quite like all of the print logic to be contained within a single function call so it's really easy to follow / understand.

For example, renaming `printBIRDResponse(...)` to `printBGPPeers(ipVersion int)`

And then, this code just looks like this:
```
// Print BGP peer information for both IPv4 and IPv6.
printBGPPeers(4)
printBGPPeers(6)
```

calicoctl/commands/node/status.go
gunjan5: Because previous print is not always printf, it's table render when there is no error.

calicoctl/commands/node/status.go
gunjan5: I think the way BIRD console works is it always errors out even if the command succeeds:
```
0001 BIRD 1.5.0 ready.
**show protocols**
2002-name     proto    table    state  since       info
1002-static1  Static   master   up     20:47:33    
 kernel1  Kernel   master   up     20:47:33    
 device1  Device   master   up     20:47:33    
 direct1  Direct   master   up     20:47:33    
0000 
```

this still returns a non-nil error for some reason

gunjan5:
Output with the restructured code:
```bash

vagrant@calico-01:~$ sudo ./lin node status
calico-node container is running. Status: Up 19 hours

IPv4 BGP status
+--------------+-------------------+-------+----------+-------------+
| PEER ADDRESS |     PEER TYPE     | STATE |  SINCE   |    INFO     |
+--------------+-------------------+-------+----------+-------------+
| 172.17.8.102 | node-to-node mesh | up    | 13:02:03 | Established |
+--------------+-------------------+-------+----------+-------------+

IPv6 BGP status
+------------------------+-------------------+-------+----------+-------------+
|      PEER ADDRESS      |     PEER TYPE     | STATE |  SINCE   |    INFO     |
+------------------------+-------------------+-------+----------+-------------+
| fd80:24e2:f998:72d7::2 | node-to-node mesh | up    | 13:02:04 | Established |
+------------------------+-------------------+-------+----------+-------------+

```

Error case output:
```bash

vagrant@calico-01:~$ sudo ./lin node status
calico-node container is running. Status: Up 13 seconds

IPv4 BGP status
+--------------+-------------------+-------+----------+-------------+
| PEER ADDRESS |     PEER TYPE     | STATE |  SINCE   |    INFO     |
+--------------+-------------------+-------+----------+-------------+
| 172.17.8.102 | node-to-node mesh | up    | 18:30:21 | Established |
+--------------+-------------------+-------+----------+-------------+

IPv6 BGP status
+------------------------+-------------------+-------+----------+--------------------------------+
|      PEER ADDRESS      |     PEER TYPE     | STATE |  SINCE   |              INFO              |
+------------------------+-------------------+-------+----------+--------------------------------+
| fd80:24e2:f998:72d7::2 | node-to-node mesh | start | 18:30:17 | Active Socket: Connection      |
|                        |                   |       |          | closed                         |
+------------------------+-------------------+-------+----------+--------------------------------+
```

calicoctl/commands/node/status.go
```diff
@@ -79,21 +78,12 @@ Description:
...
-			fmt.Println()
+			printBGPPeers("4")
+			printBGPPeers("6")
 
-			_, _ = c.Write([]byte("show protocols\n"))
-			if err != nil {
-				log.Fatal("Error writing to BIRD socket:", err)
-			}
-			printBIRDResponse(c)
+			// Loop over and try getting info about both IP versions
```
caseydavenport: This comment can go now

calicoctl/commands/node/status.go
```diff
@@ -160,5 +193,11 @@ func drawTable(birdOut string) {
 	for _, v := range data {
 		table.Append(v)
 	}
+
+	if len(data) == 0 {
+		fmt.Printf("No IPv%s address configured.\n", ipv)
```
caseydavenport: We don't have to print the _exact_ same message as the previous calicoctl version :)

It's generally better to make the message describe what is actually happening, otherwise it could be confusing to users.

In this case, we print this when no IPvX peers were found, I think.

calicoctl/commands/node/status.go
```diff
@@ -103,22 +91,41 @@ Description:
...
 
-	ipSep := "."
+	buf := make([]byte, 1024)
+
+	n, err := c.Read(buf[:])
+	if err != nil {
+		fmt.Printf("Error: %v\n", err)
+	}
+
+	birdOut := string(buf[:n])
```
caseydavenport: Could you define this guy closer to where it is used?

calicoctl/commands/node/status.go
```diff
@@ -129,7 +136,19 @@ func drawTable(birdOut string) {
 
 		if s != "" {
 			col := []string{}
-			fields := strings.Fields(line)[3:6]
+			f := strings.Fields(line)[3:]
```
caseydavenport: This is all quite confusing without comments to describe what is happening.

What is `f` supposed to represent? Do you really need to copy this data around?

gunjan5: `copy` is necessary because when you do `fields := f[0:3]` in Go they share the same underlying array, so if you modify one of those slices, the other one gets modified too! 

`f` is just a temp slice to hold the Fields (this is only when there is some connection error, since in that case there are more `fields` than what the table can hold) 

It is explained in the comment in line 143-145

caseydavenport: I think the problem here is more with this line:
```
col = append(col, fields...)
```

Where we're assuming a set number of fields.

We should probably be explicitly adding in the fields we expect:

```
// Add State, Since and Info columns.
col = append(col, fields[0])
col = append(col, fields[2])
col = append(col, fields[3]) 
```

calicoctl/commands/node/status.go
```diff
@@ -129,7 +136,19 @@ func drawTable(birdOut string) {
...
+			// For example in case of "Mesh_fd80_24e2_f998_72d7__2 BGP  master   start  17:56:21 Active Socket: Connection closed"
+			// providing only "Active" in the Info section is not enough, so we append rest of the info into the last field
+			if len(f) > 3 {
+				for _, e := range f[3:] {
+					fields[2] = fields[2] + " " + e
+				}
+			}
+
 			if strings.HasPrefix(s, "Mesh_") {
 				s = s[5:]
```
caseydavenport: s is the IP address in this case, right?  We should probably name that something more appropriate to make this code easier to follow.

calicoctl/commands/node/status.go
gunjan5: so fields is of fixed size `fields := make([]string, 3)` anything more that that gets appended to the last one (e.g. in case of connection error) similar to python version
```go
// This is to include the extra info when the "Info" field is other than "Established"
// For example in case of "Mesh_fd80_24e2_f998_72d7__2 BGP  master   start  17:56:21 Active Socket: Connection closed"
// providing only "Active" in the Info section is not enough, so we append rest of the info into the last field
if len(f) > 3 {
				for _, e := range f[3:] {
					fields[2] = fields[2] + " " + e
				}
			}
```

So even if BIRD output changes, it won't break the code

calicoctl/commands/node/status.go
gunjan5: I called is `s` because it's not exactly an IP (it's like `Mesh_fd80_24e2_f998_72d7__2`) Anyway, makes sense to name it something else, so I changed it to `ipString`

calicoctl/commands/node/status.go
caseydavenport: OK - can you at least move the comment to the top of this section (since all of this is related to that comment)?

I'm worried that, as it is, this code is relatively hard to work through, not being the one who wrote it.

gunjan5: updated the comment and moved it up

caseydavenport:
OK - let's squash it up and then LGTM.

PR #134, opened 2016-09-13
Fix `calicoctl version` command

### LINKED ISSUE #100: `calicoctl version` prints wrong information
```
macncheese:libcalico-go casey$ ./bin/calicoctl version
Version:      0.1.0-go
Build date:   __BUILD_DATE__
Git tag ref:  __GIT_DESCRIPTION__
Git commit:   __GIT_REVISION__
```

It doesn't appear to be filling in the variables for date, tag, commit.  Version should be dirty but it isnt.

doublek:
- Inject VERSION, GIT_REVISION and BUILD_DATE from Makefile
- Remove GIT_DESCRIPTION which is now redundant with the VERSION information.
- Cleanup build scripts used by release/calicoctl target.

Fixes #100


caseydavenport:
CC @robbrockbank

Dockerfile
```diff
@@ -8,12 +8,11 @@ RUN yum install -y \
 		make \
```
caseydavenport: Take a look at "Dockerfile.build" in this directory.  This is what we were using to make the build container for "build-containerized" before.

I like suffixing with ".build", because we also ship a `calico/ctl` container which should be the main Dockerfile in this repo (once we add it back in - don't have to do that in this PR!)

Makefile
```diff
@@ -24,21 +32,24 @@ force:
...
+	cd release && ln -sf calicoctl-$(CALICOCTL_VERSION) calicoctl
 
 # Build calicoctl in a container.
-build-containerized: $(BUILD_CONTAINER_MARKER)
-	mkdir -p dist
-	docker run -ti --rm --privileged --net=host \
-	-e PLUGIN=calico \
+build-containerized: force
+	mkdir -p bin
+	docker build -t calicoctl-build .
```
caseydavenport: I think I'd rather have the build-container as a make dependency for this target, rather than building the container every time this is run.  Any reason you decided to change this i.e remove the $(BUILD_CONTAINER_MARKER) dep?

calicoctl/commands/build_info.go
```diff
@@ -15,8 +15,4 @@
 // WARNING: This file is overwritten by a script in the release build process.
 package commands
 
-const (
```
caseydavenport: We can probably remove this entire file, and just stick these vars in version.go rather than splitting up the different info.  These are only used in version.go, I think.

doublek: I was thinking the same thing and they're used only in version.go. I'll combine them in version.go.

Makefile
doublek: I am right now reusing build-containerized in release/calicoctl, and when building for release/calicoctl I wanted a clean build environment which led me to recreating the container all the time. While writing this I just realized that the built container has just basic dependencies and nothing more and I can reuse the built container. I'll re-add BUILD_CONTAINER_MARKER.

caseydavenport: yep, the build container is just meant to be an environment to build in - it won't store any artifacts, etc from previous builds, so it should be good to reuse!

caseydavenport:
Enjoying the amount of red in this PR :)

Makefile
doublek: Got it.

caseydavenport:
OK - as discussed in person, we can merge the remaining Dockerfiles in another PR.

Could you squash the commits before merge?

doublek:
@caseydavenport Done.

caseydavenport:
LGTM!

PR #135, opened 2016-09-14
Use the same Dockerfile for all builds

doublek:
Essentially the old Dockerfile is now Dockerfile.build and all builds are now using the `centos:6.6` image as a base.


Makefile
```diff
@@ -1,4 +1,4 @@
-.PHONEY: all test ut update-vendor
+.PHONY: all test ut update-vendor
```
caseydavenport: Nice.

tomdee:
I don't understand this change. Why aren't we using an official Go docker image for building?

doublek:
Merging the dockerfiles and using the centos dockerfile was the result of a discussion between myself, @caseydavenport and @robbrockbank. Using centos for building comes from a specific customer req. On the other hand, we could also build all the time from official go docker images and have additional targeted builds as needed. 

Any comments @caseydavenport or @robbrockbank ?

doublek:
As discussed out of band, we'll need some more discussion on how to go about this one. Closing this until then.

Issue #85, opened 2016-08-22
Add calicoctl check system command


robbrockbank:
Another feature parity issue - I *think* we can again just port the python calicoctl command for this.

gunjan5:
Fixed in #143

PR #143, opened 2016-09-23
calicoctl checksystem command

gunjan5:
For issue #85

### DIFF
--- a/calicoctl/calicoctl.go
+++ b/calicoctl/calicoctl.go
@@ -35,6 +35,7 @@ func main() {
     get            Get a resource identified by file, stdin or resource type and name.
     version        Display the version of calicoctl.
     node           Node related commands.
+    checksystem    Check for compatibility with the host system.
 
 Options:
   -h --help  Show this screen.
@@ -48,7 +49,7 @@ Options:
 	if logLevel := arguments["--log-level"]; logLevel != nil {
 		parsedLogLevel, err := log.ParseLevel(logLevel.(string))
 		if err != nil {
-			fmt.Printf("Unknown log level: %s, expected one of: \n" +
+			fmt.Printf("Unknown log level: %s, expected one of: \n"+
 				"panic, fatal, error, warn, info, debug.\n", logLevel)
 			os.Exit(1)
 		} else {
--- a/calicoctl/commands/node/checksystem.go
+++ b/calicoctl/commands/node/checksystem.go
@@ -1,0 +1,220 @@
+// Copyright (c) 2016 Tigera, Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package node
+
+import (
+	"bufio"
+	"errors"
+	"fmt"
+	"log"
+	"os"
+	"os/exec"
+	"regexp"
+	"strings"
+
+	docopt "github.com/docopt/docopt-go"
+	goversion "github.com/mcuadros/go-version"
+)
+
+// The minimum allowed linux kernel version is 2.6.24, which introduced network
+// namespaces and veth pairs.
+const minKernelVersion = "2.6.24"
+
+// Required kernel modules to run Calico
+var requiredModules = []string{"xt_set", "ip6_tables"}
+
+// Checksystem checks host system for compatible versions
+func Checksystem() error {
+	doc := `Usage: 
+    calicoctl checksystem
+
+Description:
+    Check for compatibility with the host system
+`
+
+	// Note: This call is ignoring the error because error check happens at the level above
+	// i.e at `node.go` before it calls `node.Status`. This call is just so help message gets
+	// printed for this option
+	_, _ = docopt.Parse(doc, nil, true, "calicoctl", false, false)
+
+	// Make sure the command is run with super user priviladges
+	if os.Getuid() != 0 {
+		fmt.Println("Need super user privilages: Operation not permitted")
+		os.Exit(1)
+	}
+
+	systemOk := true
+
+	fmt.Print("Checking kernel version...\n")
+	err := checkKernelVersion()
+	if err != nil {
+		systemOk = false
+	}
+
+	fmt.Print("Checking kernel modules...\n")
+	err = checkKernelModules()
+	if err != nil {
+		systemOk = false
+	}
+
+	// If any of the checks fail, print a message and exit
+	if !systemOk {
+		fmt.Printf("%-20s\t\t\t%s", "System doesn't meet one or more minimum systems requirements to run Calico", "FAIL\n")
+		os.Exit(1)
+	}
+
+	fmt.Printf("%-20s\t\t\t%s", "System meets minimum system requirements to run Calico!", "OK\n")
+
+	return nil
+}
+
+// checkKernelVersion checks for minimum required kernel version
+func checkKernelVersion() error {
+	kernelVersion, err := exec.Command("uname", "-r").Output()
+	if err != nil {
+		log.Println(err)
+	}
+
+	// To strip the trailing `\n`
+	kernelVersionStr := strings.TrimSpace(string(kernelVersion))
+
+	// Print the version for the command output
+	fmt.Printf("\t\t%-20s", kernelVersionStr)
+
+	// Goversion normalizes the versions and compares them returns `true` if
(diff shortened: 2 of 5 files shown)


calicoctl/commands/checksystem.go
```diff
@@ -0,0 +1,257 @@
...
+			return true
+		}
+	}
+	return false
+}
+
+// checkModule is a utility function used by `checkKernelModules`
+// it opens the file provided and checks if the module passed in
+// as an argument exists for the provided kernelVersion
+func checkModule(filename, module, kernelVersion string) error {
```
matthewdupre: Would it be easier / shorter for this to just be a grep call and check the return code?

calicoctl/commands/checksystem.go
```diff
@@ -0,0 +1,257 @@
...
+	modulesLoadablePath := fmt.Sprintf("/lib/modules/%s/modules.dep", kernelVersionStr)
+
+	// File path to Builtin kernel modules
+	modulesBuiltinPath := fmt.Sprintf("/lib/modules/%s/modules.builtin", kernelVersionStr)
+
+	// Keep track of modules that are not found
+	modulesNotFound := []string{}
+
+	// Go through all the required modules and check Loadable and Builtin in order
+	for _, v := range requiredModules {
```
matthewdupre: We could delete the contains() and findAndRemove() functions as well.

calicoctl/calicoctl.go
```diff
@@ -35,6 +35,7 @@ func main() {
     get            Get a resource identified by file, stdin or resource type and name.
     version        Display the version of calicoctl.
     node           Node related commands.
+    checksystem    Check for incompatibilities on the host system.
```
gunjan5: yup! I had it worded like that before, but then I realized python version has it that way so I changed it. I'll change it back to more positive

calicoctl/commands/checksystem.go
```diff
@@ -0,0 +1,257 @@
...
+)
+
+// The minimum allowed linux kernel version is 2.6.24, which introduced network
+// namespaces and veth pairs.
+const minKernelVersion = "2.6.24"
+
+// Required kernel modules to run Calico
+var requiredModules = []string{"xt_set", "ip6_tables"}
+
+// Checksystem checks host system for compatible versions
```
gunjan5: Comment for any exported function in Go has to begin with it's name for godocs to work, that's why it's like `// Checksystem ...`.

Versions being validated are listed in comments (line 31,32)

calicoctl/commands/checksystem.go
```diff
@@ -0,0 +1,257 @@
...
+func Checksystem(args []string) error {
+	doc := `Usage: 
+    calicoctl checksystem
+
+Description:
+    Check for incompatibilities between Calico and the host system`
+
+	parsedArgs, err := docopt.Parse(doc, args, true, "calicoctl", false, false)
```
gunjan5: - `calicoctl` in version is to be consistent with all the other commands in `libcalico-go` for now. For now all the `calicoctl` commands print `calicoctl` when you pass `--version` 

- `docopts` port in Go is very hacky, but it is hittable. When `-h` is passed it treats it as no args passed O.o (as explained in the comment below)

calicoctl/commands/checksystem.go
```diff
@@ -0,0 +1,257 @@
...
+		fmt.Printf("%-20s\t\t\t%s", "System doesn't meet one or more minimum systems requirements to run Calico", "\u2717\n")
+		os.Exit(1)
+	}
+
+	fmt.Printf("%-20s\t\t\t%s", "System meets minimum system requirements to run Calico!", "\u2713\n")
+
+	return nil
+}
+
+// checkKernelVersion checks for minimum required kernel version
```
gunjan5: same as before, godocs require the comment to begin with func/field name, and version being verified is listed at the top in comment, writing it here would be redundant (also someone might forget to update it in future)

calicoctl/commands/checksystem.go
```diff
@@ -0,0 +1,257 @@
...
+
+Description:
+    Check for incompatibilities between Calico and the host system`
+
+	parsedArgs, err := docopt.Parse(doc, args, true, "calicoctl", false, false)
+	if err != nil {
+		return err
+	}
+
+	// Return if no args are passed or `-h` `--help` is passed
```
gunjan5: as I'm reading it now, it does sound strange :) I'll try to reword it

calicoctl/commands/checksystem.go
```diff
@@ -0,0 +1,257 @@
...
+
+	// Print the version for the command output
+	fmt.Printf("\t\t%-20s", kernelVersionStr)
+
+	// Goversion normalizes the versions and compares them returns `true` if
+	// running version is >= minimum version
+	if !goversion.CompareNormalized(kernelVersionStr, minKernelVersion, ">=") {
+
+		// Prints "✗" if current version is not >= minimum required version
+		fmt.Printf("\t\t\t\t\t\u2717\n")
```
gunjan5: you're right, didn't think about telnet. I'll replace ✓ with `OK.` and ✗ with `FAIL`

calicoctl/commands/checksystem.go
gunjan5: ok restructured it into a single loop but in a way `lsmod` is called only once

calicoctl/commands/checksystem.go
```diff
@@ -0,0 +1,257 @@
...
+	docopt "github.com/docopt/docopt-go"
+	goversion "github.com/mcuadros/go-version"
+)
+
+// The minimum allowed linux kernel version is 2.6.24, which introduced network
+// namespaces and veth pairs.
+const minKernelVersion = "2.6.24"
+
+// Required kernel modules to run Calico
+var requiredModules = []string{"xt_set", "ip6_tables"}
```
gunjan5: fixed!

calicoctl/commands/checksystem.go
matthewdupre: Silly godocs, but fair enough.

calicoctl/commands/checksystem.go
matthewdupre: "calicoctl" is not valid for --version; we need to fix this (everywhere) before it goes 1.0.  You can leave it and raise an issue if you like, or fix them (all ideally) here.

matthewdupre: It only squelches the -h because we tell it to in the Parse() call.

matthewdupre: (I'll reread this method once I've been through all the discussion, I really need to just look at it.)

calicoctl/commands/node/checksystem.go
```diff
@@ -0,0 +1,220 @@
...
+
+	fmt.Print("Checking kernel modules...\n")
+	err = checkKernelModules()
+	if err != nil {
+		systemOk = false
+	}
+
+	// If any of the checks fail, print a message and exit
+	if !systemOk {
+		fmt.Printf("%-20s\t\t\t%s", "System doesn't meet one or more minimum systems requirements to run Calico", "FAIL\n")
```
matthewdupre: The `-20` is superfluous here?

calicoctl/commands/node/checksystem.go
```diff
@@ -0,0 +1,220 @@
...
+		} else {
+			fmt.Printf("\t\t%-20s%s", v, "\t\t\t\t\tOK\n")
+		}
+	}
+
+	// If there are still any modules not found then return an error
+	if len(modulesNotFound) > 0 {
+
+		// ip6_tables is not a required module for ipv4 setups, so just print
+		// a warning instead of failing the systemc heck
```
matthewdupre: Space still in wrong place.

calicoctl/commands/node/checksystem.go
```diff
@@ -0,0 +1,220 @@
...
+		return errors.New("One of more kernel modules missing")
+	}
+
+	return nil
+}
+
+// checkModule is a utility function used by `checkKernelModules`
+// it opens the file provided and checks if the module passed in
+// as an argument exists for the provided kernelVersion
+func checkModule(filename, module, kernelVersion string) error {
```
matthewdupre: I still don't see the point of having a page of code to replace a single OS command, but if you really want it (and the unit testing etc. responsibilities that will go with this), then fine.

calicoctl/commands/node/checksystem.go
gunjan5: it is so the `OK` and `FAIL` line up with other `OK` and `FAIL`s

calicoctl/commands/checksystem.go
```diff
@@ -0,0 +1,227 @@
...
+		}
+	}
+
+	// If there are still any modules not found then return an error
+	if len(modulesNotFound) > 0 {
+
+		// ip6_tables is not a required module for ipv4 setups, so just print
+		// a warning instead of failing the systemc heck
+		if len(modulesNotFound) == 1 && modulesNotFound[0] == "ip6_tables" {
+			fmt.Printf("WARNING: Unable to detect ip6_tables kernel module")
```
gunjan5: fixed

Issue #94, opened 2016-08-22
Proper error reporting when ICMP Code is specified without the Type


gunjan5:
**Steps to reproduce:**
- Use this sample config yaml file: 
```yaml
- apiVersion: v1
  kind: tier
  metadata:
    name: tier1
  spec:
    order: 12.345
- apiVersion: v1
  kind: policy
  metadata:
    name: policy1
    tier: tier1
  spec:
    order: 9999.9999
    ingress:
      - action: nextTier
        icmp:
          code: 100
        source:
          tag: web
          net: 1.2.3.4/10
          selector:
          ports: [1,2,3,4]
        destination:
          tag: database
          net: 10.20.30.40/32
          selector:
          ports:
    egress:
      - action: deny
        protocol: tcp
        source:
          "!tag": abcd
          "!net": aa:bb:cc::ff/100
          "!selector": 
          "!ports": [100]
    selector:
```
- Not it has ICMP `code` specified without the `type`

**Current behavior:** 
```bash
./bin/calicoctl apply -f testprof.yaml
Error processing input file: error with field code = '100'
```

**Expeced behavior:** 
```bash
./bin/calicoctl apply -f testprof.yaml
Error processing input file: "icmpCodeWithoutType"
```

**Ref**: https://github.com/tigera/libcalico-go/blob/master/lib/api/rule.go#L62

Similar thing should be fixed for `NotICMP` too

robbrockbank:
The ICMP cross struct validation is fixed up, so closing this issue.  I will re-open another issue to make sure we handle the cross struct validation errors correctly and write out sensible error messages to the user.

robbrockbank:
See #192 for general error reporting issue.

PR #146, opened 2016-09-27
validation failure message

gunjan5:
For fixing #94 

Before: 
```bash
$./bin/calicoctl apply -f testprof.yaml
Error processing input file: error with field code = '100'
Error executing command: error with field code = '100'
```

After:
```bash
$ ./bin/calicoctl apply -f ./test/testprof.yaml
Error processing input file: error with field (!icmp Code without Type) code = '113'
Error executing command: error with field (!icmp Code without Type) code = '113'
```


robbrockbank:
Validation of ICMP fields has been handled - we still need to handle errors from struct validations but that would be a different bit of work.  Closing  this PR

Issue #138, opened 2016-09-16
status should work for non-dockerized installs.


matthewdupre:
We should implement them in a way that works for dockerized (calico-node) and non-dockerized execution of felix / BIRD.  They at least should give sane looking output in both cases.  First thought is we could use `ps` to check felix / BIRD status - that should work regardless of how they're being ran.

robbrockbank:
I think if this is required by a particular customer use case we should move up to P1

matthewdupre:
It was.  It also applies to all of OpenStack.

gunjan5:
Issue only for `status` command since `diags` doesn't really depend on a container runtime as per the discussion with @matthewdupre

gunjan5:
Behavior with non-containerized installs with only IPv4 (no `bird6` running): 
```bash
$sudo calicoctl node status
Calico process is running.

IPv4 BGP status
+--------------+-------------------+-------+----------+-------------+
| PEER ADDRESS |     PEER TYPE     | STATE |  SINCE   |    INFO     |
+--------------+-------------------+-------+----------+-------------+
| 172.17.8.102 | node-to-node mesh | up    | 19:51:08 | Established |
+--------------+-------------------+-------+----------+-------------+

INFO: BIRDv6 process: 'bird6' is not running.
```

matthewdupre:
LGTM :)

PR #149, opened 2016-09-27
status command with non-containerized Calico

gunjan5:
Fix for issue #138

### DIFF
--- a/calicoctl/commands/node/status.go
+++ b/calicoctl/commands/node/status.go
@@ -23,10 +23,8 @@ import (
 	"regexp"
 	"strings"
 
-	"github.com/docker/engine-api/client"
-	"github.com/docker/engine-api/types"
 	"github.com/docopt/docopt-go"
-	"golang.org/x/net/context"
+	gops "github.com/mitchellh/go-ps"
 
 	"github.com/olekukonko/tablewriter"
 )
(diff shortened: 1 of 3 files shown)


gunjan5:
Behavior with non-containerized installs with only IPv4 (no `bird6` running): 
```bash
$sudo calicoctl node status
Calico process is running.

IPv4 BGP status
+--------------+-------------------+-------+----------+-------------+
| PEER ADDRESS |     PEER TYPE     | STATE |  SINCE   |    INFO     |
+--------------+-------------------+-------+----------+-------------+
| 172.17.8.102 | node-to-node mesh | up    | 19:51:08 | Established |
+--------------+-------------------+-------+----------+-------------+

INFO: BIRDv6 process: 'bird6' is not running.
```

calicoctl/commands/node/status.go
```diff
@@ -48,49 +46,56 @@ Description:
...
+
+		// Have to manually enter an empty line because the table print
+		// library prints the last line, so can't insert a '\n' there
+		fmt.Println()
+		return nil
 	}
 
 	// Return and print message if calico-node is not running
-	fmt.Println("calico-node container not running")
+	fmt.Printf("Calico process is not running.\n")
```
matthewdupre: I think we should probably be OS exiting in this case: status is definitely not good, so RC 1 seems best.

It would also allow you to refactor the function to make the conditional / return structure simpler: a bunch of unnested if tests, with a couple of exit(1)s for failures, and a single return nil "OK" from the end.

gunjan5: done!

calicoctl/commands/node/status.go
```diff
@@ -51,39 +51,39 @@ Description:
 		fmt.Println(err)
 	}
 
+	if !psContains("calico-felix", processes) {
+		// Return and print message if calico-node is not running
+		fmt.Printf("Calico process is not running.\n")
+		os.Exit(1)
+	}
+
 	// Go through running processes and check if `calico-felix` and `bird` (`bird6`)
```
matthewdupre: Comment needs to move up above the if.

matthewdupre:
Just one //comment that got left behind in the move.  Please squash the commits down as well.

matthewdupre:
Just need a squash now.

gunjan5:
The fix for [1086](https://github.com/projectcalico/calico-containers/issues/1086) is still applicable even though we're not looking at the container names. We look at the calico processes instead

Issue #141, opened 2016-09-21
calicoctl won't create policy


djosborne:
```
$ bin/calicoctl create -f ./policy.yaml
Error processing input file: Unknown resource type (policy) and/or version (v1)
Error executing command: Unknown resource type (policy) and/or version (v1)
```

```
$ cat policy.yaml
apiVersion: v1
kind: policy
metadata:
  name: policy1
  tier: tier1
spec:
  order: 100
  selector: type=='database'
  ingress:
  - action: deny
```

gunjan5:
Seems like as of this change: https://github.com/tigera/libcalico-go/commit/b4fd636e860597d8ccb2f06f501135d03e169bfb#diff-fcf1fd30667a4c28c65230985a8ace13R24

the config `.yaml` file `apiVersion` has to be `"1.0"`

I will open a PR to change all the `.yaml` files to use the new `apiVersion`

CC @robbrockbank

gunjan5:
after some discussion with @robbrockbank , we are going back to `v1`, so you won't have to change the `.yaml` files and `apiVersion` can be `v1` (will open a new PR shortly)

PR #150, opened 2016-09-27
change apiVersion from v1 to 1.0 in test yaml files

gunjan5:
fix for #141 

Because of the change: https://github.com/tigera/libcalico-go/commit/b4fd636e860597d8ccb2f06f501135d03e169bfb#diff-fcf1fd30667a4c28c65230985a8ace13R24


caseydavenport:
Why is this the fix? Why not just go back to using `v1` as the version?

I don't know why that version was changed.  Can someone explain? @robbrockbank?

gunjan5:
@caseydavenport this change fixes the issue #141 however, I was planning to talk to @robbrockbank first to see if this is the right fix or should we revert back to `v1` as the version?

gunjan5:
See #153 and #141 for the fix

PR #151, opened 2016-09-28
Cross struct validation

robbrockbank:
Add cross structure validation and add UTs for validation functions.


lib/validator/validator_test.go
```diff
@@ -15,31 +15,405 @@
...
+	var V254 = 254
+	var V255 = 255
+	var V256 = 256
+
+	// Set up some values we use in various tests.
+	_ = ipv4_1.UnmarshalText([]byte("1.2.3.4"))
+	_ = ipv4_2.UnmarshalText([]byte("100.200.0.0"))
+	_ = ipv6_1.UnmarshalText([]byte("aabb:aabb::ffff"))
+	_ = ipv6_2.UnmarshalText([]byte("aabb::abcd"))
+	_, cidr, _ = gonet.ParseCIDR("1.2.3.4/32")
```
fasaxc: In my tests, I solved this initialisation problem by writing a local `mustParseCIDR(string)` function that returns the CIDR only and panics on error.

lib/validator/validator_test.go
```diff
@@ -15,31 +15,405 @@
...
+	netv6_2 = net.IPNet{*cidr}
+	_, cidr, _ = gonet.ParseCIDR("aabb:aabb::ffff/122")
+	netv6_3 = net.IPNet{*cidr}
+	_, cidr, _ = gonet.ParseCIDR("aabb:aabb::ffff/10")
+	netv6_4 = net.IPNet{*cidr}
+
+	// Perform basic validation of different fields and structures to test simple valid/invalid
+	// scenarios.  This does not test precise error strings - but does cover a lot of the validation
+	// code paths.
+	var _ = DescribeTable("Validator",
```
fasaxc: If you're in a function, I don't think you need the `var _ =` part.

lib/validator/validator.go
```diff
@@ -33,7 +34,7 @@ var validate *validator.Validate
 var (
 	nameRegex          = regexp.MustCompile("^[a-zA-Z0-9_.-]+$")
```
fasaxc: This matches Felix's regex.

lib/validator/validator.go
```diff
@@ -33,7 +34,7 @@ var validate *validator.Validate
...
 	labelRegex         = regexp.MustCompile("^[a-zA-Z_./-][a-zA-Z0-9_./-]*$")
```
fasaxc: This is less permissive than felix's regex, which allows numerics in the first char.  No harm in being more aggressive here.

lib/validator/validator.go
fasaxc: We also use this regex for validating orch/workload/endpoint IDs and tags

lib/validator/validator.go
```diff
@@ -129,43 +138,43 @@ func validateLabels(v *validator.Validate, topStruct reflect.Value, currentStruc
 
 func validateInterface(v *validator.Validate, topStruct reflect.Value, currentStructOrField reflect.Value, field reflect.Value, fieldType reflect.Type, fieldKind reflect.Kind, param string) bool {
 	b := []byte(field.String())
-	log.Infof("Validate interface: %s", b)
+	log.Debugf("Validate interface: %s", b)
 	return nameRegex.Match(b)
```
fasaxc: For linux interface names, we use a different regex in felix, limiting the length and characters: `^[a-zA-Z0-9_-]{1,15}$`

Not sure if we should allow `.` and `:` in there too; probably needs some research.

lib/api/workloadendpoint.go
```diff
@@ -15,42 +15,107 @@
...
+	Labels map[string]string `json:"labels,omitempty" validate:"omitempty,labels"`
 }
 
+// WorkloadEndpointMetadata contains the specification for a WorkloadEndpoint resource.
 type WorkloadEndpointSpec struct {
-	IPNetworks    []IPNet  `json:"ipNetworks,omitempty" validate:"omitempty"`
-	Profiles      []string `json:"profiles,omitempty" validate:"omitempty,dive,name"`
-	InterfaceName string   `json:"interfaceName,omitempty" validate:"omitempty,interface"`
-	MAC           MAC      `json:"mac,omitempty" validate:"omitempty"`
+	// IPNetworks is a list of subnets allocated to this endpoint. IP packets will only be
```
fasaxc: Missing state field?  Felix uses that field as the "admin state".

lib/validator/validator.go
```diff
@@ -129,43 +138,43 @@ func validateLabels(v *validator.Validate, topStruct reflect.Value, currentStruc
...
-	} else if p.Type == numorstring.NumOrStringString && !protocolRegex.MatchString(p.StrVal) {
-		structLevel.ReportError(reflect.ValueOf(p.StrVal), "Protocol", "protocol", "protocol name invalid")
+	log.Debugf("Validate protocol: %v %s %d", p.Type, p.StrVal, p.NumVal)
+
+	// The protocol field may be an integer 1-254, or one of the valid protocol names.
+	if num, err := p.NumValue(); err == nil {
+		if (num < 1) || (num > 255) {
+			structLevel.ReportError(reflect.ValueOf(p.NumVal), "Protocol", "protocol", "protocol number invalid")
+		}
+	} else if !protocolRegex.MatchString(p.String()) {
```
robbrockbank: That'll be handled in the previous branch.

lib/api/workloadendpoint.go
robbrockbank: We currently set this to be "active" in the client, and don't expose this on the public facing API.

lib/validator/validator.go
fasaxc: Oh, does the parser for `Protocol` objects convert `"123"` to `123`?

lib/api/workloadendpoint.go
```diff
@@ -15,42 +15,107 @@
...
+	// IPv6Gateway is the gateway IPv6 address for traffic from the workload.
+	IPv6Gateway net.IP `json:"ipv6Gateway,omitempty" validate:"omitempty,ipv6"`
+
+	// A list of security Profile resources that apply to this endpoint. Each profile is
+	// applied in the order that they appear in this list.  Profile rules are applied
+	// after the tier security policy.
+	Profiles []string `json:"profiles,omitempty" validate:"omitempty,dive,name"`
+
+	// InterfaceName the name of the Linux interface on the host: for example, tap80.
+	InterfaceName string `json:"interfaceName,omitempty" validate:"omitempty,interface"`
```
robbrockbank: k

lib/numorstring/float32orstring.go
```diff
@@ -34,7 +33,20 @@ type Float32OrString struct {
 func (f *Float32OrString) UnmarshalJSON(b []byte) error {
```
robbrockbank: will update.  I'll remove the Float32OrString since we don't use it, and update Order to be a float32

lib/validator/validator.go
robbrockbank: I'll stick with what Felix allows - since the API needs to be at least as restrictive.

lib/validator/validator.go
robbrockbank: Yup.  You can sort of work around it by setting the String field as "123" but provided you use the "NumValue()" to get the number you should be golden.

When reading from JSON we convert numerical string values to the number.

robbrockbank:
Ok, I believe that all comments are addressed.

PR #152, opened 2016-09-28
Implement host IP resource.

fasaxc:


fasaxc:
I've done some manual testing of this and it seems to work as expected.

fasaxc:
TODO:

PR #157, opened 2016-09-30
Calicoctl cmd error

### LINKED ISSUE #155: calicoctl error reporting for subcommands
```
ubuntu@ip-172-31-34-92:~$ calicoctl node
Usage:
        calicoctl node status
        calicoctl node diags [--log-dir=<LOG_DIR>]
        calicoctl node checksystem
Error executing command:
```

The error is useless (what command, what error).

```
ubuntu@ip-172-31-34-92:~$ sudo -E calicoctl node
Running Docker container with the following command:

docker run -d --restart=always --net=host --privileged --name=calico-node -e HOSTNAME=ip-172-31-34-92 -e IP= -e IP6= -e CALICO_NETWORKING=true -e AS= -e NO_DEFAULT_POOLS= -e ETCD_AUTHORITY=127.0.0.1:2379 -e ETCD_SCHEME=http -v /var/log/calico:/var/log/calico -v /lib/modules:/lib/modules -v /var/run/calico:/var/run/calico calico/node:v0.22.0

Calico node is running with id: b461dc01f7ff66420fdb8fab4730db09d1c54b2ac9b2e243744c448cacbe0f22
Waiting for successful startup
Waiting for etcd connection...
No IP provided. Using detected IP: 172.31.34.92
Calico node started successfully
```

Oh, so that's what it's trying to do...  The Usage before (non-sudo) suggested this syntax was illegal (both by it's presence, and it not containing `calicoctl node`), but now it turns out it's not illegal.  The usage statement should describe all available options (or say that it doesn't), it should say what the error was and what it was doing at the time.

```
ubuntu@ip-172-31-34-92:~$ sudo -E calicoctl node status
Usage:
  calicoctl node [--ip=<IP>] [--ip6=<IP6>] [--node-image=<DOCKER_IMAGE_NAME>]
    [--runtime=<RUNTIME>] [--as=<AS_NUM>] [--log-dir=<LOG_DIR>]
    [--detach=<DETACH>] [--no-pull]
    [(--libnetwork [--libnetwork-image=<LIBNETWORK_IMAGE_NAME>])]
  calicoctl node stop [--force]
  calicoctl node remove [--hostname=<HOSTNAME>] [--remove-endpoints]
  calicoctl node show
  calicoctl node bgp peer add <PEER_IP> as <AS_NUM>
  calicoctl node bgp peer remove <PEER_IP>
  calicoctl node bgp peer show [--ipv4 | --ipv6]
```

Wow, this is a totally different usage statement.  I'm trying to run a command the usage statement told me I could run, but instead it's giving me a totally different usage statement...

This needs to be fixed for GA.  I'd also like whoever looks at this to think about how the docopt.Parse() calls are made (where from, how many times, using what docstrings etc.).  The fact that it's gone this wrong suggests there's something wrong with the code structure.

gunjan5:
part of #155 fix

### DIFF
--- a/calicoctl/calicoctl.go
+++ b/calicoctl/calicoctl.go
@@ -16,6 +16,7 @@ package main
 
 import (
 	"fmt"
+	"strings"
 
 	"os"
 
@@ -25,7 +26,8 @@ import (
 )
 
 func main() {
-	usage := `Usage: calicoctl [options] <command> [<args>...]
+	usage := `Usage:
+    calicoctl [options] <command> [<args>...]
 
     create         Create a resource by filename or stdin.
     replace        Replace a resource by filename or stdin.
@@ -79,10 +81,11 @@ Options:
 		default:
 			fmt.Println(usage)
 		}
+
+		if err != nil {
+			fmt.Println("Error executing command. Invalid option: 'calicoctl", strings.Join(args, " "), "'. See 'calicoctl --help' to read about a specific subcommand.")
+			os.Exit(1)
+		}
 	}
 
-	if err != nil {
-		fmt.Printf("Error executing command: %s\n", err)
-		os.Exit(1)
-	}
 }
--- a/calicoctl/calicoctl.go
+++ b/calicoctl/calicoctl.go
@@ -83,7 +83,7 @@ Options:
 		}
 
 		if err != nil {
-			fmt.Println("Error executing command. Invalid option: 'calicoctl", strings.Join(args, " "), "'. See 'calicoctl --help' to read about a specific subcommand.")
+			fmt.Printf("Error executing command. Invalid option: 'calicoctl %s'. Use flag '--help' to read about a specific subcommand.\n", strings.Join(args, " "))
 			os.Exit(1)
 		}
 	}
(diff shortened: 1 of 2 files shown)


calicoctl/calicoctl.go
```diff
@@ -79,10 +81,11 @@ Options:
 		default:
 			fmt.Println(usage)
 		}
-	}
 
-	if err != nil {
-		fmt.Printf("Error executing command: %s\n", err)
-		os.Exit(1)
+		if err != nil {
+			fmt.Printf("Error executing command. Invalid option: 'calicoctl %s'. Use flag '--help' to read about a specific subcommand.\n", strings.Join(args, " "))
```
matthewdupre: What's the return code / error handling strategy for calicoctl?  If we're going to ignore the contents of the error here, we need to be sure that all reporting to the user is done inside the subcommands.

gunjan5: The errors get handled right where they happen in the subcommands

matthewdupre: In that case we shouldn't need to print any messages here (or even pass the errors up: that implies that action might be needed, whereas a contract of "succeeds, or os.exit(1)s" is nice and simple for all subcommands.

matthewdupre:
Merging this; we'll separate out refactoring calicoctl's error handling back into the issue and fix it later.

Issue #91, opened 2016-08-22
Remove "." imports


robbrockbank:
We have a bunch of:

```
import (
          . "blah"
)
```

imports.  We should really remove the . imports.

caseydavenport:
@gunjan5 :)

gunjan5:
@robbrockbank I have some spare time, so I'll work on this issue. Hope it's ok with you

gunjan5:
fixed in #159

PR #159, opened 2016-10-02
remove dot imports

gunjan5:


gunjan5:
Fixes #91

gunjan5:
@robbrockbank I think Go best practice is to not use the Dot Import for tests unless it causes circular dependency issues.

caseydavenport:
@robbrockbank @gunjan5 - I'd prefer _not_ to use dot imports wherever possible (including tests).  This LGTM!

PR #161, opened 2016-10-05
IPAM library Unit Tests + small bugfix for StrictAffinity

gunjan5:
**Note**: This also fixes a minor bug where IPAM was ignoring `StrictAffinity` config

**Test cases (AutoAssign):**
**Test 1**: AutoAssign 1 IPv4, 1 IPv6 - expect one of each to be returned.
**Test 2**: AutoAssign 256 IPv4, 256 IPv6 - expect 256 IPv4 + IPv6 addresses
**Test 3**: AutoAssign 257 IPv4, 0 IPv6 - expect 256 IPv4 addresses, no IPv6, and an error.
**Test 4**: AutoAssign 0 IPv4, 257 IPv6 - expect 256 IPv6 addresses, no IPv6, and an error.
**Test 5** (use pool of size /25 so only two blocks are contained): 
  - Assign 1 address on host A (Expect 1 address)
  - Assign 1 address on host B (Expect 1 address, different block)
  - Assign 64 more addresses on host A (Expect 63 addresses from hostA's block, 1 address from hostB's block)

**Test cases (AssignIP):**
**Test 1**: Assign 1 IPv4 from a configured pool - expect no error returned.
**Test 2**: Assign 1 IPv6 from a configured pool - expect no error returned.
**Test 3**: Assign 1 IPv4 from a non-configured pool - expect an error returned.
**Test 4**: Assign 1 IPv4 from a configured pool twice:
  - Expect no error returned while assigning the IP for the first time.
  - Expect an error returned while assigning the SAME IP again.

**Test cases (ReleaseIPs):**
**Test 1**: release an IP that's not configured in any pools - expect a slice with the same IP as unallocatedIPs and no error.
**Test 2**: release an IP that's not allocated in the pool - expect a slice with one (unallocatedIPs) and no error.
**Test 3**: Assign 1 IPv4 with AssignIP from a configured pool and then release it.
- Assign should not return an error.
- ReleaseIP should return empty slice of IPs (unallocatedIPs) and no error.

**Test 4**: Assign 66 IPs (across 2 blocks) with AutoAssign from a configured pool then release them.
- Assign should not return an error.
- ReleaseIPs should return an empty slice of IPs (unallocatedIPs) and no error.

**Test 5**: Assign 1 IPv4 address with AssignIP then try to release 2 IPs.
- Assign should not return any error.
- ReleaseIPs should return a slice with one (unallocatedIPs) and no error.

### DIFF
--- a/lib/client/ipam_test.go
+++ b/lib/client/ipam_test.go
@@ -1,0 +1,196 @@
+// Copyright (c) 2016 Tigera, Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package client_test
+
+import (
+	"fmt"
+	"log"
+	"os"
+	"os/exec"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	. "github.com/onsi/ginkgo/extensions/table"
+
+	"github.com/tigera/libcalico-go/calicoctl/commands"
+	"github.com/tigera/libcalico-go/lib/api"
+	"github.com/tigera/libcalico-go/lib/backend/etcd"
+	"github.com/tigera/libcalico-go/lib/client"
+)
+
+var etcdType api.BackendType
+
+var args = []client.AutoAssignArgs{
+	{
+		Num4: 0,
+		Num6: 20,
+		//Hostname: "notMac",
+	},
+	{
+		Num4: 4,
+		Num6: 10,
+	},
+	{
+		Num4: 1,
+		Num6: 0,
+	},
+	{
+		Num4: 0,
+		Num6: 256,
+	},
+
+	{
+		Num4:     4,
+		Num6:     254,
+		Hostname: "notMac",
+	},
+	{
+		Num4: 1,
+		Num6: 0,
+	},
+}
+
+func testIPAM(inv4, inv6 int, host string, setup bool) (int, int) {
+	fmt.Println("in func")
+
+	etcdType = "etcdv2"
+
+	etcdConfig := etcd.EtcdConfig{
+		EtcdEndpoints: "http://127.0.0.1:2379",
+	}
+	ac := api.ClientConfig{BackendType: etcdType, BackendConfig: &etcdConfig} //etcdType.NewConfig()}
+
+	bc, err := client.New(ac)
+	if err != nil {
+		panic(err)
+	}
+
+	ic := bc.IPAM()
+
+	entry := client.AutoAssignArgs{
+		Num4:     inv4,
+		Num6:     inv6,
+		Hostname: host,
+	}
+	if setup {
+		setupEnv()
+	}
+
+	v4, v6, outErr := ic.AutoAssign(entry)
+
+	if outErr != nil {
+		fmt.Print("printing error.... ")
+		fmt.Println(outErr)
+	}
+
+	return len(v4), len(v6)
+
+}
+
+var _ = Describe("IPAM", func() {
+
+	DescribeTable("requested IPs vs got IPs",
+		func(host string, setup bool, inv4, inv6, expv4, expv6 int) {
+			outv4, outv6 := testIPAM(inv4, inv6, host, setup)
+			//fmt.Println(outv4, outv6)
+			Expect(outv4).To(Equal(expv4))
+			Expect(outv6).To(Equal(expv6))
+		},
+		Entry("1 v4 1 v6", "testHost", true, 1, 1, 1, 1),
+		Entry("256 v4 256 v6", "testHost", true, 256, 256, 256, 256),
+		Entry("257 v4 0 v6", "testHost", true, 257, 0, 256, 0),
+		Entry("0 v4 257 v6", "testHost", true, 0, 257, 0, 256),
+	)
+
+})
+
+func setupEnv() {
+
+	cmd := "docker"
+	argsRm := []string{"rm", "-f", "calico-etcd"}
+	if err := exec.Command(cmd, argsRm...).Run(); err != nil {
+		log.Println(err)
+	}
+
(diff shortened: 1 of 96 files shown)


fasaxc:
TODO:

gunjan5:
I will do the `tigera` --> `projectcalico` move for these files in the end so my paths don't get messed up

lib/client/ipam_test.go
```diff
@@ -0,0 +1,300 @@
...
+	if outErr != nil {
+		log.Println(outErr)
+	}
+
+	return len(v4), len(v6), outErr
+}
+
+// setupEnv cleans up etcd if cleanEnv flag is passed and then creates IP pool based on the pool name passed to it.
+func setupEnv(pool string) {
+
```
robbrockbank: actually see note below about creating from file.

lib/client/ipam_test.go
```diff
@@ -0,0 +1,300 @@
...
+	etcdArgs := strings.Fields("rm /calico --recursive")
```
robbrockbank: nit: any reason not to use []string{"rm", "/calico", "--recursive"}  seems a bit convoluted to create a string and then split it.

lib/client/ipam_test.go
```diff
@@ -0,0 +1,300 @@
...
+	"github.com/tigera/libcalico-go/lib/client"
+	cnet "github.com/tigera/libcalico-go/lib/net"
+)
+
+// Setting BackendType to etcdv2 which is the only supported backend at the moment.
+var etcdType api.BackendType = "etcdv2"
+
+// Setting localhost as the etcd endpoint location since that's where `make run-etcd` runs it.
+var etcdConfig = etcd.EtcdConfig{
+	EtcdEndpoints: "http://127.0.0.1:2379",
```
robbrockbank: Ditto, this is the default for an empty client config.

lib/client/ipam_test.go
```diff
@@ -0,0 +1,300 @@
...
+
+	"github.com/tigera/libcalico-go/calicoctl/commands"
+	"github.com/tigera/libcalico-go/lib/api"
+	"github.com/tigera/libcalico-go/lib/backend/etcd"
+	"github.com/tigera/libcalico-go/lib/client"
+	cnet "github.com/tigera/libcalico-go/lib/net"
+)
+
+// Setting BackendType to etcdv2 which is the only supported backend at the moment.
+var etcdType api.BackendType = "etcdv2"
```
robbrockbank: This is the default for an empty client config I believe.

gunjan5: That's what I thought, but it's only when it's through parsing JSON:
`json:"datastoreType" envconfig:"DATASTORE_TYPE" default:"etcdv2"` 

Also, I think it's better to make it explicit which backend it is, so in future when we add more backends, it's clear

lib/client/ipam_test.go
gunjan5: Ditto, it doesn't work with empty config, that is it gets empty string as EtcdEndpoints instead of going default, which I think is because the the default is set for when decoding JSON:
 `json:"etcdAuthority" envconfig:"ETCD_AUTHORITY" default:"127.0.0.1:2379"`

lib/client/ipam_test.go
```diff
@@ -0,0 +1,300 @@
...
+
+		// - Assign 64 more addresses on host A (Expect 63 addresses from host A's block, 1 address from host B's block).
+		Entry("64 v4 0 v6 host-A", "host-A", false, "pool2", 64, 0, 64, 0, nil),
+	)
+
+	DescribeTable("AssignIP: requested IP vs returned error",
+		func(inIP net.IP, host string, cleanEnv bool, pool string, expError error) {
+			outError := testIPAMAssignIP(inIP, host, pool, cleanEnv)
+			if expError != nil {
+				Ω(outError).Should(HaveOccurred())
```
gunjan5: You're right, I got confused by gomega documentation.
I fixed it. It's: `Expect(outError).To(HaveOccurred())`  now

lib/client/ipam_test.go
```diff
@@ -0,0 +1,300 @@
...
+		// Test 5: Assign 1 IPv4 address with AssignIP then try to release 2 IPs.
+		// - Assign should not return an error.
+		// - ReleaseIPs should return a slice with one (unallocatedIPs) and an error???
+		Entry("Assign 1 IPv4 from a configured pool", net.IP{}, true, "pool1", net.ParseIP("192.168.1.0"), 0, []cnet.IP{}, nil),
+		Entry("Assign 1 IPv4 from a configured pool", net.ParseIP("192.168.1.1"), false, "pool1", net.IP{}, 0, []cnet.IP{cnet.IP{net.ParseIP("192.168.1.1")}}, nil),
+	)
+
+})
+
+// comment bro!
```
gunjan5: oops, I thought I fixed it 😆

lib/client/ipam_test.go
```diff
@@ -0,0 +1,300 @@
...
+
+	unallocatedIPs, outErr = ic.ReleaseIPs(inIPs)
+	if outErr != nil {
+		log.Println("######################: ", outErr)
+	}
+	return unallocatedIPs, outErr
+}
+
+// testIPAMAssignIP takes an IPv4 or IPv6 IP with a hostname and pool name and calls AssignIP.
+// When cleanEnv is passed it passes it along to wipe clean etcd and reset IPAM config.
```
gunjan5: need to update the comment, it use to be different.

lib/client/ipam_test.go
robbrockbank: Ah gotcha!  Yes you are quite right - the defaulting occurs as part of reading from file/environments.   Of course in this particular case we are explicitly setting the data.  My bad!

lib/client/ipam_test.go
gunjan5: fixed!

gunjan5:
@robbrockbank made the changes you suggested:
- creating pools through APIs now instead of calling calicoctl methods
- broke out util functions into `testuitls` package
- restructured some of the code to make things simpler
- fixed comments
- removed Affinity UTs since there are possibly 2 bugs I might have discovered with Affinity code that I need to discuss with either you or @caseydavenport so I'd leave it (Affinity UTs and possible bugfix) for a separate PR 

PTAL

lib/testutils/cleanEtcd.go
```diff
@@ -0,0 +1,28 @@
...
+package testutils
+
+import (
+	"log"
+	"os/exec"
+)
+
+// CleanEtcd is a utility function to wipe clean "/calico" recursively from etcd.
+func CleanEtcd() {
+	err := exec.Command("etcdctl", "rm", "/calico", "--recursive").Run()
```
robbrockbank: this places another requirement on running the tests no?  Is the user going to have etcdctl installed?  Spinning up etcd is handled through a container which is more reasonable, but requiring etcdctl to be installed seems more worrying.

Can we just change this to be a curl request - since we can probably do that using golang native library stuff?

How does CNI do this, or does it also use etcdctl?

PR #162, opened 2016-10-05
Add workload endpoints as a resource type

robbrockbank:
This adds workload endpoints as a resource type to calicoctl.  The workload is automatically deleted when the last endpoint is deleted.  Workloads themselves are not explicit resource types (neither are orchestrators), but they do appear as additional IDs in the workload endpoint.

Note that imports still refer to tigers.  There is a separate issue to fix up tigera imports.

### DIFF
--- a/calicoctl/commands/delete.go
+++ b/calicoctl/commands/delete.go
@@ -25,11 +25,12 @@ import (
 func Delete(args []string) error {
 	doc := EtcdIntro + `Delete a resource identified by file, stdin or resource type and name.
 
-Valid resource kinds are bgpPeer, hostEndpoint, policy, pool, profile and tier.  The <KIND>
-parameter is case insensitive and may be pluralized.
+Valid resource kinds are bgpPeer, hostEndpoint, workloadEndpoint, policy, pool, profile and tier.
+The <KIND> parameter is case insensitive and may be pluralized.
 
 Usage:
-  calicoctl delete (([--tier=<TIER>] [--hostname=<HOSTNAME>] [--scope=<SCOPE>] <KIND> <NAME>) |
+  calicoctl delete ([--tier=<TIER>] [--hostname=<HOSTNAME>] [--orchestrator=<ORCH>] [--workload=<WORKLOAD>] [--scope=<SCOPE>]
+                    (<KIND> [<NAME>])
                     --filename=<FILE>)
                    [--skip-not-exists] [--config=<CONFIG>]
 
@@ -51,6 +52,8 @@ Options:
   -f --filename=<FILENAME>     Filename to use to delete the resource.  If set to "-" loads from stdin.
   -t --tier=<TIER>             The policy tier.
   -n --hostname=<HOSTNAME>     The hostname.
+     --orchestrator=<ORCH>     The orchestrator (only used for workload endpoints).
+     --workload=<WORKLOAD>     The workload (only used for workload endpoints).
   --scope=<SCOPE>              The scope of the resource type.  One of global, node.  This is only valid
                                for BGP peers and is used to indicate whether the peer is a global peer
                                or node-specific.
--- a/calicoctl/resourcemgr/workloadendpoint.go
+++ b/calicoctl/resourcemgr/workloadendpoint.go
@@ -1,0 +1,63 @@
+// Copyright (c) 2016 Tigera, Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package resourcemgr
+
+import (
+	"github.com/projectcalico/libcalico-go/lib/api"
+	"github.com/projectcalico/libcalico-go/lib/api/unversioned"
+	"github.com/projectcalico/libcalico-go/lib/client"
+)
+
+func init() {
+	registerResource(
+		api.NewWorkloadEndpoint(),
+		api.NewWorkloadEndpointList(),
+		[]string{"HOSTNAME", "ORCHESTRATOR", "WORKLOAD", "NAME"},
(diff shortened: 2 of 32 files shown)


lib/backend/model/workloadendpoint.go
```diff
@@ -54,22 +54,28 @@ func (key WorkloadEndpointKey) defaultPath() (string, error) {
...
-		key.Hostname, key.OrchestratorID, key.WorkloadID, key.EndpointID), nil
+
+	// A workload endpoint may auto create two additional entries that need to be
+	// deleted.  This is actually dependent on the datastore, but we return these
+	// as part of the default paths.
+	workload := fmt.Sprintf("/calico/v1/host/%s/workload/%s/%s",
+		key.Hostname, key.OrchestratorID, key.WorkloadID)
+	endpoints := workload + "/endpoint"
+	endpoint := endpoints + key.EndpointID
+	return endpoint, []string{endpoints, workload}, nil
```
caseydavenport: OK, so the returned paths need to be ordered such that while iterating through, each path is always empty when being deleted.

Making a note to myself to make sure that this is documented somewhere :)

test/workload.yaml
```diff
@@ -10,8 +10,12 @@
   spec:
     interface: eth0
     profiles: [prof1, prof2]
-    ipNetworks: [1.2.3.4/32, "00:bb::aa/10"]
+    ipNetworks: ["1.2.3.4/32", "00:bb::aa/128"]
+    ipNATs:
+    - externalIP: 10.20.30.40
```
caseydavenport: Is this how we want to expose these?

What about `workloadIP` and `natIP` or something to make it more clear which IP address is which? Internal / external don't really make sense.

lib/api/workloadendpoint.go
```diff
@@ -60,10 +63,10 @@ type WorkloadEndpointSpec struct {
 	// internal IP will not have their source address changed, except when an endpoint attempts
 	// to connect one of its own external IPs. Each internal IP must be associated with the same
 	// endpoint via the configured IPNetworks.
-	IPNATs []IPNAT `json:"ipNAT,omitempty" validate:"omitempty,dive"`
+	IPNATs []IPNAT `json:"ipNATs,omitempty" validate:"omitempty,dive"`
```
caseydavenport: Can we have validation in place to ensure that any specified IPNATs match the specified IPNetworks?

calicoctl/commands/delete.go
```diff
@@ -51,6 +52,8 @@ Options:
   -f --filename=<FILENAME>     Filename to use to delete the resource.  If set to "-" loads from stdin.
   -t --tier=<TIER>             The policy tier.
   -n --hostname=<HOSTNAME>     The hostname.
+     --orchestrator=<ORCH>     The orchestrator (only used for workload endpoints).
+     --workload=<WORKLOAD>     The hostname.
```
robbrockbank: Oops.  I'll fix

calicoctl/resourcemgr/workloadendpoint.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/tigera/libcalico-go/lib/api/unversioned"
+	"github.com/tigera/libcalico-go/lib/client"
+)
+
+func init() {
+	registerResource(
+		api.NewWorkloadEndpoint(),
+		api.NewWorkloadEndpointList(),
+		[]string{"HOSTNAME", "ORCHESTRATOR", "WORKLOAD", "NAME"},
+		[]string{"HOSTNAME", "ORCHESTRATOR", "WORKLOAD", "NAME", "NETWORKS", "NATS", "INTERFACE", "PROFILES"},
```
robbrockbank: Yup.  I have a separate task to go through and rename everything in one hit.

lib/api/workloadendpoint.go
```diff
@@ -29,7 +31,8 @@ type WorkloadEndpoint struct {
 type WorkloadEndpointMetadata struct {
 	unversioned.ObjectMetadata
 
-	// The name of the endpoint.
+	// The name of the endpoint.  This may be omitted on a create, in which case an endpoint
+	// ID will be automatically created, and the endpoint ID will be included in the response.
```
robbrockbank: That is my plan.  I have a separate "renaming task".  I realize I could have just done that here too, but half of this code already existed so I was planning on doing that alongside the hostname->node and pool->ipPool stuff.  OK if I leave it out of this one?

lib/backend/etcd/etcd.go
```diff
@@ -132,7 +133,7 @@ func (c *EtcdClient) Apply(d *model.KVPair) (*model.KVPair, error) {
 
 // Delete an entry in the datastore.  This errors if the entry does not exists.
 func (c *EtcdClient) Delete(d *model.KVPair) error {
-	key, err := model.KeyToDefaultDeletePath(d.Key)
+	key, parents, err := model.KeyToDefaultDeletePaths(d.Key)
```
robbrockbank: You know I thought that as I was writing it and then I thought no because... blah... and now I can't think what blah is so I was probably thinking some nonsense or other.  I'll change - I think i'll make the interface much much cleaner!

lib/backend/model/workloadendpoint.go
robbrockbank: hahaha.  hmmm.  let me know when you find it.

lib/api/workloadendpoint.go
caseydavenport: Ok, cool.  So long as that's the plan I'm happy for it.  Is there an issue open?

lib/api/workloadendpoint.go
robbrockbank: (unless you're saying it's broken)

lib/backend/model/workloadendpoint.go
robbrockbank: by which I mean I'll make sure it's documented.

test/workload.yaml
robbrockbank: the resource is a workload endpoint so internal and external would be relative to the endpoint...

i'll do a little poll.

robbrockbank:
Imports no longer refer to tigera

lib/backend/etcd/etcd.go
robbrockbank: Done

lib/api/workloadendpoint.go
robbrockbank: Yup issue raised.  I'll work on it once I've spoken with Shaun ... just to let him know as it'll affect his stuff too.

test/workload.yaml
robbrockbank: current 4 person poll (me and you included) prefer internal/external (3:1).  I suggest leaving as is, and we can rename later if we need to.

caseydavenport:
LGTM.

PR #166, opened 2016-10-06
Update git hook to police repo location.

fasaxc:


git-hooks/pre-commit
```diff
@@ -10,6 +10,7 @@ exec 1>&2
 
 changed_files=$(git diff --cached --name-only)
 changed_go_files=$(git diff --cached --name-only | grep -E '.go$' || true)
+all_go_files=$(find -name '*.go' | grep -v /vendor/)
```
robbrockbank: I don't think it matters particularly, but this will get go files that aren't added into the repo.  But that sounds like another check we should possibly be making (i.e. all go files are part of the repo)

PR #168, opened 2016-10-07
Fix that set selectors had non-deterministic UUIDs.

fasaxc:
Replace map-based set with a binary search array.

### DIFF
--- a/lib/selector/parser/ast.go
+++ b/lib/selector/parser/ast.go
@@ -86,22 +86,22 @@ func (node LabelEqValueNode) collectFragments(fragments []string) []string {
 
 type LabelInSetNode struct {
 	LabelName string
-	Value     map[string]bool
+	Value     StringSet
 }
 
 func (node LabelInSetNode) Evaluate(labels map[string]string) bool {
 	if val, ok := labels[node.LabelName]; ok {
-		return node.Value[val]
+		return node.Value.Contains(val)
 	} else {
 		return false
 	}
 }
 
-func (node LabelInSetNode) collectFragments(fragments []string) []string {
+func collectInFragments(fragments []string, labelName, op string, values StringSet) []string {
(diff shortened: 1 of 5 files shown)


fasaxc:
@robbrockbank Please can you review?

lib/selector/parser/ast.go
```diff
@@ -86,22 +86,22 @@ func (node LabelEqValueNode) collectFragments(fragments []string) []string {
...
 	if val, ok := labels[node.LabelName]; ok {
-		return node.Value[val]
+		return node.Value.Contains(val)
 	} else {
 		return false
 	}
 }
 
-func (node LabelInSetNode) collectFragments(fragments []string) []string {
+func collectInFragments(fragments []string, labelName, op string, values StringSet) []string {
```
robbrockbank: What does the "In" refer to in the function name - is it the operator which is actually passed in as an argument?  just wondering if collectInFragments and collectFragments are the wrong way round.

lib/selector/parser/stringset.go
```diff
@@ -0,0 +1,44 @@
...
+type StringSet []string
+
+func (ss StringSet) Contains(s string) bool {
+	if len(ss) == 0 {
+		return false
+	}
+	min := 0
+	max := len(ss)
+	for min < (max - 1) {
+		partitionIdx := (min + max) / 2
```
robbrockbank: Does this work if the matching entry is the last one  in the set?  Since average will always round down I don't think the partition index will ever get as high as the top stop.

robbrockbank: As discussed a set of {1, 1, 1, 2} will have a different ID to  {1, 2} even though we're only checking when an entry is contained in the set.....

robbrockbank:
LGTM

PR #171, opened 2016-10-10
Fix delete bugs:  syntax and error handling

### LINKED ISSUE #170: `calicoctl delete -f <filename>` doesn't work
`calicoctl delete -f <filename>` does not work.  I'm seeing that it fails and the helptext says I need `calicoctl delete --filename=<filename>` instead (which does indeed work).

But this seems inconsistent with (say) `calicoctl create -f <filename>` which also works.

robbrockbank:
Fixes #170 

Fixes bug in doc string, and fixes error handling when objects do not exist.


fasaxc:
Not sure I know the context but looks fairly sane!

PR #175, opened 2016-10-11
Fix that Syncer could miss the first update.

fasaxc:
Golang API for etcdv2 differs from Python one, expects the current index, rather than the next.


robbrockbank:
LGTM

PR #176, opened 2016-10-12
Fix int accuracy of certain types, and fix bugs in marshalling

### LINKED ISSUE #137: resource help info doesn't correctly show "ports"
Ports needs to be a list, but the docs show them as comma separated string

### LINKED ISSUE #173: Compile fails on 32 bit platform(ARM) validator.go
The line in validator.go fails to compile on 32 bit ARM.
156: 	return f >= 0 && f <= 4294967295

I replaced the magic constant with math.MaxInt32 and added a math import to my version.
I'm not sure what a more appropriate fix might be.

robbrockbank:
Shaun can you review?

Tentatively Fixes #173 
by changing the ASNumber field to use an int64 rather than an int.

Fixes #137 

Fixes some bugs Lance was seeing with port and protocol not being marshalled to JSON correctly (incorrect logic in the Int32OrString type)

Note that I defined ASNumber as an num or string type.  Reason for this is we may eventually want to support dotted AS notation at some point and it feels to me that the format of the value we store in the datastore should match the format of the value we retrieve from the datastore.  WDYT - should we just store as an int32 and be done with it?  If we leave as is, I need to update some docs.

Note that I've limited the size of the various ints we use for the protocol / port and asNumber.  I could just take them all uint64 or int64.  WDYT?   I assume this will have knock on effects with your code.

Also - it feels the current representation of a port as a string or int may not be ideal -  I wonder if I should scrap the string and store as a range of ints?  WDYT?

### DIFF
--- a/docs/calicoctl/resources/bgppeer.md
+++ b/docs/calicoctl/resources/bgppeer.md
@@ -37,4 +37,4 @@ spec:
 #### Spec
 | name     | description                 | requirements               | schema  |
 |----------|-----------------------------|----------------------------|---------|
-| asNumber | The AS Number of this peer. | Must be a valid AS Number. | integer |
+| asNumber | The AS Number of this peer. | Must be a valid AS Number.  The YAML or JSON format may be a string and may be specified in dotted notation. | integer |
--- a/lib/numorstring/asnumber.go
+++ b/lib/numorstring/asnumber.go
@@ -23,6 +23,8 @@ import (
 
 type ASNumber uint32
 
+// ASNumberFromString creates an ASNumber struct from a string value.  The
+// string value may simply be a number or may be the ASN in dotted notation.
 func ASNumberFromString(s string) (ASNumber, error) {
 	if num, err := strconv.ParseUint(s, 10, 32); err == nil {
 		return ASNumber(num), nil
@@ -38,7 +40,7 @@ func ASNumberFromString(s string) (ASNumber, error) {
 	} else if num2, err := strconv.ParseUint(parts[1], 10, 16); err != nil {
 		return 0, errors.New("invalid AS Number format")
 	} else {
-		return ASNumber((num2 << 16) + num1), nil
+		return ASNumber((num1 << 16) + num2), nil
 	}
 }
 
--- a/lib/numorstring/asnumber.go
+++ b/lib/numorstring/asnumber.go
@@ -1,0 +1,67 @@
+// Copyright (c) 2016 Tigera, Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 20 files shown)


fasaxc:
Are we actually using string ASes anywhere?  Think I'd prefer to put the uint32 only in the backend datamodel then we don't need to handle both cases in confd templates etc.

I use a range of ints as the port primitive inside Felix because it's convenient to be uniform there.  Might be a bit verbose in YAML though?
```
ports: [{"min": 1, "max": 1}]
```
vs
```
ports: [1]
```

robbrockbank:
OK, so I've made ASN an int32 - the json unmarshmaling will handle a dot notation string.
I've made the ports a range, but it is marshalled and unmarshalled as a : separated value.

docs/calicoctl/resources/bgppeer.md
```diff
@@ -37,4 +37,4 @@ spec:
 #### Spec
 | name     | description                 | requirements               | schema  |
 |----------|-----------------------------|----------------------------|---------|
-| asNumber | The AS Number of this peer. | Must be a valid AS Number. | integer |
+| asNumber | The AS Number of this peer. | Must be a valid AS Number.  The YAML or JSON format may be a string and may be specified in dotted notation. | integer |
```
fasaxc: Does this work in markdown?  In RST, you have to change the length of all the other lines in the table.

lib/numorstring/asnumber.go
```diff
@@ -0,0 +1,67 @@
...
+	if len(parts) != 2 {
+		return 0, errors.New("invalid AS Number format")
+	}
+
+	if num1, err := strconv.ParseUint(parts[0], 10, 16); err != nil {
+		return 0, errors.New("invalid AS Number format")
+	} else if num2, err := strconv.ParseUint(parts[1], 10, 16); err != nil {
+		return 0, errors.New("invalid AS Number format")
+	} else {
+		return ASNumber((num2 << 16) + num1), nil
```
fasaxc: Are you shifting the right number here?  I thought `1234.5678` should map to `0x12345678`

lib/numorstring/uint8orstring.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2016 Tigera, Inc. All rights reserved.
+// Copyright (c) 208 Tigera, Inc. All rights reserved.
```
fasaxc: search+replace error ;-)

test/all.yaml
```diff
@@ -92,7 +92,7 @@
     hostname: fred
     peerIP: 1.3.5.6
   spec:
-    asNumber: 145794
+    asNumber: "1.1"
```
fasaxc: Good idea to test endianness by having different values int he two halves?

robbrockbank: I'm actually writing some UTs for this

docs/calicoctl/resources/bgppeer.md
robbrockbank: It works viewed in git (Which is currently all i'm worried about)

I ned to migrate this over to our docs - and they'll be a ton to fix up I figure - so i'm not too worried now, but I'd like the text at least to be accurate.

lib/numorstring/asnumber.go
robbrockbank: oops - nope and the comment about not having a symmetric x.x would have spotted that.  My bad.

As it happens I was actually in the middle of writing UTs so I woulda got there, but good spot!

PR #180, opened 2016-10-13
100% coverage of numorstring

robbrockbank:
Also tweaks the expect behavior of the tests - I think this makes more sense but yell if not.

And removes duplicated block


fasaxc:
LGTM, feel free to merge once it passes UT

PR #181, opened 2016-10-13
Common kv value

### LINKED ISSUE #179: Inconsistent use of pointers in backend API.
As discussed on Slack, the Syncer API returns KVPairs of the form: `KVPair{Key: k, Value: *v}`, where the rest of the API returns `*KVPair{Key: k, Value: v}`

This is causing me quite a bit of pain the work I'm doing at the moment.  

@fasaxc @robbrockbank leaving up to you guys to decide which you'd like it to be, but I'd like if we could make a decision with this _fast_ and the get on with it :)

I'm happy to take on whatever implementation work needs to be done.

robbrockbank:
Use common approach for Value in KVPair.

Value is always a pointer for structs, otherwise it is the actual element (string, boolean, slice, map)

Fixes #179

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,4 @@ bin
 release
 *.coverprofile
 vendor
-﻿calicoctl_build_container.created
+calicoctl_build_container.created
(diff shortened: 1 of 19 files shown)


fasaxc:
Please can you add comments to api.go to document what KVPairs contain?  I think that's probably where Casey started.

caseydavenport:
Looking to be about what I expected.  I'm going to merge :)

PR #184, opened 2016-10-14
Reduce log verbosity in etcd client set method.

fasaxc:
Avoids log spam in Felix when writing status reports.


robbrockbank:
cool - didn't know you could do context logs like that 👍 

LGTM

PR #188, opened 2016-10-17
Pool e2e tests + possible bug

gunjan5:
**New test structure**:

**Note**: 2 of the test cases are failing. I suspect it's a bug with IPIP substruct (IPIPConfiguration) always being stored as `nil`

**Test cases (Pool object e2e)**:
**Test 1**: Pass two fully populated PoolSpecs and expect the series of operations to succeed.
**Test 2**: Pass one partially populated PoolSpec and another fully populated PoolSpec and expect the series of operations to succeed.
**Test 3**: Pass one fully populated PoolSpec and another empty PoolSpec and expect the series of operations to succeed.
**Test 4**: Pass two fully populated PoolSpecs with two PoolMetadata (one IPv4 and another IPv6) and expect the series of operations to succeed.

**Series of operations each test goes through**:
Update meta1 - check for failure (because it doesn't exist).
Create meta1 with spec1.
Apply meta2 with spec2.
Get meta1 and meta2, compare spec1 and spec2.
Update meta1 with spec2.
Get meta1 compare spec2.
List (empty Meta) ... Get meta1 and meta2.
List (using Meta1) ... Get meta1.
Delete meta1.
Get meta1 ... fail.
Delete meta2.
List (empty Meta) ... Get no entries (should not error).

---

**Old tests**:
**Test cases (Pool object e2e)**:
**Test 1**: Create, Get and compare - expect returned pool to be the same as the input pool.
**Test 2**: Create, Apply, Get and compare - expect returned pool to be the same as the "Applied" one.
**Test 3**: Create, Update, Get and compare - expect returned pool to be the same as the "Updated" one.
**Test 4**: Create, Delete, Get and assert error.
**Test 5**: Apply, Get and compare (without creating) - expect returned pool to be the same as "Applied" one.
**Test 6**: Update, Get and assert error (without creating).
**Test 7**: Delete, Get and assert error (without creating).


gunjan5:
@robbrockbank rewrote the tests as per your suggestions, some tests are still failing because IPIP substruct is still a pointer (storing `nil` for `false`) which needs to be discussed.
But PTAL to see if anything major stands out that I need to change.

robbrockbank:
LGTM - only comment is there are number of !false and !true which should just be changed to true and false.

gunjan5:
@robbrockbank made those changes, but please don't merge it. I still have to talk to @caseydavenport about IPIPInterface being pointer - this could be a potential bug, tests are still failing (I'm investigating why CircleCI still says PASSing).

gunjan5:
prerequisite for this PR (and all the other e2e PRs) is [#215](https://github.com/projectcalico/libcalico-go/pull/215)

lib/client/pool_e2e_test.go
```diff
@@ -0,0 +1,226 @@
...
+
+			// Create a pool with meta1 and spec1.
+			_, outError = c.Pools().Create(&api.Pool{Metadata: meta1, Spec: spec1})
+
+			// Apply a pool with meta2 and spec2.
+			_, outError = c.Pools().Apply(&api.Pool{Metadata: meta2, Spec: spec2})
+
+			// Get pool with meta1.
+			outPool1, outError1 := c.Pools().Get(meta1)
+			log.Println("Out Pool object: ", outPool1)
```
robbrockbank: We should really be checking the error code after each of these calls - much like we would for code.

Personally, I'm not sure we need outError1, outError2, outError to be defined - I'd just stick with err and check it immediately after each call that we are validating.  But that's my preference - not asking you to make that change.

gunjan5: >We should really be checking the error code after each of these calls - much like we would for code.

What difference does it make if I assert the error right after vs in line 85? I think keeping them together makes it easier to understand each step we are taking for the FVs with a comment explaining what the expected result is for each step.

robbrockbank:
LGTM

PR #189, opened 2016-10-17
Policy e2e tests

gunjan5:
**New tests**:

**Test cases (Policy object e2e)**:
**Test 1**: Pass two fully populated PolicySpecs and expect the series of operations to succeed.
**Test 2**: Pass one fully populated PolicySpec and another empty PolicySpec and expect the series of operations to succeed.
**Test 3**: Pass one partially populated PolicySpec and another fully populated PolicySpec and expect the series of operations to succeed.

**Series of operations each test goes through**:
Update meta1 - check for failure (because it doesn't exist).
Create meta1 with spec1.
Apply meta2 with spec2.
Get meta1 and meta2, compare spec1 and spec2.
Update meta1 with spec2.
Get meta1 compare spec2.
List (empty Meta) ... Get meta1 and meta2.
List (using Meta1) ... Get meta1.
Delete meta1.
Get meta1 ... fail.
Delete meta2.
List (empty Meta) ... Get no entries (should not error).

---

**Old test cases**:

**Test cases (Policy object e2e)**:
**Test 1**: Create, Get and compare - expect returned policy to be the same as the input policy.
**Test 2**: Create, Apply, Get and compare - expect returned policy to be the same as the "Applied" one.
**Test 3**: Create, Update, Get and compare - expect returned policy to be the same as the "Updated" one.
**Test 4**: Create, Delete, Get and assert error.
**Test 5**: Apply, Get and compare (without creating) - expect returned policy to be the same as "Applied" one.
**Test 6**: Update, Get and assert error (without creating).
**Test 7**: Delete, Get and assert error (without creating).

### DIFF
--- a/lib/client/policy_e2e_test.go
+++ b/lib/client/policy_e2e_test.go
@@ -1,0 +1,204 @@
+// Copyright (c) 2016 Tigera, Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Test cases (Policy object e2e):
+// Test 1: Pass two fully populated PolicySpecs and expect the series of operations to succeed.
+// Test 2: Pass one fully populated PolicySpec and another empty PolicySpec and expect the series of operations to succeed.
+// Test 3: Pass one partially populated PolicySpec and another fully populated PolicySpec and expect the series of operations to succeed.
+
+// Series of operations each test goes through:
+// Update meta1 - check for failure (because it doesn't exist).
+// Create meta1 with spec1.
+// Apply meta2 with spec2.
+// Get meta1 and meta2, compare spec1 and spec2.
+// Update meta1 with spec2.
+// Get meta1 compare spec2.
+// List (empty Meta) ... Get meta1 and meta2.
+// List (using Meta1) ... Get meta1.
+// Delete meta1.
+// Get meta1 ... fail.
+// Delete meta2.
+// List (empty Meta) ... Get no entries (should not error).
+
+package client_test
+
+import (
+	"errors"
+	"log"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/ginkgo/extensions/table"
+	. "github.com/onsi/gomega"
+
+	"github.com/projectcalico/libcalico-go/lib/api"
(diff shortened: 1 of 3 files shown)


lib/client/policy_e2e_test.go
```diff
@@ -0,0 +1,206 @@
...
+			},
+			*createAPIPolicySpecObject("policy2", 22.222, "policy2-selector"),
+		),
+	)
+})
+
+// createAPIPolicySpecObject takes policy configuration options (name, order, selector),
+// creates 2 fixed set of ingress and egress rules (one with IPv4 and one with IPv6),
+// and composes & returns an api.PolicySpec object.
+func createAPIPolicySpecObject(name string, order float64, selector string) *api.PolicySpec {
```
robbrockbank: This method feels like an overkill - you're only allowing the name, order and selector to be configured and these are a tiny subset of the available fields.  By the time you extend this method to make it really useful, you'll find it'll be easier to just instantiate the structures directly.

I think you should just define upfront a few different PolicySpec objects and re-use them in your tests.

robbrockbank: Alternatively, you could define a set of Rules that you use in both the policy and profile tests.

lib/testutils/rules.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+var InRule2 = api.Rule{
+	Action:    "deny",
+	IPVersion: &ipv6,
+	Protocol:  &numProtocol1,
+	ICMP:      &icmp1,
+	Source: api.EntityRule{
+		Tag: "tag2",
+		Net: &cnet.IPNet{
```
robbrockbank: what's wrong with MustParseCIDR ?  it's in the test utils.

lib/testutils/rules.go
```diff
@@ -0,0 +1,96 @@
...
+var ipv4 = 4
+var ipv6 = 6
+var strProtocol1 = numorstring.ProtocolFromString("icmp")
+var strProtocol2 = numorstring.ProtocolFromString("udp")
+var numProtocol1 = numorstring.ProtocolFromInt(240)
+
+var icmpType1 = 100
+var icmpCode1 = 200
+
+var _, cidr1, _ = net.ParseCIDR("10.0.0.1/24")
```
robbrockbank: MustParseCIDR makes this easier.

gunjan5: yup changed it elsewhere, forgot to change it in testutils package itself

lib/client/policy_e2e_test.go
```diff
@@ -0,0 +1,204 @@
...
+			By("Updating the policy before it is created")
+			_, outError := c.Policies().Update(&api.Policy{Metadata: meta1, Spec: spec1})
+
+			// Should return an error.
+			Expect(outError.Error()).To(Equal(errors.New("resource does not exist: Policy(name=policy1)").Error()))
+
+			By("Create, Apply, Get and compare")
+
+			// Create a policy with meta1 and spec1.
+			_, outError = c.Policies().Create(&api.Policy{Metadata: meta1, Spec: spec1})
```
robbrockbank: This is true for other tests too.

robbrockbank:
lgtm

PR #206, opened 2016-10-20
Get should return a pointer to the resource

robbrockbank:


lib/backend/compat/compat.go
```diff
@@ -103,7 +103,7 @@ func (c *ModelAdaptor) Get(k model.Key) (*model.KVPair, error) {
 		}
 		d := model.KVPair{
 			Key: k,
-			Value: model.Profile{
+			Value: &model.Profile{
```
gunjan5: Line 111 needs to change to use `*model.Profile` as well.
https://github.com/projectcalico/libcalico-go/pull/206/files#diff-0244001a03450c8fe45ee4e5fa2c9a41R111

PR #209, opened 2016-10-20
WorkloadEndpoint e2e tests + bugfix

gunjan5:
**Test cases (WorkloadEndpoint object e2e)**:
**Test 1**: Pass two fully populated WorkloadEndpointSpecs and expect the series of operations to succeed.
**Test 2**: Pass one partially populated WorkloadEndpointSpec and another fully populated WorkloadEndpointSpec and expect the series of operations to succeed.
**Test 3**: Pass one fully populated WorkloadEndpointSpec and another empty WorkloadEndpointSpec and expect the series of operations to succeed.

### DIFF
--- a/lib/client/workloadendpoint.go
+++ b/lib/client/workloadendpoint.go
@@ -194,8 +194,16 @@ func (w *workloadEndpoints) convertKVPairToAPI(d *model.KVPair) (unversioned.Res
 	ah.Spec.InterfaceName = bh.Name
 	ah.Spec.MAC = bh.Mac
 	ah.Spec.Profiles = bh.ProfileIDs
-	ah.Spec.IPNetworks = nets
-	ah.Spec.IPNATs = nats
+	if len(nets) == 0 {
+		ah.Spec.IPNetworks = nil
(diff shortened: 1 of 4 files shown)


lib/client/workloadendpoint.go
```diff
@@ -194,8 +194,16 @@ func (w *workloadEndpoints) convertKVPairToAPI(d *model.KVPair) (unversioned.Res
 	ah.Spec.InterfaceName = bh.Name
 	ah.Spec.MAC = bh.Mac
 	ah.Spec.Profiles = bh.ProfileIDs
-	ah.Spec.IPNetworks = nets
-	ah.Spec.IPNATs = nats
+	if len(nets) == 0 {
```
robbrockbank: Why is this actually required?

gunjan5: because it returns empty slice instead of `nil` which is inconsistent from other fields (which return `nil` when they're not initialized). This is the one I discussed with Casey  yesterday

robbrockbank:
lgtm

PR #210, opened 2016-10-20
Profile e2e tests + bugfix

gunjan5:
**Test cases (Profile object e2e)**:
**Test 1**: Pass two fully populated ProfileSpecs and expect the series of operations to succeed.
**Test 2**: Pass one fully populated ProfileSpec and another empty ProfileSpec and expect the series of operations to succeed.
**Test 3**: Pass one partially populated ProfileSpec and another fully populated ProfileSpec and expect the series of operations to succeed.

Test case 2 fails (_fully populated ProfileSpec and another empty ProfileSpec_):
**Got:**
        `<api.ProfileSpec>: {IngressRules: nil, EgressRules: nil, Tags: []}`
**Wanted:**
       `<api.ProfileSpec>: {IngressRules: nil, EgressRules: nil, Tags: nil}`
** **This is fixed now** **


lib/client/profile_e2e_test.go
```diff
@@ -0,0 +1,240 @@
...
+			log.Println("Out Profile object: ", outProfile1)
+
+			// Get profile with meta2.
+			outProfile2, outError2 := c.Profiles().Get(meta2)
+			log.Println("Out Profile object: ", outProfile2)
+
+			// Should match spec1 & outProfile1 and outProfile2 & spec2 and errors to be nil.
+			Expect(outProfile1.Spec).To(Equal(spec1))
+			Expect(outProfile2.Spec).To(Equal(spec2))
+			Expect(outError1).To(BeNil())
```
robbrockbank: I would check the errors first as an unexpected error would certainly cause the specs not to be equal (since the return value will probably be uninitialized... or will perhaps panic since the outProfile1 and 2 are pointers and would be nil) - and in that case knowing what the error is is more important than knowing the Specs aren't equal.

robbrockbank: And I'd make sure we do that across all of the tests.

lib/client/profile_e2e_test.go
```diff
@@ -0,0 +1,240 @@
...
+			api.ProfileSpec{
+				Tags: []string{"profile1-tag1"},
+			},
+			*createAPIProfileSpecObject("profile2", []string{"profile2-tag1", "profile2-tag2"}),
+		),
+	)
+})
+
+// profileCreate takes client and api.Profile and creates a profile.
+func profileCreate(c *client.Client, p *api.Profile) (*api.Profile, error) {
```
robbrockbank: Do we need these functions?  It seems they just add a trace - but shouldn't we be expecting the return values in the actual test anyway - in which case we'd get the output we want from the test.

lib/client/profile_e2e_test.go
gunjan5: Not sure if I understand it correctly. 
Ginkgo prints what the outError was if it fails like, so we would know what outError was:
```bash
Expected
        error: "resource does not exist: ProfileTags(name=profile1)"
    to equal
        error: nil
```

One think I could do to be clearer is to use `Expect(outError1).NotTo(HaveOccurred())` instead? Not sure if the behavior changes that way but it's clearer.

lib/client/profile_e2e_test.go
```diff
@@ -0,0 +1,204 @@
...
+			},
+			*createAPIProfileSpecObject("profile2", []string{"profile2-tag1", "profile2-tag2"}),
+		),
+	)
+})
+
+// createAPIProfileSpecObject takes profile configuration options (name, order, selector),
+// creates 2 fixed set of ingress and egress rules (one with IPv4 and one with IPv6),
+// and composes & returns an api.ProfileSpec object.
+func createAPIProfileSpecObject(name string, tags []string) *api.ProfileSpec {
```
robbrockbank: See comment about this in the policy tests

lib/client/profile.go
```diff
@@ -133,7 +133,11 @@ func (h *profiles) convertKVPairToAPI(d *model.KVPair) (unversioned.Resource, er
 	ap.Metadata.Labels = bp.Labels
 	ap.Spec.IngressRules = rulesBackendToAPI(bp.Rules.InboundRules)
 	ap.Spec.EgressRules = rulesBackendToAPI(bp.Rules.OutboundRules)
-	ap.Spec.Tags = bp.Tags
+	if len(bp.Tags) == 0 {
```
robbrockbank: This is required because of the "opposite" code in compat.go::toTagsLabelsRules.

IIRC that code was required because the Python version of Felix didn't handle nil values of labels and tags.  Once we move over to golang Felix we should be able to get rid of the conversion code I assume.

robbrockbank: In the meantime, could you move the code that is in combat.go into the convertAPIToKVPair() method above - that way the forward and reverse conversions are in the same place.

robbrockbank:
One comment - which we could leave it's a pain to move the suggested code.

Then LGTM.  Could you rebase onto the latest master, then squash and merge.

gunjan5:
added label change to the tests, I'd rather do the compat.go changes in a separate PR

PR #211, opened 2016-10-20
ClaimAffinity bugfix + UTs

gunjan5:
**Bugs this patch should fix**: 
- Claim/Release claims/releases one block less than requested (off by one) 
- `largerThanBlock` always returns `true` for most cases
- ClaimAffinity doesn't return `failed` blocks when one IP is claimed in a block, also it returns wrong `claimed` blocks

**Test cases (ClaimAffinity)**:
**Test 1**: claim affinity for an unclaimed IPNet of size 64 - expect 1 claimed blocks, 0 failed and expect no error.
**Test 2**: claim affinity for an unclaimed IPNet of size smaller than 64 - expect 0 claimed blocks, 0 failed and expect an error error.

**Test 3**: claim affinity for a IPNet that has an IP already assigned to another host.
- Assign an IP with AssignIP to "host-A" from a configured pool - expect 0 claimed blocks, 0 failed and expect no error.
- Claim affinity for "Host-B" to the block that IP belongs to - expect 3 claimed blocks and 1 failed.

**Test 4**: claim affinity to a block twice from different hosts.
- Claim affinity to an unclaimed block for "Host-A" - expect 4 claimed blocks, 0 failed and expect no error.
- Claim affinity to the same block again but for "host-B" this time - expect 0 claimed blocks, 4 failed and expect no error.

### DIFF
--- a/lib/client/ipam_block.go
+++ b/lib/client/ipam_block.go
@@ -361,11 +361,10 @@ func getIPVersion(ip cnet.IP) ipVersion {
 	return ipv4
 }
 
-func largerThanBlock(blockCIDR cnet.IPNet) bool {
-	ones, bits := blockCIDR.Mask.Size()
-	prefixLength := bits - ones
+func largerThanOrEqualToBlock(blockCIDR cnet.IPNet) bool {
+	ones, _ := blockCIDR.Mask.Size()
 	ipVersion := getIPVersion(cnet.IP{blockCIDR.IP})
-	return prefixLength < ipVersion.BlockPrefixLength
+	return ones <= ipVersion.BlockPrefixLength
 }
 
 func intInSlice(searchInt int, slice []int) bool {
--- a/lib/client/ipam_test.go
+++ b/lib/client/ipam_test.go
@@ -193,10 +193,6 @@ var _ = Describe("IPAM tests", func() {
 		func(args testArgsClaimAff) {
 			inIPNet := testutils.MustParseCIDR(args.inNet)
 
-			if args.inNet == "" {
-				inIPNet = cnet.IPNet{}
-			}
-
 			// Wipe clean etcd, create a new client, and pools when cleanEnv flag is true.
 			if args.cleanEnv {
 				testutils.CleanEtcd()
--- a/lib/client/ipam_test.go
+++ b/lib/client/ipam_test.go
@@ -81,6 +81,15 @@ var etcdConfig = etcd.EtcdConfig{
 	EtcdEndpoints: "http://127.0.0.1:2379",
 }
 
+type testArgsClaimAff struct {
+	inNet, host                 string
+	cleanEnv                    bool
+	pool                        []string
+	assignIP                    net.IP
+	expClaimedIPs, expFailedIPs int
+	expError                    error
+}
+
 var _ = Describe("IPAM tests", func() {
 
 	DescribeTable("AutoAssign: requested IPs vs returned IPs",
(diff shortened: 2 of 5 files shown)


lib/client/ipam_block.go
```diff
@@ -361,11 +361,10 @@ func getIPVersion(ip cnet.IP) ipVersion {
 	return ipv4
 }
 
-func largerThanBlock(blockCIDR cnet.IPNet) bool {
-	ones, bits := blockCIDR.Mask.Size()
-	prefixLength := bits - ones
```
caseydavenport: What is wrong about this calculation of prefixLength?

lib/client/ipam_test.go
```diff
@@ -170,8 +180,86 @@ var _ = Describe("IPAM tests", func() {
...
+		// Test 4: claim affinity to a block twice from different hosts.
+		// - Claim affinity to an unclaimed block for "Host-A" - expect 4 claimed blocks, 0 failed and expect no error.
+		Entry("Claim affinity to an unclaimed block for Host-A", "10.0.0.0/24", "host-A", true, []string{"10.0.0.0/24", "fd80:24e2:f998:72d6::/120"}, net.IP{}, 4, 0, nil),
+
+		// - Claim affinity to the same block again but for "host-B" this time - expect 0 claimed blocks, 4 failed and expect no error.
+		Entry("Claim affinity to the same block again but for Host-B this time", "10.0.0.0/24", "host-B", false, []string{"10.0.0.0/24", "fd80:24e2:f998:72d6::/120"}, net.IP{}, 0, 4, nil),
+	)
 })
 
+// testIPAMClaimAffinity thakes inNet which is the CIDR to claim affinity to. It also takes host, poolSubnets to configure a new pool,
```
caseydavenport: typo: thakes

lib/client/ipam_test.go
```diff
@@ -170,8 +180,86 @@ var _ = Describe("IPAM tests", func() {
 		Entry("Assign 1 IPv4 address with AssignIP then try to release 2 IPs (release a second one)", net.ParseIP("192.168.1.1"), false, []string{"192.168.1.0/24", "fd80:24e2:f998:72d6::/120"}, net.IP{}, 0, []cnet.IP{cnet.IP{net.ParseIP("192.168.1.1")}}, nil),
 	)
 
+	DescribeTable("ClaimAffinity: claim IPNet vs actual number of blocks claimed",
+		func(inNet string, host string, cleanEnv bool, pool []string, assignIP net.IP, expClaimedIPs, expFailedIPs int, expError error) {
```
caseydavenport: In general, I think once you've got this many args to a function, it is better to wrap them up in a struct so that they're easier to parse.

This will make reading the individual test cases easier as well.

lib/client/ipam_test.go
```diff
@@ -170,8 +180,86 @@ var _ = Describe("IPAM tests", func() {
...
+	if cleanEnv {
+		testutils.CleanEtcd()
+		c, _ := testutils.NewClient("")
+		for _, v := range poolSubnet {
+			testutils.CreateNewPool(*c, v, false, false, true)
+		}
+	}
+	ic := setupIPAMClient(cleanEnv)
+
+	if len(assignIP) != 0 {
```
caseydavenport: I'm not sure why something with claimAffinity in the name is doing an IP address assignment - seems confusing to me.

lib/client/ipam_block.go
gunjan5: prefixLength is just `ones` when you do `bits - ones`, for example for subnet of `/24` it will return prefixLength of `8`  which is incorrect, it should be `24`
https://play.golang.org/p/D2a2bKhOdE

lib/client/ipam_test.go
gunjan5: hmm.. sounds like a good idea, but that also means: 
1. Creating custom struct type for every function under test, and dereferencing every variable.
2. Changing all the other tests to be consistent. 
What do you say? I'm ok with doing it either way.

lib/client/ipam_test.go
gunjan5: I agree this is rather hacky, but it is for the test case where one IP is assigned from a block to host-A and then host-B tries to claim all the blocks, and AssignIP needs to happen after creating the client and before claimingAffinity.

What I can do is make a utility function that gets called instead of having this AssignIP login inside claimAffinity function. 
Does that sound ok or you have something else in mind?

lib/client/ipam_block.go
caseydavenport: Gotcha.

lib/client/ipam_test.go
caseydavenport: > Changing all the other tests to be consistent. 

It definitely does not imply this :)  It only implies it for the current code review.

> Creating custom struct type for every function under test, and dereferencing every variable.

Not sure what you mean by this.  It turns this:

```
func(inNet string, host string, cleanEnv bool, pool []string, assignIP net.IP, expClaimedIPs, expFailedIPs int, expError error) {
```

Into this:

```
type TestArgs struct {
. . .
}

func(args TestArgs) {
```

lib/client/ipam_test.go
caseydavenport: Yeah - splitting this out into its own utility and calling from the main test function sounds right.

lib/client/ipam_test.go
gunjan5: that's what I mean, but I was saying I'd have to have different `type testArgs` for every test which is ok, but wouldn't it look weird to have it only for this test? I guess I can do it only for this one in this PR

caseydavenport: You shouldn't need different type TestArgs for each test case within a file, but its fine to have a TestArgs per test file.

I'm not really worried about it looking weird - this test has plenty of other stuff that isn't in any other test (it uses ClaimAffinity, for example) and there is absolutely no requirement that we update the other tests every time we add new ones.

lib/client/ipam_test.go
```diff
@@ -170,8 +189,73 @@ var _ = Describe("IPAM tests", func() {
...
+
+		// - Claim affinity to the same block again but for "host-B" this time - expect 0 claimed blocks, 4 failed and expect no error.
+		Entry("Claim affinity to the same block again but for Host-B this time", testArgsClaimAff{"10.0.0.0/24", "host-B", false, []string{"10.0.0.0/24", "fd80:24e2:f998:72d6::/120"}, net.IP{}, 0, 4, nil}),
+	)
 })
 
+// testIPAMClaimAffinity takes inNet which is the CIDR to claim affinity to. It also takes host, poolSubnets to configure a new pool,
+// when assignIP is not empty it will assign that IP and return zero-value for the return values, and cleanEnv is to clear the datastore etc.
+func testIPAMClaimAffinity(inNet net.IPNet, host string, poolSubnet []string, assignIP net.IP, cleanEnv bool) (claimed []cnet.IPNet, failed []cnet.IPNet, outErr error) {
+	if cleanEnv {
```
caseydavenport: A bit weird that this is here.  I think this belongs in the main Table function

lib/client/ipam_test.go
```diff
@@ -170,8 +189,73 @@ var _ = Describe("IPAM tests", func() {
...
+		testutils.CleanEtcd()
+		c, _ := testutils.NewClient("")
+		for _, v := range poolSubnet {
+			testutils.CreateNewPool(*c, v, false, false, true)
+		}
+	}
+	ic := setupIPAMClient(cleanEnv)
+
+	assignIPutil(ic, assignIP, "Host-A")
```
caseydavenport: Why are you splitting into a `testIPAMClaimAffinity` - can't you just put all this code into the main table test function? 

It feels like not that much occurs here that couldn't just be in that function - will make this a lot easier to follow.

lib/client/ipam_test.go
```diff
@@ -170,6 +189,63 @@ var _ = Describe("IPAM tests", func() {
 		Entry("Assign 1 IPv4 address with AssignIP then try to release 2 IPs (release a second one)", net.ParseIP("192.168.1.1"), false, []string{"192.168.1.0/24", "fd80:24e2:f998:72d6::/120"}, net.IP{}, 0, []cnet.IP{cnet.IP{net.ParseIP("192.168.1.1")}}, nil),
 	)
 
+	DescribeTable("ClaimAffinity: claim IPNet vs actual number of blocks claimed",
+		func(args testArgsClaimAff) {
+			inIPNet := testutils.MustParseCIDR(args.inNet)
+
+			if args.inNet == "" {
```
caseydavenport: Won't testutils.MustParseCIDR fail if inNet is blank?

gunjan5: just realized it's not even needed since AssignIP is a separate func now.

caseydavenport:
LGTM - @gunjan5 could you please squash and merge?

PR #212, opened 2016-10-20
Client Pool UTs + bug

gunjan5:
**Test cases**:
**API to KVPair**:
**Test 1**: IPv4 with non-default values for IPIP, NATOut, IPAM - assert the returned values, expect IPIPInterface to be "tunl0", and no error.
**Test 2**: IPv4 with default values for IPIP, NATOut, IPAM - assert the returned values, expect IPIPInterface to be "", and no error.
**Test 3**: IPv6 with non-default bools for IPIP, NATOut, IPAM - assert the returned values, expect IPIPInterface to be "tunl0", and no error.

**KVPair to API**:
**Test 1**: IPv4 with non-default values for masquerade, disabled and ipipInterface - assert the returned Metadata and Spec values, expect IPIP.Enabled to be true and no error.
**Test 2**: IPv4 with default values for masquerade, disabled and ipipInterface - assert the returned Metadata and Spec values, expect IPIP.Enabled to be false and no error.
**Test 3**: IPv6 with non-default bools for masquerade, disabled and ipipInterface - assert the returned Metadata and Spec values, expect IPIP.Enabled to be true and no error.

**Note**:  one of the test cases is failing which I think is a bug (also surfaced in pool e2e tests)

### DIFF
--- a/lib/client/pool_test.go
+++ b/lib/client/pool_test.go
@@ -13,218 +13,162 @@
 // limitations under the License.
 
 // Test cases:
-// Test 1: IPv4 with non-default values for IPIP, NATOut, IPAM - assert the returned Metadata and Spec values.
-// Test 2: IPv4 with default values for IPIP, NATOut, IPAM - assert the returned Metadata and Spec values.
-// Test 3: IPv6 with non-default bools for IPIP, NATOut, IPAM - assert the returned Metadata and Spec values.
+// API to KVPair:
+// Test 1: IPv4 with non-default values for IPIP, NATOut, IPAM - assert the returned values, expect IPIPInterface to be "tunl0", and no error.
+// Test 2: IPv4 with default values for IPIP, NATOut, IPAM - assert the returned values, expect IPIPInterface to be "", and no error.
+// Test 3: IPv6 with non-default bools for IPIP, NATOut, IPAM - assert the returned values, expect IPIPInterface to be "tunl0", and no error.
+
+// KVPair to API:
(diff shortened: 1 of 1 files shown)


lib/client/pool_test.go
```diff
@@ -0,0 +1,244 @@
...
+			client, _ := newClient("")
+
+			p := pools{
+				c: client,
+			}
+
+			// Call function under test with test arguments and get the result.
+			out, err := p.convertKVPairToAPI(&kvp)
+
+			// Assert output to the expected values.
```
robbrockbank: Can we assert the Spec and Metadata equal each other as well?  I'm worried that when we add a new field we may not remember to add these explicit checks - so having a struct comparison would avoid that ... perhaps there is a test utility that does this?

lib/client/pool_test.go
```diff
@@ -0,0 +1,244 @@
...
+			Disabled:              true,
+			expectedIPIPInterface: "tunl0",
+		},
+	}
+
+	// Go through all the values in table and test them one at a time.
+	for _, v := range table {
+		fmt.Fprintf(GinkgoWriter, "Testing APItoKVPair: %s\n", v.description)
+
+		It("", func() {
```
robbrockbank: Should this be an empty string?

robbrockbank: Should the contents of Printf be moved into the It?

robbrockbank:
Does this cover any code paths that would not be hit by the e2e style of test?   I would expect the e2e tests would cover the two conversion methods *and* also test the actual IPPool interface.

gunjan5:
Closing in favor of #188 (Pool FVs)

Issue #198, opened 2016-10-19
chown: Permission denied while running build-containerized or test-containerized targets


mihgen:
Running on Mac with Docker app Version 1.12.1 (build: 12133):
....

> total:										(statements)			95.8%
github.com/projectcalico/libcalico-go/lib/selector/selector.go:26:	Parse		0.0%
total:									(statements)	0.0%
github.com/projectcalico/libcalico-go/lib/selector/tokenizer/tokenizer.go:72:	Tokenize	68.8%
total:										(statements)	68.8%
chown: ./.git/objects/pack/pack-6c377b3d1e1b3092e0ccc1e93ba7582dff573781.idx: Permission denied
chown: ./.git/objects/pack/pack-6c377b3d1e1b3092e0ccc1e93ba7582dff573781.pack: Permission denied
make: *** [test-containerized] Error 1

From container:
> bash-4.3# ls -la ./.git/objects/pack/pack-6c377b3d1e1b3092e0ccc1e93ba7582dff573781.pack
-r--r--r--    1 root     root        703723 Oct 18 19:38 ./.git/objects/pack/pack-6c377b3d1e1b3092e0ccc1e93ba7582dff573781.pack

> bash-4.3# chown 502:20 ./.git/objects/pack/pack-6c377b3d1e1b3092e0ccc1e93ba7582dff573781.pack
chown: ./.git/objects/pack/pack-6c377b3d1e1b3092e0ccc1e93ba7582dff573781.pack: Permission denied

> bash-4.3# chmod +w ./.git/objects/pack/pack-6c377b3d1e1b3092e0ccc1e93ba7582dff573781.pack
bash-4.3# chown 502:20 ./.git/objects/pack/pack-6c377b3d1e1b3092e0ccc1e93ba7582dff573781.pack

> bash-4.3# ls -la ./.git/objects/pack/pack-6c377b3d1e1b3092e0ccc1e93ba7582dff573781.pack
-rw-r--r--    1 502      dialout     703723 Oct 18 19:38 ./.git/objects/pack/pack-6c377b3d1e1b3092e0ccc1e93ba7582dff573781.pack

I don't know why I had read-only access to git object. I didn't modify access rights, and just fetched repo from github (master branch).

[...]

PR #215, opened 2016-10-23
Fix: CircleCI always receives 0 Exit code for test-containerized

gunjan5:
```bash
bash -c 'make ut; \
	chown $(shell id -u):$(shell id -g) -R ./vendor'
``` 
if the tests fail in `make ut`, but second operation after the `;` succeeds, it returns an Exit code of `0` so CircleCI doesn't catch failed tests


gunjan5:
related to #198

robbrockbank:
lgtm

robbrockbank:
As discussed - ideally we want to always make sure we do the chown step to make sure the vendor directory is left in a decent state.  However - let's get this in and worry if it starts causing issues.  It's only an issue when the UTs are failing.

PR #217, opened 2016-10-25
Poll etcd's cluster ID to ensure we spot if the cluster is rebuilt.

fasaxc:
Closes an annoying corner case where the etcd cluster is rebuilt (for example in OpenStack where we have recommended running a transient etcd in a RAM disk).


robbrockbank:
LGTM

PR #219, opened 2016-10-25
Add missing DeleteSubtree() call to HWM tracker.

fasaxc:


robbrockbank:
lgtm although I don't really fully grok what's going on here.

PR #220, opened 2016-10-25
Change Syncer API to return an Update object with additional metadata.

fasaxc:
Felix uses the new metadata to count keys as they churn.

### DIFF
--- a/lib/backend/model/keys.go
+++ b/lib/backend/model/keys.go
@@ -87,21 +87,6 @@ type KVPair struct {
 	TTL      time.Duration // For writes, if non-zero, key has a TTL.
 }
 
-// Update from the Syncer.  A KV pair plus extra metadata.
-type Update struct {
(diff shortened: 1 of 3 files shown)


lib/backend/model/keys.go
```diff
@@ -87,6 +87,21 @@ type KVPair struct {
 	TTL      time.Duration // For writes, if non-zero, key has a TTL.
 }
 
+// Update from the Syncer.  A KV pair plus extra metadata.
+type Update struct {
```
robbrockbank: wdyt?

robbrockbank:
One comment/question/suggestion in line.  But nothing I feel super strongly about, so lgtm if you don't want to change it.

fasaxc:
Done, back to you @robbrockbank

robbrockbank:
lgtm

PR #221, opened 2016-10-25
Fixes and improvements to HWM tracker + UTs.

fasaxc:
- Add missing DeleteSubtree() call to HWM tracker.
- Handle tracking of nested deletions.
- Add UTs.

### DIFF
--- a/lib/hwm/hwm.go
+++ b/lib/hwm/hwm.go
@@ -45,7 +45,7 @@ func (trie *HighWatermarkTracker) StopTrackingDeletions() {
 	trie.deletionHwm = 0
 }
 
-func (trie *HighWatermarkTracker) StoreUpdate(key string, newModIdx uint64) uint64 {
+func (trie *HighWatermarkTracker) StoreUpdate(key string, newModIdx uint64) (oldModIdx uint64) {
 	prefix := keyToPrefix(key)
 	if trie.deletionHwms != nil {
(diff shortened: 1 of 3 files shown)


lib/hwm/hwm.go
```diff
@@ -120,6 +139,15 @@ func (trie *HighWatermarkTracker) DeleteOldKeys(hwmLimit uint64) []string {
 	return deletedKeys
 }
 
+func (trie *HighWatermarkTracker) ToMap() map[string]uint64 {
```
matthewdupre: It's sad that this function is only used by the tests.  Does it have to be here?

fasaxc: Alternative is to put the tests in the same package and poke into private state.  I think this is a bit cleaner.

caseydavenport: Hm, I thought the Go test philosophy was to include tests in the same package as the thing they tested anyway (though I could be wrong!)

fasaxc: There are two types of tests: you can put tests in the same package `foo`, say or in a special test package in the same directory `foo_test`.  The latter type test using your public API, so more of a black box test.  Both sorts of test are favoured.  In this case, it's borderline but adding a new (perfectly reasonable) public method seemed a lot cleaner than poking into the awkward internals.

robbrockbank:
lgtm - let's get this merged.

PR #229, opened 2016-10-31
Config and node

robbrockbank:
This PR adds a couple of related bits of function:

Ability to configure a node resource.  This is the configuration that the user may specify when creating a new Calico node and currently just consists of the node-specific BGP info (IPs and ASN).

Ability to configure system level defaults or low level node-specific config.  This includes global IPIP settings, node IPIP tunnel address, log level (I'm combining felix and BGP log levels for a simpler API), default ASN, Node-to-node mesh.


lib/client/config.go
```diff
@@ -0,0 +1,343 @@
...
+	return err
+}
+
+// SetDefaultASNumber gets the default AS Number used by the BGP agent running
+// on each node.  See SetDefaultASNumber for more details.
+func (c *config) GetDefaultASNumber() (uint64, error) {
+	if s, err := c.getValue(model.GlobalBGPConfigKey{Name: "as_num"}); err != nil {
+		return 0, err
+	} else if s == nil {
+		return 64511, nil
```
fasaxc: Is now a good time to fix this default?  64511 is in the wrong block IIRC.

It should at least be one const, I think.

lib/client/config.go
```diff
@@ -0,0 +1,343 @@
...
+	} else if err = ip.UnmarshalText([]byte(*s)); err != nil {
+		return nil, err
+	} else {
+		return ip, nil
+	}
+}
+
+// SetDefaultLogLevel sets the system default log level used by the node.  This
+// may be overridden on a per-node basis.
+func (c *config) SetDefaultLogLevel(level LogLevel) error {
```
fasaxc: Think I'd prefer Global instead of Default (since Default, to me, means "what you get if you don't have any value set").

fasaxc: Neil suggests writing the docs as an acid test

lib/backend/model/felixconfig.go
```diff
@@ -80,45 +78,12 @@ func (key GlobalConfigKey) String() string {
 	return fmt.Sprintf("GlobalConfig(name=%s)", key.Name)
 }
 
-type GlobalConfigListOptions struct {
```
robbrockbank: Reinstated

lib/backend/model/felixconfig.go
```diff
@@ -138,43 +107,5 @@ func (key HostConfigKey) valueType() reflect.Type {
 }
 
 func (key HostConfigKey) String() string {
-	return fmt.Sprintf("HostConfig(hostame=%s,name=%s)", key.Hostname, key.Name)
-}
-
-type HostConfigListOptions struct {
```
robbrockbank: Reinstated

lib/backend/model/hostip.go
```diff
@@ -1,50 +0,0 @@
...
-	"github.com/projectcalico/libcalico-go/lib/net"
-)
-
-var (
-	matchHostIp = regexp.MustCompile(`^/?calico/v1/host/([^/]+)/bird_ip`)
-	typeHostIp  = reflect.TypeOf(net.IP{})
-)
-
-// TODO find a place to put this
-type HostIPKey struct {
```
robbrockbank: Moved this to the node file - since it is now part of the node resource.

lib/backend/model/node.go
```diff
@@ -0,0 +1,228 @@
...
+	log "github.com/Sirupsen/logrus"
+	"github.com/projectcalico/libcalico-go/lib/errors"
+	"github.com/projectcalico/libcalico-go/lib/net"
+	"github.com/projectcalico/libcalico-go/lib/numorstring"
+)
+
+var (
+	typeNode          = reflect.TypeOf(Node{})
+	typeHostMetadata  = reflect.TypeOf(HostMetadata{})
+	typeHostAsn       = reflect.TypeOf(new(numorstring.ASNumber))
```
robbrockbank: yup fair point

lib/client/config.go
robbrockbank: Hmmm well it needs updates to the node container image - so probably not.

lib/backend/compat/compat.go
```diff
@@ -151,17 +322,70 @@ func toTagsLabelsRules(d *model.KVPair) (t, l, r *model.KVPair) {
...
-	if p.Labels == nil {
-		log.Info("Labels is nil - convert to empty map for backend")
-		l.Value = map[string]string{}
+	return t, l, r
+}
+
+// toNodeComponents converts a Node KVPair to separate KVPair types that make up the
+// Node.  This includes:
+// -  The host metadata (the primary component)
+// -  The host IPv4 address
```
robbrockbank: Not of this one - it's the one used by Felix for IPIP filtering.

lib/backend/etcd/etcd.go
```diff
@@ -300,3 +326,79 @@ func convertEtcdError(err error, key model.Key) error {
...
+		}
+	}
+
+	// No hostname specified, so enumerate the directories directly under
+	// the host tree, return no entries if the host directory does not exist.
+	log.Info("Listing all host metadatas")
+	key := "/calico/v1/host"
+	if results, err := c.etcdKeysAPI.Get(context.Background(), key, etcdListChildrenOpts); err != nil {
+		// If the root key does not exist - that's fine, return no list entries.
+		log.Info("Error enumerating host directories")
```
fasaxc: `log.WithErr(err).Info`?

lib/backend/etcd/etcd.go
```diff
@@ -300,3 +326,79 @@ func convertEtcdError(err error, key model.Key) error {
...
+		case errors.ErrorResourceDoesNotExist:
+			return []*model.KVPair{}, nil
+		default:
+			return nil, err
+		}
+	} else {
+		// TODO:  Since the host metadata is currently empty, we don't need
+		// to perform an additional get here, but in the future when the metadata
+		// may contain fields, we would need to perform a get.
+		log.Info("Parse host directories.")
```
fasaxc: I'd probably use `Debug` to avoid spam if something like Felix does a read, WDYT?

lib/backend/etcd/etcd.go
robbrockbank: i forgot about that.   i'll change

lib/backend/etcd/etcd.go
robbrockbank: sounds good!

PR #232, opened 2016-10-31
Add Kubernetes Datastore Driver

caseydavenport:
@fasaxc Looks like this PR also has a number of your yet unmerged fixes in it - hopefully you can spot which changes those are :)

It also appears to have a couple conflicts that I'll fix up when I get into the office in a little bit.

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -1,8 +1,15 @@
-hash: f789d5a91746d174a834e03fefc78c62d8a418753ab8e19ef6d2f595505c7db1
-updated: 2016-10-25T10:29:40.766893026+01:00
+hash: 99c9202c85f8f078481428fd55193e84c4e36119769e8792140d86954dd60726
+updated: 2016-10-31T14:24:41.90248438-07:00
 imports:
+- name: cloud.google.com/go
+  version: 3b1ae45394a234c385be014e9a488f2bb6eef821
+  subpackages:
+  - compute/metadata
+  - internal
+- name: github.com/blang/semver
+  version: 31b736133b98f26d5e078ec9eb591666edfd091f
 - name: github.com/coreos/etcd
-  version: 8e5f34fd97c03025c531fa547e654838ca490f04
+  version: ea057115224138376622d63a51b699133310ea31
   subpackages:
   - client
   - pkg/fileutil
@@ -10,24 +17,85 @@ imports:
   - pkg/tlsutil
   - pkg/transport
   - pkg/types
-- name: github.com/coreos/go-systemd
-  version: 61c7ee68bddeb1e66b491c3e1045c4f72b2cd6e3
+- name: github.com/coreos/go-oidc
+  version: 5644a2f50e2d2d5ba0b474bc5bc55fea1925936d
   subpackages:
+  - http
+  - jose
+  - key
+  - oauth2
+  - oidc
+- name: github.com/coreos/go-systemd
+  version: bfdc81d0d7e0fb19447b08571f63b774495251ce
+  subpackages:
+  - daemon
   - journal
+  - util
 - name: github.com/coreos/pkg
-  version: 3ac0863d7acf3bc44daf49afef8919af12f704ef
+  version: fa29b1d70f0beaddd4c7021607cc3c3be8ce94b8
   subpackages:
   - capnslog
+  - health
+  - httputil
+  - timeutil
+- name: github.com/davecgh/go-spew
+  version: 5215b55f46b2b919f50a1df0eaa5886afe4e3b3d
+  subpackages:
+  - spew
+- name: github.com/docker/distribution
+  version: cd27f179f2c10c5d300e6d09025b538c475b0d51
+  subpackages:
+  - digest
+  - reference
 - name: github.com/docker/engine-api
   version: 3d1601b9d2436a70b0dfc045a23f6503d19195df
 - name: github.com/docopt/docopt-go
   version: 784ddc588536785e7299f7272f39101f7faccc3f
+- name: github.com/emicklei/go-restful
+  version: 89ef8af493ab468a45a42bb0d89a06fccdd2fb22
+  subpackages:
+  - log
+  - swagger
 - name: github.com/ghodss/yaml
   version: bea76d6a4713e18b7f5321a2b020738552def3ea
+- name: github.com/go-openapi/jsonpointer
+  version: 46af16f9f7b149af66e5d1bd010e3574dc06de98
+- name: github.com/go-openapi/jsonreference
+  version: 13c6e3589ad90f49bd3e3bbe2c2cb3d7a4142272
+- name: github.com/go-openapi/spec
+  version: 6aced65f8501fe1217321abf0749d354824ba2ff
+- name: github.com/go-openapi/swag
+  version: 1d0bd113de87027671077d3c71eb3ac5d7dbba72
+- name: github.com/gogo/protobuf
+  version: e18d7aa8f8c624c915db340349aad4c49b10d173
+  subpackages:
+  - proto
+  - sortkeys
+- name: github.com/golang/glog
+  version: 44145f04b68cf362d9c4df2182967c2275eaefed
+- name: github.com/golang/protobuf
+  version: 8616e8ee5e20a1704615e6c8d7afcdac06087a67
+  subpackages:
+  - jsonpb
+  - proto
+- name: github.com/google/gofuzz
+  version: bbcb9da2d746f8bdbd6a936686a0a6067ada0ec5
+- name: github.com/howeyc/gopass
+  version: 3ca23474a7c7203e0a0a070fd33508f6efdb9b3d
+- name: github.com/imdario/mergo
+  version: 6633656539c1639d9d78127b7d47c622b5d7b6dc
+- name: github.com/jonboulle/clockwork
+  version: 2eee05ed794112d45db504eb05aa693efd2b8b09
+- name: github.com/juju/ratelimit
+  version: 77ed1c8a01217656d2080ad51981f6e99adaa177
 - name: github.com/kelseyhightower/envconfig
   version: 9aca109c9aec4633fced9717c4a09ecab3d33111
-- name: github.com/mattn/go-runewidth
-  version: 737072b4e32b7a5018b4a7125da8d12de90e8045
+- name: github.com/mailru/easyjson
+  version: d5b7844b561a7bc640052f1b935f7b800330d7e0
+  subpackages:
+  - buffer
+  - jlexer
+  - jwriter
 - name: github.com/mcuadros/go-version
   version: d52711f8d6bea8dc01efafdb68ad95a4e2606630
 - name: github.com/mitchellh/go-ps
@@ -35,10 +103,17 @@ imports:
 - name: github.com/olekukonko/tablewriter
   version: bdcc175572fd7abece6c831e643891b9331bc9e7
 - name: github.com/onsi/ginkgo
-  version: a8ccd817611666211fdf6e6a752174a0db8374a8
+  version: 74c678d97c305753605c338c6c78c49ec104b5e7
   subpackages:
   - config
   - extensions/table
+  - ginkgo
+  - ginkgo/convert
+  - ginkgo/interrupthandler
+  - ginkgo/nodot
+  - ginkgo/testrunner
+  - ginkgo/testsuite
+  - ginkgo/watch
   - internal/codelocation
   - internal/containernode
   - internal/failer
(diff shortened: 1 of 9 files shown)


lib/backend/compat/compat.go
```diff
@@ -143,12 +143,14 @@ func toTagsLabelsRules(d *model.KVPair) (t, l, r *model.KVPair) {
...
+		Key:      model.ProfileLabelsKey{pk},
+		Value:    p.Labels,
+		Revision: d.Revision,
 	}
 	r = &model.KVPair{
-		Key:   model.ProfileRulesKey{pk},
-		Value: &p.Rules,
+		Key:      model.ProfileRulesKey{pk},
+		Value:    &p.Rules,
+		Revision: d.Revision,
```
caseydavenport: Ah, here's a hack.

lib/backend/k8s/syncer.go
```diff
@@ -0,0 +1,536 @@
...
+	log.Info("Starting readSnapshot worker")
+	// Indicate we're starting a resync.
+	statusUpdates <- api.ResyncInProgress
+
+	// Perform an initial snapshot, and send the latest versions to the
+	// watcher routine.
+	initialVersions := resourceVersions{}
+	updateChan <- syn.performSnapshot(&initialVersions)
+
+	// Indicate we're in sync for the first time.
```
fasaxc: Thought there might be a race here if the InSync message gets delayed but looks like you avoid that because the channel isn't buffered. IIUC, that means the send is a rendezvous point for the two threads so you know the InSync was received before you send the initialVersions to the watcher.

Probably worth a comment that you're relying on statusUpdates being unbuffered?

lib/backend/k8s/syncer.go
```diff
@@ -0,0 +1,536 @@
...
+
+func (syn *kubeSyncer) Start() {
+	// Channel for receiving updates from a snapshot.
+	snapshotUpdates := make(chan *[]model.Update)
+
+	// Channel for receiving updates from the watcher.
+	watchUpdates := make(chan *model.Update)
+
+	// Channel used by the watcher to trigger a re-sync.
+	triggerResync := make(chan *resourceVersions, 5)
```
fasaxc: I recommend switching these from `*resourceVersions` to `resourceVersions` to rule out bugs caused by accidentally re-using the struct after sending it on the channel.

lib/backend/k8s/syncer.go
```diff
@@ -0,0 +1,536 @@
...
+	newStatus := api.WaitForDatastore
+	syn.callbacks.OnStatusUpdated(api.WaitForDatastore)
+	for {
+		select {
+		case updates = <-snapshotUpdates:
+			log.Debugf("Snapshot update: %+v", updates)
+			if currentStatus != api.ResyncInProgress {
+				panic("Recieved snapshot update while not resyncing")
+			}
+			syn.callbacks.OnUpdates(*updates)
```
fasaxc: I think you're missing handling of deletions here.  If you lose sync with k8s and then start a new snapshot, some pods may have been deleted in the meantime. You need to keep some sort of datastructure (say a set of all known Key objects) to spot that case.

You don't need a HWM tracker (a la etcd) because your resync algorithm is simpler.

lib/backend/k8s/syncer.go
```diff
@@ -0,0 +1,536 @@
...
+					continue
+				}
+				updateChan <- k
+				latestVersions.namespaceVersion = k.KVPair.Revision.(string)
+			}
+			continue
+		case event = <-poChan:
+			log.Debugf("Incoming Pod watch event. Type=%s", event.Type)
+			if needsResync = syn.eventTriggersResync(event); needsResync {
+				// We need to resync.  Break out of the inner for loop, back
```
fasaxc: I tink this actually breaks out of the select, which confused me the first time I used break within a select!

lib/backend/k8s/syncer.go
```diff
@@ -0,0 +1,536 @@
...
+	case watch.Deleted:
+		// For deletes, we need to nil out the Value part of the KVPair.
+		log.Debugf("Delete for pod %s/%s", pod.ObjectMeta.Namespace, pod.ObjectMeta.Name)
+		kvp.Value = nil
+
+		// Remove it from the cache, if it is there.
+		workload := kvp.Key.(model.WorkloadEndpointKey).WorkloadID
+		delete(labelCache, workload)
+	default:
+		// Adds and modifies are treated the same.  First, if the pod doesn't have an
```
fasaxc: Can the IP be removed?  If so, you need to treat removal like a deletion.

lib/backend/k8s/conversion.go
```diff
@@ -0,0 +1,359 @@
...
+		Key: model.WorkloadEndpointKey{
+			Hostname:       pod.Spec.NodeName,
+			OrchestratorID: "k8s",
+			WorkloadID:     workload,
+			EndpointID:     "eth0",
+		},
+		Value: &model.WorkloadEndpoint{
+			State:      "active",
+			Name:       interfaceName,
+			Mac:        &cnet.MAC{HardwareAddr: mac},
```
caseydavenport: Ooh, that's nice.

If it's omitted, then it just doesn't do the src MAC check in iptables?

glide.lock
```diff
@@ -1,97 +0,0 @@
-hash: f789d5a91746d174a834e03fefc78c62d8a418753ab8e19ef6d2f595505c7db1
```
caseydavenport: Nope.

lib/backend/k8s/conversion.go
```diff
@@ -0,0 +1,359 @@
...
+	return pod.Status.PodIP != ""
+}
+
+func (c converter) podToWorkloadEndpoint(pod *k8sapi.Pod) (*model.KVPair, error) {
+	// Pull out the profile and workload ID based on pod name and Namespace.
+	profile := fmt.Sprintf("k8s_ns.%s", pod.ObjectMeta.Namespace)
+	workload := fmt.Sprintf("%s.%s", pod.ObjectMeta.Namespace, pod.ObjectMeta.Name)
+
+	// If the pod doesn't have an IP address yet, then it hasn't gone through CNI.
+	// Treat this as if it didn't exist.
```
caseydavenport: Hmm, pretty sure this can't happen.

lib/backend/k8s/conversion.go
```diff
@@ -0,0 +1,359 @@
...
+
+	// Neither is defined - return an empty selector.
+	return ""
+}
+
+func (c converter) parsePolicyPort(port extensions.NetworkPolicyPort) []numorstring.Port {
+	if port.Port != nil {
+		p, err := numorstring.PortFromString(port.Port.String())
+		if err != nil {
+			panic(fmt.Sprintf("Invalid port %+v: %s", port.Port, err))
```
caseydavenport: Will do.

And yeah, pretty sure you can print out err with a %s formatter.

lib/backend/k8s/syncer.go
caseydavenport: @fasaxc does Felix treat a snapshot different than any other batch of updates it might receive?

I would have thought that Felix should treat a snapshot as a "this is the current state", wiping out any any state it's storing internally and replacing it with the snapshot (so if an update isn't in the snapshot, but was known by Felix before the snapshot, it is essentially deleted from Felix upon receiving the snapshot)

For the Kubernetes case, if a resource is deleted _during_ a snapshot, the Kubernetes API will inform us of any deletes that occurred when we start the watches below, since we start with the provided `resourceVersion` from the snapshot for each resource type.

From Felix's point of view this means it will get a snapshot with a bunch of stuff in it, and then subsequent modify / delete events immediately following.

If the snapshot takes so long that the resourceVersions are immediately stale, then a new snapshot will be triggered right away.

Am I missing something?

lib/backend/k8s/conversion.go
```diff
@@ -0,0 +1,359 @@
...
+	// If there no peers, or no ports, represent that as nil.
+	if len(peers) == 0 {
+		peers = []*extensions.NetworkPolicyPeer{nil}
+	}
+	if len(ports) == 0 {
+		ports = []*extensions.NetworkPolicyPort{nil}
+	}
+
+	// Combine desintations with sources to generate rules.
+	for _, port := range ports {
```
caseydavenport: I'm going to leave this as-is for now, since it seems to work and we're a bit crunched for time.  Let me know if you think there's a reason I should do it now instead.

caseydavenport:
@fasaxc OK - I fixed a number of comments and left a few questions for you on others.  Could you take another look - mostly just at the questions I had :)

lib/backend/k8s/conversion.go
fasaxc: Yep.

lib/backend/k8s/conversion.go
fasaxc: Happy to defer

lib/backend/k8s/syncer.go
fasaxc: It's the Syncer's job to understand the semantics of the snapshot and to generate the appropriate deletes.  Felix doesn't treat the OnUpdates call differently for a snapshot.  (The etcd driver doesn't even send one big update, it streams them out.)

lib/backend/k8s/syncer.go
```diff
@@ -67,12 +68,50 @@ func (syn *kubeSyncer) Start() {
...
+	syn.callbacks.OnUpdates(updates)
+	syn.updateTracker(updates)
+}
+
+// updateTracker updates the global object tracker with the given update.
+// updateTracker should be called after sending a update to the OnUpdates callback.
+func (syn *kubeSyncer) updateTracker(updates []api.Update) {
+	for _, upd := range updates {
+		if upd.UpdateType == api.UpdateTypeKVDeleted {
+			log.Debugf("Delete from tracker: %+v", upd.KVPair.Key)
```
fasaxc: FYI, you can just do `upd.Key` as a shortcut for `update.KVPair.Key`

lib/backend/k8s/syncer.go
```diff
@@ -67,12 +68,50 @@ func (syn *kubeSyncer) Start() {
...
+			delete(syn.tracker, upd.KVPair.Key)
+		} else {
+			log.Debugf("Update tracker: %+v: %+v", upd.KVPair.Key, upd.KVPair.Revision)
+			syn.tracker[upd.KVPair.Key] = upd.KVPair.Revision
+		}
+	}
+}
+
+// isUpdated returns true if the given KVPair is either new or has changed
+// since it was last seen, and false if it hasn't changed.
```
fasaxc: Method doesn't look to be used, but I don't think the comment matches the method.  You're returning true if the key is new or if it has the same value as before; false iff the key is known and the value is different.

lib/backend/k8s/syncer.go
```diff
@@ -174,16 +218,42 @@ func (syn *kubeSyncer) readFromKubernetesAPI() {
...
+		return api.UpdateTypeKVUpdated
+	}
+}
+
+func (syn *kubeSyncer) performSnapshotDeletes(exists map[model.Key]bool) {
+	log.Info("Checking for any deletes for snapshot")
+	deletes := []api.Update{}
+	for cachedKey, _ := range syn.tracker {
+		// Check each cached key to see if it exists in the snapshot.  If it doesn't,
+		// we need to send a delete for it.
```
fasaxc: You also need to delete it from the tracker or replace the tracker completely with the new map.

lib/backend/k8s/syncer.go
caseydavenport: I think I'd prefer to leave it as `update.KVPair.Key` - it's less obscured.

lib/backend/k8s/syncer.go
caseydavenport: Yeah, that's a straggler from an older iteration - will remove!

lib/backend/k8s/syncer.go
caseydavenport: The way I've got it right now, that always occurs in `sendUpdates`. 

Tried to push all the tracker modification / sending of updates through one place so that the behavior is always the same.  

This method just builds the set of delete updates to send.

caseydavenport:
@fasaxc OK - I think I've addressed all your comments! Let me know if I'm wrong :)

caseydavenport:
@fasaxc 

So, I think Felix is still expecting the endpoint to have a mac address in at least one place.  I'm hitting this in my most recent tests:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x4c67f0]

goroutine 27 [running]:
panic(0x14968e0, 0xc42000c070)
        /usr/local/go/src/runtime/panic.go:500 +0x1a1
github.com/projectcalico/felix/go/felix/calc.(*EventBuffer).OnEndpointTierUpdate(0xc420352f80, 0x1f03e40, 0xc4204a4640, 0x13c9aa0, 0xc42033c300, 0x1f754d0, 0x0, 0x0)
        /go/src/github.com/projectcalico/felix/go/felix/calc/event_buffer.go:274 +0x880
github.com/projectcalico/felix/go/felix/calc.(*PolicyResolver).sendEndpointUpdate(0xc4200bd2d0, 0x15753e0, 0xc4204a4640, 0xc4205ffc90, 0xc4205ffc88)
        /go/src/github.com/projectcalico/felix/go/felix/calc/policy_resolver.go:166 +0x793
github.com/projectcalico/felix/go/felix/calc.(*PolicyResolver).(github.com/projectcalico/felix/go/felix/calc.sendEndpointUpdate)-fm(0x15753e0, 0xc4204a4640, 0xc4205ffc70, 0xffffffffffffffff)
        /go/src/github.com/projectcalico/felix/go/felix/calc/policy_resolver.go:131 +0x3e
github.com/projectcalico/felix/go/felix/set.mapSet.Iter(0xc420212cc0, 0xc420232ff0)
        /go/src/github.com/projectcalico/felix/go/felix/set/set.go:61 +0x9c
```

Is this something I should just fix up? Or was this on your radar?

On a similar note, I was planning on removing (and already have removed) all of the mac stuff from the k8s backend.  I'm assuming that this is OK and we're not actually losing any "features" by doing so?

fasaxc:
Nope, that looks like a bug, I'll fix up.

lib/numorstring/uint8orstring.go
```diff
@@ -54,10 +54,10 @@ func (i *Uint8OrString) UnmarshalJSON(b []byte) error {
 
 // MarshalJSON implements the json.Marshaller interface.
 func (i Uint8OrString) MarshalJSON() ([]byte, error) {
-	if num, err := i.NumValue(); err == nil {
-		return json.Marshal(num)
-	} else {
+	if i.Type == NumOrStringString {
```
fasaxc: What's this change for?  I think @robbrockbank's version is intended to normalise `{"foo": "123"}` to `{"foo": 123}`.  Is there somewhere that you don't want that behaviour?

He says there might be some subtleties with the programatic API so probably worth touching base with him.

caseydavenport: Will do - I forgot that I made this change, it was a loooong time ago.  Can't recall immediately why but it was causing me some trouble.

lib/backend/k8s/syncer.go
```diff
@@ -28,17 +28,21 @@ import (
 )
 
 func newSyncer(kc KubeClient, callbacks api.SyncerCallbacks) *kubeSyncer {
-	return &kubeSyncer{
+	syn := &kubeSyncer{
 		kc:        kc,
 		callbacks: callbacks,
 	}
+	syn.tracker = map[model.Key]interface{}{}
```
fasaxc: Any reason not to init these inline in the struct?

caseydavenport: Just because I'm silly and didn't think of it.

PR #236, opened 2016-10-31
validate port range to be between 0 and 65536

### LINKED ISSUE #223: Port numbers not validated correctly
calicoctl is accepting port ranges and lists containing port 0 and port 65536.

gunjan5:
Fixes #223


robbrockbank:
LGTM.  Can you rebase and merge?

robbrockbank:
Actually .... the port field is a uint16 -- can it really be > 65535?

Also this needs rebasing.

lib/validator/validator.go
```diff
@@ -178,6 +178,10 @@ func validatePort(v *validator.Validate, structLevel *validator.StructLevel) {
 	if p.MinPort > p.MaxPort {
 		structLevel.ReportError(reflect.ValueOf(p.MaxPort), "Port", "port", "port range invalid")
 	}
+
+	if p.MinPort < 1 || p.MaxPort < 1 {
```
caseydavenport: This should (I know, I know) probably have a comment which says we don't need to check the upper bound because we use a uint16.

caseydavenport:
One _comment_ then LGTM :)

PR #241, opened 2016-11-02
host affinity backwards compatiblity

### LINKED ISSUE #226: Mismatched affinity representation with Python libcalico
As discussed here: https://github.com/projectcalico/calico-containers/issues/1194#issuecomment-256553868

Looks like the Go libcalico uses `hostAffinity: hostname` whereas the Python libcalico uses `affinity: host=hostname`.

This is pretty bad, and doubly so since the CNI IPAM plugin has been writing data in the former way since v1.4.0.  Changing this will likely have compatibility problems.

gunjan5:
Fixes #226

### DIFF
--- a/lib/backend/compat/compat.go
+++ b/lib/backend/compat/compat.go
@@ -30,9 +30,9 @@ type ModelAdaptor struct {
 
 var _ api.Client = (*ModelAdaptor)(nil)
 
-// affinedTo variable represents what the resource (host in
+// affineTo variable represents what the resource (host in
 // this case) is affined to.
-var affinedTo = "host"
+var affineTo = "host"
 
 func NewAdaptor(c api.Client) *ModelAdaptor {
 	return &ModelAdaptor{client: c}
@@ -189,7 +189,7 @@ func (c *ModelAdaptor) Get(k model.Key) (*model.KVPair, error) {
 	case model.NodeKey:
 		return c.getNode(kt)
 	case model.BlockKey:
-		return c.getBlockAffinity(k)
+		return c.getBlock(k)
 	default:
 		return c.client.Get(k)
 	}
@@ -202,7 +202,7 @@ func (c *ModelAdaptor) List(l model.ListInterface) ([]*model.KVPair, error) {
 	case model.NodeListOptions:
 		return c.listNodes(lt)
 	case model.BlockListOptions:
-		return c.listBlockAffinity(lt)
+		return c.listBlock(lt)
 	default:
 		return c.client.List(l)
 	}
@@ -239,10 +239,10 @@ func (c *ModelAdaptor) getProfile(k model.Key) (*model.KVPair, error) {
 	return &d, nil
 }
 
-// getBlockAffinity gets KVPair for Block. It gets the block value first,
+// getBlock gets KVPair for Block. It gets the block value first,
 // then checks for `Affinity` field first, then `HostAffinity` as a backup.
 // For more details see: https://github.com/projectcalico/libcalico-go/issues/226
-func (c *ModelAdaptor) getBlockAffinity(k model.Key) (*model.KVPair, error) {
+func (c *ModelAdaptor) getBlock(k model.Key) (*model.KVPair, error) {
 	bk := k.(model.BlockKey)
 
 	v, err := c.client.Get(model.BlockKey{CIDR: bk.CIDR})
@@ -254,16 +254,17 @@ func (c *ModelAdaptor) getBlockAffinity(k model.Key) (*model.KVPair, error) {
 	// Check for `Affinity` field first (this is to make sure we're
 	// compatible with Python version etcd data-model).
 	if val.Affinity == nil {
-
 		// Check for hostAffinity if Affinity is nil.
 		if val.HostAffinity != nil {
 			// Convert HostAffinity=hostname into Affinity=host:hostname format.
-			hostAffinityStr := affinedTo + ":" + *val.HostAffinity
+			hostAffinityStr := affineTo + ":" + *val.HostAffinity
 			val.Affinity = &hostAffinityStr
+
+			// Set AllocationBlock.HostAffinity to nil so it's never non-nil for the clients.
+			val.HostAffinity = nil
 		}
 	}
 	return &model.KVPair{Key: v.Key, Value: val}, nil
-
 }
 
 // getNode gets the composite node by getting the individual components
@@ -290,7 +291,7 @@ func (c *ModelAdaptor) getNode(nk model.NodeKey) (*model.KVPair, error) {
 // make sure the deprecated HostAffinity field is not used.
 func validateBlockValue(kvp *model.KVPair) error {
 	if kvp.Value.(*model.AllocationBlock).HostAffinity != nil {
-		return goerrors.New("HostAffinity field is deprecated, use Affinity")
+		return goerrors.New("AllocationBlock.HostAffinity is deprecated, please use Affinity instead.")
 	}
 	return nil
 }
@@ -327,9 +328,9 @@ func (c *ModelAdaptor) listNodes(l model.NodeListOptions) ([]*model.KVPair, erro
 	return results, nil
 }
 
-// listBlockAffinity returns list of KVPairs for Block, includes making sure
-// backwards compatiblity. See getBlockAffinity for more details.
-func (c *ModelAdaptor) listBlockAffinity(l model.BlockListOptions) ([]*model.KVPair, error) {
+// listBlock returns list of KVPairs for Block, includes making sure
+// backwards compatiblity. See getBlock for more details.
+func (c *ModelAdaptor) listBlock(l model.BlockListOptions) ([]*model.KVPair, error) {
 
 	// Get a list of block KVPairs.
 	bal, err := c.client.List(l)
@@ -342,9 +343,9 @@ func (c *ModelAdaptor) listBlockAffinity(l model.BlockListOptions) ([]*model.KVP
 
 	// Go through the list to make sure Affinity field has a proper value,
 	// and map the value to Affinity if deprecated HostAffinity field is used
-	// by calling getBlockAffinity, and populate the KVPair slice to return.
+	// by calling getBlock, and populate the KVPair slice to return.
 	for idx, bakv := range bal {
-		kvp, err := c.getBlockAffinity(bakv.Key)
+		kvp, err := c.getBlock(bakv.Key)
 		if err != nil {
 			return nil, err
 		}
--- a/lib/backend/compat/compat.go
+++ b/lib/backend/compat/compat.go
@@ -30,10 +30,6 @@ type ModelAdaptor struct {
 
 var _ api.Client = (*ModelAdaptor)(nil)
 
-// affineTo variable represents what the resource (host in
-// this case) is affined to.
-var affineTo = "host"
-
 func NewAdaptor(c api.Client) *ModelAdaptor {
 	return &ModelAdaptor{client: c}
 }
@@ -257,7 +253,7 @@ func (c *ModelAdaptor) getBlock(k model.Key) (*model.KVPair, error) {
 		// Check for hostAffinity if Affinity is nil.
 		if val.HostAffinity != nil {
 			// Convert HostAffinity=hostname into Affinity=host:hostname format.
-			hostAffinityStr := affineTo + ":" + *val.HostAffinity
+			hostAffinityStr := "host:" + *val.HostAffinity
 			val.Affinity = &hostAffinityStr
 
 			// Set AllocationBlock.HostAffinity to nil so it's never non-nil for the clients.
--- a/lib/backend/compat/compat.go
+++ b/lib/backend/compat/compat.go
@@ -15,6 +15,8 @@
 package compat
 
 import (
+	goerrors "errors"
+
 	"github.com/projectcalico/libcalico-go/lib/backend/api"
 	"github.com/projectcalico/libcalico-go/lib/backend/model"
 	"github.com/projectcalico/libcalico-go/lib/errors"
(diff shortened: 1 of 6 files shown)


gunjan5:
@caseydavenport I need to fix a panic (that I plan to get to in the morning ☕️ ) but you can start reviewing if you want.

gunjan5:
this is what it looks like in etcd now:
```bash
$ etcdctl get /calico/ipam/v2/assignment/ipv4/block/10.0.0.64-26
{"cidr":"10.0.0.64/26","affinity":"host:host-A","strictAffinity":false,"allocations":[0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0],"unallocated":[],"attributes":[{"handle_id":null,"secondary":null}]}
```

lib/backend/compat/compat.go
```diff
@@ -199,6 +239,33 @@ func (c *ModelAdaptor) getProfile(k model.Key) (*model.KVPair, error) {
 	return &d, nil
 }
 
+// getBlockAffinity gets KVPair for Block. It gets the block value first,
+// then checks for `Affinity` field first, then `HostAffinity` as a backup.
+// For more details see: https://github.com/projectcalico/libcalico-go/issues/226
+func (c *ModelAdaptor) getBlockAffinity(k model.Key) (*model.KVPair, error) {
```
caseydavenport: I think this should just be called `getBlock`

A `blockAffinity` is a different thing :)

lib/backend/compat/compat.go
```diff
@@ -199,6 +239,33 @@ func (c *ModelAdaptor) getProfile(k model.Key) (*model.KVPair, error) {
...
+	v, err := c.client.Get(model.BlockKey{CIDR: bk.CIDR})
+	if err != nil {
+		return nil, err
+	}
+	val := v.Value.(*model.AllocationBlock)
+
+	// Check for `Affinity` field first (this is to make sure we're
+	// compatible with Python version etcd data-model).
+	if val.Affinity == nil {
+
```
caseydavenport: Extra newline here too

lib/backend/compat/compat.go
```diff
@@ -199,6 +239,33 @@ func (c *ModelAdaptor) getProfile(k model.Key) (*model.KVPair, error) {
...
+		// Check for hostAffinity if Affinity is nil.
+		if val.HostAffinity != nil {
+			// Convert HostAffinity=hostname into Affinity=host:hostname format.
+			hostAffinityStr := affinedTo + ":" + *val.HostAffinity
```
caseydavenport: You should rename `affinedTo` to `affineTo`.

lib/backend/compat/compat.go
```diff
@@ -199,6 +239,33 @@ func (c *ModelAdaptor) getProfile(k model.Key) (*model.KVPair, error) {
...
+			val.Affinity = &hostAffinityStr
```
caseydavenport: I think we also want to do `val.HostAffinity = nil` here so that it's never non-nil for clients.

lib/backend/compat/compat.go
```diff
@@ -251,6 +327,34 @@ func (c *ModelAdaptor) listNodes(l model.NodeListOptions) ([]*model.KVPair, erro
 	return results, nil
 }
 
+// listBlockAffinity returns list of KVPairs for Block, includes making sure
+// backwards compatiblity. See getBlockAffinity for more details.
+func (c *ModelAdaptor) listBlockAffinity(l model.BlockListOptions) ([]*model.KVPair, error) {
```
caseydavenport: Same rename here - should be `listBlocks`

lib/backend/compat/compat.go
```diff
@@ -219,6 +286,15 @@ func (c *ModelAdaptor) getNode(nk model.NodeKey) (*model.KVPair, error) {
 	return &model.KVPair{Key: nk, Value: &nv}, nil
 }
 
+// validateBlockValue validates the AllocationBlock fields (specifically Affinity) to
+// make sure the deprecated HostAffinity field is not used.
+func validateBlockValue(kvp *model.KVPair) error {
+	if kvp.Value.(*model.AllocationBlock).HostAffinity != nil {
+		return goerrors.New("HostAffinity field is deprecated, use Affinity")
```
caseydavenport: Tweak the wording of this error.

"_AllocationBlock_.HostAffinity is deprecated, _please use_ Affinity _instead_."

lib/client/ipam.go
```diff
@@ -486,7 +486,7 @@ func (c ipams) assignFromExistingBlock(
 
 		// Update the block using CAS by passing back the original
 		// KVPair.
-		obj.Value = &b.AllocationBlock
+		obj.Value = b.AllocationBlock
```
caseydavenport: You should undo this change.  It's currently expected that KVPair.Value fields are pointers since https://github.com/projectcalico/libcalico-go/pull/181

lib/client/ipam_block.go
```diff
@@ -77,9 +77,9 @@ func (b *allocationBlock) autoAssign(
 
 	// Determine if we need to check for affinity.
 	checkAffinity := b.StrictAffinity || affinityCheck
-	if checkAffinity && b.HostAffinity != nil && host != *b.HostAffinity {
+	if checkAffinity && b.Affinity != nil && host != *b.Affinity {
```
caseydavenport: This isn't going to work.

`host` is still going to be of the form `hostname`, and `b.Affinity` is going to be of the form `host: hostname`

You probably want a small helper function for all of these cases.  e.g

```
// hostAffinityMatches returns true if the provided host matches the provided affinity.
func hostAffinityMatches(host, affinity) bool {
}
```

lib/client/ipam_block_reader_writer.go
```diff
@@ -30,6 +30,10 @@ import (
 	cnet "github.com/projectcalico/libcalico-go/lib/net"
 )
 
+// affinedTo variable represents what the resource (host in
+// this case) is affined to.
+var affinedTo = "host"
```
caseydavenport: Don't think there's much point in having this defined like this -  when we do add more types of affinity, it's going to be way more significant a change than just changing the value of a constant.

Let's just inline the "host" string.

lib/client/ipam_block_reader_writer.go
```diff
@@ -194,9 +205,16 @@ func (rw blockReaderWriter) releaseBlockAffinity(host string, blockCIDR cnet.IPN
...
+		if host == "" {
+			log.Errorf("Hostname can't be empty")
+			return goerrors.New("Hostname must be sepcified to release block affinity")
+		}
+		affinityKeyStr := affinedTo + ":" + host
+
 		// Check that the block affinity matches the given affinity.
-		if b.HostAffinity != nil && *b.HostAffinity != host {
-			log.Errorf("Mismatched affinity: %s != %s", *b.HostAffinity, host)
+		if b.Affinity != nil && *b.Affinity != affinityKeyStr {
```
caseydavenport: This would also use the helper mentioned above.

lib/backend/compat/compat.go
gunjan5: I don't think it's a good idea to put "host" inline, it would be difficult to understand for anyone reading this code as in why we do that. That's why I stick a comment where I'm declaring the variable to explain what it is.

lib/backend/compat/compat.go
gunjan5: hmmm using "please" inside an error value. Interesting 🤔 ...

lib/backend/compat/compat.go
caseydavenport: It's especially weird because we've got two constants in two different places defined to the same thing.  I think it is more clear for people reading the code to see what value is actually being put there, rather than having to go look at that variable to see what it is.

lib/client/ipam.go
gunjan5: It is already a pointer since `b.AllocationBlock` is a pointer.
`fmt.Printf("%T", obj.Value)` prints `*model.AllocationBlock`

gunjan5:
@caseydavenport made those changes except inlining `host`, I'm still in favor of not inlining it for code readability PoV, but I can be convinced otherwise.

lib/client/ipam_block.go
```diff
@@ -133,6 +133,12 @@ func (b *allocationBlock) assign(address cnet.IP, handleID *string, attrs map[st
 	return nil
 }
 
+// hostAffinityMatches checks if the provided host matches the provided affinity.
+func hostAffinityMatches(host string, block *model.AllocationBlock) bool {
+	affinityKeyStr := affinedTo + ":" + host
+	return *block.Affinity != affinityKeyStr
```
caseydavenport: I think you want this to be `block.Affinity == affinityKeyStr` right?

lib/client/ipam_block_reader_writer.go
```diff
@@ -210,11 +210,10 @@ func (rw blockReaderWriter) releaseBlockAffinity(host string, blockCIDR cnet.IPN
 			log.Errorf("Hostname can't be empty")
 			return goerrors.New("Hostname must be sepcified to release block affinity")
 		}
-		affinityKeyStr := affinedTo + ":" + host
 
 		// Check that the block affinity matches the given affinity.
-		if b.Affinity != nil && *b.Affinity != affinityKeyStr {
-			log.Errorf("Mismatched affinity: %s != %s", *b.Affinity, affinityKeyStr)
+		if b.Affinity != nil && hostAffinityMatches(host, b.AllocationBlock) {
+			log.Errorf("Mismatched affinity: %s != %s", *b.Affinity, affinedTo+":"+host)
```
caseydavenport: Weird spacing around the `+`

gunjan5: weird gofmt didn't fix this

gunjan5:
@caseydavenport made those changes including inlining "host:"
PTAL

lib/backend/compat/compat.go
```diff
@@ -199,6 +235,34 @@ func (c *ModelAdaptor) getProfile(k model.Key) (*model.KVPair, error) {
...
+
+	v, err := c.client.Get(model.BlockKey{CIDR: bk.CIDR})
+	if err != nil {
+		return nil, err
+	}
+	val := v.Value.(*model.AllocationBlock)
+
+	// Check for `Affinity` field first (this is to make sure we're
+	// compatible with Python version etcd data-model).
+	if val.Affinity == nil {
```
caseydavenport: Might want to consider flattening this into a single `if`

`if val.Affinity == nil && val.HostAffinity != nil`

lib/backend/compat/compat.go
```diff
@@ -251,6 +324,34 @@ func (c *ModelAdaptor) listNodes(l model.NodeListOptions) ([]*model.KVPair, erro
...
+	}
+
+	// Create an empty slice of KVPair.
+	results := make([]*model.KVPair, len(bal))
+
+	// Go through the list to make sure Affinity field has a proper value,
+	// and map the value to Affinity if deprecated HostAffinity field is used
+	// by calling getBlock, and populate the KVPair slice to return.
+	for idx, bakv := range bal {
+		kvp, err := c.getBlock(bakv.Key)
```
caseydavenport: Not a big fan of doing a `List()`, and then doing individual `Get()`s for each item in the list.  This is a lot of round trips to the datastore.

We should just have a `ensureBlockAffinity` type utility which can be applied to the items returned by the above `List()` so we don't need to do so many gets.

lib/client/ipam_block.go
```diff
@@ -103,7 +103,7 @@ func (b *allocationBlock) autoAssign(
 }
 
 func (b *allocationBlock) assign(address cnet.IP, handleID *string, attrs map[string]string, host string) error {
-	if b.StrictAffinity && b.HostAffinity != nil && host != *b.HostAffinity {
+	if b.StrictAffinity && b.Affinity != nil && host != *b.Affinity {
```
caseydavenport: This guy is still wrong - `host != *b.Affinity`

lib/backend/compat/compat.go
gunjan5: done!

lib/backend/compat/compat.go
```diff
@@ -251,6 +313,44 @@ func (c *ModelAdaptor) listNodes(l model.NodeListOptions) ([]*model.KVPair, erro
...
+	bal, err := c.client.List(l)
+	if err != nil {
+		return nil, err
+	}
+
+	// Go through the list to make sure Affinity field has a proper value,
+	// and map the value to Affinity if deprecated HostAffinity field is used
+	// by calling ensureBlockAffinity, and update the KVPair slice to return.
+	for idx, bakv := range bal {
+		bal[idx] = ensureBlockAffinity(bakv)
```
caseydavenport: I think this is OK, but be really careful about modifying arrays while you're iterating over them - especially if the array size changes.

I'd feel more comfortable if you generated a new array here rather than re-using the `bal` array.

Also, what does `bal` and `bakv` stand for?  Slightly more descriptive names would be nice :)

lib/backend/compat/compat.go
```diff
@@ -251,6 +313,44 @@ func (c *ModelAdaptor) listNodes(l model.NodeListOptions) ([]*model.KVPair, erro
...
+	}
+
+	return bal, nil
+}
+
+// ensureBlockAffinity ensures Affinity field has a proper value,
+// and map the value to Affinity if deprecated HostAffinity field is used.
```
caseydavenport: `and _maps_ the value to Affinity if _the_ deprecated HostAffinity field is used`

caseydavenport:
OK looking pretty good.  A couple more comments - almost there!


I'm also a bit worried about the testing we should be doing on this.  Let's chat.

lib/backend/compat/compat.go
gunjan5: +1 to having a separate slice.
variable names are short because generally go best practice is to use shorter variable names for variables with short scope (more descriptive names as the scope grows) 
Ref: https://github.com/golang/go/wiki/CodeReviewComments#variable-names

PR #246, opened 2016-11-03
Fix and simplify the etcdv2 Syncer

fasaxc:
I found a potential race and a real bug in the etcdv2 Syncer.  This change reworks the Syncer to fix the issue and simplify the snapshot start processing.

The bug was that, after the watcher went out-of-sync, it wouldn't start a new snapshot until it received the next event.  Hence, if nothing was going on, you could be out-of-sync indefinitely.  The watcher now explicitly establishes the current etcd index by reading the Ready flag (which has the nice side effect of stalling the watcher until the Ready flag appears).

The snapshot initiation logic was based on a dubious triangle of channels and I couldn't convince myself that there wasn't a race lurking in there.  I've simplified that so that the watcher thread tells the merge thread that it needs a new snapshot; then the merge thread sends the message to the snapshot thread.  That way there can't be a race between merge thread being told a snapshot is starting and receiving the first update from the snapshot thread.

Along the way, I also cleaned up the channels to use real message types rather than a single struct for everything.  That makes the unpacking code much cleaner.

### DIFF
--- a/lib/backend/model/ippool.go
+++ b/lib/backend/model/ippool.go
@@ -28,7 +28,7 @@ import (
 
 var (
 	matchIPPool = regexp.MustCompile("^/?calico/v1/ipam/v./pool/([^/]+)$")
-	typeIPPool = reflect.TypeOf(IPPool{})
+	typeIPPool  = reflect.TypeOf(IPPool{})
 )
 
 type IPPoolKey struct {
--- a/lib/api/ippool.go
+++ b/lib/api/ippool.go
@@ -74,7 +74,7 @@ func NewIPPool() *IPPool {
 type IPPoolList struct {
 	unversioned.TypeMetadata
 	Metadata unversioned.ListMetadata `json:"metadata,omitempty"`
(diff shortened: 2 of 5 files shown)


fasaxc:
This has passed the OpenStack CI.  I think we should merge it soon so that it has some time to bake before the GA.

fasaxc:
@robbrockbank did you say you'd take a crack at reviewing this?

lib/backend/model/ippool.go
```diff
@@ -28,7 +28,7 @@ import (
 
 var (
 	matchIPPool = regexp.MustCompile("^/?calico/v1/ipam/v./pool/([^/]+)$")
-	typeIPPool = reflect.TypeOf(IPPool{})
+	typeIPPool  = reflect.TypeOf(IPPool{})
```
caseydavenport: extra space here

fasaxc: That's a `go fmt` cleanup so I'll leave as-is.

PR #250, opened 2016-11-04
Add test

caseydavenport:


lib/client/ipam_test.go
```diff
@@ -103,6 +103,39 @@ type testArgsClaimAff struct {
 
 var _ = Describe("IPAM tests", func() {
 
+	Describe("IPAM Assign from any pool", func() {
```
gunjan5: Assign -> AutoAssign

lib/client/ipam_test.go
```diff
@@ -103,6 +103,39 @@ type testArgsClaimAff struct {
...
+			// Call once in order to assign an IP address and create a block.
+			v4, _, outErr := ic.AutoAssign(args)
+			It("should have assigned an IP address with no error", func() {
+				Expect(outErr).NotTo(HaveOccurred())
+				Expect(len(v4) == 1).To(BeTrue())
+			})
+
+			// Call again to trigger an assignment from the newly created block.
+			v4, _, outErr = ic.AutoAssign(args)
+			It("should have assigned an IP address with no error", func() {
```
gunjan5: in general we are supposed to wrap `It(...)` with a `Context(...)` so what you wrote in the comment ` Call again to trigger an assignment from the newly created block.` would be in the `Context("...")` so when the test fails it prints what was the context and what was the expected behavior.

lib/client/ipam_test.go
```diff
@@ -103,6 +103,39 @@ type testArgsClaimAff struct {
...
+				Expect(outErr).NotTo(HaveOccurred())
+				Expect(len(v4) == 1).To(BeTrue())
+			})
+		})
+
```
gunjan5: extra newline :)

Issue #251, opened 2016-11-05
Tags should be metadata on profiles


caseydavenport:
tags should be part of the metadata for the API structs, _not_ part of the spec.  They're basically simpler labels.

As discussed with @lxpollitt

caseydavenport:
Won't make this for 2.0.0-beta, but this should go into v2.0.0

robbrockbank:
Yup funny was thinking that should be the case today....

robbrockbank:
[ ] Update libcalico-go
[ ] Update libnetwork-plugin
[ ] Update CNI plugin
[ ] Update docs
[ ] Update calicoctl

PR #254, opened 2016-11-10
Move tags into the profile metadata

robbrockbank:


robbrockbank:
Part of the fix for #251

lib/api/profile.go
```diff
@@ -36,6 +36,9 @@ type ProfileMetadata struct {
 	// The name of the endpoint.
 	Name string `json:"name,omitempty" validate:"omitempty,name"`
 
+	// A list of tags that are applied to each endpoint that references this profile.
```
caseydavenport: We should mark this as deprecated in the code, yeah?

lib/client/profile_e2e_test.go
```diff
@@ -48,14 +48,14 @@ import (
 var profileSpec1 = api.ProfileSpec{
 	IngressRules: []api.Rule{testutils.InRule1, testutils.InRule2},
 	EgressRules:  []api.Rule{testutils.EgressRule1, testutils.EgressRule2},
-	Tags:         []string{"profile1-tag1", "profile1-tag2"},
 }
+var tags1 = []string{"profile1-tag1", "profile1-tag2"}
```
caseydavenport: Any reason we can't inline these in the tests below rather than creating these special variables?

caseydavenport:
Couple of minor comments, then LGTM.

caseydavenport:
I'm assuming that old profiles will currently just be silently allowed by calicoctl without any warning that tags have moved, but that this will be fixed by another PR in calicoctl to add validation for malformed / unsupported fields?

robbrockbank:
@caseydavenport re your last point.  Yes old profile YAML will silently ignore the old tag field until I add in the updated validation.

PR #255, opened 2016-11-12
Fix revision number update bug + tests for it

gunjan5:
fixes [#1295](https://github.com/projectcalico/calico-containers/issues/1295)

### DIFF
--- a/lib/backend/backend_test.go
+++ b/lib/backend/backend_test.go
@@ -1,0 +1,78 @@
+// Copyright (c) 2016 Tigera, Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package backend_test
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/projectcalico/libcalico-go/lib/backend/model"
+
+	"github.com/projectcalico/libcalico-go/lib/backend/etcd"
+	"github.com/projectcalico/libcalico-go/lib/testutils"
+)
+
+// Setting localhost as the etcd endpoint location since that's where `make run-etcd` runs it.
+var etcdConfig = &etcd.EtcdConfig{
+	EtcdEndpoints: "http://127.0.0.1:2379",
+}
+
+var _ = Describe("Backend tests", func() {
+	Describe("etcd GET/List Revision values", func() {
+		testutils.CleanEtcd()
+
+		etcdClient, _ := etcd.NewEtcdClient(etcdConfig)
+
+		Context("CREATE a Block", func() {
+			block := &model.KVPair{
+				Key: model.BlockKey{
+					CIDR: testutils.MustParseCIDR("10.0.0.0/26"),
+				},
+				Value: model.AllocationBlock{
+					CIDR: testutils.MustParseCIDR("10.0.0.0/26"),
+				},
+			}
+
+			_, cErr := etcdClient.Create(block)
+
+			It("should succeed without an error", func() {
+				Expect(cErr).NotTo(HaveOccurred())
+			})
+		})
+
+		Context("GET BlockKey", func() {
+			key := model.BlockKey{
+				CIDR: testutils.MustParseCIDR("10.0.0.0/26"),
+			}
+			b, bErr := etcdClient.Get(key)
+			It("Revision field should not be nil", func() {
+				Expect(bErr).NotTo(HaveOccurred())
+				Expect(b.Revision).NotTo(BeNil())
+			})
+		})
+
+		Context("LIST BlockKey", func() {
+			blockListOpt := model.BlockListOptions{
+				IPVersion: 4,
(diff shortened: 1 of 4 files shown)


lib/backend/backend_test.go
```diff
@@ -0,0 +1,145 @@
...
+var etcdConfig = &etcd.EtcdConfig{
+	EtcdEndpoints: "http://127.0.0.1:2379",
+}
+
+var _ = Describe("Backend tests", func() {
+
+	FDescribe("etcd GET/List Revision values", func() {
+		testutils.CleanEtcd()
+		c, _ := testutils.NewClient("")
+		ic := testutils.NewIPAMClient(true, etcdType, etcdConfig)
```
caseydavenport: I'm not a big fan of these tests mixing the use of the frontend client + backend client.

These are really backend FVs.  Instead of creating an IPAMClient, let's just use the backend.etcd client to create objects, read them back, and ensure a version is present.

Each test can be:

- Create
- Get
- List
- Delete


You can easily grab a backend client like this:
```
cfg := api.ClientConfig{BackendType: "etcdv2", BackendConfig: etcdConfig}
client, _:= client.New(cfg)
backendClient := client.backend
```

lib/backend/compat/compat.go
```diff
@@ -352,7 +353,7 @@ func ensureBlockAffinity(kvp *model.KVPair) *model.KVPair {
 		// Set AllocationBlock.HostAffinity to nil so it's never non-nil for the clients.
 		val.HostAffinity = nil
 	}
-	return &model.KVPair{Key: kvp.Key, Value: val}
+	return &model.KVPair{Key: kvp.Key, Value: val, Revision: kvp.Revision, TTL: kvp.TTL}
```
caseydavenport: Not sure TTL is valid to set on get / list operation?

https://github.com/projectcalico/libcalico-go/blob/master/lib/backend/model/keys.go#L90

lib/testutils/ipam.go
```diff
@@ -0,0 +1,41 @@
...
+
+package testutils
+
+import (
+	"github.com/projectcalico/libcalico-go/lib/api"
+	"github.com/projectcalico/libcalico-go/lib/backend/etcd"
+	"github.com/projectcalico/libcalico-go/lib/client"
+)
+
+// NewIPAMClient sets up a client, and returns IPAMInterface.
```
caseydavenport: Once we just use the backend client in the tests above we can get rid of this whole file

lib/backend/backend_test.go
```diff
@@ -0,0 +1,145 @@
...
+	"github.com/projectcalico/libcalico-go/lib/backend/model"
+
+	"github.com/projectcalico/libcalico-go/lib/api"
+	"github.com/projectcalico/libcalico-go/lib/backend/etcd"
+	"github.com/projectcalico/libcalico-go/lib/client"
+	"github.com/projectcalico/libcalico-go/lib/testutils"
+)
+
+// Setting BackendType to etcdv2.
+var etcdType api.BackendType = "etcdv2"
```
caseydavenport: This is only used in once place - better to inline the string than to define a var.

lib/backend/backend_test.go
gunjan5: @caseydavenport do you suggest I do the backend FVs similar to what we have for client FVs? separate test file for each object type that does the series of operations (CREATE, GET, LIST, DELETE) ? or all in one file (since we can't have the new subtests with Ginkgo)

caseydavenport: I think that would be better.

For getting this merged let's try to keep it at the bare minimum (i.e let's just ensure the revision is set properly on the block stuff, and any other resource types that you had to change).

lib/backend/backend_test.go
gunjan5: it's gone now

lib/backend/compat/compat.go
gunjan5: TTL is not being set here, this is just passing the TTL value along.
If etcd has a TTL value for a key, `etcd.Get()` call will get it and pass it to the client. Now compat is just adding a layer so it should be transparent and pass TTL and Revision along with Key and Value it gets from `etcd.Get()` https://github.com/projectcalico/libcalico-go/blob/master/lib/backend/compat/compat.go#L245

lib/backend/etcd/etcd.go
```diff
@@ -117,6 +117,7 @@ func (c *EtcdClient) Update(d *model.KVPair) (*model.KVPair, error) {
 	// If the request includes a revision, set it as the etcd previous index.
 	options := etcd.SetOptions{PrevExist: etcd.PrevExist}
 	if d.Revision != nil {
+		log.Infof("Updating etcd revision from: %v to: %v", options.PrevIndex, d.Revision.(uint64))
```
caseydavenport: I think this log should be taken out.

gunjan5: curious why do we want to remove this one but keep the one in `compat`? Only some resoures/objects go through the extra layer of `compat`, so I think it would look inconsistent if we only log the revision update for the ones that go through `compat` (Block, Node and Profile as of now)

lib/backend/backend_test.go
```diff
@@ -0,0 +1,72 @@
...
+		block := &model.KVPair{
+			Key: model.BlockKey{
+				CIDR: testutils.MustParseCIDR("10.0.0.0/26"),
+			},
+			Value: model.AllocationBlock{
+				CIDR: testutils.MustParseCIDR("10.0.0.0/26"),
+			},
+		}
+
+		etcdClient.Create(block)
```
caseydavenport: - Should be in a Context
- Should handle any errors that Create returns.

gunjan5:
done! back to you @caseydavenport

caseydavenport:
LGTM - squash and merge baby.

PR #257, opened 2016-11-14
Update to use k8s client, and improve Makefile

### LINKED ISSUE #256: Use kubernetes/client-go instead of kubernetes/kubernetes
Looks like we should now be able to use client-go instead of vendoring in the entire k8s repository.

robbrockbank:
Fixes #256


robbrockbank:
@caseydavenport:  The size of the k8s repo was proving to be a bit of a pain over the weekend so I had a bash at imported the k8s client instead.  Unless I'm mistaken, the NetworkPolicy objects are present, but they are not yet added to the client interface so I don't think they can be used.

Wanna take a look too, see if I'm just missing it?

You can ignore the Makefile changes in this if you like - I'll probably pull those out as a separate PR.

Makefile
```diff
@@ -1,31 +1,65 @@
...
+TEST_CONTAINER_MARKER=libcalico_test_container.created
+VENDOR_CREATED=vendor/.vendor.created
 
 GO_FILES:=$(shell find lib -name '*.go')
 
 default: all
 all: test
 test: ut
 
+.PHONY: vendor
```
insequent: Was about to add a PR for this as well! :+1:

caseydavenport: Kinda seems wierd to have vendor be PHONY, given it's actually a real directory that we want to build.  What's the reason for this?

caseydavenport:
> Unless I'm mistaken, the NetworkPolicy objects are present, but they are not yet added to the client interface so I don't think they can be used.

I'd like to get this merged, even without NetworkPolicy support, and we can come back and figure out how to move NetworkPolicy over to the client-go repo at a later date.

I see three issues with what we've got at the moment:
- Most API objects can just use client-go (e.g Pods, Namespaces, etc)
- NetworkPolicy currently only exists in the internal clients (or could be done w/ restclient, but requires extra code to make watches work)
- ThirdPartyResource instances can only be done in restclient at the moment (and need extra work to make watches work).

- This PR moves all of the first bullet point to client-go.  
- NetworkPolicy continues to use the internal clients in the kubernetes repo - we can submit an upstream fix for client-go and then switch it over.
- All of the third bullet point (once we need TPRs, which is soon) can _also_ be done with client-go/rest

Makefile
caseydavenport: Gotcha.  I tend to use the glide commands directly for those situations :)

caseydavenport:
@robbrockbank closing for #283 instead.

The makefile changes should be a separate PR :)

PR #266, opened 2016-11-20
Store ASN as string

robbrockbank:
The ASN number in the backend datastore must be stored as a string - this is because confd uses floating point e-notation when writing out large integers which breaks the BIRD templates.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ release
 *.coverprofile
 vendor
 libcalico_test_container.created
+.idea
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -3,4 +3,5 @@ bin
 release
 *.coverprofile
 vendor
-libcalico_test_container.created
\ No newline at end of file
+libcalico_test_container.created
+.idea
```
gunjan5: what is `.idea` for?

PR #268, opened 2016-11-22
Add per-node felix config and global and per-node BGP config

robbrockbank:


lib/client/config.go
```diff
@@ -254,15 +256,65 @@ func (c *config) GetNodeLogLevel(node string) (string, ConfigLocation, error) {
...
 // GetFelixConfig provides a mechanism for getting arbitrary Felix configuration
-// in the datastore.
-func (c *config) GetFelixConfig(name string) (*string, error) {
-	return c.getValue(model.GlobalConfigKey{Name: name})
+// in the datastore.  An empty value for the node will get the global
+// configuration, a non-empty node will get the per-node configuration.
+func (c *config) GetFelixConfig(name, node string) (*string, error) {
+	if node == "" {
+		return c.getValue(model.GlobalConfigKey{Name: name})
+	} else {
```
gunjan5: no need for an else here if you're returning when node==""

lib/client/config.go
```diff
@@ -254,15 +256,65 @@ func (c *config) GetNodeLogLevel(node string) (string, ConfigLocation, error) {
...
+// Caution should be observed using this method as no data validation is
+// performed, and setting incorrect values may have unexpected consequences.
+func (c *config) SetFelixConfig(name, node string, value *string) error {
+	var key model.Key
+	if node == "" {
+		key = model.GlobalConfigKey{Name: name}
+	} else {
+		key = model.HostConfigKey{Hostname: node, Name: name}
+	}
+	if value == nil {
```
gunjan5: I think moving this to the top (Line 277) makes the flow of logic more readable.
```
if val == nil{ 
 return
} 

var key model.Key
// logic for setting the config
...
...
```

lib/client/config.go
```diff
@@ -254,15 +256,65 @@ func (c *config) GetNodeLogLevel(node string) (string, ConfigLocation, error) {
...
+	} else {
+		return c.getValue(model.HostConfigKey{Hostname: node, Name: name})
+	}
 }
 
 // SetFelixConfig provides a mechanism for setting arbitrary Felix configuration
 // in the datastore.  A nil value will remove the entry from the datastore.
-func (c *config) SetFelixConfig(name string, value *string) error {
-	key := model.GlobalConfigKey{Name: name}
+// An empty value for the node will set the global configuration, a non-empty
```
gunjan5: Setting value = nil for delete, "" for global and anything else for node doesn't seem very intuitive. 
Is it possible to add a parameter (maybe a type) for the operation {delete, global, node}
```
type opDelete string
type opGlobal string
type opNode string
```
and then that would be an argument you pass in?

lib/client/config.go
```diff
@@ -254,15 +256,65 @@ func (c *config) GetNodeLogLevel(node string) (string, ConfigLocation, error) {
...
+	}
+	_, err := c.c.backend.Apply(&model.KVPair{
+		Key:   key,
+		Value: *value,
+	})
+	return err
+}
+
+// GetBGPConfig provides a mechanism for getting arbitrary BGP configuration
+// in the datastore.  An empty value for the node will get the global
```
gunjan5: Same as above, adding an operation parameter would make the API more user friendly, using `nil`, `""` etc to suggest the operation seems weird

lib/client/config.go
```diff
@@ -254,15 +256,65 @@ func (c *config) GetNodeLogLevel(node string) (string, ConfigLocation, error) {
...
+// configuration, a non-empty node will get the per-node configuration.
+func (c *config) GetBGPConfig(name, node string) (*string, error) {
+	if node == "" {
+		return c.getValue(model.GlobalBGPConfigKey{Name: name})
+	} else {
```
gunjan5: no need for an `else` here

lib/client/config.go
```diff
@@ -254,15 +256,65 @@ func (c *config) GetNodeLogLevel(node string) (string, ConfigLocation, error) {
...
+// Caution should be observed using this method as no data validation is
+// performed, and setting incorrect values may have unexpected consequences.
+func (c *config) SetBGPConfig(name, node string, value *string) error {
+	var key model.Key
+	if node == "" {
+		key = model.GlobalBGPConfigKey{Name: name}
+	} else {
+		key = model.HostBGPConfigKey{Hostname: node, Name: name}
+	}
 	if value == nil {
```
gunjan5: this guy can move to Line 312 and else on line 315 can go if that happens

lib/client/config.go
robbrockbank: umm when val==nil we are deleting using the key .... so I can't move this to the top.

lib/client/config.go
robbrockbank: I'm not sure I follow your suggestion here.

lib/client/config.go
```diff
@@ -254,25 +258,101 @@ func (c *config) GetNodeLogLevel(node string) (string, ConfigLocation, error) {
...
+// and changing arbitrary configuration may have unexpected consequences.
+func (c *config) UnsetBGPConfig(name, node string) error {
+	return c.deleteConfig(getBGPConfigKey(name, node))
+}
+
+// getFelixConfigKey returns the model.Key interface for the Felix config.
+func getFelixConfigKey(name, node string) model.Key {
+	if node == "" {
+		return model.GlobalConfigKey{Name: name}
+	} else {
```
gunjan5: else is unnecessary here

lib/client/config.go
```diff
@@ -254,25 +258,101 @@ func (c *config) GetNodeLogLevel(node string) (string, ConfigLocation, error) {
...
+		return model.HostConfigKey{Hostname: node, Name: name}
+	}
+}
+
+// getBGPConfigKey returns the model.Key interface for the BGP config.
+func getBGPConfigKey(name, node string) model.Key {
+	if node == "" {
+		return model.GlobalBGPConfigKey{Name: name}
+	} else {
```
gunjan5: else is unnecessary here

PR #271, opened 2016-11-23
Update AutoAssignArgs to support a list of pools

duglin:
This addresses the first bullet of https://github.com/projectcalico/calico-cni/issues/212

I'm assuming we need to still support the old variant (single pool) of AutoAssignArgs too, otherwise we'd break existing users.  Should we deprecate the single flavor though or just keep both going forward?

/cc @lxpollitt @mikespreitzer

Signed-off-by: Doug Davis <dug@us.ibm.com>

### DIFF
--- a/lib/client/ipam.go
+++ b/lib/client/ipam.go
@@ -125,7 +125,7 @@ func (c ipams) AutoAssign(args AutoAssignArgs) ([]net.IP, []net.IP, error) {
 	if args.Num4 != 0 {
 		// Assign IPv4 addresses.
 		log.Debugf("Assigning IPv4 addresses")
-		v4list, err = c.autoAssign(args.Num4, args.HandleID, args.Attrs, args.IPv4Pool, ipv4, hostname)
+		v4list, err = c.autoAssign(args.Num4, args.HandleID, args.Attrs, args.IPv4Pools, ipv4, hostname)
 		if err != nil {
 			log.Errorf("Error assigning IPV4 addresses: %s", err)
 			return nil, nil, err
@@ -135,7 +135,7 @@ func (c ipams) AutoAssign(args AutoAssignArgs) ([]net.IP, []net.IP, error) {
 	if args.Num6 != 0 {
 		// If no err assigning V4, try to assign any V6.
 		log.Debugf("Assigning IPv6 addresses")
-		v6list, err = c.autoAssign(args.Num6, args.HandleID, args.Attrs, args.IPv6Pool, ipv6, hostname)
+		v6list, err = c.autoAssign(args.Num6, args.HandleID, args.Attrs, args.IPv6Pools, ipv6, hostname)
 		if err != nil {
 			log.Errorf("Error assigning IPV6 addresses: %s", err)
 			return nil, nil, err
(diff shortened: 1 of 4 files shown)


caseydavenport:
@robbrockbank assigning you as a joint reviewer.

@duglin my take would be that we don't need to continue supporting the old IPv4Pool field, since we're still pre v1.0.0.

duglin:
@caseydavenport 

Can easily remove - just want to make sure no one objects over the day or so....

lib/client/ipam.go
```diff
@@ -243,19 +250,16 @@ func (c ipams) autoAssign(num int, handleID *string, attrs map[string]string, po
...
-			// Use the given pool.
-			pools = []net.IPNet{*pool}
 		}
 
 		// Iterate over pools and assign addresses until we either run out of pools,
 		// or the request has been satisfied.
 		for _, p := range pools {
 			log.Debugf("Assigning from random blocks in pool %s", p.String())
-			newBlock := randomBlockGenerator(p)
+			newBlock := randomBlockGenerator(*p)
```
caseydavenport: Still thinking about it, but I suspect we probably want to change pools from being a `[]*net.IPNet` to `[]net.IPNet`.

I don't think we need them to be pointers. WDYT?

lib/client/ipam_block_reader_writer.go
```diff
@@ -56,31 +56,41 @@ func (rw blockReaderWriter) getAffineBlocks(host string, ver ipVersion, pool *cn
...
 
-		// Add the block if no IP pool was specified, or if an IP pool was specified
-		// and the block falls within the given IP pool.
-		if pool == nil || pool.Contains(k.CIDR.IPNet.IP) {
+		// Add the block if no IP pools were specified, or if IP pools were specified
+		// and the block falls within the given IP pools.
+		if len(pools) == 0 {
 			ids = append(ids, k.CIDR)
+			continue
+		}
```
caseydavenport: I think I'd like this better if we used an if/else here rather than the if / continue.

lib/client/ipam_test.go
```diff
@@ -335,6 +335,81 @@ var _ = Describe("IPAM tests", func() {
...
+		pool2 := testutils.MustParseCIDR("20.0.0.0/24")
+		pool3 := testutils.MustParseCIDR("30.0.0.0/24")
+
+		testutils.CreateNewIPPool(*c, "10.0.0.0/24", false, false, true)
+		testutils.CreateNewIPPool(*c, "20.0.0.0/24", false, false, true)
+		testutils.CreateNewIPPool(*c, "30.0.0.0/24", false, false, true)
+
+		// Step-1: AutoAssign 300 IPs from 2 pools
+		// Expect that the IPs are from both pools
+		Context("AutoAssign 1 IP from pool1", func() {
```
caseydavenport: Think this Context log needs updating

caseydavenport:
@duglin Thanks for this!

Looking pretty good to me. Stuck a few comments on.

Biggest one is to make sure nobody objects to getting rid of the IPv4Pool / IPv6Pool args and then do that.

lib/client/ipam.go
duglin: I don't have a strong opinion on this one since the struct is small - I'll make the switch.

duglin: this is a change from how it is today though - IPv4Pool is defined as:
```
IPv4Pool *net.IPNet
```
But if backwards compat isn't an issue...

duglin:
Ok. I updated the first commit with the minor syntax changes.

I also pushed another commit with the removing of the old IPv*Pool param and got rid of the pointers.

Once the maintainers decide if backwards compat matters or not I can either squash the 2 commits or delete the 2nd one.

robbrockbank:
@duglin  -  LGTM.

I agree with @caseydavenport with the removal the IPv4Pool and IPv6Pool options - since the API was only Beta at this stage.

duglin:
ok I squashed the two commits down to just one since there's agreement on removing the old singleton IPv*pool args.

@robbrockbank re: other repo's impact ... I'm working on the 2nd(CNI) bullet of https://github.com/projectcalico/calico-cni/issues/212 and am running into testing issues... will ask on slack or in that repo.

robbrockbank:
LGTM

@duglin - we'll be looking out for the CNI update.   Let me know if you plan i updating the libnetwork-plugin too (otherwise we'll look into that one).

duglin:
yes I'm working on the libnetwork-plugin one right now and I may have found a bug in this PRs code but still investigating.... keep tuned...

duglin:
hmm never mind - now that this PR is in master I think my libnetwork-plugin PR is working.

PR #272, opened 2016-11-23
Lots of data validation fixes

### LINKED ISSUE #222: Validate should probably limit field sizes
As summary.  We should reject attempts to create really really long string values.  

Its unlikely that particularly long strings would be legitimate config - attempts to do so are more likely to be either bugs in automated config generation or done with malicious intent.

robbrockbank:
Fixes #222

This also provides more information to the user, and allows the validation tag to be used to return a useful message to the user (which before we could not)

### DIFF
--- a/lib/errors/errors.go
+++ b/lib/errors/errors.go
@@ -65,23 +65,33 @@ type ErrorValidation struct {
 }
 
 type ErroredField struct {
-	Name  string
-	Value interface{}
+	Name   string
+	Value  interface{}
+	Reason string
 }
 
 func (e ErrorValidation) Error() string {
 	if len(e.ErroredFields) == 0 {
 		return "unknown validation error"
 	} else if len(e.ErroredFields) == 1 {
-		return fmt.Sprintf("error with field %s = '%v'",
-			e.ErroredFields[0].Name,
-			e.ErroredFields[0].Value)
+		f := e.ErroredFields[0]
+		if f.Reason == "" {
+			return fmt.Sprintf("error with field %s = '%v'",
+				f.Name, f.Value)
+		} else {
+			return fmt.Sprintf("error with field %s = '%v' (%s)",
+				f.Name, f.Value, f.Reason)
+		}
 	} else {
 		s := "error with the following fields:\n"
 		for _, f := range e.ErroredFields {
-			s = s + fmt.Sprintf("-  %s = '%v'\n",
-				f.Name,
-				f.Value)
+			if f.Reason == "" {
+				s = s + fmt.Sprintf("-  %s = '%v'\n",
+					f.Name, f.Value)
+			} else {
+				s = s + fmt.Sprintf("-  %s = '%v' (%s)\n",
+					f.Name, f.Value, f.Reason)
+			}
 		}
 		return s
 	}
--- a/lib/numorstring/port.go
+++ b/lib/numorstring/port.go
@@ -36,7 +36,8 @@ func SinglePort(port uint16) Port {
 func PortFromRange(minPort, maxPort uint16) (Port, error) {
 	port := Port{minPort, maxPort}
 	if minPort > maxPort {
-		return port, errors.New("minimum port number is greater than maximum port number in port range")
+		msg := fmt.Sprintf("minimum port number (%d) is greater than maximum port number (%d) in port range", minPort, maxPort)
+		return port, errors.New(msg)
 	}
 	return port, nil
 }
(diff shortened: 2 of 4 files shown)


lib/numorstring/port.go
```diff
@@ -50,13 +51,16 @@ func PortFromString(s string) (Port, error) {
...
 
 	if pmin, err := strconv.ParseUint(parts[0], 10, 16); err != nil {
-		return Port{}, errors.New("invalid minimum port number in range")
+		msg := fmt.Sprintf("invalid minimum port number in range (%s)", s)
+		return Port{}, errors.New(msg)
 	} else if pmax, err := strconv.ParseUint(parts[1], 10, 16); err != nil {
-		return Port{}, errors.New("invalid maximum port number in range")
+		msg := fmt.Sprintf("invalid maximum port number in range (%s)", s)
+		return Port{}, errors.New(msg)
 	} else {
```
gunjan5: unnecessary `else`

lib/numorstring/port.go
```diff
@@ -50,13 +51,16 @@ func PortFromString(s string) (Port, error) {
 
 	parts := strings.Split(s, ":")
 	if len(parts) != 2 {
-		return Port{}, errors.New("invalid port format")
+		msg := fmt.Sprintf("invalid port format (%s)", s)
+		return Port{}, errors.New(msg)
```
gunjan5: just curious why can't the error message be inlined? like `errors.New(fmt.Sprintf("invalid port format (%s)", s))`

lib/validator/validator.go
```diff
@@ -30,14 +32,35 @@ import (
 var validate *validator.Validate
 
 var (
-	nameRegex          = regexp.MustCompile("^[a-zA-Z0-9_.-]+$")
+	nameRegex          = regexp.MustCompile("^[a-zA-Z0-9_.-]{1,128}$")
 	interfaceRegex     = regexp.MustCompile("^[a-zA-Z0-9_-]{1,15}$")
-	labelRegex         = regexp.MustCompile("^[a-zA-Z_./-][a-zA-Z0-9_./-]*$")
+	labelRegex         = regexp.MustCompile("^[a-zA-Z_./-][a-zA-Z0-9_./-]{0,127}$")
```
gunjan5: why is this `{0,127}` starting from `0` compared to `nameRegex`? does that mean we allow empty labels?

lib/validator/validator.go
```diff
@@ -66,6 +89,23 @@ func init() {
...
+func reason(r string) string {
+	return reasonString + r
+}
+
+// extractReason extracts the error reason from the field tag in a validator
+// field error.
+func extractReason(tag string) string {
+	if strings.HasPrefix(tag, reasonString) {
+		return strings.TrimPrefix(tag, reasonString)
+	} else {
```
gunjan5: unnecessary else

lib/validator/validator.go
```diff
@@ -66,6 +89,23 @@ func init() {
...
+		return ""
```
gunjan5: would there ever be a case where `reasonString` is not prefixed? if not we should be returning an error since returning an empty string is a silent failure

lib/validator/validator.go
```diff
@@ -162,10 +188,12 @@ func validateProtocol(v *validator.Validate, structLevel *validator.StructLevel)
 	// names.
 	if num, err := p.NumValue(); err == nil {
 		if num == 0 {
-			structLevel.ReportError(reflect.ValueOf(p.NumVal), "Protocol", "protocol", "protocol number invalid")
+			structLevel.ReportError(reflect.ValueOf(p.NumVal),
+				"Protocol", "", reason("protocol number invalid"))
```
gunjan5: for my understanding, why did you change `"protocol"` field to `""` in this one and all the following `ReportError`?

lib/numorstring/port.go
robbrockbank: it can - i think having it on a separate line is easier to read tho.

I'll change it if its bugging you or if it is against golang suggested practices.  Let me know.

lib/validator/validator.go
robbrockbank: Nope the label regex has an initial char [a-zA-Z_./-] followed by 0-127 [a-zA-Z0-9_./-]

Basically, the first char cannot be a number - the 0,127 applies to the second character

lib/validator/validator.go
robbrockbank: Well that depends on the validator that someone has written - I agree that it should always be set, but better to  have it error something sensible rather than panic or similar.   This won't cause a silent failure - it just returns less information to the user.

lib/errors/errors.go
```diff
@@ -65,23 +65,33 @@ type ErrorValidation struct {
...
 	if len(e.ErroredFields) == 0 {
 		return "unknown validation error"
 	} else if len(e.ErroredFields) == 1 {
-		return fmt.Sprintf("error with field %s = '%v'",
-			e.ErroredFields[0].Name,
-			e.ErroredFields[0].Value)
+		f := e.ErroredFields[0]
+		if f.Reason == "" {
+			return fmt.Sprintf("error with field %s = '%v'",
+				f.Name, f.Value)
```
robbrockbank: Good question,  it is possible that the Value could be nil I guess (if someone specified null in the json)

But Sprintf handles it gracefully (it writes out `<nil>`)

lib/numorstring/port.go
gunjan5: its not bugging me, I was just curious. Don't think there's a golang best practice for this, I was just wondering about the extra variable creation in memory

PR #273, opened 2016-11-24
Change global default AS number to be 64512

robbrockbank:


gunjan5:
Reference: https://github.com/projectcalico/libcalico/issues/140

PR #276, opened 2016-11-25
Add 'untracked' property to policy data model

neiljerram:


lib/api/policy.go
```diff
@@ -93,6 +93,10 @@ type PolicySpec struct {
 	// 	deployment != "dev"
 	// 	! has(label_name)
 	Selector string `json:"selector" validate:"selector"`
+
+	// Indicates, if True, that the rules in this policy should be applied before any data plane
+	// connection tracking, and that packets allowed by these rules should not be tracked.
```
caseydavenport: GoDoc should begin with the name of the field / method / etc that it is describing.

e.g "Untracked indicates whether or not the rules in this policy should go through connection tracking.  If True, ..."

lib/api/policy.go
```diff
@@ -93,6 +93,10 @@ type PolicySpec struct {
...
+	Untracked bool `json:"untracked,omitempty" validate:"omitempty"`
```
caseydavenport: Minor point, but I think as far as naming goes we wan't to keep fields declarative.

Untracked feels more like a status rather than a declaration of intent.  My preference would be for something like 

```go
DontTrack bool
```

or 

```go
DoNotTrack bool
```

or similar

caseydavenport: Note this is just for the client API.

etcd datamodel can definitely stay as it is.

neiljerram: Very happy to make that change, but should I wait for a view from @robbrockbank as well?

neiljerram: Of the two possibilities that you've given, my preference would be DoNotTrack.

lib/api/policy.go
robbrockbank: Yup @caseydavenport is right and you can slap me for not adhering to this format - but we should do - i'll make sure there is an issue to fix up existing doc strings, but might as well start getting it right :-)

lib/api/policy.go
```diff
@@ -93,6 +93,12 @@ type PolicySpec struct {
 	// 	deployment != "dev"
 	// 	! has(label_name)
 	Selector string `json:"selector" validate:"selector"`
+
+	// Untracked indicates whether packets matched by the rules in this policy should go through
+	// the data plane's connection tracking, such as Linux conntrack.  If True, the rules in
+	// this policy are applied before any data plane connection tracking, and packets allowed by
+	// this policy are marked as not to be tracked.
+	Untracked bool `json:"untracked,omitempty" validate:"omitempty"`
```
robbrockbank: not sure the validate is really necessary for a bool field.

lib/api/policy.go
robbrockbank: I'm happy with DoNotTrack for the API.

lib/api/policy.go
neiljerram: Could you be more definitive? (As I'm not yet familiar with this detail myself.)

lib/api/policy.go
neiljerram: Done.

neiljerram:
@caseydavenport @robbrockbank Thanks for your reviews.  With new commits the PR now addresses everything, I believe, except for the point about whether omitempty is needed for a boolean field - please can you give me more guidance about that?

robbrockbank:
@neiljerram 

The validate tag is used to validate the structs.  Since the field is a boolean, it always has a value: false or true (in Go the zero-struct value is false) - so effectively the field is never empty.

neiljerram:
Thanks @robbrockbank.  I've removed the validation and it still passes `make ut` for me locally - so will merge if the CI also still passes now.

PR #280, opened 2016-11-27
Fix bad log level field names

robbrockbank:
Fixes https://github.com/projectcalico/calico-containers/issues/1341


caseydavenport:
LGTM

PR #283, opened 2016-11-29
k8s backend support for GlobalConfig

caseydavenport:
Adds support for GlobalConfig via a ThirdPartyResource in the k8s API.

This also attempts to switch over to client-go instead of the kubernetes/pkg/client code.


lib/backend/k8s/k8s.go
```diff
@@ -79,12 +98,91 @@ func NewKubeClient(kc *KubeConfig) (*KubeClient, error) {
...
+			Name:      "global-config.projectcalico.org",
+			Namespace: "kube-system",
+		},
+		Description: "Calico Global Configuration",
+		Versions:    []extensions.APIVersion{{Name: "v1"}},
+	}
+	_, err := c.clientSet.Extensions().ThirdPartyResources().Get(tpr.ObjectMeta.Name)
+	if err != nil {
+		// The resource needs to be created.
+		_, err := c.clientSet.Extensions().ThirdPartyResources().Create(&tpr)
```
robbrockbank: Rather than getting and then creating, it's better to just create and handle the appropriate error that indicates it's already created.  It's one less round trip on the first create, but more importantly closes the window where two concurrent threads could attempt to create the resource and expect to pass.

lib/backend/k8s/k8s.go
```diff
@@ -93,27 +191,41 @@ func (c *KubeClient) Syncer(callbacks api.SyncerCallbacks) api.Syncer {
...
-	log.Warn("Attempt to 'Create' using kubernetes backend is not supported.")
-	return nil, errors.ErrorResourceDoesNotExist{
-		Err:        goerrors.New("Resource does not exist"),
-		Identifier: d.Key,
+	switch d.Key.(type) {
+	case model.GlobalConfigKey:
+		return c.createGlobalConfig(d)
+	default:
+		log.Warn("Attempt to 'Create' using kubernetes backend is not supported.")
+		return nil, errors.ErrorResourceDoesNotExist{
```
robbrockbank: Feels like we should have an ErrorActionNotSupported   or some such thing rather than a resource does not exist error.

lib/backend/k8s/k8s.go
robbrockbank: Are we "handling" any of the errors we get from the k8s client at the moment?  I think we should be looking at the returned error and converting a known error type for our API (in particular resource already exists, or resource not found etc) .... At the moment it looks as though k8s errors are passed straight up which means a user of the API has to understand k8s errors as well as etcd etc.  Perhaps not something to worry about for v1.0.0 - although I would prefer to see some error handling for that release.

lib/backend/k8s/thirdparty/global_config.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+// Required to satisfy ListMetaAccessor interface
+func (el *GlobalConfigList) GetListMeta() unversioned.List {
+	return &el.Metadata
+}
+
+// The code below is used only to work around a known problem with third-party
+// resources and ugorji. If/when these issues are resolved, the code below
+// should no longer be required.
```
robbrockbank: I assume from below that the problem is that they are unmarshmaling into a non-zeroed struct?

lib/backend/k8s/syncer.go
```diff
@@ -285,7 +293,12 @@ func (syn *kubeSyncer) performSnapshot() ([]model.KVPair, map[model.Key]bool, re
 
 		// Get NetworkPolicies (Policies)
 		log.Info("Syncing NetworkPolicy")
-		npList, err := syn.kc.clientSet.NetworkPolicies("").List(opts)
+		npList := extensions.NetworkPolicyList{}
+		err = syn.kc.clientSet.Extensions().RESTClient().
+			Get().
+			Resource("networkpolicies").
+			Timeout(10 * time.Second).
```
robbrockbank: Might be nice to have these times declared as costs above rather than buried here.

robbrockbank: same applies elsewhere

robbrockbank:
Given we'll want to be testing this as much as possible it probably makes sense just to get this in, but there are a number of changes that I think we should sort out over the next few RCs:
-  We seem to just be passing errors up from the k8s datastore rather than handling and returning as Calico client error types.
-  We need to fill out a bunch of missing function comments
-  There are a lot of inline timeouts and resource names e.g "networkpolicies" that are repeated a lot and would probably benefit from defining as constants.

lib/backend/k8s/thirdparty/global_config.go
caseydavenport: Honestly, I'm not entirely sure. 

I copied this directly from the examples provided here: https://github.com/kubernetes/client-go/blob/master/examples/third-party-resources/types.go

It doesn't work without it :)

caseydavenport:
Also needs:
- Watch support in the syncer.
- FV tests.

robbrockbank:
This is still pulling in k8s.io/kubernetes which is one of the things we were hoping to avoid no?

caseydavenport:
> This is still pulling in k8s.io/kubernetes which is one of the things we were hoping to avoid no?

Ideally yes.  If client-go needs it then there isn't much we can do!  I don't think it should though.. I'll check.

lib/backend/k8s/k8s.go
```diff
@@ -86,7 +104,89 @@ func NewKubeClient(kc *KubeConfig) (*KubeClient, error) {
...
+		Description: "Calico Global Configuration",
+		Versions:    []extensions.APIVersion{{Name: "v1"}},
+	}
+	_, err := c.clientSet.Extensions().ThirdPartyResources().Create(&tpr)
+	if err != nil {
+		// Don't care if it already exists.
+		if !kerrors.IsAlreadyExists(err) {
+			return goerrors.New(fmt.Sprintf("failed to create ThirdPartyResource %s: %s", tpr.ObjectMeta.Name, err))
+		}
+	}
```
caseydavenport: I still need to figure out what to do here.

A watch on global configs fails until an instance of the TPR is created.  I tried to create an instance here at startup, but it's not only yucky it seems to not work. idk if the requests are too fast, but adding a 5 second sleep between creating the TPR and the GlobalConfig makes it work.  Retries don't seem to help...

We should talk tomorrow - this needs to either go in or be punted until v2.1.

If it's punted, there are two key bugfixes in this PR that need to go into v2.0

lib/backend/k8s/thirdparty/global_config.go
```diff
@@ -0,0 +1,78 @@
...
+
+	"k8s.io/client-go/pkg/api"
+	"k8s.io/client-go/pkg/api/meta"
+	"k8s.io/client-go/pkg/api/unversioned"
+	"k8s.io/client-go/pkg/runtime/schema"
+)
+
+type GlobalConfigSpec struct {
+	Name  string `json:"name"`
+	Value string `json:"value"`
```
caseydavenport: Is it OK that Value is a string? Do we support GlobalConfig that isn't a string?

lib/errors/errors.go
```diff
@@ -39,6 +39,15 @@ func (e ErrorResourceDoesNotExist) Error() string {
 	return fmt.Sprintf("resource does not exist: %s", e.Identifier)
 }
 
+// Error indicating an operation is not supported.
+type ErrorOperationNotSupported struct {
+	Err error
```
caseydavenport: TODO: Fix this up to include an identifier.

lib/backend/k8s/k8s_fv_test.go
```diff
@@ -312,4 +318,31 @@ var _ = Describe("Test Syncer API for Kubernetes backend", func() {
...
+			},
+			Value: "someguid",
+		}
+		updGC, err := c.Create(gc)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(updGC.Value.(string)).To(Equal("someguid"))
+		Expect(updGC.Key.(model.GlobalConfigKey).Name).To(Equal("ClusterGUID"))
+		Expect(updGC.Revision).NotTo(BeNil())
+
+		updGC, err = c.Get(gc.Key)
```
caseydavenport: TODO: Check apply/update here as well.

robbrockbank: I assume apply/update should work - in which case we should just add these additional tests....

lib/backend/k8s/k8s_fv_test.go
```diff
@@ -312,4 +318,31 @@ var _ = Describe("Test Syncer API for Kubernetes backend", func() {
...
 	})
+
+	It("should support setting and getting GlobalConfig", func() {
+		gc := &model.KVPair{
+			Key: model.GlobalConfigKey{
+				Name: "ClusterGUID",
+			},
+			Value: "someguid",
+		}
+		updGC, err := c.Create(gc)
```
robbrockbank: Nit: Should this be split up with some By() blocks just to indicate what step we're actually doing.

lib/backend/k8s/thirdparty/global_config.go
robbrockbank: All of our global config is a string - so that should be fine.   It's up to Felix to interpret as required.

The mapping to a specific data type is handled in the front end side of the config (or at least will be once we map through all of the possible values to the front end).

lib/backend/k8s/syncer.go
```diff
@@ -472,3 +511,21 @@ func (syn *kubeSyncer) parseNetworkPolicyEvent(e watch.Event) *model.KVPair {
 	}
 	return kvp
 }
+
+func (syn *kubeSyncer) parseGlobalConfigEvent(e watch.Event) *model.KVPair {
+	log.Debug("Parsing GlobalConfig watch event")
+	// First, check the event type.
+	gc, ok := e.Object.(*thirdparty.GlobalConfig)
+	if !ok {
+		log.Panicf("Invalid GlobalConfig event. Type: %s, Object: %+v", e.Type, e.Object)
```
robbrockbank: Is a panic the right thing here?  Seems a little severe if some bad data has somehow gotten into k8s

lib/backend/k8s/k8s_fv_test.go
caseydavenport: Yep - fixed.

lib/backend/k8s/k8s_fv_test.go
caseydavenport: Yep, fixed.

lib/backend/k8s/syncer.go
caseydavenport: Not sure that this is a case of "bad data got into k8s" - this is us receiving an event for something that isn't a GlobalConfig over the GlobalConfig watch.

I think a panic is OK here.

PR #290, opened 2016-12-04
Support setting IP address on wep for k8s backend

caseydavenport:
By supporting an Apply action, the CNI plugin now triggers an update to Felix much sooner than would have before, meaning policy is programmed sooner and time to first connectivity is significantly shorter.

### DIFF
--- a/lib/backend/k8s/k8s.go
+++ b/lib/backend/k8s/k8s.go
@@ -114,8 +114,13 @@ func (c *KubeClient) Update(d *model.KVPair) (*model.KVPair, error) {
 // Set an existing entry in the datastore.  This ignores whether an entry already
 // exists.
 func (c *KubeClient) Apply(d *model.KVPair) (*model.KVPair, error) {
-	log.Infof("Ignoring 'Apply' for %s", d.Key)
-	return d, nil
+	switch d.Key.(type) {
+	case model.WorkloadEndpointKey:
+		return c.applyWorkloadEndpoint(d)
+	default:
+		log.Infof("Ignoring 'Apply' for %s", d.Key)
+		return d, nil
+	}
 }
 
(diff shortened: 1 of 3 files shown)


lib/backend/k8s/k8s.go
```diff
@@ -214,6 +219,29 @@ func (c *KubeClient) getProfile(k model.ProfileKey) (*model.KVPair, error) {
...
+	if len(ips) > 0 {
+		log.Debugf("Applying workload with IPs: %+v", ips)
+		ns, name := c.converter.parseWorkloadID(k.Key.(model.WorkloadEndpointKey).WorkloadID)
+		pod, err := c.clientSet.Pods(ns).Get(name)
+		if err != nil {
+			return nil, err
+		}
+		pod.Status.PodIP = ips[0].IP.String()
+		log.Debugf("Pod IP: %+v", pod.Status.PodIP)
+		pod, err = c.clientSet.Pods(ns).UpdateStatus(pod)
```
robbrockbank: I think we should be converting the error here to a known error type for the Calico API.   In particular it feels that we should explicitly handle a resource update conflict in the CNI plugin (so that if k8s sorts itself out and magically updates before we do then we could hit concurrent update issues).

robbrockbank: Should we be checking that the IP is not set and no-op if it is already set to the correct value?

caseydavenport: The k8s API won't care if we're setting this field on something that already has an address.

We actually want to change it anyway - if an IP already exists it means the pod has restarted and might be getting a new IP address.

caseydavenport: Probably worth this module having a k8sErrToCalicoErr type helper.

robbrockbank:
lgtm

PR #293, opened 2016-12-07
Re-create glide files for RC4

robbrockbank:


glide.yaml
```diff
@@ -1,29 +1,40 @@
 package: github.com/projectcalico/libcalico-go
-homepage: https://projectcalico.org
-license: Apache-2.0
```
neiljerram: Does it matter that the homepage and license have been lost here?

glide.yaml
```diff
@@ -1,29 +1,40 @@
...
-owners:
-- name: Rob Brockbank
-  email: rob@tigera.io
```
neiljerram: And the owner!

glide.yaml
tomdee: No - none of our other glide files have them

glide.yaml
tomdee: As above :smile:

PR #295, opened 2016-12-08
hostname based hashing

### LINKED ISSUE #294: IPAM assignment uses random seed for affine block selection
It should use a hash of the hostname to seed the random selection.  This way it avoids multiple affine blocks being allocated to a host if there are multiple concurrent IP requests at start of day (or when one affine block is full)

gunjan5:
Fixes #294

### DIFF
--- a/lib/client/ipam_block_reader_writer.go
+++ b/lib/client/ipam_block_reader_writer.go
@@ -17,11 +17,12 @@ package client
 import (
 	goerrors "errors"
 	"fmt"
+	"hash/fnv"
 	"math/big"
 	"math/rand"
 	"net"
+	"os"
 	"reflect"
-	"time"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/projectcalico/libcalico-go/lib/api"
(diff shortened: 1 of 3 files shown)


lib/client/ipam_block_reader_writer.go
```diff
@@ -340,8 +341,17 @@ func randomBlockGenerator(pool cnet.IPNet) func() *cnet.IPNet {
 	numBlocks := new(big.Int)
 	numBlocks.Div(numIP, big.NewInt(blockSize))
 
-	// Start at a random offset index
-	source := rand.NewSource(time.Now().UnixNano())
+	// Create a random number generator seed based on the hostname.
+	// This is to avoid assigning multiple blocks when multiple
+	// workloads request IPs around the same time.
+	hostName, err := os.Hostname()
```
robbrockbank: You should be passing in the hostname through to this function and using that.

robbrockbank:
This will improve things, so I think we should just merge.

However, the random block algorithm is still a little off since we really want to randomly select blocks across randomly shuffled pools.  So we're still missing the shuffling of IP Pools - which may be extremely relevant to customers with lots of small pools - I think that is more work and worthy of a different issue.  Can you raise please?   For example see the _random_subnets_from_cidr() method in the old python ipam code.

gunjan5:
For reference, the hashing algorithm used is `FNV-1` (See the speed/collision comparison below): 

```
Hash           Lowercase      Random UUID  Numbers
=============  =============  ===========  ==============
Murmur            145 ns      259 ns          92 ns
                    6 collis    5 collis       0 collis
FNV-1a            152 ns      504 ns          86 ns
                    4 collis    4 collis       0 collis
FNV-1             184 ns      730 ns          92 ns
                    1 collis    5 collis       0 collis▪
DBJ2a             158 ns      443 ns          91 ns
                    5 collis    6 collis       0 collis▪▪▪
DJB2              156 ns      437 ns          93 ns
                    7 collis    6 collis       0 collis▪▪▪
SDBM              148 ns      484 ns          90 ns
                    4 collis    6 collis       0 collis**
SuperFastHash     164 ns      344 ns         118 ns
                   85 collis    4 collis   18742 collis
CRC32             250 ns      946 ns         130 ns
                    2 collis    0 collis       0 collis
LoseLose          338 ns        -             -
               215178 collis
```
Ref: http://softwareengineering.stackexchange.com/questions/49550/which-hashing-algorithm-is-best-for-uniqueness-and-speed

PR #298, opened 2016-12-10
Use projectcalico/go-yaml-wrapper fork and UnmarshalStrict 

gunjan5:
This is the current behavior (with config.cfg being invalid yaml) :
```bash
$ calicoctl apply -c config.cfg -f pool.yaml
No resources specified in file
``` 
With this change, error looks like:
```bash
$ calicoctl apply -c config.cfg -f pool.yaml
Failed to Calico API client: error parsing document: fields in document are not recognized or are in the wrong location: etcdCACertFile, etcdCertFile, etcdEndpoints, etcdKeyFile
exit status 1
``` 

which is much more clear!


gunjan5:
moving to #302

PR #301, opened 2016-12-12
Relax the unmarshaling of the block affinity

robbrockbank:
Convert the AffinityBlock to be a raw string type.

Python code handles any value being assigned for the block, so it will be fine with pre GA releases of golang IPAM.

It was not sufficient to just add an unmarshal method because the unmarshaling code hooks into the son unmarshaller which does not handle empty string values.   Migrating to be a raw string type is a much safer approach.


robbrockbank:
One detail that might be worth doing.... rather than writing in an empty string, we could write in a {} value.  Python handles that just fine and it means that new golang and old golang will be compatible.

robbrockbank:
oops didn't mean to close

caseydavenport:
OK.  This LGTM.

PR #304, opened 2016-12-13
Use glide cache to build go code

SergK:
Use cache as much as possible from CI user (~/.glide) also with
fixing permissions

We need to have an ability to run calico build in isolated environment
(without access to 3rd part repos) and cache usage is one step to this
direction


tomdee:
Glide cache is now in master (in another PR) so closing.

PR #305, opened 2016-12-13
Removed label value validation and related tests

heschlie:
The Label value validation was causing us to be unable to delete
workload endpoints via calicoctl.

### DIFF
--- a/lib/selector/tokenizer/tokenizer.go
+++ b/lib/selector/tokenizer/tokenizer.go
@@ -54,7 +54,7 @@ type Token struct {
 }
 
 const (
-	identifierExpr = `[a-zA-Z_./-][a-zA-Z0-9_./-]*`
+	identifierExpr = `([a-zA-Z0-9.-]{0,253}/)?[a-zA-Z0-9]([a-zA-Z0-9_.-]{0,61}[a-zA-Z0-9])*`
 	hasExpr        = `has\(\s*(` + identifierExpr + `)\s*\)`
 	allExpr        = `all\(\s*\)`
 	notInExpr      = `not\s*in\b`
--- a/lib/validator/validator.go
+++ b/lib/validator/validator.go
@@ -34,7 +34,8 @@ var validate *validator.Validate
 var (
 	nameRegex          = regexp.MustCompile("^[a-zA-Z0-9_.-]{1,128}$")
 	interfaceRegex     = regexp.MustCompile("^[a-zA-Z0-9_-]{1,15}$")
-	labelRegex         = regexp.MustCompile("^[a-zA-Z_./-][a-zA-Z0-9_./-]{0,127}$")
+	labelRegex         = regexp.MustCompile("^([a-zA-Z0-9.-]{0,253}/)?[a-zA-Z0-9]([a-zA-Z0-9_.-]{0,61}[a-zA-Z0-9])?$")
+	labelValueRegex    = regexp.MustCompile("^[a-zA-Z0-9]?([a-zA-Z0-9_.-]{0,61}[a-zA-Z0-9])?$")
 	actionRegex        = regexp.MustCompile("^(allow|deny|log|pass)$")
 	backendActionRegex = regexp.MustCompile("^(allow|deny|log|next-tier|)$")
 	protocolRegex      = regexp.MustCompile("^(tcp|udp|icmp|icmpv6|sctp|udplite)$")
@@ -166,7 +167,7 @@ func validateLabels(v *validator.Validate, topStruct reflect.Value, currentStruc
 	l := field.Interface().(map[string]string)
 	log.Debugf("Validate labels: %s", l)
 	for k, v := range l {
-		if !labelRegex.MatchString(k) || !labelRegex.MatchString(v) {
+		if !labelRegex.MatchString(k) || !labelValueRegex.MatchString(v) {
 			return false
 		}
 	}
--- a/lib/validator/validator_test.go
+++ b/lib/validator/validator_test.go
@@ -110,10 +110,19 @@ func init() {
 
 		// (API) Labels.
 		Entry("should accept a valid label", api.HostEndpointMetadata{Labels: map[string]string{"rank_.0-9": "gold._0-9"}}, true),
-		Entry("should reject label key starting with 0-9", api.HostEndpointMetadata{Labels: map[string]string{"2rank": "gold"}}, false),
-		Entry("should reject label value starting with 0-9", api.HostEndpointMetadata{Labels: map[string]string{"rank": "2gold"}}, false),
+		Entry("should accept label key starting with 0-9", api.HostEndpointMetadata{Labels: map[string]string{"2rank": "gold"}}, true),
+		Entry("should accept label value starting with 0-9", api.HostEndpointMetadata{Labels: map[string]string{"rank": "2gold"}}, true),
+		Entry("should accept label key with dns prefix", api.HostEndpointMetadata{Labels: map[string]string{"calico/k8s_ns": "kube-system"}}, true),
+		Entry("should accept label key where prefix is 253 characters", api.HostEndpointMetadata{Labels: map[string]string{"Projectcalico.org.projectcalico.org.projectcalico.org.projectcalico.org.projectcalico.org.projectcalico.org.projectcalico.org.projectcalico.org.projectcalico.org.projectcalico.org.projectcalico.org.projectcalico.org.projectcalico.org.projectcalico.org12/k8s_ns": "gold"}}, true),
+		Entry("should accept label key where name segment is 63 character", api.HostEndpointMetadata{Labels: map[string]string{"projectcalico.org/gold._0-9gold._0-9gold._0-9gold._0-9gold._0-9gold._0-9gold._012": "gold"}}, true),
 		Entry("should reject label key with !", api.HostEndpointMetadata{Labels: map[string]string{"rank!": "gold"}}, false),
-		Entry("should reject label value with !", api.HostEndpointMetadata{Labels: map[string]string{"rank": "gold!"}}, false),
+		Entry("should reject label key starting with ~", api.HostEndpointMetadata{Labels: map[string]string{"~rank_.0-9": "gold"}}, false),
+		Entry("should reject label key ending with ~", api.HostEndpointMetadata{Labels: map[string]string{"rank_.0-9~": "gold"}}, false),
+		Entry("should reject label key with multiple /", api.HostEndpointMetadata{Labels: map[string]string{"projectcalico.org/a/b": "gold"}}, false),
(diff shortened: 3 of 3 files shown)


lib/validator/validator.go
```diff
@@ -165,8 +165,8 @@ func validateTag(v *validator.Validate, topStruct reflect.Value, currentStructOr
 func validateLabels(v *validator.Validate, topStruct reflect.Value, currentStructOrField reflect.Value, field reflect.Value, fieldType reflect.Type, fieldKind reflect.Kind, param string) bool {
 	l := field.Interface().(map[string]string)
 	log.Debugf("Validate labels: %s", l)
-	for k, v := range l {
-		if !labelRegex.MatchString(k) || !labelRegex.MatchString(v) {
+	for k := range l {
```
gunjan5: shouldn't this be `k, _ := range l` ? range returns 2 values AFAIK

gunjan5: BTW, it's not wrong, it's just more clear and idiomatic Go :)

lib/validator/validator.go
```diff
@@ -165,8 +165,8 @@ func validateTag(v *validator.Validate, topStruct reflect.Value, currentStructOr
...
+		if !labelRegex.MatchString(k) {
```
gunjan5: why aren't we validating the value of the label?

robbrockbank: Speaking to Shaun, I don't think values need validating - although we should probably limit their length to match the length limit used by k8s.

robbrockbank:
I don't think the value actually needs to be validated - although we probably want to apply a sensible length limit.

Also, from the k8s docs:

Labels are key value pairs. Valid label keys have two segments: an optional prefix and name, separated by a slash (/). The name segment is required and must be 63 characters or less, beginning and ending with an alphanumeric character ([a-z0-9A-Z]) with dashes (-), underscores (_), dots (.), and alphanumerics between. The prefix is optional. If specified, the prefix must be a DNS subdomain: a series of DNS labels separated by dots (.), not longer than 253 characters in total, followed by a slash (/). If the prefix is omitted, the label key is presumed to be private to the user. Automated system components (e.g. kube-scheduler, kube-controller-manager, kube-apiserver, kubectl, or other third-party automation) which add labels to end-user objects must specify a prefix. The kubernetes.io/ prefix is reserved for Kubernetes core components.
Valid label values must be 63 characters or less and must be empty or begin and end with an alphanumeric character ([a-z0-9A-Z]) with dashes (-), underscores (_), dots (.), and alphanumerics between.

Might be worth chatting to @caseydavenport regarding what that means for us.  For example we currently don't support labels that start with a number.  Does that mean someone using such a label things will fail silently?

caseydavenport:
> For example we currently don't support labels that start with a number

@robbrockbank Is this a fundamental limit somewhere or did we just decide that for libcalico?

robbrockbank:
Not including a number at the start of the label was a specific decision made for the selector code some time ago.  The reason being was forward thinking to the possibility of needing to parse numbers ... if you allow labels to be numeric you can't really do that.

However, we may just have to relax that restriction.  See above note for comment from Shaun regarding where the code is in the selector parsing.  And the validator matches the selector.

heschlie:
I was just adding a new regex for the label values, I'm happy to make similar changes to the label key if we want to relax that restriction now.

robbrockbank:
Cool i just realize I'd misread that blob of text and thought that values weren't validated, but they do need to be.  Glad you spotted that!

heschlie:
I've added the new regex for the Label values and updated the Label key regex, I've also added a couple of tests to validate the changes, thoughts?

robbrockbank:
It's missing the equivalent change in tokenizer.go (see comment above).   That'll also need some UT updates.

lib/validator/validator.go
```diff
@@ -34,7 +34,8 @@ var validate *validator.Validate
 var (
 	nameRegex          = regexp.MustCompile("^[a-zA-Z0-9_.-]{1,128}$")
 	interfaceRegex     = regexp.MustCompile("^[a-zA-Z0-9_-]{1,15}$")
-	labelRegex         = regexp.MustCompile("^[a-zA-Z_./-][a-zA-Z0-9_./-]{0,127}$")
+	labelRegex         = regexp.MustCompile("^([a-zA-Z0-9.-]{0,253}/)?[a-zA-Z0-9]([a-zA-Z0-9_./-]{0,61}[a-zA-Z0-9])?$")
```
robbrockbank:
```
"^([a-zA-Z0-9.-]{0,253}/)?[a-zA-Z0-9]([a-zA-Z0-9_./-]{0,61}[a-zA-Z0-9])?$"
                                                                                         ^don't this this slash should be here
```

lib/validator/validator.go
```diff
@@ -34,7 +34,8 @@ var validate *validator.Validate
...
+	labelValueRegex    = regexp.MustCompile("^[a-zA-Z0-9]?([a-zA-Z0-9_./-]{0,61}[a-zA-Z0-9])?$")
```
robbrockbank: Again I don't think the `/` should be in the regex

heschlie: ah good catch

lib/selector/tokenizer/tokenizer.go
```diff
@@ -54,7 +54,7 @@ type Token struct {
 }
 
 const (
-	identifierExpr = `[a-zA-Z_./-][a-zA-Z0-9_./-]*`
+	identifierExpr = `[a-zA-Z0-9]?([a-zA-Z0-9_.-]{0,61}[a-zA-Z0-9])*`
```
robbrockbank: I think this should match the labelRegex no?  (i.e. the full thing)  We don't, unless I'm missing something, strip the subdomain prefix... or does the subdomain prefix never get passed to us?

@caseydavenport what say you?

Also, I think we need to make sure this regex is UT'd  -  I believe there are plenty of UTs in the selector codebase

caseydavenport: > or does the subdomain prefix never get passed to us?

We don't do anything with the labels provided to k8s.  As in, we don't modify them..  they come to us unchanged.

heschlie: I'll update the regex, as you are probably right and we'll need to see the prefixes as well.  Looking into adding UTs for the regex.

lib/validator/validator_test.go
```diff
@@ -110,10 +110,19 @@ func init() {
...
+		Entry("should reject label key ending with ~", api.HostEndpointMetadata{Labels: map[string]string{"rank_.0-9~": "gold"}}, false),
+		Entry("should reject label key with multiple /", api.HostEndpointMetadata{Labels: map[string]string{"tigera.io/a/b": "gold"}}, false),
+		Entry("should reject label key where prefix > 253 characters", api.HostEndpointMetadata{Labels: map[string]string{"tigera.iotigera.iotigera.iotigera.iotigera.iotigera.iotigera.iotigera.iotigera.iotigera.iotigera.iotigera.iotigera.iotigera.iotigera.iotigera.iotigera.iotigera.iotigera.iotigera.iotigera.iotigera.iotigera.iotigera.iotigera.iotigera.iotigera.iotigera.iotigera.io/rank_.0-9": "gold"}}, false),
+		Entry("should reject label key where name segment > 63 character", api.HostEndpointMetadata{Labels: map[string]string{"tigera.io/gold._0-9gold._0-9gold._0-9gold._0-9gold._0-9gold._0-9gold._0123": "gold"}}, false),
+		Entry("should reject label value starting with ~", api.HostEndpointMetadata{Labels: map[string]string{"rank_.0-9": "~gold"}}, false),
+		Entry("should reject label value ending with ~", api.HostEndpointMetadata{Labels: map[string]string{"rank_.0-9": "gold~"}}, false),
```
caseydavenport: Can you add a UT for `calico/k8s_ns: kube-system`

lib/validator/validator_test.go
```diff
@@ -110,10 +110,19 @@ func init() {
 
 		// (API) Labels.
 		Entry("should accept a valid label", api.HostEndpointMetadata{Labels: map[string]string{"rank_.0-9": "gold._0-9"}}, true),
-		Entry("should reject label key starting with 0-9", api.HostEndpointMetadata{Labels: map[string]string{"2rank": "gold"}}, false),
-		Entry("should reject label value starting with 0-9", api.HostEndpointMetadata{Labels: map[string]string{"rank": "2gold"}}, false),
+		Entry("should accept label key starting with 0-9", api.HostEndpointMetadata{Labels: map[string]string{"2rank": "gold"}}, true),
+		Entry("should accept label value starting with 0-9", api.HostEndpointMetadata{Labels: map[string]string{"rank": "2gold"}}, true),
+		Entry("should accept label key with dns prefix", api.HostEndpointMetadata{Labels: map[string]string{"tigera.io/rank_.0-9": "gold"}}, true),
```
caseydavenport: We should switch all the `tigera.io` stuff to `projectcalico.org`

lib/validator/validator_test.go
```diff
@@ -110,10 +110,19 @@ func init() {
...
-		Entry("should reject label value starting with 0-9", api.HostEndpointMetadata{Labels: map[string]string{"rank": "2gold"}}, false),
+		Entry("should accept label key starting with 0-9", api.HostEndpointMetadata{Labels: map[string]string{"2rank": "gold"}}, true),
+		Entry("should accept label value starting with 0-9", api.HostEndpointMetadata{Labels: map[string]string{"rank": "2gold"}}, true),
+		Entry("should accept label key with dns prefix", api.HostEndpointMetadata{Labels: map[string]string{"calico/k8s_ns": "kube-system"}}, true),
+		Entry("should accept label key where prefix is 253 characters", api.HostEndpointMetadata{Labels: map[string]string{"Projectcalico.org.projectcalico.org.projectcalico.org.projectcalico.org.projectcalico.org.projectcalico.org.projectcalico.org.projectcalico.org.projectcalico.org.projectcalico.org.projectcalico.org.projectcalico.org.projectcalico.org.projectcalico.org12/k8s_ns": "gold"}}, true),
+		Entry("should accept label key where name segment is 63 character", api.HostEndpointMetadata{Labels: map[string]string{"tigera.io/gold._0-9gold._0-9gold._0-9gold._0-9gold._0-9gold._0-9gold._012": "gold"}}, true),
```
gunjan5: nit: still got 2 more places with `tigera.io` :)

heschlie: There were 3 actually >_<

caseydavenport:
OK LGTM - @robbrockbank?

robbrockbank:
Yup LGTM

gunjan5:
I don't see the CI ✅  on the list, or is it just me?

robbrockbank:
it's still building

gunjan5:
It had disappeared from the list for some reason, it's back now (building...)

PR #309, opened 2016-12-15
Add IP Pool support to k8s backend

caseydavenport:
Also investigating moving the various list/get/create/delete methods into their own packages so we don't end up with a giant k8s.go with a whole bunch of logic in it.

We can also move the "conversion" methods out into their own resource specific directories to break up that monolith as well.

### DIFF
--- a/lib/backend/k8s/k8s.go
+++ b/lib/backend/k8s/k8s.go
@@ -23,7 +23,9 @@ import (
 	log "github.com/Sirupsen/logrus"
 
 	"github.com/projectcalico/libcalico-go/lib/backend/api"
+	"github.com/projectcalico/libcalico-go/lib/backend/k8s/resources"
 	"github.com/projectcalico/libcalico-go/lib/backend/k8s/thirdparty"
+	"github.com/projectcalico/libcalico-go/lib/backend/k8s/utils"
 	"github.com/projectcalico/libcalico-go/lib/backend/model"
 	"github.com/projectcalico/libcalico-go/lib/errors"
 
@@ -51,6 +53,9 @@ type KubeClient struct {
 	// Contains methods for converting Kubernetes resources to
 	// Calico resources.
 	converter converter
+
+	// Clients for interacting with Calico resources.
+	ipPoolClient api.Client
 }
 
 type KubeConfig struct {
@@ -97,13 +102,13 @@ func NewKubeClient(kc *KubeConfig) (*KubeClient, error) {
 	config, err := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
 		&loadingRules, configOverrides).ClientConfig()
 	if err != nil {
-		return nil, k8sErrorToCalico(err, nil)
+		return nil, utils.K8sErrorToCalico(err, nil)
 	}
 
 	// Create the clientset
 	cs, err := kubernetes.NewForConfig(config)
 	if err != nil {
-		return nil, k8sErrorToCalico(err, nil)
+		return nil, utils.K8sErrorToCalico(err, nil)
 	}
 	log.Debugf("Created k8s clientSet: %+v", cs)
 
@@ -115,6 +120,9 @@ func NewKubeClient(kc *KubeConfig) (*KubeClient, error) {
 		clientSet: cs,
 		tprClient: tprClient,
 	}
+
+	// Create the Calico sub-clients.
+	kubeClient.ipPoolClient = resources.NewIPPools(cs, tprClient)
 
 	return kubeClient, nil
 }
@@ -150,6 +158,7 @@ func (c *KubeClient) ensureThirdPartyResources() error {
 // do not already exist.
 func (c *KubeClient) createThirdPartyResources() (bool, error) {
 	// Ensure a resource exists for Calico global configuration.
+	log.Info("Ensuring GlobalConfig ThirdPartyResource exists")
 	tpr := extensions.ThirdPartyResource{
 		ObjectMeta: kapiv1.ObjectMeta{
 			Name:      "global-config.projectcalico.org",
@@ -164,6 +173,12 @@ func (c *KubeClient) createThirdPartyResources() (bool, error) {
 		if !kerrors.IsAlreadyExists(err) {
 			return false, goerrors.New(fmt.Sprintf("failed to create ThirdPartyResource %s: %s", tpr.ObjectMeta.Name, err))
 		}
+	}
+
+	// Ensure the IP Pool TPR exists.
+	err = c.ipPoolClient.EnsureInitialized()
+	if err != nil {
+		return false, err
 	}
 	return true, nil
 }
@@ -240,6 +255,8 @@ func buildTPRClient(baseConfig *rest.Config) (*rest.RESTClient, error) {
 				*cfg.GroupVersion,
 				&thirdparty.GlobalConfig{},
 				&thirdparty.GlobalConfigList{},
+				&thirdparty.IpPool{},
+				&thirdparty.IpPoolList{},
 				&kapiv1.ListOptions{},
 				&kapiv1.DeleteOptions{},
 			)
@@ -259,6 +276,8 @@ func (c *KubeClient) Create(d *model.KVPair) (*model.KVPair, error) {
 	switch d.Key.(type) {
 	case model.GlobalConfigKey:
 		return c.createGlobalConfig(d)
+	case model.IPPoolKey:
+		return c.ipPoolClient.Create(d)
 	default:
 		log.Warn("Attempt to 'Create' using kubernetes backend is not supported.")
 		return nil, errors.ErrorOperationNotSupported{
@@ -274,6 +293,8 @@ func (c *KubeClient) Update(d *model.KVPair) (*model.KVPair, error) {
 	switch d.Key.(type) {
 	case model.GlobalConfigKey:
 		return c.updateGlobalConfig(d)
+	case model.IPPoolKey:
+		return c.ipPoolClient.Update(d)
 	default:
 		// If the resource isn't supported, then this is a no-op.
 		log.Infof("'Update' for %+v is no-op", d.Key)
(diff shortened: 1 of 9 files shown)


caseydavenport:
@robbrockbank could you take a look at this?

lib/backend/k8s/syncer.go
```diff
@@ -267,20 +284,33 @@ func (syn *kubeSyncer) readFromKubernetesAPI() {
...
+				// We need to resync.  Break out into the sync loop.
+				log.Warn("Event triggered resync: %+v", event)
+				continue
+			}
+
+			// Event is OK - parse it and send it over the channel.
+			kvp = syn.parseIPPoolEvent(event)
+			latestVersions.poolVersion = kvp.Revision.(string)
+			syn.sendUpdates([]model.KVPair{*kvp})
+
```
caseydavenport: Extra newline

lib/backend/k8s/syncer.go
```diff
@@ -46,7 +48,7 @@ type kubeSyncer struct {
 	kc         KubeClient
 	callbacks  api.SyncerCallbacks
 	OneShot    bool
-	tracker    map[model.Key]interface{}
+	tracker    map[string]model.KVPair
```
caseydavenport: To do that I'd need to marshal the keys as json for the map keys, so we can unmarshal and delete them below.

At a minimum, just store the keys.

lib/backend/k8s/k8s.go
```diff
@@ -96,13 +101,13 @@ func NewKubeClient(kc *KubeConfig) (*KubeClient, error) {
 	config, err := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
 		&loadingRules, configOverrides).ClientConfig()
 	if err != nil {
-		return nil, k8sErrorToCalico(err, nil)
+		return nil, utils.K8sErrorToCalico(err, nil)
```
robbrockbank: aren't general purpose "utils" modules discouraged?

lib/backend/k8s/resources/ippools.go
```diff
@@ -0,0 +1,177 @@
...
+		// Don't care if it already exists.
+		if !kerrors.IsAlreadyExists(err) {
+			return goerrors.New(fmt.Sprintf("failed to create ThirdPartyResource %s: %s", tpr.ObjectMeta.Name, err))
+		}
+	}
+	return nil
+}
+
+// Syncer is not implemented for ip pools.
+func (c *client) Syncer(callbacks api.SyncerCallbacks) api.Syncer {
```
robbrockbank: I assume this is temporary - Felix will need to have Ip Pool notifications for NAT logic.

lib/backend/k8s/k8s.go
```diff
@@ -258,6 +275,8 @@ func (c *KubeClient) Create(d *model.KVPair) (*model.KVPair, error) {
 	switch d.Key.(type) {
 	case model.GlobalConfigKey:
 		return c.createGlobalConfig(d)
+	case model.IPPoolKey:
```
robbrockbank: Is there much of a cost in creating a client?  Just wondering if ipPoolClient should be a function which only instantiates a client when necessary.

`c.ipPools

robbrockbank:
Also.... merge conflicts.

lib/backend/k8s/resources/ippools.go
caseydavenport: I considered that briefly.

Unfortunately, I think the way the Syncer API works means it would be a lot of work to do this and would require the other resources to also implement.

Ultimately, I'd like to add a `Watch()` method to the Client interface which can be implemented by the resources.

lib/backend/k8s/k8s.go
```diff
@@ -149,6 +157,7 @@ func (c *KubeClient) ensureThirdPartyResources() error {
 // do not already exist.
 func (c *KubeClient) createThirdPartyResources() (bool, error) {
```
caseydavenport: It's so we can use the wait.PollImmediate above, but you're right we get all the information we need from the error, so we can simplify.

lib/backend/k8s/k8s.go
caseydavenport: There isn't much cost, but I don't think it matters much either way.  I like the way I've got it more though :)

lib/backend/k8s/k8s.go
```diff
@@ -162,10 +173,16 @@ func (c *KubeClient) createThirdPartyResources() (bool, error) {
...
 		}
 	}
-	return true, nil
+
+	// Ensure the IP Pool TPR exists.
+	err = c.ipPoolClient.EnsureInitialized()
+	if err != nil {
+		return err
+	}
+	return nil
```
robbrockbank: all this can be written:

`return c.ipPoolClient.EnsureInitialized()`

lib/backend/k8s/syncer.go
```diff
@@ -325,9 +354,9 @@ func (syn *kubeSyncer) performSnapshot() ([]model.KVPair, map[model.Key]bool, re
 			labels.Revision = profile.Revision
 
 			snap = append(snap, *rules, *tags, *labels)
-			keys[rules.Key] = true
-			keys[tags.Key] = true
-			keys[labels.Key] = true
+			keys[fmt.Sprintf("%s", rules.Key)] = true
```
robbrockbank: If we are stating here that the Key must be representable by a unique string then we should be explicit about that with the Key interface.

You either need to:
-  Add the Stringer interface to the Key interface with a clear comment that the string representation of the Key should be a unique identifier for the object within the cluster.... or
-  or, you could use the defaultPath() as the identifier ... although that is really for KV stores.
-  or add another method ... 

personally I'd plump for solidifying the requirement for each Key type to implement a unique String() representation.

lib/backend/k8s/syncer.go
```diff
@@ -379,7 +408,20 @@ func (syn *kubeSyncer) performSnapshot() ([]model.KVPair, map[model.Key]bool, re
 
 		for _, c := range confList {
 			snap = append(snap, *c)
-			keys[c.Key] = true
+			keys[fmt.Sprintf("%s", c.Key)] = true
+		}
+
+		// Sync IP Pools.
+		poolList, err := syn.kc.List(model.IPPoolListOptions{})
+		if err != nil {
```
robbrockbank: What error are you expecting here that a 1s sleep and retry will be sufficient?  It's not clear from the code/comments.

robbrockbank: Ah I see the comment at the top of the loop.  Should we be checking for a specific error type here?  Or perhaps the message "Error syncing" is misleading.  It's more "Error querying pools - waiting for k8s datastore to initialize" or something like that.

lib/backend/k8s/resources/ippools_conversion.go
```diff
@@ -0,0 +1,72 @@
...
+	kapi "k8s.io/client-go/pkg/api"
+)
+
+// ThirdPartyToIPPool takes a Kubrnetes ThirdPartyResource representation
+// of a Calico IP Pool and returns the equivalen IPPool object.
+func ThirdPartyToIPPool(t *thirdparty.IpPool) *model.KVPair {
+	v := model.IPPool{}
+	err := json.Unmarshal([]byte(t.Spec.Value), &v)
+	if err != nil {
+		panic(err)
```
robbrockbank: Should we panic here?  Better to raise a suitable error no?

lib/backend/k8s/thirdparty/ippool.go
```diff
@@ -0,0 +1,77 @@
...
+	"k8s.io/client-go/pkg/api/meta"
+	"k8s.io/client-go/pkg/api/unversioned"
+	"k8s.io/client-go/pkg/runtime/schema"
+)
+
+type IpPoolSpec struct {
+	Value string `json:"value"`
+}
+
+type IpPool struct {
```
robbrockbank: A lot of missing comments and non-standard Golang comments.

lib/backend/k8s/resources/ippools.go
```diff
@@ -0,0 +1,172 @@
...
+
+func (c *client) Apply(kvp *model.KVPair) (*model.KVPair, error) {
+	updated, err := c.Update(kvp)
+	if err != nil {
+		if _, ok := err.(errors.ErrorResourceDoesNotExist); !ok {
+			return nil, err
+		}
+
+		// It doesn't exist - create it.
+		updated, err = c.Create(kvp)
```
robbrockbank: Is there no equivalent to Apply in the k8s client?

Also, for future, is there a way to specify revision numbers to prevent concurrent updates - we'll want to expose that in the API eventually.

lib/backend/k8s/syncer.go
caseydavenport: Yeah, can do.  I don't really like this change at all.

I'd much prefer to make the requirement that a Key needs to be hashable, but IPPoolKey has a net.IPNet in it which breaks that.

We could tweak to use a pointer to a CIDR instead, but I that wouldn't work since two equal keys wouldn't hash to the same thing.

Maybe we could implement a hash method or something to make it hashable? I can't seem to find any examples of that.

lib/backend/k8s/syncer.go
caseydavenport: I think this is right the way it is.  We basically want to always retry here until we get a success.

 I can tweak the msg

lib/backend/k8s/resources/ippools_conversion.go
caseydavenport: I've switched to a fatal log that gives more detail.  I think its still ok to shutdown here, since it's basically "someone wrote something wrong to the datastore, this should never happen".

lib/backend/k8s/resources/ippools.go
caseydavenport: Nope, at least not in the restclient used for ThirdPartyResources.

This code already sets the revision numbers based on kvp.Revision.  See `IPPoolToThirdParty`

caseydavenport:
Huh, why isn't Semaphore running tests on this guy?

caseydavenport:
I've re-opened this here https://github.com/projectcalico/libcalico-go/pull/321

PR #313, opened 2016-12-21
Add Node calls to k8s backend

heschlie:
First go at getting Node related calls working in the k8s backend.

- Translates k8s Node to Calico Node
- Watches k8s API for Node changes
- Read only for Nodes (only supports list and get)

### DIFF
--- a/lib/backend/k8s/conversion_test.go
+++ b/lib/backend/k8s/conversion_test.go
@@ -25,6 +25,8 @@ import (
 	extensions "k8s.io/client-go/pkg/apis/extensions/v1beta1"
 	metav1 "k8s.io/client-go/pkg/apis/meta/v1"
 	"k8s.io/client-go/pkg/util/intstr"
+	"github.com/projectcalico/libcalico-go/lib/backend/k8s/resources"
+	"github.com/projectcalico/libcalico-go/lib/net"
 )
 
 var _ = Describe("Test parsing strings", func() {
@@ -671,5 +673,117 @@ var _ = Describe("Test Namespace conversion", func() {
 			Expect(err).NotTo(HaveOccurred())
 		})
 
+	})
+})
+
+var _ = Describe("Test Node conversion", func() {
+
+	It("should parse a k8s Node to a Calico Node", func() {
+		l := map[string]string{"net.beta.kubernetes.io/role": "master"}
+		node := k8sapi.Node{
+			ObjectMeta: k8sapi.ObjectMeta{
+				Name: "TestNode",
+				Labels: l,
+				ResourceVersion: "1234",
+			},
+			Spec: k8sapi.NodeSpec{},
+			Status: k8sapi.NodeStatus{
+				Addresses: []k8sapi.NodeAddress{
+					k8sapi.NodeAddress{
+						Type: k8sapi.NodeInternalIP,
+						Address: "172.17.17.10",
+					},
+					k8sapi.NodeAddress{
+						Type: k8sapi.NodeExternalIP,
+						Address: "192.168.1.100",
+					},
+					k8sapi.NodeAddress{
+						Type: k8sapi.NodeHostName,
+						Address: "172-17-17-10",
+					},
+				},
+
+			},
+
+		}
+
+		n, err := resources.K8sNodeToCalico(&node)
+		Expect(err).NotTo(HaveOccurred())
+
+		// Ensure we got the correct values
+		felixAddress := n.Value.(*model.Node).FelixIPv4
+		bgpAddress := n.Value.(*model.Node).BGPIPv4Addr
+		bgpAddressV6 := n.Value.(*model.Node).BGPIPv6Addr
+		labels := n.Value.(*model.Node).Labels
+
+		ip := net.ParseIP("172.17.17.10")
+
+		Expect(felixAddress).To(Equal(ip))
+		Expect(bgpAddress).To(Equal(ip))
+		Expect(*bgpAddressV6).To(Equal(net.IP{}))
+		Expect(labels).To(Equal(l))
+	})
+
+	It("should error on an invalid IP", func() {
+		l := map[string]string{"net.beta.kubernetes.io/role": "master"}
+		node := k8sapi.Node{
+			ObjectMeta: k8sapi.ObjectMeta{
+				Name: "TestNode",
+				Labels: l,
+				ResourceVersion: "1234",
+			},
+			Spec: k8sapi.NodeSpec{},
+			Status: k8sapi.NodeStatus{
+				Addresses: []k8sapi.NodeAddress{k8sapi.NodeAddress{
+					Type: k8sapi.NodeInternalIP,
+					Address: "1720.17.17.10",
+				}},
+
+			},
+
+		}
+
+		_, err := resources.K8sNodeToCalico(&node)
+		Expect(err).To(HaveOccurred())
+	})
+
+	It("should error on missing NodeInternalIP", func() {
+		l := map[string]string{"net.beta.kubernetes.io/role": "master"}
+		node := k8sapi.Node{
+			ObjectMeta: k8sapi.ObjectMeta{
+				Name: "TestNode",
+				Labels: l,
+				ResourceVersion: "1234",
+			},
+			Spec: k8sapi.NodeSpec{},
+			Status: k8sapi.NodeStatus{},
+
+		}
+
+		_, err := resources.K8sNodeToCalico(&node)
+		Expect(err).To(HaveOccurred())
+	})
+
+	It("should error when only NodeExternalIP is provided", func() {
+		l := map[string]string{"net.beta.kubernetes.io/role": "master"}
+		node := k8sapi.Node{
+			ObjectMeta: k8sapi.ObjectMeta{
+				Name: "TestNode",
+				Labels: l,
+				ResourceVersion: "1234",
+			},
+			Spec: k8sapi.NodeSpec{},
+			Status: k8sapi.NodeStatus{
+				Addresses: []k8sapi.NodeAddress{k8sapi.NodeAddress{
+					Type: k8sapi.NodeExternalIP,
+					Address: "172.17.17.10",
+				}},
+
+			},
+
+		}
+
+		_, err := resources.K8sNodeToCalico(&node)
+		Expect(err).To(HaveOccurred())
 	})
 })
--- a/lib/backend/k8s/k8s.go
+++ b/lib/backend/k8s/k8s.go
@@ -55,6 +55,7 @@ type KubeClient struct {
 
 	// Clients for interacting with Calico resources.
 	ipPoolClient api.Client
+	nodeClient api.Client
 }
 
 type KubeConfig struct {
@@ -122,6 +123,7 @@ func NewKubeClient(kc *KubeConfig) (*KubeClient, error) {
 
 	// Create the Calico sub-clients.
 	kubeClient.ipPoolClient = resources.NewIPPools(cs, tprClient)
+	kubeClient.nodeClient = resources.NewNodeClient(cs, tprClient)
 
 	return kubeClient, nil
 }
@@ -281,6 +283,8 @@ func (c *KubeClient) Create(d *model.KVPair) (*model.KVPair, error) {
 		return c.createGlobalConfig(d)
 	case model.IPPoolKey:
 		return c.ipPoolClient.Create(d)
+	case model.NodeKey:
+		return c.nodeClient.Create(d)
 	default:
 		log.Warn("Attempt to 'Create' using kubernetes backend is not supported.")
 		return nil, errors.ErrorOperationNotSupported{
@@ -298,6 +302,8 @@ func (c *KubeClient) Update(d *model.KVPair) (*model.KVPair, error) {
 		return c.updateGlobalConfig(d)
 	case model.IPPoolKey:
 		return c.ipPoolClient.Update(d)
+	case model.NodeKey:
+		return c.nodeClient.Update(d)
 	default:
 		// If the resource isn't supported, then this is a no-op.
 		log.Infof("'Update' for %+v is no-op", d.Key)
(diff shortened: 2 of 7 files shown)


caseydavenport:
@heschlie I've got another WIP that should be going in soon here: https://github.com/projectcalico/libcalico-go/pull/309

It adds IP pools, but does so in a slightly different way than has been done for other resources.  I think we'll want to update this PR to follow the same format.

lib/backend/k8s/conversion.go
```diff
@@ -361,3 +362,29 @@ func (c converter) k8sPortToCalico(port extensions.NetworkPolicyPort) []numorstr
...
+	// Get the internal IP of the node, should be 1-1 with calico node IP
+	nodeIP := ""
+	for _, address := range node.Status.Addresses {
+		if address.Type == kapiv1.NodeInternalIP {
+			nodeIP = address.Address
+			break
+		}
+	}
+
+	ip, _, _ := net.ParseCIDR(nodeIP)
```
caseydavenport: We should be handling the error here rather than ignoring it :)

lib/backend/k8s/conversion.go
```diff
@@ -361,3 +362,29 @@ func (c converter) k8sPortToCalico(port extensions.NetworkPolicyPort) []numorstr
...
+
+func (c converter) k8sNodeToCalico(node *kapiv1.Node) *model.KVPair {
+	// Get the internal IP of the node, should be 1-1 with calico node IP
+	nodeIP := ""
+	for _, address := range node.Status.Addresses {
+		if address.Type == kapiv1.NodeInternalIP {
+			nodeIP = address.Address
+			break
+		}
+	}
```
caseydavenport: What happens if there is no NodeInternalIP specified?

We can probably do one of a few things.

Maybe, if no NodeInternalIP, then:
- use a NodeExternalIP, if none exists then
- use any other configured IP, if none exists then
- treat as "node does not exist"

I wonder if we should be doing anything in the case there are multiple NodeInternalIP addresses specified.  Not sure what we could do though.

lib/backend/k8s/conversion.go
```diff
@@ -361,3 +362,29 @@ func (c converter) k8sPortToCalico(port extensions.NetworkPolicyPort) []numorstr
...
+
+	ip, _, _ := net.ParseCIDR(nodeIP)
+	asn := numorstring.ASNumber(64512)
+	return &model.KVPair{
+		Key:    model.NodeKey{
+			Hostname: node.Name,
+		},
+		Value:  &model.Node{
+			FelixIPv4:   ip,
```
caseydavenport: Indentation looks a little off to me... does this pass gofmt?

lib/backend/k8s/k8s.go
```diff
@@ -655,3 +659,38 @@ func (c *KubeClient) getHostConfig(k model.HostConfigKey) (*model.KVPair, error)
...
+
+func (c *KubeClient) listNodes(k model.NodeListOptions) ([]*model.KVPair, error) {
+	nodes, err := c.clientSet.Nodes().List(kapiv1.ListOptions{})
+	if err != nil {
+		k8sErrorToCalico(err, k)
+	}
+
+	kvp := []*model.KVPair{}
+	for _, node := range nodes.Items {
+		n, err := c.getNode(model.NodeKey{Hostname: node.Name})
```
caseydavenport: This is going to trigger a whole bunch of API queries to get each node (potentially thousands).

The Nodes().List() call will already return all the nodes, so we should just use those :)

lib/backend/k8s/syncer.go
```diff
@@ -437,6 +477,22 @@ func (syn *kubeSyncer) parseNamespaceEvent(e watch.Event) []model.KVPair {
...
+	node, ok := e.Object.(*k8sapi.Node)
+	if !ok {
+		log.Panicf("Invalid node event. Type: %s, Object: %+v", e.Type, e.Object)
+	}
+
+	kvp := syn.kc.converter.k8sNodeToCalico(node)
+
+	switch e.Type {
+	default:
+		syn.labelCache[kvp.Key.(model.NodeKey).Hostname] = kvp.Value.(*model.Node).Labels
```
caseydavenport: Hmm, a switch with only a default seems a bit weird!

We need to handle deletes by setting the `kvp.Value` to nil.

Also, the `labelCache` field is meant for Pods only - we don't need to do the equivalent for nodes since we care about all updates to the Node object.

`parseNetworkPolicyEvent` is probably the closest to what you're looking for.

lib/backend/k8s/k8s.go
```diff
@@ -655,3 +659,38 @@ func (c *KubeClient) getHostConfig(k model.HostConfigKey) (*model.KVPair, error)
...
 	return []*model.KVPair{}, nil
 }
+
+func (c *KubeClient) listNodes(k model.NodeListOptions) ([]*model.KVPair, error) {
+	nodes, err := c.clientSet.Nodes().List(kapiv1.ListOptions{})
+	if err != nil {
+		k8sErrorToCalico(err, k)
+	}
+
+	kvp := []*model.KVPair{}
```
caseydavenport: I'd make this a plural `kvps` since it's a slice

heschlie:
@caseydavenport This still needs work, which calls for Node do we need to replicate?  Are we just looking to fetch info?  Does the node need to support EnsureInitialized?

caseydavenport:
@heschlie Yeah, we need read-only support right now (Get, List, Syncer).  The rest can be no-ops / errors.

heschlie:
Great!  I'll get this passing, should have this ready for merge today.

heschlie:
My tests running `make test-containerized` are passing locally, it looks like semaphoreci is complaining about missing etcdctl in `$PATH` did we make changes to the build container recently?

heschlie:
Looks like the builds are failing due to an upstream issue with Alpine.  They updated to etcd v3.0.15 and their build logs show it is not compiling the binaries right (they made this change less than 2 days ago) I've filed a bug with them to get it resolved.

lib/backend/k8s/resources/node.go
```diff
@@ -0,0 +1,131 @@
...
+	return &nodeClient{
+		clientSet: c,
+		tprClient: r,
+	}
+}
+
+// Implements the api.Client interface for Nodes.
+type nodeClient struct{
+	clientSet *kubernetes.Clientset
+	tprClient *rest.RESTClient
```
caseydavenport: No need for the tprClient here.  It's not used for nodes since they're not ThirdPartyResources.

lib/backend/k8s/resources/node.go
```diff
@@ -0,0 +1,131 @@
...
+func (c *nodeClient) Update(kvp *model.KVPair) (*model.KVPair, error){
+	log.Warn("Operation Update no supported on Node type")
+	return nil, errors.ErrorOperationNotSupported{
+		Identifier: kvp.Key,
+		Operation: "Update",
+	}
+}
+
+func (c *nodeClient) Apply(kvp *model.KVPair) (*model.KVPair, error){
+	log.Warn("Operation Apply no supported on Node type")
```
caseydavenport: Typo in all of these messages: "no" should be "not".  

"type" should also be changed "resource"

Why not through a simple "is" in while you're at it :)

"Operation 'apply' is not supported for the Node resource"

lib/backend/k8s/resources/node.go
```diff
@@ -0,0 +1,131 @@
...
+func (c *nodeClient) Get(key model.Key) (*model.KVPair, error){
+	log.Debug("Received Get request on Node type")
+	node, err := c.clientSet.Nodes().Get(key.(model.NodeKey).Hostname, metav1.GetOptions{})
+	if err != nil {
+		return nil, K8sErrorToCalico(err, key)
+	}
+
+	cn := K8sNodeToCalico(node)
+
+	kvp := &model.KVPair{
```
caseydavenport: Hm, K8sNodeToCalico() already returns a *model.KVPair.  Why do you need to re-create it here?

lib/backend/k8s/resources/node.go
```diff
@@ -0,0 +1,131 @@
...
+	if err != nil {
+		K8sErrorToCalico(err, list)
+	}
+
+	kvps := []*model.KVPair{}
+
+	for _, node := range nodes.Items {
+		n := K8sNodeToCalico(&node)
+
+		kvp := &model.KVPair{
```
caseydavenport: Same question here.

lib/backend/k8s/resources/node.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	return kvps, nil
+}
+
+func (c *nodeClient) EnsureInitialized() error{
+	return nil
+}
+
+func (c *nodeClient)EnsureCalicoNodeInitialized(node string) error {
```
caseydavenport: Weird spacing here

lib/net/ip.go
```diff
@@ -42,6 +42,12 @@ func (i *IP) UnmarshalJSON(b []byte) error {
 	return i.UnmarshalText([]byte(s))
 }
 
+// ParseIP returns an IP from a string
```
caseydavenport: What does this do for an invalid IP address?

heschlie: The base call will return `nil` if it is invalid, I could do a nil check and throw an error so the problem doesn't make it further down the pipe before being caught, WDYT?

caseydavenport: It's nice to keep the behavior consistent with the `net` package.  We should just be making sure to do the nil check when we call this function.

lib/backend/k8s/conversion_test.go
```diff
@@ -429,3 +431,38 @@ var _ = Describe("Test Namespace conversion", func() {
...
+		l := map[string]string{"net.beta.kubernetes.io/role": "master"}
+		node := k8sapi.Node{
+			ObjectMeta: k8sapi.ObjectMeta{
+				Name: "TestNode",
+				Labels: l,
+				ResourceVersion: "1234",
+			},
+			Spec: k8sapi.NodeSpec{},
+			Status: k8sapi.NodeStatus{
+				Addresses: []k8sapi.NodeAddress{k8sapi.NodeAddress{
```
caseydavenport: Can we add a test for:

- when there is an ExternalIP and an InternalIP provided?
- when there is just an ExternalIP provided?
- when there are no IPs provided?

lib/backend/k8s/conversion_test.go
```diff
@@ -429,3 +431,38 @@ var _ = Describe("Test Namespace conversion", func() {
...
+
+			},
+
+		}
+
+		n, err := resources.K8sNodeToCalico(&node)
+		Expect(err).NotTo(HaveOccurred())
+
+		// Ensure we got the correct values
+		address := n.Value.(*model.Node).FelixIPv4
```
caseydavenport: There are two IP fields on the model.Node object right? We should check both.

lib/backend/k8s/k8s_fv_test.go
```diff
@@ -453,4 +453,19 @@ var _ = Describe("Test Syncer API for Kubernetes backend", func() {
...
+		node := ""
+		By("Listing all Nodes", func() {
+			nodes, err := c.List(model.NodeListOptions{})
+			Expect(err).NotTo(HaveOccurred())
+			// Get the hostname so we can make a Get call
+			node = nodes[0].Key.(model.NodeKey).Hostname
+		})
+
+		By("Getting a specific node", func() {
+			_, err := c.Get(model.NodeKey{Hostname: node})
```
caseydavenport: We should assert that the returned node is the right one!

lib/backend/k8s/k8s_fv_test.go
```diff
@@ -453,4 +453,19 @@ var _ = Describe("Test Syncer API for Kubernetes backend", func() {
...
+	It("Should support getting and listing Nodes", func() {
+		node := ""
+		By("Listing all Nodes", func() {
+			nodes, err := c.List(model.NodeListOptions{})
+			Expect(err).NotTo(HaveOccurred())
+			// Get the hostname so we can make a Get call
+			node = nodes[0].Key.(model.NodeKey).Hostname
+		})
+
+		By("Getting a specific node", func() {
```
caseydavenport: What about tests for Apply, Delete, etc?

lib/backend/k8s/resources/node_conversion.go
```diff
@@ -0,0 +1,54 @@
...
+		if address.Type == kapiv1.NodeInternalIP {
+			nodeIP = address.Address
+			break
+		}
+	}
+	log.Debugf("Node IP is %s", nodeIP)
+
+	ip := net.ParseIP(nodeIP)
+	if ip == nil {
+		return nil, errors.New("Invalid IP received from k8s for Node")
```
caseydavenport: We should log out what the value of `nodeIP` is here.

caseydavenport:
A few comments - mostly around testing, but looking pretty good!

heschlie:
@caseydavenport Some better tests hopefully.  I also set the `BGPIPv6` to an empty `net.IP{}` struct pointer, which may or may not be the best idea?

caseydavenport:
OK @heschlie this LGTM.

Could you please rebase / squash to a single commit with a descriptive commit message?  Then I'll merge!

heschlie:
@caseydavenport squashed!

heschlie:
@caseydavenport fixed the conflict, should be good to go again

caseydavenport:
@heschlie You've got some build failures

```
Failed to compile backend:

# github.com/projectcalico/libcalico-go/lib/backend/k8s/resources
lib/backend/lib/backend/k8s/resources/node_conversion.go:50: unknown model.Node field 'BGPIPv4' in struct literal
lib/backend/lib/backend/k8s/resources/node_conversion.go:51: unknown model.Node field 'BGPIPv6' in struct literal
```

caseydavenport:
@heschlie I think we'll need to wait for v2.1.0 feature freeze to be over before we merge this.

PR #316, opened 2016-12-21
Added new iface EnsureInitialized to client

robbrockbank:
Set GUID in EnsureInitialized

### DIFF
--- a/lib/client/client.go
+++ b/lib/client/client.go
@@ -15,10 +15,10 @@
 package client
 
 import (
+	"encoding/hex"
+	goerrors "errors"
 	"io/ioutil"
 	"reflect"
-
-	"errors"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/kelseyhightower/envconfig"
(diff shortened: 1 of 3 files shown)


lib/client/client.go
```diff
@@ -47,6 +49,35 @@ func New(config api.CalicoAPIConfig) (*Client, error) {
 	return &cc, err
 }
 
+// EnsureInitialized is used to ensure the backend datastore is correctly
+// initialized for use by Calico.  This method is idempotent, it may be called
+// multiple times, and will have no effect if the datastore is already
+// correctly initialized.  Generally, explicit invocation of this method is not
+// required, as use of the calico/node container, or management of Node
+// resources will invoke this.
```
robbrockbank: See comment - I'm actually wondering if we need this to be public at all....I can't think of a time when an application will need to explicitly call this since the calico/node will end up invoking node resource management commands and therefore implicitly initialize the datastore.

WDYT?

caseydavenport:
Cool! LGTM.

2blesd2bstresd:
👍

PR #317, opened 2016-12-22
Move stop-etcd to separate target

SergK:
We need to run "teardown procedure" after finishing work with
libcalico-go code, we have stop for kubernetes master, so let's
move stop-etcd to separate target as well


gunjan5:
LGTM! Thanks for the contribution @SergK !

PR #318, opened 2016-12-23
Add global init

robbrockbank:


lib/client/client.go
```diff
@@ -288,3 +316,95 @@ func (c *Client) list(metadata unversioned.ResourceMetadata, helper conversionHe
...
+			return err
+		}
+		log.WithField(key, defaultValue).Info("Using previously configured global Felix config")
+	}
+	return nil
+}
+
+// ensureGlobalFelixConfig ensures the specific global BGP configuration has
+// been initialized, setting to the default value if not.
+func (c *Client) ensureGlobalBGPConfig(key string, defaultValue interface{}) error {
```
caseydavenport: Would it be worthwhile to combine this and ensureGlobalFelixConfig to something like this:

```go
func (c *Client) ensureGlobalConfig(kvp *model.KVPair) error
```

They look almost the same, with the main difference just being the kvp that gets used.

lib/client/client.go
```diff
@@ -288,3 +316,95 @@ func (c *Client) list(metadata unversioned.ResourceMetadata, helper conversionHe
...
+			return err
+		}
+		log.WithField(key, defaultValue).Info("Using previously configured global BGP config")
+	}
+	return nil
+}
+
+// ensureGlobalFelixInterfacePrefix ensures the global felix interface prefix
+// contains at least the prefix used by calico/node.
+func (c *Client) ensureGlobalFelixInterfacePrefix() error {
```
caseydavenport: It looks like Felix will default this, I think... https://github.com/projectcalico/felix/blob/c725a96a566f3116c419e48aa63c9ba6b6027541/go/felix/config/config_params.go#L116

Anyway, it feels pretty wrong to me that we set this here if this is `calico/node` specific.

lib/client/config.go
```diff
@@ -44,10 +44,11 @@ var logToBgp = map[string]string{
 }
 
 const (
-	GlobalDefaultASNumber       = 64512
-	GlobalDefaultLogLevel       = "info"
-	GlobalDefaultIPIP           = false
-	GlobalDefaultNodeToNodeMesh = true
+	GlobalDefaultASNumber        = 64512
+	GlobalDefaultLogLevel        = "none"
```
caseydavenport: "none" means we log nothing?  Why the switch from "info"?

lib/client/client.go
```diff
@@ -108,20 +111,45 @@ func (c *Client) EnsureInitialized() error {
...
+	// -  Log severity for file (LogSeverityFile)
+	// -  Log file path (LogFilePath)
+	guid := hex.EncodeToString(uuid.NewV4().Bytes())
+	if err := c.ensureGlobalFelixConfig("ClusterGUID", guid); err != nil {
+		return err
+	} else if err = c.ensureGlobalFelixConfig("ReportingIntervalSecs", "0"); err != nil {
+		return err
+	} else if err = c.ensureGlobalFelixConfig("IpInIpEnabled", GlobalDefaultIPIP); err != nil {
+		return err
+	} else if err = c.ensureGlobalFelixConfig("LogSeverityScreen", GlobalDefaultLogLevel); err != nil {
```
caseydavenport: Really I feel that `EnsureInitialized` should not be a place where we throw a lot of defaults for things.  It's meant to make sure that the datastore is initialized to a minimal level such that things won't break (e.g the right etcd directories, or k8s thirdpartyresources exist).

If calico/node wants certain defaults, then it should configure them itself.

lib/client/client.go
```diff
@@ -108,20 +111,45 @@ func (c *Client) EnsureInitialized() error {
...
-			log.WithError(err).Warnf("Failed to set ClusterGUID")
-			return err
-		}
-		log.Info("Using previously configured ClusterGUID")
+
+	// Ensure a number of global BGP config values are configured:
+	// -  Node to node mesh (node_mesh)
+	// -  Log severity (loglevel)
+	// -  Default node AS number
+	nm, _ := json.Marshal(nodeToNodeMesh{Enabled: GlobalDefaultNodeToNodeMesh})
```
caseydavenport: Presumably confd falls over if these don't exist?  That's probably fine for now then.

lib/backend/etcd/etcd.go
```diff
@@ -120,6 +121,35 @@ func (c *EtcdClient) EnsureInitialized() error {
...
+		return err
+	}
+
+	return nil
+}
+
+// EnsureInitialized makes sure that the etcd data is initialized for use by
+// Calico.
+func (c *EtcdClient) EnsureNodeInitialized(node string) error {
+	// The confd agent in calico/node requires the IPAM affinity directories
```
caseydavenport: confd isn't just for calico/node - confd will need this even when running outside of a container.

lib/backend/api/api.go
```diff
@@ -105,6 +105,10 @@ type Client interface {
 	// EnsureInitialized ensures that the backend is initialized
 	// any ready to be used.
 	EnsureInitialized() error
+
+	// EnsureNodeInitialized ensures that the backend is initialized
```
caseydavenport: Would be good to go into detail w.r.t when this should be called.  Before / after creating the node resource?

lib/backend/etcd/etcd.go
```diff
@@ -120,6 +121,35 @@ func (c *EtcdClient) EnsureInitialized() error {
 		log.Info("Ready flag is already set")
 	}
 
+	// The confd agent in calico/node requires a few directories to exist
+	// to be able to monitor correctly:
+	// -  IP pools parent directories
+	// -  BGP custom filters directories.
+	if err := c.ensureDirectory("/calico/v1/ipam/v4/pool"); err != nil {
```
robbrockbank: Exactly!  This is only because confd watches directories :-(

Ideally we only really want to do this configuration when using confd.  One thing I'd considered was to have a "sneaky" back door "backend" API that we hook into just for confd - in fact most of this process is only really required when using calico/node and confd.  it's a little naughty, but would be a temporary measure until we expand our API sufficiently to be able to drop confd.   WDYT?

lib/backend/etcd/etcd.go
robbrockbank: True (not that we document how to do that yet :-) )

lib/client/client.go
robbrockbank: I agree wholeheartedly and I was in two minds about how to do this.

One reason i didn't do it in the node was because the current node and global config is configurable using set/unset/get commands there is no concept of assign a value if not already assigned - so perhaps what I should really be doing is updating the calico/node image to handle values not yet being assigned, and make sure they use the global default value.  That in itself a bit of a PITA, partly because it requires a bunch of rework to the templates, and partly because it means we have to store the "default" values in both the golang code and the templates.

But that said, maybe we should bit the bullet and update the templates to handle not having default values assigned.  WDYT?

lib/client/client.go
robbrockbank: Yeah this is for confd.   Again though, to be honest I don't really like this being here.  Probably the "correct" thing to do here is update the confd templates to handle these values being missing, and to use the global default when they are missing.

lib/client/client.go
```diff
@@ -108,20 +111,45 @@ func (c *Client) EnsureInitialized() error {
...
+	if err := c.ensureGlobalBGPConfig("node_mesh", string(nm)); err != nil {
+		return err
+	} else if err = c.ensureGlobalBGPConfig("loglevel", GlobalDefaultLogLevel); err != nil {
+		return err
+	} else if err = c.ensureGlobalBGPConfig("as_num", GlobalDefaultASNumber); err != nil {
+		return err
+	}
+
+	// Ensure the global interface prefix config contains the standard
+	// prefix used by calico/node.
```
robbrockbank: well the interface prefix can be anything - just we happen to use "cali" for the container side of things.  There is nothing stopping anyone assigning another prefix prior to starting the node, for example ... so this method really should append the cali prefix if it isn't already there.  That said - perhaps this should also just be a feature of calico/node

lib/client/client.go
robbrockbank: Maybe - it's a little more bulky but perhaps if I'm removing a bunch of the defaults then it's not a lot of code.

lib/client/client.go
robbrockbank: Hmmm,  I needed to add this to get the node to work.... so perhaps that config parm is not currently in use.

lib/client/config.go
robbrockbank: because the default in calico/node was actually setting it to info.

robbrockbank: actually i don't think that was true .... set back to info.

lib/client/client.go
robbrockbank: speaking to Shaun - the default should work, so perhaps I was doing something wrong... I'll retry.  If I do need to include it (temporarily then I'll raise an issue with Felix and remove it once it's fixed).

caseydavenport:
This looks a lot better now :)

LGTM

PR #319, opened 2016-12-26
Add basic consul wrapper.

aensidhe:
Summary of changes:
- changes in Makefile to allow to exclude k8s tests, because of bug in Docker for mac. I can't map anything from /var on host to container, so I can't run kubelet and any tests dependent on it.
- changes in Makefile and run_uts to allow to specify exactly what config use for e2e tests.
- changes in tests: all e2e tests are now parameterized with at least one parameter: config to backend. Infrastructure for ease of use is included. As I can't run k8s kubelet I can't check if any of tests will run on it.
- consul backend.
- consul watcher. 

Unfortunately, there are no way to ask Consul to send a diff from specified Index to you. `consul watch` receives entire snapshot. That also means that we do not have notifications about deletion of keys. To mitigate this I'll update Ready key with same value. It increases it's version and we'll get notified about deletions.


lib/backend/backend_test.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2016 Tigera, Inc. All rights reserved.
+// Copyright (c) 2017 Tigera, Inc. All rights reserved.
```
matthewdupre: I think the original year needs to be left in too; so 2016-2017.

Makefile
```diff
@@ -23,11 +23,19 @@ ut: vendor
...
-test-containerized: $(BUILD_CONTAINER_MARKER) run-kubernetes-master 
+test-containerized: $(BUILD_CONTAINER_MARKER) run-kubernetes-master
 	docker run --rm --privileged --net=host \
 	-e PLUGIN=calico \
 	-v ${PWD}:/go/src/github.com/projectcalico/libcalico-go:rw \
-	$(BUILD_CONTAINER_NAME) bash -c 'make WHAT=$(WHAT) ut && chown $(shell id -u):$(shell id -g) -R ./vendor'
+	$(BUILD_CONTAINER_NAME) bash -c 'make WHAT=$(WHAT) SKIP=$(SKIP) ut && chown $(shell id -u):$(shell id -g) -R ./vendor'
+
+.PHONY: test-containerized-mac-docker
+## Run the tests in a conteiner without k8s. https://github.com/vyshane/kid/issues/14
```
matthewdupre: s/conteiner/container/

Makefile
```diff
@@ -47,7 +55,14 @@ run-etcd:
 	--advertise-client-urls "http://$(LOCAL_IP_ENV):2379,http://127.0.0.1:2379,http://$(LOCAL_IP_ENV):4001,http://127.0.0.1:4001" \
 	--listen-client-urls "http://0.0.0.0:2379,http://0.0.0.0:4001"
 
-run-kubernetes-master: stop-kubernetes-master run-etcd
+## Run consul as a container
+run-consul: run-etcd
```
matthewdupre: I might be missing something, but wouldn't it make more sense for run-kubernetes-master to depend on both run-consul and run-etcd than through this intermediate step?  Unless consul somehow depends on etcd?

lib/backend/backend_test.go
fasaxc: @aensidhe Please can you use this format for copyright statements in updated files: 

`// Copyright (c) <old year>-2017 Tigera, Inc. and other authors. All rights reserved.`  

The `and other authors` part give you credit without adding a whole new `(c)` line for every author.

If you add a new file, please use your own copyright statement (if desired) and the Apache license stanza (required).

aensidhe: Ok, I fix it. Thank you for clarification.

git-hooks/pre-commit
```diff
@@ -10,7 +10,13 @@ exec 1>&2
 
 changed_files=$(git diff --cached --name-only)
 changed_go_files=$(git diff --cached --name-only | grep -E '.go$' || true)
-all_go_files=$(find -name '*.go' | grep -v /vendor/)
+
+find_cmd="find"
+unamestr=`uname`
+if [[ "$unamestr" == 'Darwin' ]]; then
```
aensidhe: This implies that mac developer uses brew to install findutils which contains gfind.

fasaxc:
@aensidhe Thanks a lot for this contribution, we haven't forgotten about this, we're just trying to find enough time from someone with the right expertise to review such a large change.

fasaxc:
Should resolve the Consul part of https://github.com/projectcalico/felix/issues/782

lib/backend/consul/syncer.go
```diff
@@ -0,0 +1,178 @@
...
+
+	syn.callbacks.OnStatusUpdated(api.ResyncInProgress)
+	updates := make([]api.Update, len(results))
+	tombstones := map[model.Key]bool{}
+	for i, x := range results {
+		key := model.KeyFromDefaultPath(x.Key)
+		value, err := model.ParseValue(key, x.Value)
+		updateType := api.UpdateTypeKVNew
+
+		if tombstones[key] {
```
matthewdupre: I've probably missed something really obvious here, but isn't tombstones always empty?  I don't see anything actually creating keys.

lib/api/apiconfig.go
```diff
@@ -25,6 +26,7 @@ type DatastoreType string
 const (
 	EtcdV2     DatastoreType = "etcdv2"
 	Kubernetes DatastoreType = "kubernetes"
+	Consul     DatastoreType = "consul"
```
matthewdupre: Do you think we should call this consulv1 for potential forward compatibility?  I'm not a Consul expert, so I don't know how likely it is that there'd be an incompatible consulv2 in the near future.

lib/backend/consul/consul.go
```diff
@@ -0,0 +1,613 @@
...
+	}
+	deleteOp := &consulapi.KVTxnOp{Key: path, Verb: consulapi.KVDeleteTree}
+	if d.Revision != nil {
+		deleteOp.Index = d.Revision.(uint64)
+	}
+	log.Debugf("Delete Key: %s", path)
+
+	kv := c.Client.KV()
+
+	readyFlagPair := getReadyFlagPair()
```
matthewdupre: (I've looked through the rest of the code / the PR itself, but it should be documented in the comments.)

Dockerfile
```diff
@@ -5,7 +5,7 @@ MAINTAINER Casey Davenport <casey@tigera.io>
 # Install libc (to run calicoctl) and various build tools to build the vendored packages.
 ENV GLIBC_VERSION 2.23-r1
 RUN apk add --update bash git iproute2 curl make gcc alpine-sdk linux-headers && \
-  apk add --update-cache --repository http://dl-cdn.alpinelinux.org/alpine/edge/testing etcd && \
+  apk add --update-cache --repository http://dl-cdn.alpinelinux.org/alpine/edge/testing etcd consul && \
```
aensidhe: I thought that this is used only for testing and at the moment of writing etcd was made same. Difference is 22.1 Mb. I'll remove this dependency and rewrite some test code :)

lib/backend/consul/consul.go
```diff
@@ -0,0 +1,613 @@
...
+		return convertConsulError(err, d.Key)
+	}
+
+	if !ok || len(response.Errors) > 0 {
+		return convertArrayOfErrorsToError(convertTxnErrors(d.Key, ops, response.Errors))
+	}
+
+	// If there are parents to be deleted, delete these as well provided there
+	// are no more children.
+	// TODO: investigate if we need to do it in consul
```
aensidhe: As far as I can understand we need this in situation when there were keys `/a/b/c` and `/a/b/d`. So, children (c and d) got deleted and we want to clean up `/a/b'. For consul this is not a case, because if we didn't create `/a/b` explicitly it will return 404 on `/a/b`.
```
/ # consul kv put calico/v1/a/b/c 1
Success! Data written to: calico/v1/a/b/c
/ # consul kv get calico/v1/a/b
Error! No key exists at: calico/v1/a/b
```

But if we do create parents explicitly and logic of calico requires them to be cleaned up we will need to cleanup.

lib/backend/consul/syncer.go
aensidhe: No, this is a bug, sorry. We need to populate tombstones from current state initially.

lib/api/apiconfig.go
aensidhe: Let's do the versioning from the start. Just in case. Thank you.

lib/backend/consul/syncer.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+}
+
+func getReadFromNonServerOpts(clusterIndex uint64) *consulapi.QueryOptions {
+	return &consulapi.QueryOptions{WaitIndex: clusterIndex, AllowStale: true, RequireConsistent: false}
+}
+
+func repeatableSync(syn *consulSyncer, state *ClusterState, clusterIndex uint64) error {
+	results, meta, err := syn.client.KV().List(
+		"/calico/v1/",
```
aensidhe: @matthewdupre I just change the key for sync for "/calico".

lib/backend/consul/syncer.go
```diff
@@ -0,0 +1,178 @@
...
+			if err != nil {
+				log.WithError(err).Warn("Failed to get snapshot from consul")
+				break watchLoop
+			}
+		}
+	}
+}
+
+func getReadFromNonServerOpts(clusterIndex uint64) *consulapi.QueryOptions {
+	return &consulapi.QueryOptions{WaitIndex: clusterIndex, AllowStale: true, RequireConsistent: false}
```
aensidhe: This is a good question. Those options mean to force calico node to connect to nearest agent, not to the master of cluster to reduce load on them. And we can't receive a snapshot that has index smaller then we already have. But if you insist we can force connection to masters of cluster. 

About the question: I don't know, but of etcd2 syncer has code to workaround reading data from stale replica, so I decided to connect syncer to a replica too.

lib/backend/consul/consul.go
```diff
@@ -0,0 +1,613 @@
...
+	}
+
+	for _, parent := range parents {
+		log.Debugf("Delete empty Key: %s", parent)
+		_, err2 := kv.Delete(parent, nil)
+		if err2 != nil {
+			log.Debugf("Unable to delete parent: %s", err2)
+			break
+		}
+	}
```
aensidhe: This code was copy-pasted from etcd for consistency sake.

lib/backend/consul/consul.go
matthewdupre: OK yeah, I think this is alright then: you can't have keys under other keys in etcdv2.

lib/backend/consul/syncer.go
matthewdupre: I've looked at the meanings; and it seems RequireConsistent isn't as scary as it sounds; I think this is fine.

matthewdupre:
Thanks for the markups :) - looks good to me.

Remaining thing that would be nice to add is little documentation (in the projectcalico/calico repo).  Doesn't need to be much, and we should refer to it as experimental for a while.

I think we'll want to hold off from merging until 2.1 is out - @tomdee?

tomdee:
Yes, this won't make it into "2.1" but we can move libcalico-go development off master for that release if we want to get this merged.

lib/testutils/backends.go
```diff
@@ -0,0 +1,145 @@
...
+			Scheme:  config.Spec.ConsulScheme,
+		}
+		c, _ := consulapi.NewClient(&cfg)
+		kv := c.KV()
+		ops := consulapi.KVTxnOps{
+			&consulapi.KVTxnOp{Key: "calico", Verb: consulapi.KVDelete},
+			&consulapi.KVTxnOp{Key: "calico/", Verb: consulapi.KVDeleteTree},
+		}
+		_, _, _, err = kv.Txn(ops, nil)
+	case api.Kubernetes:
```
robbrockbank: I think we should either error here just so this doesn't get used by k8s tests without fixing it up.

robbrockbank:
Rather that the enhanceWithConfigs() approach, I think it would be easier to create a test runner that provides an interface to run tests using a specific  API config.  Something like the following:

```
import (
	"fmt"
	"github.com/projectcalico/libcalico-go/lib/api"
	"github.com/projectcalico/libcalico-go/lib/client"

	. "github.com/onsi/ginkgo"
	. "github.com/onsi/ginkgo/extensions/table"
)

func ClientTestRunner(description string, testRunner func(config *api.CalicoAPIConfig)) bool {
	for _, configFileName := range getConfigFileNames() {
		config, err := client.LoadClientConfig(configFileName)
		if err != nil {
			panic(err)
 		}
		Describe(fmt.Sprintf("%s (backend: %s)", description, config.Spec.DatastoreType),
			func(){testRunner(config)})
	}

	return true
}

var _ = ClientTestRunner("IPPool tests", func(apiConfig *api.CalicoAPIConfig) {

	DescribeTable("IPPool e2e tests",
		func(meta1, meta2 api.IPPoolMetadata, spec1, spec2 api.IPPoolSpec) {

			// Erase etcd clean.
			testutils.CleanBackend(apiConfig)

			// Create a new client.
			c, err := client.New(apiConfig)
			if err != nil {
				panic(err)
			}

	.... rest of test
```

So in the above I've taken the IP pool test and replaced the first `Describe` with a call to ClientTestRunner instead.  That makes multiple Describe calls one for each client.

I think that would have the smallest impact on the code and is clearer to see.

@matthewdupre @aensidhe  WDYT

westsouthnight:
When we can see that in master?

fasaxc:
Assigning to @robbrockbank who is refactoring the datastore API.

wjoel:
@fasaxc @robbrockbank this would be very useful for organizations using Mesos and Consul, but not etcd, and decrease the operational burden of adopting Calico. Is there anything I can help you with to get this merged? It seems to have stalled, despite positive reviews.

mpenate:
Hi guys, what's the status for this? Any thoughts on when we can see this in master?

beddari:
Very VERY interested in status updates, yes ...

robbrockbank:
Quick update.  I'm afraid this is currently stalled on some refactoring/data model updates that several of us are actively working on.  Current estimate for completing this work is mid-Nov - so a little ways out I'm afraid.

beddari:
Thanks for the update, appreciate it. Running etcd "just" for Calico is somewhat of a catch considered what else its model brings :) Landing this will be great for simplifying integration of Calico into existing environments. Can't wait 👍

zrml:
Much needed here too echoing @beddari 
Thank you guys

2014qirong:
Dear all,

When can you support the consul  in the official release ?

Oloremo:
Any updates?

fasaxc:
Unfortunately, looks like this has been abandoned by the original contributor (and our backend datastore architecture has changed a lot since it was written so it's not a trivial merge).  I'm going to close this but if anyone picks it back up, feel free to open a new PR.

mpenate:
The response time has been... quite long? And in the middle of that time a huge refactor appeared. It seems that this work needs to be started from scratch isn’t it?

PR #321, opened 2017-01-06
Add IP Pool support to k8s backend

caseydavenport:
Reopening https://github.com/projectcalico/libcalico-go/pull/309 after accidentally closing.

### DIFF
--- a/lib/backend/k8s/syncer.go
+++ b/lib/backend/k8s/syncer.go
@@ -22,6 +22,7 @@ import (
 	log "github.com/Sirupsen/logrus"
 	"github.com/projectcalico/libcalico-go/lib/backend/api"
 	"github.com/projectcalico/libcalico-go/lib/backend/compat"
+	"github.com/projectcalico/libcalico-go/lib/backend/k8s/resources"
 	"github.com/projectcalico/libcalico-go/lib/backend/k8s/thirdparty"
 	"github.com/projectcalico/libcalico-go/lib/backend/model"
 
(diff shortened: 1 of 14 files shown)


lib/backend/k8s/syncer.go
```diff
@@ -325,9 +354,9 @@ func (syn *kubeSyncer) performSnapshot() ([]model.KVPair, map[model.Key]bool, re
 			labels.Revision = profile.Revision
 
 			snap = append(snap, *rules, *tags, *labels)
-			keys[rules.Key] = true
-			keys[tags.Key] = true
-			keys[labels.Key] = true
+			keys[fmt.Sprintf("%s", rules.Key)] = true
```
robbrockbank: This can just be rules.Key.String() no?

lib/backend/k8s/syncer.go
```diff
@@ -348,14 +377,14 @@ func (syn *kubeSyncer) performSnapshot() ([]model.KVPair, map[model.Key]bool, re
 		for _, np := range npList.Items {
 			pol, _ := syn.kc.converter.networkPolicyToPolicy(&np)
 			snap = append(snap, *pol)
-			keys[pol.Key] = true
+			keys[fmt.Sprintf("%s", pol.Key)] = true
```
robbrockbank: sim.

lib/backend/k8s/syncer.go
```diff
@@ -365,32 +394,45 @@ func (syn *kubeSyncer) performSnapshot() ([]model.KVPair, map[model.Key]bool, re
 			wep, _ := syn.kc.converter.podToWorkloadEndpoint(&po)
 			if wep != nil {
 				snap = append(snap, *wep)
-				keys[wep.Key] = true
+				keys[fmt.Sprintf("%s", wep.Key)] = true
```
robbrockbank: ditto

robbrockbank: and other places too

robbrockbank:
okey dokey LGTM just a couple more comments... don't need to see again,

I'm still a little worried about Apply() since there are always timing windows when you split into two operations, but unless you start adding lots of retry mechanisms it probably isn't worth sorting out - instead just let it fail and the user can retry.

Issue #322, opened 2017-01-11
validator rejects labels with more than one `/`


gunjan5:
This commit line: https://github.com/projectcalico/libcalico-go/commit/9930a7e60f8ffd7b74bc9ced86637a93450f9b3f#diff-fe058abafc6b5378c4afc418edd45ef3L37

It seems to break our stars demo at: http://docs.projectcalico.org/v2.0/getting-started/kubernetes/tutorials/stars-policy/

with the following felix errors: 
```bash
root@node-01# k logs calico-node-66z7x -c calico-node -n kube-system | grep validation
2017-01-11 22:06:28.658 [WARNING][195] validation_filter.go 53: Validation failed; treating as missing error=error with field src_selector = 'k8s_ns/label/role == 'management-ui'' key=Policy(name=stars.allow-ui) value=order:1000,selector:"calico/k8s_ns == 'stars'",inbound:allow from selector "k8s_ns/label/role == 'management-ui'",outbound:allow
2017-01-11 22:06:53.382 [WARNING][195] validation_filter.go 53: Validation failed; treating as missing error=error with field src_selector = 'k8s_ns/label/role == 'management-ui'' key=Policy(name=client.allow-ui) value=order:1000,selector:"calico/k8s_ns == 'client'",inbound:allow from selector "k8s_ns/label/role == 'management-ui'",outbound:allow
2017-01-11 22:16:04.476 [WARNING][195] validation_filter.go 53: Validation failed; treating as missing error=error with field src_selector = 'k8s_ns/label/role == 'client'' key=Policy(name=stars.frontend-policy) value=order:1000,selector:"calico/k8s_ns == 'stars' && role == 'frontend'",inbound:allow tcp from selector "k8s_ns/label/role == 'client'" to ports 80,outbound:allow
``` 

Need to:
- [x] Fix libcalico-go validator code to not reject multiple `/`s 
- [x] Do a felix release with that commit pinned 
- [x] Need to add this to Calico `v2.0.1` release

caseydavenport:
This breaks NamespaceSelector in Kubernetes NetworkPolicy, so this is pretty serious.

caseydavenport:
A potential workaround for users would be using `calicoctl` to interact with Calico policy objects directly (instead of using `kubectl` and `NetworkPolicy`), and to use labels without more than one `/`.

[...]

Issue #307, opened 2016-12-13
Commonize label matching code in validator and tokenizer


robbrockbank:
The validator and the tokenizer both have their own regex for matching labels.   We should move this to a common place.  We validate selector strings using the selector parser, we should probably do something similar for validating labels.

gunjan5:
Fixed in #323

PR #323, opened 2017-01-12
Relax the label key validation

gunjan5:
- allows 0 to 253 chars
- alphanumeric, `-`, `_`, `/`, and `.`
- end with only alphanumeric char

### DIFF
--- a/lib/validator/validator.go
+++ b/lib/validator/validator.go
@@ -26,16 +26,18 @@ import (
 	"github.com/projectcalico/libcalico-go/lib/numorstring"
 	"github.com/projectcalico/libcalico-go/lib/scope"
 	"github.com/projectcalico/libcalico-go/lib/selector"
+	"github.com/projectcalico/libcalico-go/lib/selector/tokenizer"
+
 	"gopkg.in/go-playground/validator.v8"
 )
 
 var validate *validator.Validate
 
 var (
+	labelRegex         = regexp.MustCompile(`^` + tokenizer.LabelKeyMatcher + `$`)
+	labelValueRegex    = regexp.MustCompile("^[a-zA-Z0-9]?([a-zA-Z0-9_.-]{0,61}[a-zA-Z0-9])?$")
 	nameRegex          = regexp.MustCompile("^[a-zA-Z0-9_.-]{1,128}$")
 	interfaceRegex     = regexp.MustCompile("^[a-zA-Z0-9_-]{1,15}$")
-	labelRegex         = regexp.MustCompile("^([a-zA-Z0-9.-]{0,253}/)?[a-zA-Z0-9]([a-zA-Z0-9_.-]{0,61}[a-zA-Z0-9])?$")
-	labelValueRegex    = regexp.MustCompile("^[a-zA-Z0-9]?([a-zA-Z0-9_.-]{0,61}[a-zA-Z0-9])?$")
 	actionRegex        = regexp.MustCompile("^(allow|deny|log|pass)$")
 	backendActionRegex = regexp.MustCompile("^(allow|deny|log|next-tier|)$")
 	protocolRegex      = regexp.MustCompile("^(tcp|udp|icmp|icmpv6|sctp|udplite)$")
--- a/lib/validator/validator_test.go
+++ b/lib/validator/validator_test.go
@@ -101,6 +101,8 @@ func init() {
 		// (API) Selectors.  Selectors themselves are thorougly UT'd so only need to test simple
 		// accept and reject cases here.
 		Entry("should accept valid selector", api.EntityRule{Selector: "foo == \"bar\""}, true),
+		Entry("should accept valid selector with 'has' and a '/'", api.EntityRule{Selector: "has(calico/k8s_ns)"}, true),
+		Entry("should accept valid selector with 'has'and two '/'", api.EntityRule{Selector: "has(calico/k8s_ns/role)"}, true),
 		Entry("should reject invalid selector", api.EntityRule{Selector: "thing=hello &"}, false),
 
 		// (API) Tags.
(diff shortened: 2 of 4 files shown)


gunjan5:
Fixes #322

gunjan5:
CC @robbrockbank @heschlie

lib/validator/validator.go
```diff
@@ -34,7 +34,7 @@ var validate *validator.Validate
 var (
 	nameRegex          = regexp.MustCompile("^[a-zA-Z0-9_.-]{1,128}$")
 	interfaceRegex     = regexp.MustCompile("^[a-zA-Z0-9_-]{1,15}$")
-	labelRegex         = regexp.MustCompile("^([a-zA-Z0-9.-]{0,253}/)?[a-zA-Z0-9]([a-zA-Z0-9_.-]{0,61}[a-zA-Z0-9])?$")
+	labelRegex         = regexp.MustCompile(`^([a-zA-Z0-9\_\.\-\/]{0,252}[a-zA-Z0-9])?$`)
```
caseydavenport: Why did you change from quotes to backticks?

Also, do we really need to escape all of those?

caseydavenport: Also, why the switch to 252 characters?

heschlie:
@gunjan5 I believe you need to make a similar change here as well https://github.com/projectcalico/libcalico-go/blob/master/lib/selector/tokenizer/tokenizer.go#L57

lib/validator/validator.go
gunjan5: > Why did you change from quotes to backticks?

because you get a compile time error with some of those escape chars, so have to use runes.

> Also, do we really need to escape all of those?

yeah the only one that could have been without escape was `/` but I did it to be consistent.

> Also, why the switch to 252 characters?

so it's 252 (alphanumeric+specials) + 1 only alphanumeric so it doesn't end with a special char

caseydavenport: > yeah the only one that could have been without escape was / but I did it to be consistent.

Interesting, what is the behavior of `_`, for example, when not escaped?

Also, no need to be consistent - it's more confusing the unnecessarily escape something (might come back later and be confused by it)

caseydavenport: As should `-`, I think.

Also, looks like `.` doesn't need to be escaped either: https://regex101.com/r/PH8aQM/1

caseydavenport: > so it's 252 (alphanumeric+specials) + 1 only alphanumeric so it doesn't end with a special char

The old format was 253 alphanumeric+specials followed by a slash + 61  alphanumric+specials followed by an alphanumeric.  Right?

So, we've gone from total length 253+1+61+1 to 252+1?

caseydavenport: What about this regex?
```
^([a-zA-Z0-9_.-/]{0,253}/)?([a-zA-Z0-9][a-zA-Z0-9_.-]{0,61}[a-zA-Z0-9])?$
```

This is essentially the same as before, but it supports slashes and underscores in the prefix, and has two splits:
- The prefix.
- The value after the prefix.

For comparison:

```
^([a-zA-Z0-9_.-/]{0,253}/)?([a-zA-Z0-9][a-zA-Z0-9_.-]{0,61}[a-zA-Z0-9])?$
^([a-zA-Z0-9.-]{0,253}/)?[a-zA-Z0-9]([a-zA-Z0-9_.-]{0,61}[a-zA-Z0-9])?
```

gunjan5:
Fixes #307 as well

lib/validator/validator.go
```diff
@@ -26,16 +26,18 @@ import (
...
 	"github.com/projectcalico/libcalico-go/lib/selector"
+	"github.com/projectcalico/libcalico-go/lib/selector/tokenizer"
+
 	"gopkg.in/go-playground/validator.v8"
 )
 
 var validate *validator.Validate
 
 var (
+	labelRegex         = regexp.MustCompile(`^` + tokenizer.LabelKeyMatcher + `?$`)
```
caseydavenport: Hm. I think the `?` belongs in the LabelKeyMatcher.  Right?

The `?` means that the trailing `([a-zA-Z0-9_.-]{0,61}[a-zA-Z0-9])` can occur once or not at all.

lib/validator/validator_test.go
```diff
@@ -113,14 +113,12 @@ func init() {
...
 		Entry("should accept label key with dns prefix", api.HostEndpointMetadata{Labels: map[string]string{"calico/k8s_ns": "kube-system"}}, true),
-		Entry("should accept label key where prefix is 253 characters", api.HostEndpointMetadata{Labels: map[string]string{"Projectcalico.org.projectcalico.org.projectcalico.org.projectcalico.org.projectcalico.org.projectcalico.org.projectcalico.org.projectcalico.org.projectcalico.org.projectcalico.org.projectcalico.org.projectcalico.org.projectcalico.org.projectcalico.org12/k8s_ns": "gold"}}, true),
-		Entry("should accept label key where name segment is 63 character", api.HostEndpointMetadata{Labels: map[string]string{"projectcalico.org/gold._0-9gold._0-9gold._0-9gold._0-9gold._0-9gold._0-9gold._012": "gold"}}, true),
+		Entry("should accept label key where prefix is 253 characters", api.HostEndpointMetadata{Labels: map[string]string{"projectcalico.org.projectcalico.org.projectcalico.org.projectcalico.org.projectcalico.org.projectcalico.org.projectcalico.org.projectcalico.org.projectcalico.org.projectcalico.org.projectcalico.org.projectcalico.org.projectcalico.org.calico12345/k8s_ns": "gold"}}, true),
```
caseydavenport: Looks like we no longer test a key that starts with a capital?

gunjan5: not sure if that makes any difference. Most keys are not capitalized anyway

lib/validator/validator.go
gunjan5: you're absolutely right, changed it

lib/validator/validator_test.go
caseydavenport: I think it does.  We don't want to accidentally break the regex such that it doesn't accept those types.

Should be easy enough to add another entry.

gunjan5: ok done

gunjan5:
ok I finally got it working with felix master. Calico node was using an old felix hash somehow. Tested it with stars and advanced policy demos. @caseydavenport should we merge this and pin it for a felix RC? or cut a libcalico-go v1.0.1 and then felix?

gunjan5:
/CC @fasaxc or @neiljerram can do a Felix release once I get a 👍 from Casey, and this is merged

caseydavenport:
LGTM

PR #325, opened 2017-01-18
Improves Backend API Client tests

luizbafilho:
This PR improves a bit the backend api client coverage by the missing methods. I did not add the `EnsureCalicoNodeInitialized` and `EnsureInitialized` because they are backend dependent, and the idea is after the Consul code gets merged, in different PR we could loop through it backend and run the same tests.

### DIFF
--- a/lib/backend/backend_suite_test.go
+++ b/lib/backend/backend_suite_test.go
@@ -1,4 +1,4 @@
-package backend
+package backend_test
 
 import (
 	. "github.com/onsi/ginkgo"
--- a/lib/backend/backend_test.go
+++ b/lib/backend/backend_test.go
@@ -17,9 +17,9 @@ package backend_test
 import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
-	"github.com/projectcalico/libcalico-go/lib/backend/model"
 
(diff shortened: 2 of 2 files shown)


tomdee:
Rebuilding to see if that fixes CI

glide.yaml
```diff
@@ -43,3 +43,4 @@ import:
   - tools/clientcmd
 testImport:
 - package: github.com/onsi/gomega
+- package: github.com/onsi/ginkgo/ginkgo
```
tomdee: What is this added for?

luizbafilho: to run the tests

https://github.com/projectcalico/libcalico-go/blob/master/run-uts#L7

tomdee: I don't think adding ginkgo to the glide.yaml helps with the run-uts script

luizbafilho: You are right. Just tried to remove ginkgo and run glide, I thought that would make it download ginkgo. I hope we get a package manager as soon as possible; every single tool works in a different way.

I'll remove this line.

tomdee: Cool, if you could squash your commits into one, then I can merge.

luizbafilho:
@tomdee squash done. It should be good to go now.

PR #326, opened 2017-01-19
glide.lock: Update with missing test imports

tomdee:

PR #327, opened 2017-01-19
Makefile: Start using calico/go-build container

tomdee:
- Vendoring
- Running tests


tomdee:
Depends on #328 and #326

PR #328, opened 2017-01-19
testutils: remove etcdctl dependency

tomdee:


lib/testutils/cleanEtcd.go
```diff
@@ -15,15 +15,26 @@
...
+}
+
 // CleanEtcd is a utility function to wipe clean "/calico" recursively from etcd.
 func CleanEtcd() {
-	err := exec.Command("etcdctl", "rm", "/calico", "--recursive").Run()
-	if err != nil {
-		log.Println(err)
+	_, err := kapi.Delete(context.Background(), "/calico", &etcdclient.DeleteOptions{Dir: true, Recursive: true})
+	if err != nil && !etcdclient.IsKeyNotFound(err) {
+		panic(err)
```
gunjan5: can we use `log.Fatalf()` or something similar here instead of panicing?

lib/testutils/cleanEtcd.go
```diff
@@ -15,15 +15,26 @@
...
+	cfg := etcdclient.Config{Endpoints: []string{"http://127.0.0.1:2379"}}
+	c, _ := etcdclient.New(cfg)
+	kapi = etcdclient.NewKeysAPI(c)
+}
+
 // CleanEtcd is a utility function to wipe clean "/calico" recursively from etcd.
 func CleanEtcd() {
-	err := exec.Command("etcdctl", "rm", "/calico", "--recursive").Run()
-	if err != nil {
-		log.Println(err)
```
gunjan5: I think it's still worth logging even when the error is `KeyNotFound`. WDYT?

tomdee:
OK, updated so behavior is the same as before.

gunjan5:
@tomdee LGTM! squash and merge

PR #335, opened 2017-01-29
Fix bug with multiple rules in Kubernetes datastore driver

caseydavenport:
We were building rules by doing the cross product of peers * ports.  However, the lists we generated used pointers to the from / ports objects so that it's possible to represent "not present".

This meant that then there were multiple rules, we'd build the list using a pointer to the same struct, so all the rules were generated the same....

This is a quick fix which copies the struct for the rule calculation.  Once this goes in we should look at re-organizing the rule building code so that it no longer does a product of peers * ports (i.e two peers, two ports -> four resulting rules)

### DIFF
--- a/lib/backend/k8s/conversion.go
+++ b/lib/backend/k8s/conversion.go
@@ -32,6 +32,7 @@ import (
 	kapiv1 "k8s.io/client-go/pkg/api/v1"
 	extensions "k8s.io/client-go/pkg/apis/extensions/v1beta1"
 	metav1 "k8s.io/client-go/pkg/apis/meta/v1"
+	"k8s.io/client-go/pkg/util/intstr"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


lib/backend/k8s/conversion.go
```diff
@@ -337,6 +354,8 @@ func (c converter) k8sIngressRuleToCalico(r extensions.NetworkPolicyIngressRule,
 	}
 
 	// Combine desintations with sources to generate rules.
+	// TODO: This currently creates a lot of rules by making every combination of from / ports
+	// into a rule.  We can combine these so that we don't need as many rules!
```
matthewdupre: There are two possible compactions here.
- Instead of one port per rule, you should be able to do one protocol per rule.
- You could just ()||() all the selectors together (although we don't have a data model reference for anything >1.4; I'm assuming we haven't modified the allowed syntax).

caseydavenport: Yep, for the selector we'd need this I think:

```
nsSelector  || (namespace == 'x' && (podSelector || podSelector))
```

And then do a rule per-protocol using that selector.

PR #336, opened 2017-02-05
Implements etcd v3 backend

luizbafilho:
This PR implements the Etcd V3 backend.

I didn't implement anything related to run multiple backends because that is already done on Consul PR, but all tests are passing if you change it to use the v3.

Another concern I had was testing the Syncer, I could find any unit test to it, so I added one.

as discussed with @tomdee I chose to duplicate most of the code, instead of trying to share with v2 version, so I wouldn't have to touch a stable and critical code.

### DIFF
--- a/lib/backend/etcd/etcdv3.go
+++ b/lib/backend/etcd/etcdv3.go
@@ -1,0 +1,452 @@
+// Copyright (c) 2016 Tigera, Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


xiaods:
any update?

lib/backend/etcd/etcdv3.go
```diff
@@ -0,0 +1,452 @@
...
+// Apply an existing entry in the datastore.  This ignores whether an entry already
+// exists.
+func (c *EtcdV3Client) Apply(d *model.KVPair) (*model.KVPair, error) {
+	logCxt := log.WithFields(log.Fields{"key": d.Key, "value": d.Value, "ttl": d.TTL, "rev": d.Revision})
+	return c.apply(d, logCxt)
+}
+
+// Delete an entry in the datastore.  This errors if the entry does not exists.
+func (c *EtcdV3Client) Delete(d *model.KVPair) error {
+	key, err := model.KeyToDefaultDeletePath(d.Key)
```
fasaxc: Unless it's handled somewhere else that I didn't spot, this is bugged.  Since etcdv3 doesn't have a concept of a directory, deleting `foo/bar` may delete `foo/barrel` too.

caseydavenport:
Thanks for the effort on this - master now contains etcdv3 support, so I'm going to close this PR.

PR #337, opened 2017-02-07
Making etcd.go idiomatic

luizbafilho:
This PR makes the code more idiomatic. To do so, I remove the use of `else` statements which are mostly unnecessary given that most of the times there is a `return` statement in an `if` block.

This pattern comes directly from `Go` docs (https://golang.org/doc/effective_go.html#if). You can also verify when running a `linter` (https://goreportcard.com/report/github.com/projectcalico/libcalico-go#golint)

I think the Rule of thumb is, when the function returns only an error, it is ok to verify it in the same line:
```Go
if err := someFunction(); err != nil {
    //do something
}
```

But if returns multiple values, the error checking goes one line down:
```Go
value, err := anotherFunction(); 
if err != nil {
    //do something
}
```

By doing that, there is no need to use `if... else... if...` statements to have access to `value` variable in the rest of the code.

I have not touched in anything related to the business logic; I just removed `else` statements when they were not necessary.

If you agree with this, I would like to make this change in other files, always one PR per file, to make it easy to review.

### DIFF
--- a/lib/backend/etcd/etcd.go
+++ b/lib/backend/etcd/etcd.go
@@ -124,7 +124,7 @@ func (c *EtcdClient) EnsureInitialized() error {
 	return nil
 }
 
-// EnsureCalicoNodeInitialized() performs additional initialization required
+// EnsureCalicoNodeInitialized performs additional initialization required
 // by the calico/node components [startup/ipip-allocation/confd].  This is a
 // temporary requirement until the calico/node components are updated to not
 // require special etcd setup, or until the global and per-node config is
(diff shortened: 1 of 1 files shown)


tomdee:
LGTM

lib/backend/etcd/etcd.go
```diff
@@ -134,19 +134,21 @@ func (c *EtcdClient) EnsureCalicoNodeInitialized(node string) error {
...
+	dirs := []string{
+		"/calico/v1/ipam/v4/pool",
+		"/calico/v1/ipam/v6/pool",
+		"/calico/bgp/v1/global/custom_filters/v4",
+		"/calico/bgp/v1/global/custom_filters/v6",
+		"/calico/ipam/v2/host/" + node + "/ipv4/block",
+		"/calico/ipam/v2/host/" + node + "/ipv6/block",
+	}
+
+	for _, d := range dirs {
```
neiljerram: I guess the definition of dirs could even be inlined here.  Would that be more or less idiomatic?

lib/backend/etcd/etcd.go
```diff
@@ -218,8 +220,10 @@ func (c *EtcdClient) Get(k model.Key) (*model.KVPair, error) {
 	if err != nil {
 		return nil, err
 	}
+
 	log.Debugf("Get Key: %s", key)
-	if r, err := c.etcdKeysAPI.Get(context.Background(), key, etcdGetOpts); err != nil {
+	r, err := c.etcdKeysAPI.Get(context.Background(), key, etcdGetOpts)
+	if err != nil {
```
neiljerram: I think it was idiomatic to combine those onto a single line as we were doing?

lib/backend/etcd/etcd.go
```diff
@@ -233,11 +237,14 @@ func (c *EtcdClient) Get(k model.Key) (*model.KVPair, error) {
...
+	}
+
+	v, err := model.ParseValue(k, []byte(r.Node.Value))
+	if err != nil {
 		return nil, err
-	} else {
-		return &model.KVPair{Key: k, Value: v, Revision: r.Node.ModifiedIndex}, nil
 	}
+
+	return &model.KVPair{Key: k, Value: v, Revision: r.Node.ModifiedIndex}, nil
```
neiljerram: Agree with losing the 'else's though.

luizbafilho:
@neiljerram doing the error checking one line down it should only happen when we have **multiple** assignments, when only an error it's being returned, it is ok to do it in one line.

By doing that we avoid the use of `else` just to have access to that value. 

And I personally have never seen multiple assignments and error checking in a single line in any other go code I have read.

PR #341, opened 2017-02-14
Some tweaks to logging in kdd

caseydavenport:


lib/backend/k8s/k8s.go
```diff
@@ -55,7 +55,7 @@ type KubeClient struct {
 
 	// Clients for interacting with Calico resources.
 	ipPoolClient api.Client
-	nodeClient api.Client
+	nodeClient   api.Client
```
djosborne: ...okay

lib/backend/k8s/k8s.go
```diff
@@ -223,6 +223,7 @@ func (c *KubeClient) ensureClusterType() (bool, error) {
 		}
 		value = existingValue
 	}
+	log.Debugf("Setting ClusterType to '%s'", value)
```
djosborne: Would now be a good time for us to start making use of [logrus' `withField`](https://github.com/sirupsen/logrus#example)?

```
log.WithField("ClusterType", value).Debug("Setting ClusterType")
```

lib/backend/k8s/resources/node_conversion.go
```diff
@@ -37,7 +39,7 @@ func K8sNodeToCalico(node *kapiv1.Node) (*model.KVPair, error) {
 	ip := net.ParseIP(nodeIP)
 	log.Debugf("Node IP is %s", ip)
```
djosborne: Maybe should this go in an `else` or be removed

lib/backend/k8s/k8s.go
caseydavenport: `gofmt` fixed this.

caseydavenport:
@djosborne any chance we could get this in? I'd like to have the log level tweaks.

lib/backend/k8s/k8s.go
```diff
@@ -278,6 +279,7 @@ func (c *KubeClient) Syncer(callbacks api.SyncerCallbacks) api.Syncer {
 
 // Create an entry in the datastore.  This errors if the entry already exists.
 func (c *KubeClient) Create(d *model.KVPair) (*model.KVPair, error) {
+	log.Debugf("Performing 'Create' for %+v", d)
```
robbrockbank: Better to use WithField() here

lib/backend/k8s/syncer.go
```diff
@@ -274,7 +275,7 @@ func (syn *kubeSyncer) readFromKubernetesAPI() {
 			log.Debugf("Incoming NetworkPolicy watch event. Type=%s", event.Type)
 			if needsResync = syn.eventTriggersResync(event); needsResync {
 				// We need to resync.  Break out into the sync loop.
-				log.Warn("Event triggered resync: %+v", event)
+				log.Warnf("Event triggered resync: %+v", event)
```
robbrockbank: Again, these should really use WithField() I think.

caseydavenport: Does WithField give you %+v formatting?

robbrockbank: Ah I see - no idea TBH.  Probably not, in which case LGTM.

robbrockbank: I'm not sure - but always doing the formatting when we're not logging doesn't feel right to me.  I guess it won't add much overhead so I'd be ok with merging this.

Would it be better though to add a String() method to the KVPair and then use WithField()

caseydavenport: I'm not too fussed about this one.

Looking at the logging code, the String formatting doesn't actually get done unless the correct log level is set, so it won't even have a perf impact.  Plus, this is a low-frequency warning log.

robbrockbank: Oh, of course!   Yes - in that case LGTM.

PR #343, opened 2017-02-21
Prevent tight loop due to repeated net errors.

### LINKED ISSUE #342: etcd Syncer can tight loop on repeated network failures
E.g. if etcd's port is refusing connections, we see this:
```
2017-02-20 17:00:49.441 [WARNING][20972] syncer.go 345: Net error from etcd error=dial tcp 127.0.0.1:2379: getsockopt: connection refused
2017-02-20 17:00:49.442 [WARNING][20972] syncer.go 345: Net error from etcd error=dial tcp 127.0.0.1:2379: getsockopt: connection refused
2017-02-20 17:00:49.442 [WARNING][20972] syncer.go 345: Net error from etcd error=dial tcp 127.0.0.1:2379: getsockopt: connection refused
...
```

fasaxc:
Can be caused by etcd closing its port, for example.

Fixes #342

### DIFF
--- a/lib/backend/etcd/syncer.go
+++ b/lib/backend/etcd/syncer.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2016 Tigera, Inc. All rights reserved.
+// Copyright (c) 2016-2017 Tigera, Inc. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -245,6 +245,7 @@ func sendSnapshotNode(node *client.Node, snapshotUpdates chan<- interface{}, res
 // so that the merge goroutine can trigger a new resync via snapshot.
 func (syn *etcdSyncer) watchEtcd(watcherUpdateC chan<- interface{}) {
 	log.Info("etcd watch thread started.")
+	var timeOfLastError time.Time
 	// Each trip around the outer loop establishes the current etcd index
 	// of the cluster, triggers a new snapshot read from that index (via
 	// message to the merge goroutine) and starts watching from that index.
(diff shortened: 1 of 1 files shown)


fasaxc:
Assigning to Rob since the doc says he owns libcalico-go ;-)

lib/backend/etcd/syncer.go
```diff
@@ -278,6 +279,13 @@ func (syn *etcdSyncer) watchEtcd(watcherUpdateC chan<- interface{}) {
...
 			resp, err := watcher.Next(context.Background())
 			if err != nil {
+				// Prevent a tight loop if etcd is repeatedly failing.
+				if time.Since(timeOfLastError) < 1*time.Second {
+					log.Warning("May be tight looping, throttling retries.")
+					time.Sleep(1 * time.Second)
+				}
+				timeOfLastError = time.Now()
+
 				if !retryableWatcherError(err) {
```
robbrockbank: Should this branch come before the throttling branch?

robbrockbank: Is there a risk that retryableWatcherError() returns the wrong value and we keep retrying a Watcher that is just broken and we should just exit instead?

robbrockbank:
couple of questions..  back to you.

lib/backend/etcd/syncer.go
fasaxc: Yes, that's what I wanted to guard against.  Since we're comparing with the last error time we'll only trigger the sleep if we hit two errors within 1s.

robbrockbank:
So, as discussed - this improves things, so I'm happy to merge - but I wonder if it goes far enough.

I'm worried about an internal error case where the watcher doesn't recover and we should exit the loop to resync.   @fasaxc - on your list to give that some more thought.

PR #344, opened 2017-02-21
Validate IPIP Mode

tmjd:
I was wanting to use the validator.Validate in calicoctl/calico_node/startup/startup.go but noticed it does not validate the IPIP Mode.  I think this is a valid check to include here.

### DIFF
--- a/lib/validator/validator.go
+++ b/lib/validator/validator.go
@@ -41,6 +41,7 @@ var (
 	actionRegex        = regexp.MustCompile("^(allow|deny|log|pass)$")
 	backendActionRegex = regexp.MustCompile("^(allow|deny|log|next-tier|)$")
 	protocolRegex      = regexp.MustCompile("^(tcp|udp|icmp|icmpv6|sctp|udplite)$")
+	ipipModeRegex      = regexp.MustCompile("^(always|cross-subnet|)$")
 	reasonString       = "Reason: "
 )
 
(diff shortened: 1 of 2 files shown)


lib/validator/validator.go
```diff
@@ -301,6 +302,21 @@ func validatePoolMetadata(v *validator.Validate, structLevel *validator.StructLe
 	}
 }
```
robbrockbank: If we are making this change - we should perhaps make the change to disallow IPIP for IPv6 pools.

lib/validator/validator.go
```diff
@@ -301,6 +302,21 @@ func validatePoolMetadata(v *validator.Validate, structLevel *validator.StructLe
...
 
+func validateIPIPConfiguration(v *validator.Validate, structLevel *validator.StructLevel) {
+	ipip := structLevel.CurrentStruct.Interface().(api.IPIPConfiguration)
+
+	log.Debugf("IPIP: Enabled %b, Mode: %s", ipip.Enabled, ipip.Mode)
```
robbrockbank: This could actually be done with a field level validator (see examples that just use a regex for checking values) at the top of the file.

lib/validator/validator.go
robbrockbank: You'd need to do this by adding a validator for the IPPool since you'd need to validate across the Metadata and the Spec.

robbrockbank:
LGTM

PR #347, opened 2017-02-22
Validate API data 

### LINKED ISSUE #346: Need to validate the API data
Currently we are only validating data that is passed in through calicoctl

I think we should get this in for v2.1.

robbrockbank:
Fixes #346

Will also fix https://github.com/projectcalico/calicoctl/issues/1526 once calicoctl glide is rev'd.

### DIFF
--- a/lib/api/unversioned/types.go
+++ b/lib/api/unversioned/types.go
@@ -30,8 +30,8 @@ var (
 //
 // All resource and resource lists embed a TypeMetadata as an anonymous field.
 type TypeMetadata struct {
-	Kind       string `json:"kind" validate:"required"`
-	APIVersion string `json:"apiVersion" validate:"required"`
+	Kind       string `json:"kind"`
+	APIVersion string `json:"apiVersion"`
 }
 
 func (md TypeMetadata) GetTypeMetadata() TypeMetadata {
--- a/lib/client/ippool_e2e_test.go
+++ b/lib/client/ippool_e2e_test.go
@@ -233,7 +233,7 @@ var _ = Describe("IPPool tests", func() {
 		// Test 5: Test starting with IPIP (cross subnet mode) and moving to no IPIP
 		Entry("IPIP (cross subnet mode) and moving to no IPIP",
 			api.IPPoolMetadata{CIDR: testutils.MustParseNetwork("10.0.0.0/24")},
-			api.IPPoolMetadata{CIDR: testutils.MustParseNetwork("10.10.10.0/24")},
+			api.IPPoolMetadata{CIDR: testutils.MustParseNetwork("fe80::00/120")},
 			api.IPPoolSpec{
 				IPIP: &api.IPIPConfiguration{
 					Enabled: true,
--- a/lib/client/ippool_e2e_test.go
+++ b/lib/client/ippool_e2e_test.go
@@ -270,7 +270,7 @@ var _ = Describe("IPPool tests", func() {
 
 		// Step-1: Test data validation occurs on create.
 		It("should invoke validation failure", func() {
-			By("Creating a pool with small CIDR")
+			By("Creating a pool with small CIDR (< /26)")
 			_, err = c.IPPools().Create(&api.IPPool{
 				Metadata: api.IPPoolMetadata{CIDR: testutils.MustParseCIDR("10.10.10.0/30")},
 				Spec:     api.IPPoolSpec{},
@@ -282,7 +282,7 @@ var _ = Describe("IPPool tests", func() {
 
 		// Step-2: Test data validation occurs on apply.
 		It("should invoke validation failure", func() {
-			By("Applying a pool with small CIDR")
+			By("Applying a pool with small CIDR (< /122)")
 			_, err = c.IPPools().Apply(&api.IPPool{
 				Metadata: api.IPPoolMetadata{CIDR: testutils.MustParseCIDR("aa:bb::cc/125")},
 				Spec:     api.IPPoolSpec{},
@@ -300,7 +300,7 @@ var _ = Describe("IPPool tests", func() {
 			})
 			Expect(err).NotTo(HaveOccurred())
 
-			By("Updating the pool using invalid settings")
+			By("Updating the pool using invalid settings (IPIP on IPv6 pool)")
 			_, err = c.IPPools().Update(&api.IPPool{
 				Metadata: api.IPPoolMetadata{CIDR: testutils.MustParseCIDR("aa:bb::cc/120")},
 				Spec:     api.IPPoolSpec{IPIP: &api.IPIPConfiguration{Enabled: true}},
--- a/lib/client/ippool_e2e_test.go
+++ b/lib/client/ippool_e2e_test.go
@@ -37,12 +37,14 @@ package client_test
 import (
 	"errors"
 	"log"
+	"reflect"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/ginkgo/extensions/table"
 	. "github.com/onsi/gomega"
 
 	"github.com/projectcalico/libcalico-go/lib/api"
+	cerrors "github.com/projectcalico/libcalico-go/lib/errors"
 	"github.com/projectcalico/libcalico-go/lib/ipip"
 	"github.com/projectcalico/libcalico-go/lib/testutils"
 )
@@ -219,16 +221,10 @@ var _ = Describe("IPPool tests", func() {
 			api.IPPoolMetadata{CIDR: testutils.MustParseNetwork("10.0.0.0/24")},
 			api.IPPoolMetadata{CIDR: testutils.MustParseNetwork("fe80::00/120")},
 			api.IPPoolSpec{
-				IPIP: &api.IPIPConfiguration{
-					Enabled: true,
-				},
 				NATOutgoing: true,
 				Disabled:    true,
 			},
 			api.IPPoolSpec{
-				IPIP: &api.IPIPConfiguration{
-					Enabled: true,
-				},
 				NATOutgoing: false,
 				Disabled:    false,
 			},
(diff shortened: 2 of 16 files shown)


lib/validator/validator.go
```diff
@@ -286,16 +285,28 @@ func validateHostEndpointSpec(v *validator.Validate, structLevel *validator.Stru
...
+	}
 
-	// The Calico IPAM places restrictions on the minimum IP pool size, check that the
-	// pool is at least the minimum size.
-	if pool.CIDR.IP != nil {
-		ones, bits := pool.CIDR.Mask.Size()
-		log.Debugf("Pool CIDR: %s, num bits: %d", pool.CIDR, bits-ones)
+	// The Calico IPAM places restrictions on the minimum IP pool size.  If
+	// the pool is enabled, check that the pool is at least the minimum size.
+	if !pool.Spec.Disabled {
```
tmjd: I wish the PoolMetadata check could have remained separate but I see that this is also changing the behavior to only apply the CIDR size check when the pool is Enabled.
In the previous check, `CIDR.IP` was checked for not null before checking the CIDR Mask, I don't understand the reasoning for that, but should that be done here too?

lib/api/unversioned/types.go
```diff
@@ -30,8 +30,8 @@ var (
 //
 // All resource and resource lists embed a TypeMetadata as an anonymous field.
 type TypeMetadata struct {
-	Kind       string `json:"kind" validate:"required"`
-	APIVersion string `json:"apiVersion" validate:"required"`
+	Kind       string `json:"kind"`
+	APIVersion string `json:"apiVersion"`
```
tmjd: Removing the validate here removes a check because some uses would not have the values.  Are there cases where `Kind` and `APIVersion` should be present and tests should be added in explicitly?

lib/client/ippool_e2e_test.go
```diff
@@ -237,7 +233,7 @@ var _ = Describe("IPPool tests", func() {
 		// Test 5: Test starting with IPIP (cross subnet mode) and moving to no IPIP
 		Entry("IPIP (cross subnet mode) and moving to no IPIP",
 			api.IPPoolMetadata{CIDR: testutils.MustParseNetwork("10.0.0.0/24")},
-			api.IPPoolMetadata{CIDR: testutils.MustParseNetwork("fe80::00/120")},
+			api.IPPoolMetadata{CIDR: testutils.MustParseNetwork("10.10.10.0/24")},
```
tmjd: Was this change and the same one below necessary?  I could understand if the 'spec1' was applied with 'meta2' but I did not see that in the test.

lib/validator/validator.go
robbrockbank: @caseydavenport asked for that additional change at the last minute :-)

Since there is cross Spec/Meta checking it didn't make sense to keep the PoolMetadata check.
We are still checking CIDR.IP - just that I moved it to the start of the validation method returning if the IP is nil.  On hindsight I prefer encasing the checks in that branch so I'll change it back - I think it's clearer that way and it's less error prone if people add more checks in the future.

lib/api/unversioned/types.go
robbrockbank: The Kind and APIVersion are only required for YAML/JSON input.  Since it is a typed interface the kind and version are implied by the structure types - generally, users of the programatic API won't have those fields filled in so validation will always fail - so we need to stop validating them.

calicoctl which does use them performs an explicit check on those fields.

lib/client/ippool_e2e_test.go
robbrockbank: oh for some reason i thought we swapped Specs with the two pools - maybe we don't.  It's not really testing anything new though having the IPv6 CIDR so I'm ok with it remaining like this - we do have another test that explicitly tests IPv6 CIDR.

robbrockbank: Nope not necessary - I reverted.  For some reason I thought we swapped specs but we only update the 1st resource.

tmjd:
So this LGTM but I still feel like someone that understands the places this could impact should have a look too.
Looks like this picked up a conflict from my IPIP change too.

lib/validator/validator.go
```diff
@@ -286,19 +285,32 @@ func validateHostEndpointSpec(v *validator.Validate, structLevel *validator.Stru
...
+		}
+
+		// The Calico IPAM places restrictions on the minimum IP pool size.  If
+		// the pool is enabled, check that the pool is at least the minimum size.
+		if !pool.Spec.Disabled {
+			ones, bits := pool.Metadata.CIDR.Mask.Size()
+			log.Debugf("Pool CIDR: %s, num bits: %d", pool.Metadata.CIDR, bits-ones)
+			if bits-ones < 6 {
+				structLevel.ReportError(reflect.ValueOf(pool.Metadata.CIDR),
+					"CIDR", "", reason("IP pool is too small"))
```
caseydavenport: Might be useful if the error message indicates that it's too small for an IPAM enabled Pool.

caseydavenport:
@robbrockbank I've taken a quick skim (haven't looked at tests) and it LGTM.

lib/validator/validator.go
```diff
@@ -288,31 +293,32 @@ func validateHostEndpointSpec(v *validator.Validate, structLevel *validator.Stru
...
-			structLevel.ReportError(reflect.ValueOf(ipip.Mode),
-				"Mode", "", reason("When IPIP Enabled, Mode should be always, <empty string>, or cross-subnet"))
+		// The Calico IPAM places restrictions on the minimum IP pool size.  If
+		// the pool is enabled, check that the pool is at least the minimum size.
+		if !pool.Spec.Disabled {
+			ones, bits := pool.Metadata.CIDR.Mask.Size()
+			log.Debugf("Pool CIDR: %s, num bits: %d", pool.Metadata.CIDR, bits-ones)
+			if bits-ones < 6 {
+				structLevel.ReportError(reflect.ValueOf(pool.Metadata.CIDR),
+					"CIDR", "", reason("IP pool CIDR is too small to an IPAM enabled pool"))
```
gunjan5: I'm confused by this error message wording. What do you mean by `an IPAM enabled pool`? Also, we should mention the minimum pool size in the message so user can know what is the minimum size if they see this error.

lib/validator/validator_test.go
```diff
@@ -261,16 +261,33 @@ func init() {
...
-		Entry("should accept IP pool with IPv4 CIDR /10", api.IPPoolMetadata{CIDR: netv4_4}, true),
-		Entry("should accept IP pool with IPv6 CIDR /122", api.IPPoolMetadata{CIDR: netv6_3}, true),
-		Entry("should accept IP pool with IPv6 CIDR /10", api.IPPoolMetadata{CIDR: netv6_4}, true),
-		Entry("should reject IP pool with IPv4 CIDR /27", api.IPPoolMetadata{CIDR: netv4_5}, false),
-		Entry("should reject IP pool with IPv6 CIDR /128", api.IPPoolMetadata{CIDR: netv6_1}, false),
+		Entry("should accept IP pool with IPv4 CIDR /26", api.IPPool{Metadata: api.IPPoolMetadata{CIDR: netv4_3}}, true),
+		Entry("should accept IP pool with IPv4 CIDR /10", api.IPPool{Metadata: api.IPPoolMetadata{CIDR: netv4_4}}, true),
+		Entry("should accept IP pool with IPv6 CIDR /122", api.IPPool{Metadata: api.IPPoolMetadata{CIDR: netv6_3}}, true),
+		Entry("should accept IP pool with IPv6 CIDR /10", api.IPPool{Metadata: api.IPPoolMetadata{CIDR: netv6_4}}, true),
+		Entry("should accept a disabled IP pool with IPv4 CIDR /27",
```
gunjan5: curious, why do we accept invalid pool size when they are disabled?

lib/client/ippool_e2e_test.go
```diff
@@ -265,4 +261,52 @@ var _ = Describe("IPPool tests", func() {
...
+
+		// Step-1: Test data validation occurs on create.
+		It("should invoke validation failure", func() {
+			By("Creating a pool with small CIDR")
+			_, err = c.IPPools().Create(&api.IPPool{
+				Metadata: api.IPPoolMetadata{CIDR: testutils.MustParseCIDR("10.10.10.0/30")},
+				Spec:     api.IPPoolSpec{},
+			})
+
+			Expect(err).To(HaveOccurred())
```
robbrockbank: Misleading though - since we reflect.TypeOf(nil) .... 

Better to be explicit.

lib/client/ippool_e2e_test.go
```diff
@@ -265,4 +261,52 @@ var _ = Describe("IPPool tests", func() {
...
+			By("Applying a pool with small CIDR")
+			_, err = c.IPPools().Apply(&api.IPPool{
+				Metadata: api.IPPoolMetadata{CIDR: testutils.MustParseCIDR("aa:bb::cc/125")},
+				Spec:     api.IPPoolSpec{},
+			})
+			Expect(err).To(HaveOccurred())
+			Expect(reflect.TypeOf(err)).To(Equal(valErrorType))
+		})
+
+		// Step-3: Test data validation occurs on update.
```
robbrockbank: Because there is no validation of IP pool metadata so it didn't seem like a very good end to end test.

I added those test in the node e2e file.

lib/validator/validator_test.go
robbrockbank: ask casey

lib/validator/validator.go
robbrockbank: an IPAM-enabled pool ... meaning a pool for which IPAM is enabled.

I could add the minimum pool size, but the message and it's construction will be a bit wieldy.

lib/validator/validator_test.go
gunjan5: ^ @caseydavenport

caseydavenport: It's not really an invalid pool size is why :)

We only limit to a `/26` because that is the minimum block size in Calico IPAM.  If the Pool isn't being used for IPAM, then there's no reason to require a certain size.

PR #350, opened 2017-02-24
Don't send hostNetwork pods through in snapshots.

### LINKED ISSUE #349: Kubernetes datastore driver sending host-networked pods
Saw this in the Felix logs, indicating that somehow the KDD is sending the calico/node Pod through to Felix even though it is a host-networked Pod (which it shouldn't do according to the code)

```
2017-02-16 19:46:00.975 [INFO][32] calc_graph.go 244: Local endpoint updated id=WorkloadEndpoint(node=gke-frank-gke-master-251-default-pool-8cbec5f5-1m7p, orchestrator=k8s, workload=kube-system.calico-node-dvjwp, name=eth0)
```

CC @frnkdny @fasaxc

caseydavenport:
Fixes #349

### DIFF
--- a/lib/backend/k8s/syncer.go
+++ b/lib/backend/k8s/syncer.go
@@ -361,6 +361,8 @@ func (syn *kubeSyncer) performSnapshot() ([]model.KVPair, map[string]bool, resou
 			time.Sleep(1 * time.Second)
 			continue
 		}
+		log.Info("Received Namespace List() response")
+
 		versions.namespaceVersion = nsList.ListMeta.ResourceVersion
 		for _, ns := range nsList.Items {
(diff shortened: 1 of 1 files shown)


robbrockbank:
Looks fine.

One related-isa question though:  Why are isCalicoPod, isHostNetworked and hasIPAddress methods on the converter rather than just plain old functions?

caseydavenport:
> One related-isa question though: Why are isCalicoPod, isHostNetworked and hasIPAddress methods on the converter rather than just plain old functions?

Yep, they don't need to be.  Not going to change in this PR though!

caseydavenport:
@tomdee is there a chance I could get this cherry-picked into v2.1.0?  Not critical, but would be useful for me.

Issue #345, opened 2017-02-22
Missing node information for WorkloadEndpoint using KDD 


caseydavenport:
Seen in Felix logs using the Kubernetes datastore driver.  Looks like there are some cases where we send through updates without a Node field populated.

```
10232 2017-02-16 20:59:04.579 [WARNING][32] stats_collector.go 110: Failed to get hostname key=WorkloadEndpoint(node=, orchestrator=k8s, workload=cali-ns-0.client0-cali-ns-0-765639624-00z3d, name=eth0)
10233 2017-02-16 20:59:04.579 [WARNING][32] stats_collector.go 110: Failed to get hostname key=WorkloadEndpoint(node=, orchestrator=k8s, workload=cali-ns-0.client0-cali-ns-0-765639624-01n3c, name=eth0)
10234 2017-02-16 20:59:04.581 [WARNING][32] stats_collector.go 110: Failed to get hostname key=WorkloadEndpoint(node=, orchestrator=k8s, workload=cali-ns-0.client0-cali-ns-0-765639624-05m2v, name=eth0)
10235 2017-02-16 20:59:04.582 [WARNING][32] stats_collector.go 110: Failed to get hostname key=WorkloadEndpoint(node=, orchestrator=k8s, workload=cali-ns-0.client0-cali-ns-0-765639624-06nkt, name=eth0)
```

Note that when this happened, we saw that one of the Kubernetes nodes had gone into a failed state (can't remember which state exactly), which might be related.

CC @frnkdny

caseydavenport:
@frnkdny do you have the `kubectl get nodes` output from this test?  I'd like to see exactly what state the bad node was in, and which node actually had a problem so I can correlate in the logs.

frnkdny:
node: ```gke-frank-gke-master-251-default-pool-8cbec5f5-rl8q``` was in a NotReady   (1h)

```
$ kubectl get nodes
NAME                                                  STATUS     AGE
gke-frank-gke-master-251-default-pool-8cbec5f5-rl8q   NotReady   1h
<snip>
```
Other resources with ```gke-frank-gke-master-251-default-pool-8cbec5f5-rl8q``` in the name:
```
$ kubectl get pods --all-namespaces
kube-system   fluentd-cloud-logging-gke-frank-gke-master-251-default-pool-8cbec5f5-rl8q   1/1       Running     2          1m
<snip>
[...]

PR #352, opened 2017-02-24
Log out if we ever get a missing NodeName

caseydavenport:
This will help diagnose #345 in the future.


lib/backend/k8s/conversion.go
```diff
@@ -219,6 +219,11 @@ func (c converter) podToWorkloadEndpoint(pod *kapiv1.Pod) (*model.KVPair, error)
 		ipNets = []cnet.IPNet{*ipNet}
 	}
 
+	// Check to see if the Pod has a Node.
+	if len(pod.Spec.NodeName) == 0 {
```
robbrockbank: If we do hit this - should we error out - the rest of the processing seems a little worrisome if we don't have a node name.

robbrockbank:
LGTM once the tests pass

PR #353, opened 2017-02-27
Add multiple clients for e2e testing

robbrockbank:
Add test infrastructure for running e2e tests using different client backends.

This PR also fixes a bug in the k8s IP pools List processing that was preventing IP pools from being Listed when the CIDR was specified in the request.

### DIFF
--- a/lib/testutils/client.go
+++ b/lib/testutils/client.go
@@ -82,12 +82,13 @@ func CreateNewIPPool(c client.Client, poolSubnet string, ipip, natOut, ipam bool
 
 // CleanIPPools removes all IP pool configuration from the datastore.
 func CleanIPPools(c *client.Client) {
-	// Delete IP Pools from the datastore
-	pools, err := c.IPPools().List(api.IPPoolMetadata{})
-	if err != nil {
+	if pools, err := c.IPPools().List(api.IPPoolMetadata{}); err == nil {
+		for _, pool := range pools.Items {
+			if err := c.IPPools().Delete(pool.Metadata); err != nil {
+				panic(err)
+			}
+		}
+	} else {
 		panic(err)
-	}
-	for _, pool := range pools.Items {
-		c.IPPools().Delete(pool.Metadata)
 	}
 }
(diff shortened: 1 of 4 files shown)


lib/testutils/e2e_describe.go
```diff
@@ -0,0 +1,56 @@
...
+)
+
+// E2eDescribe is a replacement for ginkgo.Describe which invoke Describe
+// multiple times for one or more different backend clients - passing in the
+// Calico API configuration as a parameter to the test function.  This allows
+// easy construction of end-to-end tests covering multiple different backends.
+//
+// The *clients* parameter is a bit-wise OR of the required client/backend
+// types that will be tested.
+func E2eDescribe(description string, clients int, body func(config api.CalicoAPIConfig)) bool {
```
matthewdupre: This is a function about datastore drivers, but it's called E2eDescribe() - that seems a little obfuscatey to me.

lib/testutils/client.go
```diff
@@ -79,3 +79,15 @@ func CreateNewIPPool(c client.Client, poolSubnet string, ipip, natOut, ipam bool
...
+
+// CleanIPPools removes all IP pool configuration from the datastore.
+func CleanIPPools(c *client.Client) {
+	// Delete IP Pools from the datastore
+	pools, err := c.IPPools().List(api.IPPoolMetadata{})
+	if err != nil {
+		panic(err)
+	}
+	for _, pool := range pools.Items {
+		c.IPPools().Delete(pool.Metadata)
```
matthewdupre: (Can this error?)

robbrockbank: Yup - probably should panic - given it's a test function.  Will fix!

lib/testutils/e2e_describe.go
robbrockbank: Renamed, and tweaked the wording to talk about datastore drivers rather than backends.

caseydavenport:
LGTM

PR #354, opened 2017-02-28
K8s datastore node Update call

heschlie:
- Added the `Node.Update` and `Node.Apply` calls to the k8s datastore driver
- Added Conversion of Node resource from Calico to k8s representation
  - Also updated the conversion mechanics from k8s to Calico to pull from new Annotations instead of default values
- Added BGP related fields as Annotations to the k8s Node

### DIFF
--- a/lib/backend/compat/compat.go
+++ b/lib/backend/compat/compat.go
@@ -102,6 +102,12 @@ func (c *ModelAdaptor) Update(d *model.KVPair) (*model.KVPair, error) {
 			return d, nil
 		}
 	case model.NodeKey:
+		// Quick fix for k8s Node call.  The k8s Node.Update call requires a complete model.Node object,
+		// therefore we do not need to break down the node object.
+		node, err := c.client.Update(d); if err == nil {
+		         return node, nil
+		}
+
 		p, o := toNodeComponents(d)
 		if p, err = c.client.Update(p); err != nil {
 			return nil, err
@@ -144,6 +150,12 @@ func (c *ModelAdaptor) Apply(d *model.KVPair) (*model.KVPair, error) {
 			return d, nil
 		}
 	case model.NodeKey:
+		// Quick fix for k8s Node call.  The k8s Node.Apply call requires a complete model.Node object,
+		// therefore we do not need to break down the node object.  The etcd API will raise an error
+		// if we try to hand it the composite object, hence we then try to apply the separate components.
+		node, err := c.client.Apply(d); if err == nil {
+			return node, nil
+		}
 		p, o := toNodeComponents(d)
 		if p, err = c.client.Apply(p); err != nil {
 			return nil, err
@@ -266,6 +278,12 @@ func (c *ModelAdaptor) getBlock(k model.Key) (*model.KVPair, error) {
 func (c *ModelAdaptor) getNode(nk model.NodeKey) (*model.KVPair, error) {
 	var err error
 
+	// Quick fix for k8s Node call.  The k8s Node.Get call returns a complete model.Node object, therefore we do
+	// not need to find each piece individually.
+	node, err := c.client.Get(nk); if err == nil {
+		return node, nil
+	}
+
 	// Fill in the Metadata specific part of the node configuration.  At the
 	// moment, there is nothing to fill in.
 	if _, err = c.client.Get(model.HostMetadataKey{nk.Hostname}); err != nil {
@@ -296,6 +314,15 @@ func validateBlockValue(kvp *model.KVPair) error {
 // because of the way we do our list queries - we'll enumerate all endpoints on
 // host as well!
 func (c *ModelAdaptor) listNodes(l model.NodeListOptions) ([]*model.KVPair, error) {
+	// The k8s List returns the complete list and values we're interested in, so we make a
+	// call and check if the list has anything in it, returning if it does.
+	nodes, err := c.client.List(l)
+	if err != nil {
+		return nil, err
+	} else if len(nodes) > 0 {
+		return nodes, nil
+	}
+	
 	hml := model.HostMetadataListOptions{Hostname: l.Hostname}
 	hmr, err := c.client.List(hml)
 	if err != nil {
--- a/lib/backend/k8s/conversion_test.go
+++ b/lib/backend/k8s/conversion_test.go
@@ -709,8 +709,11 @@ var _ = Describe("Test Node conversion", func() {
 				Name:            "TestNode",
 				Labels:          l,
 				ResourceVersion: "1234",
+				Annotations: map[string]string{
+					"projectcalico.org/IPv4Address": "172.17.17.10/24",
+					"projectcalico.org/ASNumber": "2546",
+				},
 			},
-			Spec: k8sapi.NodeSpec{},
 			Status: k8sapi.NodeStatus{
 				Addresses: []k8sapi.NodeAddress{
 					k8sapi.NodeAddress{
@@ -727,23 +730,26 @@ var _ = Describe("Test Node conversion", func() {
 					},
 				},
 			},
+			Spec: k8sapi.NodeSpec{},
 		}
 
 		n, err := resources.K8sNodeToCalico(&node)
 		Expect(err).NotTo(HaveOccurred())
 
-		// Ensure we got the correct values
-		felixAddress := n.Value.(*model.Node).FelixIPv4
-		bgpAddress := n.Value.(*model.Node).BGPIPv4Addr
-		bgpAddressV6 := n.Value.(*model.Node).BGPIPv6Addr
+		// Ensure we got the correct values.
+		felixAddress := *n.Value.(*model.Node).FelixIPv4
+		bgpAddress := *n.Value.(*model.Node).BGPIPv4Addr
+		bgpNet := *n.Value.(*model.Node).BGPIPv4Net
 		labels := n.Value.(*model.Node).Labels
+		asn := n.Value.(*model.Node).BGPASNumber
 
-		ip := net.ParseIP("172.17.17.10")
+		ip, ipNet, _ := net.ParseCIDR("172.17.17.10/24")
 
-		Expect(felixAddress).To(Equal(ip))
-		Expect(bgpAddress).To(Equal(ip))
-		Expect(*bgpAddressV6).To(Equal(net.IP{}))
+		Expect(felixAddress).To(Equal(*ip))
+		Expect(bgpAddress).To(Equal(*ip))
+		Expect(bgpNet).To(Equal(*ipNet))
 		Expect(labels).To(Equal(l))
+		Expect(asn.String()).To(Equal("2546"))
 	})
 
 	It("should error on an invalid IP", func() {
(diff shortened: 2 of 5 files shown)


heschlie:
Still need to add some tests for this as well

lib/backend/k8s/resources/node_conversion.go
```diff
@@ -24,23 +24,29 @@ import (
...
+	_, bgpV4Cidr, err := net.ParseCIDR(annotations["projectcalico.org/BGPIPv4Net"])
+	if err != nil {
+		log.Warnf("Could not parse BGPIPv4CIDR from k8s annotation: %s",
+			annotations["projectcalico.org/BGPIPv4Net"])
+	}
+
+	bgpAsn, err := numorstring.ASNumberFromString(annotations["projectcalico.org/BGPASNumber"])
+	if err != nil {
+		log.Warnf("Could not parse ASNumber from k8s annotation: %s",
+			annotations["projectcalico.org/BGPASNumber"])
```
caseydavenport: If we pull the values out of the annotations at the top, then we don't have to keep accessing them from the map like this all over.  I think I'd like that better.

lib/backend/k8s/resources/node_conversion.go
```diff
@@ -49,9 +55,31 @@ func K8sNodeToCalico(node *kapiv1.Node) (*model.KVPair, error) {
...
+			BGPASNumber: &bgpAsn,
+			BGPIPv4Net:  bgpV4Cidr,
 		},
 		Revision: node.ObjectMeta.ResourceVersion,
 	}, nil
 }
+
+// Convert a Calico Node to Kubernetes, place BGP configuration info in annotations
+func CalicoToK8sNode(kvp *model.KVPair) (*kapiv1.Node, error) {
+	vals := kvp.Value.(model.Node)
```
caseydavenport: vals is a weird name for this - it should be `calicoNode` or similar.

lib/backend/k8s/resources/node_conversion.go
```diff
@@ -49,9 +55,31 @@ func K8sNodeToCalico(node *kapiv1.Node) (*model.KVPair, error) {
...
+	annotations := map[string]string {
+		"projectcalico.org/BGPIPv4Addr": vals.BGPIPv4Addr.String(),
+		"projectcalico.org/BGPASNumber": vals.BGPASNumber.String(),
```
caseydavenport: If there's no `node.BGPASNumber`, then I think we shouldn't be applying the annotation (rather than populating it with a nil value).

Same goes for the other fields.

lib/backend/k8s/resources/node_conversion.go
```diff
@@ -49,9 +55,31 @@ func K8sNodeToCalico(node *kapiv1.Node) (*model.KVPair, error) {
...
+		"projectcalico.org/BGPIPv4Net":  vals.BGPIPv4Net.String(),
+	}
+
+	nodeMeta := kapiv1.ObjectMeta{
+		Name:        kvp.Key.(model.NodeKey).Hostname,
+		Annotations: annotations,
+		Labels:      vals.Labels,
+	}
+
+	node := &kapiv1.Node{
```
caseydavenport: Hmm, I don't think this is going to work.

We're returning a pretty much empty Node object.  When we write this to the API, it's probably going to delete a whole bunch of information on the Node.

For example, there's no `Spec` here.  Also, what happens if there are other annotations on the Node?

I think, unfortunately, we need to make the Update() method do a Get() from the k8s API for this node, then populate the fields we care about and write it back to the API. There's not enough information in a Calico "Node" resource to regenerate the entire Kubernetes "Node" object.

lib/backend/k8s/resources/node_conversion.go
```diff
@@ -15,43 +15,87 @@
...
+		ObjectMeta: nodeMeta,
+	}
+
+	return node, nil
+}
+
+// We take a k8s node and a Calico node and push the values from the Calico node into the k8s node
+func MakeK8sNode(kvp *model.KVPair, node *kapiv1.Node) (*kapiv1.Node, error) {
```
caseydavenport: Let's not allow people to pass in whatever KVP they want.  In general, best not to let people pass things that will break stuff.

So, this function should take a `model.Node` instead of a `model.KVPair`.

lib/backend/k8s/resources/node_conversion.go
```diff
@@ -15,43 +15,87 @@
...
+	calicoNode := kvp.Value.(model.Node)
+	node.Annotations["projectcalico.org/IPv4Address"] = calicoNode.BGPIPv4Net.String()
```
caseydavenport: `calicoNode.BGPIPv4Net` is a pointer - we need to do a nil pointer check before we try to call the `String()` method on it.

lib/backend/k8s/resources/node_conversion.go
```diff
@@ -15,43 +15,87 @@
...
+
+	// Don't set the ASNumber if it is nil
+	if calicoNode.BGPASNumber != nil {
```
caseydavenport: If the AS number is nil, we should probably be deleting the `projectcalico.org/ASNumber` annotation if it exists.

lib/backend/k8s/resources/node_conversion.go
```diff
@@ -15,43 +15,87 @@
...
+		node.Annotations["projectcalico.org/ASNumber"] = calicoNode.BGPASNumber.String()
+	}
+
+	// Add all of the calicoNode labels into the k8s node
+	for k, v := range calicoNode.Labels {
```
caseydavenport: This is a tiny bit scary.  I'm a little bit worried about us messing with labels on the underlying Node.

It's possible that it's the right thing to do, but also could lead to people unknowingly modifying their Nodes in ways they didn't expect.

On the other hand, users might expect to be able to modify the labels on a Node and have that change stick, so when it doesn't that could also be confusing.

WDYT?

CC @robbrockbank

caseydavenport:
@heschlie looks like you've also need to rebase.

lib/backend/k8s/resources/node_conversion.go
heschlie: I had a similar thought, and went with the expecting to see it come up on the k8s Node, it does seem like there is a definite risk of accidentally doing something naughty.  I assume the alternative would be to just not push Labels on to the k8s Node, would we still want to pull them off of the k8s node?

lib/backend/k8s/resources/node_conversion.go
```diff
@@ -15,43 +15,70 @@
...
+	// In order to make sure we always end up with a CIDR that has the IP and not just network
+	// we assemble the CIDR from BGPIPv4 and FelixIPv4.
+	subnet, _ := calicoNode.BGPIPv4Net.Mask.Size()
+	ipCidr := fmt.Sprintf("%s/%d", calicoNode.BGPIPv4Addr.String(), subnet)
+	k8sNode.Annotations["projectcalico.org/IPv4Address"] = ipCidr
+
+	// Don't set the ASNumber if it is nil.
+	if calicoNode.BGPASNumber != nil {
+		k8sNode.Annotations["projectcalico.org/ASNumber"] = calicoNode.BGPASNumber.String()
+	}
```
caseydavenport: I think we still need to do something here if the BGPASNumber is nil.

i.e if the node has the ASNumber annotation, but the provided Calico node does not, then we need to remove the annotation.

lib/backend/k8s/resources/node_conversion.go
```diff
@@ -15,43 +15,70 @@
...
+
+	// Add all of the calicoNode labels into the k8s node.
```
caseydavenport: Let's make labels read-only for now.

So, don't update labels on the node when doing an Apply / Update.

lib/backend/k8s/resources/node_conversion.go
```diff
@@ -15,43 +15,70 @@
...
-			BGPIPv4Addr: ip,
-			BGPIPv6Addr: &net.IP{},
-			BGPASNumber: &asn,
-		},
 		Revision: node.ObjectMeta.ResourceVersion,
-	}, nil
+	}
+
+	calicoNode := model.Node{}
+
```
caseydavenport: Unnecessary newlines here.

heschlie:
Another round of changes pushed for ya @caseydavenport, working on the FV tests now.

lib/backend/compat/compat.go
```diff
@@ -144,6 +150,11 @@ func (c *ModelAdaptor) Apply(d *model.KVPair) (*model.KVPair, error) {
 			return d, nil
 		}
 	case model.NodeKey:
+		// Quick fix for k8s Node call.  The k8s Node.Apply call requires a complete model.Node object,
+		// therefore we do not need to break down the node object.
```
robbrockbank: Probably worth adding to this comment that etcd will raise an error attempting to apply the composite Node model, hence we then attempt to apply the separate components.

robbrockbank: Just a thought - what's the behavior if the k8s call fails (e.g. transient connection issue)?  At that point we'll drop into trying to apply the separate components....  which I assume k8s will fail because it won't recognize the individual component types?

heschlie: Correct the KDD API will complain about not supporting those API calls with some no-op messages in the debug logs.

lib/backend/k8s/k8s_fv_test.go
```diff
@@ -532,13 +533,45 @@ var _ = Describe("Test Syncer API for Kubernetes backend", func() {
 		})
 
 		By("Applying changes to a node", func() {
```
caseydavenport: Can we do another Get() on the node after the Apply and assert that the right fields are present?

heschlie:
@caseydavenport I've fixed up the tests, they should be closer to what we want.

lib/backend/k8s/conversion_test.go
```diff
@@ -702,41 +702,31 @@ var _ = Describe("Test Node conversion", func() {
 				Name:            "TestNode",
 				Labels:          l,
 				ResourceVersion: "1234",
-			},
-			Spec: k8sapi.NodeSpec{},
-			Status: k8sapi.NodeStatus{
-				Addresses: []k8sapi.NodeAddress{
```
caseydavenport: Might want to leave some of these other IPs in, just for funsies, and to make sure we don't do anything weird if they're present (it's more realistic)

lib/backend/k8s/resources/node_conversion.go
```diff
@@ -15,43 +15,65 @@
...
+	kvp.Value = &calicoNode
+
+	return &kvp, nil
+}
+
+// mergeCalicoK8sNode takes a k8s node and a Calico node and push the values from the Calico
+// node into the k8s node.
+func mergeCalicoK8sNode(calicoNode *model.Node, k8sNode *kapiv1.Node) (*kapiv1.Node, error) {
+	// In order to make sure we always end up with a CIDR that has the IP and not just network
+	// we assemble the CIDR from BGPIPv4 and FelixIPv4.
```
caseydavenport: Comment is wrong - we don't use FelixIPv4 here.

lib/backend/k8s/resources/node_conversion.go
```diff
@@ -15,43 +15,65 @@
...
+	kvp.Value = &calicoNode
+
+	return &kvp, nil
+}
+
+// mergeCalicoK8sNode takes a k8s node and a Calico node and push the values from the Calico
+// node into the k8s node.
+func mergeCalicoK8sNode(calicoNode *model.Node, k8sNode *kapiv1.Node) (*kapiv1.Node, error) {
+	// In order to make sure we always end up with a CIDR that has the IP and not just network
+	// we assemble the CIDR from BGPIPv4 and BGPIPvNet.
```
caseydavenport: > from BGPIPv4 and BGPIPvNet

Redundant much? ;)

caseydavenport: `BGPIPv4Addr` I think, right?

heschlie: Oh for crying out loud I was looking right at it too...  should also be `BGPIPv4Net`

caseydavenport:
@heschlie OK this LGTM.

Once you squash all your commits then we can merge.

PR #356, opened 2017-03-01
Support InsecureSkipTLSVerify

neiljerram:
Allow k8s datastore users to set an environmental
variable (K8S_INSECURE_SKIP_TLS_VERIFY) that means to skip verification
of the API server's TLS certificate.

### DIFF
--- a/lib/backend/k8s/k8s.go
+++ b/lib/backend/k8s/k8s.go
@@ -59,12 +59,13 @@ type KubeClient struct {
 }
 
 type KubeConfig struct {
-	Kubeconfig     string `json:"kubeconfig" envconfig:"KUBECONFIG" default:""`
-	K8sAPIEndpoint string `json:"k8sAPIEndpoint" envconfig:"K8S_API_ENDPOINT" default:""`
-	K8sKeyFile     string `json:"k8sKeyFile" envconfig:"K8S_KEY_FILE" default:""`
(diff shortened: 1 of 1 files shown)


lib/backend/k8s/k8s.go
```diff
@@ -94,6 +95,9 @@ func NewKubeClient(kc *KubeConfig) (*KubeClient, error) {
 			*override.variable = override.value
 		}
 	}
+	if kc.K8sInsecureSkipTLSVerify {
```
caseydavenport: For consistency, probably best to put this in the existing overrides mechanism above.

https://github.com/projectcalico/libcalico-go/pull/356/files#diff-cfc4259ee76e5c99f93b10127d9736beR78

caseydavenport: Ah I see, it's a bool and that bit above is just strings.  I suppose this is fine then.

neiljerram: Yes indeed, that's why I did it this way.  (After trying the other way first :smile: )

No need for this to be merged soon, though; I only need it for my scale testing work, so best to wait until after we've released Calico 2.1.0.

neiljerram:
@caseydavenport Given that there's now a separate branch for 2.1 work, I think this PR could be merged - WDYT?

caseydavenport:
Yep, LGTM!

PR #360, opened 2017-03-07
Introduce Labels interface

fasaxc:
Allows for different implementations of labels rather than just `map[string]string`.  Allows for various occupancy improvements in Felix by calculating combined labels on-the-fly.

Existing interface is maintained and new `EvaluateLabels()` method added for back compatibility. 

Targeting 2.1.1 for occupancy improvements so this isn't needed for 2.1.0.

### DIFF
--- a/lib/selector/parser/ast.go
+++ b/lib/selector/parser/ast.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2016 Tigera, Inc. All rights reserved.
+// Copyright (c) 2016-2017 Tigera, Inc. All rights reserved.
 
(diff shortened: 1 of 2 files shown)


robbrockbank:
Looks fine - but do we really need to support both Evaluate() and EvalutateLabels() though?

fasaxc:
It made  my dev work a lot easier to be able to switch branches without swapping out libcalico-go and there might be other components relying on that interface such as calicoq so I don't think it's crazy.

robbrockbank:
LGTM

PR #361, opened 2017-03-07
Refactor all tests to e2edescribe

aensidhe:
I need help with cleaning and dumping kubernetes. It will help to remove cleanIpPools method and move k8s test to the same paradigm.


lib/backend/backend_test.go
```diff
@@ -170,15 +165,13 @@ var _ = Describe("Backend tests", func() {
...
 		}
 
-		bl, err := etcdClient.List(blockListOpt)
+		It("List", func() {
+			bl, err := client.List(blockListOpt)
 
-		for _, blv := range bl {
-			It("sets revision field", func() {
+			for _, blv := range bl {
 				Expect(err).NotTo(HaveOccurred())
```
robbrockbank: `Expect(err).NotTo(HaveOccurred())` should be outside the loop

robbrockbank: (I realize this was already in the code :-) )

lib/backend/backend_test.go
```diff
@@ -18,54 +18,49 @@ import (
...
 	Describe("Create", func() {
 
 		It("persists a new kv pair", func() {
-			kv, err := etcdClient.Get(block.Key)
+			kv, err := client.Get(block.Key)
 			Expect(err).NotTo(HaveOccurred())
 
 			expectedCIDR := kv.Value.(*model.AllocationBlock).CIDR
-			persitedCIRD := block.Value.(model.AllocationBlock).CIDR
+			persitedCIRD := block.Value.(*model.AllocationBlock).CIDR
```
robbrockbank: I know this is not related to your change.... but would it be possible to rename 

`persitedCIRD -> persistedCIDR`

aensidhe:
@robbrockbank done.

lib/testutils/client.go
```diff
@@ -77,18 +63,88 @@ func CreateNewIPPool(c client.Client, poolSubnet string, ipip, natOut, ipam bool
...
+	default:
+		err = errors.New(fmt.Sprintf("Unknown datastore type: %v", config.Spec.DatastoreType))
+	}
+
+	if err != nil {
 		panic(err)
 	}
 }
+
+// CleanBackend is a utility function to wipe clean "/calico" recursively from backend.
```
robbrockbank: Wrong comment

lib/testutils/client.go
```diff
@@ -77,18 +63,88 @@ func CreateNewIPPool(c client.Client, poolSubnet string, ipip, natOut, ipam bool
...
+}
+
+func CleanDatastore(config api.CalicoAPIConfig) {
+	var err error
+
+	log.Println(fmt.Sprintf("Cleaning datastore: %v", config.Spec.DatastoreType))
+
+	switch config.Spec.DatastoreType {
+	case api.EtcdV2:
+		cfg := etcdclient.Config{
```
robbrockbank: config already contains the etcd configuration, so no need to create it here  (config.EtcdConfig)

lib/testutils/client.go
```diff
@@ -77,18 +63,88 @@ func CreateNewIPPool(c client.Client, poolSubnet string, ipip, natOut, ipam bool
...
+
+// CleanBackend is a utility function to wipe clean "/calico" recursively from backend.
+func CreateCleanClient(config api.CalicoAPIConfig) *client.Client {
+	CleanDatastore(config)
+
+	return CreateClient(config)
+}
+
+// DumpBackend prints out a recursive dump of the contents of backend.
+func DumpBackend(config api.CalicoAPIConfig) error {
```
robbrockbank: Since we're tidying this up - I think renaming it to DumpDatastore might be better.  WDYT?

Makefile
```diff
@@ -44,7 +44,14 @@ run-etcd: stop-etcd
 	--advertise-client-urls "http://$(LOCAL_IP_ENV):2379,http://127.0.0.1:2379,http://$(LOCAL_IP_ENV):4001,http://127.0.0.1:4001" \
 	--listen-client-urls "http://0.0.0.0:2379,http://0.0.0.0:4001"
 
-run-kubernetes-master: stop-kubernetes-master run-etcd
+## Run consul as a container
```
robbrockbank: I don't think this should be part of this PR.

git-hooks/pre-commit
```diff
@@ -10,7 +10,13 @@ exec 1>&2
 
 changed_files=$(git diff --cached --name-only)
 changed_go_files=$(git diff --cached --name-only | grep -E '.go$' || true)
-all_go_files=$(find -name '*.go' | grep -v /vendor/)
+
+find_cmd="find"
```
robbrockbank: I think this is quite environment specific - for example on my system I have find but not gfind.

Perhaps better to check if `find` exists and if not try `gfind`.

git-hooks/pre-commit
```diff
@@ -10,7 +10,13 @@ exec 1>&2
...
+unamestr=`uname`
+if [[ "$unamestr" == 'Darwin' ]]; then
+   find_cmd='gfind'
+fi
+all_go_files=$($find_cmd -name '*.go' | grep -v /vendor/)
```
robbrockbank: I think this might be environment specific.  For example on my Mac I have `find` but not `gfind` - however the syntax for `find` is not quite correct since it needs to have the path explicitly included:

`find . -name '*.go'`

So I think this hook is missing the "." then it should work ok on Mac.

lib/testutils/client.go
aensidhe: @robbrockbank config.EtcdConfig is config for our backend which is transformed to config of etcd client inside of etcd backend. I need to duplicate that logic here, because there is no option to get backend specific config from api config or from configured client.

lib/testutils/client.go
aensidhe: Done

lib/testutils/client.go
aensidhe: Fixed

robbrockbank:
Hi @aensidhe - this is looking good.  You had a question about how to handle tidy up of the k8s datastore.  Been thinking about this and I reckon the following would be good.

For k8s because we can't easily "delete everything", I would do the following:
-  Do a list query for each resource type and then trigger delete on each resource type (basically what I was doing in the delete pools function).
-  Do another list query to ensure things have been deleted.

We'd need to do that for each resource type for which full CRUD operations are supported on the k8s backend.  At the moment - just doing it for the pools is fine.

I think what this means for this iteration would be:
-  Update the pools test to call CleanDatastore or CreateCleanClient etc.
-  In CleanDatastore in the k8s branch just call through to DeletePools()
-  In DeletePools() add another branch to do a List after and verify that it's empty.

I'd be fine doing that later - but if you want to do that refactor that would be awesome too - let me know either way.

lib/testutils/client.go
robbrockbank: Oh I'm being daft!  For some reason I thought you were instantiating a backend driver - but you're actually creating an etcd.Client.  I think I'm now erring on adding Clean/Dump methods to the datastore API ... but we can save that for another day :-)

robbrockbank:
Awesome!  Thanks for doing this refactor!

LGTM.

PR #362, opened 2017-03-08
Handle missing node better

### LINKED ISSUE #345: Missing node information for WorkloadEndpoint using KDD

caseydavenport:
Fix #345 

We receive Pods with a missing NodeName when they are first created in the k8s API, but not yet scheduled.

This PR makes the syncer ignore any such updates, and updates the code to be consistent in using `isCalicoPod` for determining if a Pod should be shown through the Calico API as a workloadEndpoint.

### DIFF
--- a/lib/backend/k8s/conversion.go
+++ b/lib/backend/k8s/conversion.go
@@ -192,7 +192,21 @@ func (c converter) globalConfigToTPR(kvp *model.KVPair) thirdparty.GlobalConfig 
 // isCalicoPod returns true if the pod should be shown as a workloadEndpoint
 // in the Calico API and false otherwise.
 func (c converter) isCalicoPod(pod *kapiv1.Pod) bool {
-	return !c.isHostNetworked(pod) && c.hasIPAddress(pod)
+	if c.isHostNetworked(pod) {
+		log.WithField("pod", pod.Name).Debug("Pod is host networked.")
+		return false
+	} else if !c.hasIPAddress(pod) {
+		log.WithField("pod", pod.Name).Debug("Pod does not have an IP address.")
+		return false
+	} else if !c.isScheduled(pod) {
+		log.WithField("pod", pod.Name).Debug("Pod is not scheduled.")
+		return false
+	}
+	return true
+}
+
+func (c converter) isScheduled(pod *kapiv1.Pod) bool {
+	return pod.Spec.NodeName != ""
 }
 
 func (c converter) isHostNetworked(pod *kapiv1.Pod) bool {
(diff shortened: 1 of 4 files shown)


lib/backend/k8s/conversion.go
```diff
@@ -209,21 +223,21 @@ func (c converter) podToWorkloadEndpoint(pod *kapiv1.Pod) (*model.KVPair, error)
...
 	// If the pod doesn't have an IP address yet, then it hasn't gone through CNI.
-	ipNets := []cnet.IPNet{}
-	if c.hasIPAddress(pod) {
-		// Parse the Pod's IP address.
-		_, ipNet, err := cnet.ParseCIDR(fmt.Sprintf("%s/32", pod.Status.PodIP))
-		if err != nil {
-			return nil, err
-		}
-		ipNets = []cnet.IPNet{*ipNet}
+	if !c.hasIPAddress(pod) {
```
robbrockbank: Should we just be calling our to isCalicoPod() here?

Also, it feels like this will cause errors to trickle up in the list processing and syncer?  Is that not the case?    e.g. in the syncer there is a call to pofdToWorkloadEndpoint() that ignores the errors - is there a chance it will get called for non calico pods and thus error resulting in a panic?

robbrockbank: It feels like the requirement to check if it is a Calico pod is split between the podToWorkloadEndpoint() code and also by some of the code that calls in to podToWorkloadEndpoint() - I just wonder if the interface should be defined that podToWorkloadEndpoint() should not be called for non-Calico pods (and maybe panic if it is).... dunno just thinking out loud.

caseydavenport: Yeah - I agree.

I think we can safely make the contract "Only call podToWorkloadEndpoint on something that passes isCalicoPod, and expect a panic otherwise."

caseydavenport: Actually, I don't want the result to be a panic.

robbrockbank: Yeah I wasn't sure either.....  probably a better for podToWorkloadEndpoint to return a specific error or summint and handle that error - but then is it worth it?

caseydavenport: OK - I've updated so that the error is returned if not a calico pod, and all instances handle the error (and/or check isCalicoPod beforehand).

lib/backend/k8s/syncer.go
```diff
@@ -424,18 +424,20 @@ func (syn *kubeSyncer) performSnapshot() ([]model.KVPair, map[string]bool, resou
...
 				continue
 			}
 
 			// Convert to a workload endpoint.
-			wep, _ := syn.kc.converter.podToWorkloadEndpoint(&po)
-			if wep != nil {
-				snap = append(snap, *wep)
-				keys[wep.Key.String()] = true
+			wep, err := syn.kc.converter.podToWorkloadEndpoint(&po)
+			if err == nil {
```
robbrockbank: !=

robbrockbank: are we missing a test here that we didn't spot that?

caseydavenport: d'oh, yes.  I'll fix.

caseydavenport: @robbrockbank I've added some FV infrastructure to this PR which validates that updates actually do show up in the syncer.

It doesn't go all the way, in that:
- I haven't added it to all of the different resources tests yet.
- It doesn't check the content yet, just that an update was received.

It catches the above `==` vs `!=` bug.  I think this is good enough for now, but we should fix those two bullets in the near future to make these tests more robust.

lib/backend/k8s/k8s_fv_test.go
```diff
@@ -71,10 +77,48 @@ func (c cb) OnUpdates(updates []api.Update) {
...
+}
+
+func (c cb) ProcessUpdates() {
+	for u := range c.updateChan {
+		// Store off the update so it can be checked by the test.
+		// Use a mutex for safe cross-goroutine reads/writes.
+		c.Lock.Lock()
+		if u.UpdateType == api.UpdateTypeKVUnknown {
+			// We should never get this!
+			log.Panic("Recieved Unkown update type")
```
robbrockbank: Unkown

robbrockbank: Recieved

caseydavenport:
@tomdee @robbrockbank what are your thoughts on this for v2.1? 

It's not tiny, but without it we produce some ERROR logs during mainline operation which are benign.

I could make a much smaller stopgap PR which just downgrades the log level for v2.1.0 and merge this after?

robbrockbank:
LGTM (by the way).

A large chunk of this PR is testing, and the remainder adds improved testing of the pod and adds more logging.  If you think the logging improvements will themselves be beneficial for debugging any potential issues then I'm personally ok with this going in to v2.1.  - I don't personally have a feel for that though.

PR #363, opened 2017-03-08
Add validation for backend Rules.

fasaxc:
Defend against ports with no protocol.

Prevents felix crash in https://github.com/projectcalico/felix/issues/1375 because it will validate the rules coming out of the KDD and drop any that fail validation (but not a complete fix).


robbrockbank:
LGTM

Just to check though - Felix is invoking the validation on the backend data correct?  If so, then merge away!

fasaxc:
Yes, it does.

PR #365, opened 2017-03-08
Changing k8s cluster type name

heschlie:
Updating the k8s GlobalConfig ClusterType to be shorter, "KDD" instead
of "kubernetes,k8sdatastoredriver"


tomdee:
LGTM

PR #366, opened 2017-03-13
ENV client and GoDoc update

### LINKED ISSUE #359: Client example in the docs is broken
The [client package docs](https://godoc.org/github.com/projectcalico/libcalico-go/lib/client#Client) contain an example that tells you to initialize a Calico client in the following way:
```go
config := api.ClientConfig{}
client := New(&config)
```
However, this doesn't actually work as the ClientConfig struct [does not seem to exist](https://github.com/projectcalico/libcalico-go/search?utf8=%E2%9C%93&q=ClientConfig).

The way I ended up making it work was by using the following code:
```go
config := api.NewCalicoAPIConfig()
config.Spec.DatastoreType = api.EtcdV2
config.Spec.EtcdEndpoints = "http://localhost:2379"
client, err := client.New(*config)
```

Not sure if that's "the" way to do it though.

gunjan5:
Fixes #359

### DIFF
--- a/lib/client/client.go
+++ b/lib/client/client.go
@@ -43,9 +43,8 @@ type Client struct {
 	Backend bapi.Client
 }
 
-// New returns a connected Client.  This is the only mechanism by which to create a
-// Client.  The ClientConfig can either be created explicitly, or can be loaded from
-// a config file or environment variables using the LoadClientConfig() function.
+// New returns a connected Client. The ClientConfig can either be created explicitly,
+// or can be loaded from a config file or environment variables using the LoadClientConfig() function.
 func New(config api.CalicoAPIConfig) (*Client, error) {
 	var err error
 	cc := Client{}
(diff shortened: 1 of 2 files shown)


lib/client/client.go
```diff
@@ -55,6 +54,23 @@ func New(config api.CalicoAPIConfig) (*Client, error) {
...
 
+// NewFromEnv loads the config from ENV variables and returns a connected Client.
+func NewFromEnv() (*Client, error) {
+
+	config, err := LoadClientConfigFromEnvironment()
+	if err != nil {
+		return nil, err
+	}
+
+	cc := Client{}
```
tmjd: I think you should use the `func New(config...` that is defined above for actually creating the client.

gunjan5: Good point! I had more logic there at one point, but reverted back and forgot to change it to `New()`

robbrockbank:
One minor comment, but otherwise LGTM

PR #372, opened 2017-03-22
Adds functionality to merge datastore config with precedence

tmjd:
- When config file is specified (or use default) load the config
  and then merge with config from Env Variables with precedence being
  with the Env Variables
- Added unit tests too


tmjd:
Part of effort to address https://github.com/projectcalico/calico/issues/563

caseydavenport:
This is pretty old - closing for now

PR #373, opened 2017-03-23
Check for conflicting Source/Dest CIDR type

shatrugna:
Shouldn't allow for creation of rules with mixed
IPv4/v6 source/dest types

### DIFF
--- a/lib/validator/validator.go
+++ b/lib/validator/validator.go
@@ -364,6 +364,26 @@ func validateRule(v *validator.Validate, structLevel *validator.StructLevel) {
 				"Destination.NotPorts", "", reason("protocol does not support ports"))
 		}
 	}
+
+	// Check for mismatch in IP versions
+	if rule.Source.Net != nil && rule.IPVersion != nil {
+		if rule.Source.Net.Version() != *(rule.IPVersion) {
+			structLevel.ReportError(reflect.ValueOf(rule.Source.Net), "Source.Net",
+				"", reason("rule contains an IP version that does not match src CIDR version"))
+		}
+	}
+	if rule.Destination.Net != nil && rule.IPVersion != nil {
+		if rule.Destination.Net.Version() != *(rule.IPVersion) {
+			structLevel.ReportError(reflect.ValueOf(rule.Destination.Net), "Destination.Net",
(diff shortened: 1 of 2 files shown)


lib/validator/validator.go
```diff
@@ -364,6 +364,14 @@ func validateRule(v *validator.Validate, structLevel *validator.StructLevel) {
 				"Destination.NotPorts", "", reason("protocol does not support ports"))
 		}
 	}
+
+	if rule.Source.Net != nil && rule.Destination.Net != nil {
+		if (rule.Source.Net.IP.To4() == nil && rule.Destination.Net.IP.To4() != nil) ||
```
robbrockbank: I think you can just do  `rule.Source.Net.Version() != rule.Desintation.Net.Version()`

lib/validator/validator_test.go
```diff
@@ -484,6 +506,16 @@ func init() {
 	)
 }
 
+func v4IPNet() *net.IPNet {
```
robbrockbank: There's a whole bunch of these defined at the top of the file.

robbrockbank:
I wonder if this is going far enough.  We also have an optional IPVersion field in the Rule.  It's main use case is to match on a specific version when we aren't matching on CIDR.  I think we should also beef this up to validate that there is no conflict between the IPVersion and the Net fields.

lib/validator/validator.go
```diff
@@ -364,6 +364,26 @@ func validateRule(v *validator.Validate, structLevel *validator.StructLevel) {
 				"Destination.NotPorts", "", reason("protocol does not support ports"))
 		}
 	}
+
+	// Check for mismatch in IP versions
+	if rule.Source.Net != nil && rule.IPVersion != nil {
+		if rule.Source.Net.Version() != *(rule.IPVersion) {
+			structLevel.ReportError(reflect.ValueOf(rule.Source.Net), "Source.Net",
+				"", reason("rule does not support mixing of IPv4/v6 CIDRs"))
```
robbrockbank: I think we can say that the rule contains an IP version match that does not match the source/dest CIDR version.

robbrockbank:
LGTM.

PR #375, opened 2017-03-23
Fix pod deletion in KDD

### LINKED ISSUE #374: KDD Doesn't handle Pod IP changes
The Kubernetes datastore driver currently assumes that a Pod's IP address will never change. 

When using the etcd datastore, this is actually the case, because we check if an endpoint already exists and use its existing IP address if it does.  See this bit of code: https://github.com/projectcalico/cni-plugin/blob/master/calico.go#L93

The KDD should be hitting this same bit of code, but for some reasons it isn't detecting that an endpoint exists and has an IP address, and thus it's assigning a new one.

However, the KDD syncer ignores the update when the Pod IP address is changed, and so Felix never updates the route to the Pod.

caseydavenport:
Fixes #374

### DIFF
--- a/lib/backend/k8s/conversion.go
+++ b/lib/backend/k8s/conversion.go
@@ -191,9 +191,9 @@ func (c converter) globalConfigToTPR(kvp *model.KVPair) thirdparty.GlobalConfig 
 	return tpr
 }
 
-// isCalicoPod returns true if the pod should be shown as a workloadEndpoint
+// isReadyCalicoPod returns true if the pod should be shown as a workloadEndpoint
 // in the Calico API and false otherwise.
-func (c converter) isCalicoPod(pod *kapiv1.Pod) bool {
+func (c converter) isReadyCalicoPod(pod *kapiv1.Pod) bool {
 	if c.isHostNetworked(pod) {
 		log.WithField("pod", pod.Name).Debug("Pod is host networked.")
 		return false
@@ -219,23 +219,24 @@ func (c converter) hasIPAddress(pod *kapiv1.Pod) bool {
 	return pod.Status.PodIP != ""
 }
 
+// podToWorkloadEndpoint converts a Pod to a WorkloadEndpoint.  It assumes the calling code
+// has verified that the provided Pod is valid to convert to a WorkloadEndpoint.
 func (c converter) podToWorkloadEndpoint(pod *kapiv1.Pod) (*model.KVPair, error) {
 	// Pull out the profile and workload ID based on pod name and Namespace.
 	profile := fmt.Sprintf("ns.projectcalico.org/%s", pod.ObjectMeta.Namespace)
 	workload := fmt.Sprintf("%s.%s", pod.ObjectMeta.Namespace, pod.ObjectMeta.Name)
 
-	// If the pod isn't ready, we can't parse it.
-	if !c.isCalicoPod(pod) {
-		return nil, fmt.Errorf("Pod is not ready / valid. pod=%s", pod.Name)
+	// We do, in some circumstances, want to parse Pods without an IP address.  For example,
+	// a DELETE update will not include an IP.
+	ipNets := []cnet.IPNet{}
+	if c.hasIPAddress(pod) {
+		_, ipNet, err := cnet.ParseCIDR(fmt.Sprintf("%s/32", pod.Status.PodIP))
+		if err != nil {
+			log.WithFields(log.Fields{"ip": pod.Status.PodIP, "pod": pod.Name}).WithError(err).Error("Failed to parse pod IP")
+			return nil, err
+		}
+		ipNets = append(ipNets, *ipNet)
 	}
-
-	// Parse the Pod's IP address.
-	_, ipNet, err := cnet.ParseCIDR(fmt.Sprintf("%s/32", pod.Status.PodIP))
-	if err != nil {
-		log.WithFields(log.Fields{"ip": pod.Status.PodIP, "pod": pod.Name}).WithError(err).Error("Failed to parse pod IP")
-		return nil, err
-	}
-	ipNets := []cnet.IPNet{*ipNet}
 
 	// Generate the interface name and MAC based on workload.  This must match
 	// the host-side veth configured by the CNI plugin.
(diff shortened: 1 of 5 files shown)


lib/backend/k8s/conversion.go
```diff
@@ -224,18 +224,16 @@ func (c converter) podToWorkloadEndpoint(pod *kapiv1.Pod) (*model.KVPair, error)
...
-		return nil, fmt.Errorf("Pod is not ready / valid. pod=%s", pod.Name)
-	}
-
 	// Parse the Pod's IP address.
-	_, ipNet, err := cnet.ParseCIDR(fmt.Sprintf("%s/32", pod.Status.PodIP))
-	if err != nil {
-		log.WithFields(log.Fields{"ip": pod.Status.PodIP, "pod": pod.Name}).WithError(err).Error("Failed to parse pod IP")
-		return nil, err
+	ipNets := []cnet.IPNet{}
+	if c.hasIPAddress(pod) {
```
robbrockbank: why is `isCalicoPod()` not suitable here?  I assume we need `isScheduled()` to return `true`, so does that mean we don't care about what value `isHostNetworked()` returns? ... in which case is our definition of `isCalicoPod()` wrong?

robbrockbank: or does this method only get called when isCalicoPod() is true?   In which case we should beef up (or add) a method doc block.

lib/backend/k8s/k8s_fv_test.go
```diff
@@ -106,16 +109,59 @@ func (c cb) ProcessUpdates() {
...
+			update, ok := c.State[kvp.Key.String()]
+			exists = ok
+			c.Lock.Unlock()
+
+			log.Infof("[TEST] Key exists? %t: %+v", ok, update)
+			if ok {
+				// Expected key to not exist, and it does.
+				return false, nil
+			} else {
+				// Key does not yet exist.
```
robbrockbank: wrong comment?

lib/backend/k8s/syncer.go
```diff
@@ -588,10 +588,21 @@ func (syn *kubeSyncer) parsePodEvent(e watch.Event) *model.KVPair {
...
+			log.WithField("pod", pod.Name).Debug("Pod is host networked.")
+			log.Debugf("Skipping pod %s/%s", pod.ObjectMeta.Namespace, pod.ObjectMeta.Name)
+			return nil
+		}
+	default:
+		// Ignore add/modify updates for Pods that aren't ready / valid.
+		if !syn.kc.converter.isCalicoPod(pod) {
+			log.Debugf("Skipping pod %s/%s", pod.ObjectMeta.Namespace, pod.ObjectMeta.Name)
+			return nil
+		}
```
robbrockbank: We do still care about isScheduled() no?

lib/backend/k8s/syncer.go
```diff
@@ -588,10 +588,21 @@ func (syn *kubeSyncer) parsePodEvent(e watch.Event) *model.KVPair {
...
 
-	// Ignore updates for Pods that aren't ready / valid.
-	if !syn.kc.converter.isCalicoPod(pod) {
-		log.Debugf("Skipping pod %s/%s", pod.ObjectMeta.Namespace, pod.ObjectMeta.Name)
-		return nil
+	switch e.Type {
+	case watch.Deleted:
+		// For deletes, the validity conditions are different.  We only care if the update
+		// is not for a host-networked Pods, but don't care about IP / scheduled state.
+		if syn.kc.converter.isHostNetworked(pod) {
```
robbrockbank: do we no care about isScheduled() too?

lib/backend/k8s/conversion.go
```diff
@@ -224,18 +224,16 @@ func (c converter) podToWorkloadEndpoint(pod *kapiv1.Pod) (*model.KVPair, error)
 	profile := fmt.Sprintf("ns.projectcalico.org/%s", pod.ObjectMeta.Namespace)
 	workload := fmt.Sprintf("%s.%s", pod.ObjectMeta.Namespace, pod.ObjectMeta.Name)
 
-	// If the pod isn't ready, we can't parse it.
```
robbrockbank: Shouldn't we still check `isScheduled()` and `isHostNetworked()` ?

It feels as though `isCalicoPod()` should be a combination of the above two and not include the IP address.

Perhaps with a `isReadyCalicoPod()` to also include the IP address.

lib/backend/k8s/k8s_fv_test.go
caseydavenport: Yep.

lib/backend/k8s/syncer.go
caseydavenport: For a delete, not really.

caseydavenport:
> Feels like we haven't quite got the definition and use of the isCalicoPod() etc. down yet

Yeah, I sort of agree.  It's because there really isn't one true set of conditions for every Pod that indicate whether we should send it through the Calico API.  There are a few difference scenarios:

Syncer:
- We receive an ADD/MODIFY (Requires IP address, scheduled, not host networked)
- We receive a DELETE (Requires not host networked)

Not-syncer:
- GET / LIST operationrs (Require an IP address, scheduled, not host networked)

So, we need to be able to convert a Pod -> WorkloadEndpoint for any non-host-networked Pod because of the Delete case.  However, we need to make sure that things we show for ADD/UPDATE/GET/LIST have an IP address and Node present.

It's not really a matter of "isCalicoReadyPod vs isNotReadyCalicoPod", it's really just a matter of the DELETE having a separate set of requirements than ADD/UPDATE/LIST/GET.

I think that in order to make the code clear:
- `podToWorkloadEndpoint` shouldn't care about IP / host networked / node and should just do its best. Calling code needs to perform the correct checks based on what it is trying to do.
- `isCalicoPod` should probably be renamed to `showInCalicoAPI` (or potentially just removed and the code can use the individual checks directly)
- Comments should be updated to make it clear what checks are required when and whose responsibility it is to perform the check.

lib/backend/k8s/conversion.go
caseydavenport: I don't think so - this is just a conversion method, and I don't think it should make assumptions about whether or not the calling code is doing the right thing.  The calling code should know what conditions it needs to check, but this bit of code should just do its best to turn a Pod into a WorkloadEndpoint.

robbrockbank:
LGTM.  Can you squash please.

PR #376, opened 2017-03-23
validate ip pool CIDR overlap with link local range

### LINKED ISSUE #224: link local IP pools should not be allowed
At present, calicoctl allows the configuration of a /0 pool - both for IPv4 and IPv6(!).  

There should probably be a maximum pool size.

gunjan5:
Fixes #224

### DIFF
--- a/lib/validator/validator.go
+++ b/lib/validator/validator.go
@@ -15,6 +15,7 @@
 package validator
 
 import (
+	"net"
 	"reflect"
 	"regexp"
 	"strings"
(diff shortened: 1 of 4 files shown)


lib/validator/validator.go
```diff
@@ -326,8 +329,26 @@ func validateIPPool(v *validator.Validate, structLevel *validator.StructLevel) {
...
+
+		if pool.Metadata.CIDR.Version() == 6 && isNetOverlap(&pool.Metadata.CIDR.IPNet, ipv6LinkLocalNet) {
+			structLevel.ReportError(reflect.ValueOf(pool.Metadata.CIDR),
+				"CIDR", "", reason(overlapsV6LinkLocal))
+		}
 	}
+}
 
+// isNetOverlap is a utility function that returns true if the two subnet have an overlap.
+func isNetOverlap(n1, n2 *net.IPNet) bool {
```
robbrockbank: Shouldn't this just be another method on the net.IPNet structure?

lib/validator/validator_test.go
```diff
@@ -304,6 +304,8 @@ func init() {
 					IPIP: &api.IPIPConfiguration{Enabled: true},
 				},
 			}, false),
+		Entry("should reject IPv4 pool with a CIDR rnage overlapping with Link Local range", api.IPPool{Metadata: api.IPPoolMetadata{CIDR: testutils.MustParseCIDR("169.254.5.0/24")}}, false),
```
robbrockbank: rnage -> range

lib/validator/validator.go
```diff
@@ -326,8 +329,26 @@ func validateIPPool(v *validator.Validate, structLevel *validator.StructLevel) {
 				}
 			}
 		}
+
+		// IP Pool CIDR cannot overlap with IPv4 or IPv6 link local address range.
+		_, ipv4LinkLocalNet, _ := net.ParseCIDR("169.254.0.0/16")
```
robbrockbank: These should be constants.

robbrockbank:
@gunjan5 :  just pinging on the status of this - would be good to get it closed down.

lib/validator/validator.go
gunjan5: don't really like the whole idea of layering on top of `net.IPNet` but I guess since we have it, I can do it for this one. In future, if we move away from this layering on top of `net`, this will come back here :)

lib/validator/validator.go
gunjan5: can't have this type as a const see: https://golang.org/ref/spec#Constants so putting it in the `var` section

gunjan5:
@robbrockbank pong. back to you

lib/validator/validator_test.go
```diff
@@ -304,6 +304,8 @@ func init() {
 					IPIP: &api.IPIPConfiguration{Enabled: true},
 				},
 			}, false),
+		Entry("should reject IPv4 pool with a CIDR range overlapping with Link Local range", api.IPPool{Metadata: api.IPPoolMetadata{CIDR: testutils.MustParseCIDR("169.254.5.0/24")}}, false),
```
robbrockbank: These are very long lines.  Can you split up please.

robbrockbank:
One minor comment -  LGTM.

gunjan5:
@robbrockbank ok I split it up!

PR #385, opened 2017-04-05
Remove label cache from k8s syncer.

fasaxc:
Allows duplicate updates through to Felix (which are handled anyway) in return for a substantial occupancy reduction (50-60MB if my calculations are correct).


caseydavenport:
Back to you - feel free to merge when you're comfortable with your occupancy testing.

fasaxc:
Tested at 10k endpoint scale using the new k8sfv framework and see an ~8MB reduction at that scale, which is a pretty good win.

PR #386, opened 2017-04-11
K8s host config

heschlie:
Fleshed out portion of the `HostConfig` object in KDD to return an IP address for an IPIP tunnel if requested.

As we are currently using `host-local` for the IPAM provider in KDD we can assume that the first two IPs are available for use, as `host-local` reserves those for use.  This change will query the Node for its pod CIDR and grab the first IP from it to return.

fixes projectcalico/calicoctl#1599

### DIFF
--- a/lib/backend/k8s/k8s.go
+++ b/lib/backend/k8s/k8s.go
@@ -40,6 +40,8 @@ import (
 	"k8s.io/client-go/pkg/util/wait"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
+	"github.com/projectcalico/libcalico-go/lib/net"
+	v1 "k8s.io/client-go/pkg/api/v1"
 )
 
 type KubeClient struct {
@@ -739,9 +741,79 @@ func (c *KubeClient) deleteGlobalConfig(k *model.KVPair) error {
 }
 
 func (c *KubeClient) getHostConfig(k model.HostConfigKey) (*model.KVPair, error) {
+	if k.Name == "IpInIpTunnelAddr" {
+		n, err := c.clientSet.Nodes().Get(k.Hostname, metav1.GetOptions{})
+		if err != nil {
+			return nil, resources.K8sErrorToCalico(err, k)
+		}
+
+		kvp, err := getTunIp(n)
+		if err != nil {
+			return nil, err
+		}
+
+		return kvp, nil
+	}
+
 	return nil, errors.ErrorResourceDoesNotExist{Identifier: k}
 }
 
 func (c *KubeClient) listHostConfig(l model.HostConfigListOptions) ([]*model.KVPair, error) {
-	return []*model.KVPair{}, nil
+	var kvps = []*model.KVPair{}
+
+	// Short circuit if they aren't asking for information we can provide.
+	if l.Name == "" || l.Name == "IpInIpTunnelAddr" {
+		// First see if we were handed a specific host, if not list all Nodes
+		if l.Hostname == "" {
+			nodes, err := c.clientSet.Nodes().List(v1.ListOptions{})
+			if err != nil {
+				return nil, resources.K8sErrorToCalico(err, l)
+			}
+
+			for _, node := range nodes.Items {
+				kvp, err := getTunIp(&node)
+				if err != nil {
+					log.Warnf("Invalid IP for HostConfig: %s, %s", node.Name, node.Spec.PodCIDR)
+					continue
+				}
+
+				kvps = append(kvps, kvp)
+			}
+		} else {
+			node, err := c.clientSet.Nodes().Get(l.Hostname, metav1.GetOptions{})
+			if err != nil {
+				return nil, resources.K8sErrorToCalico(err, l)
+			}
+
+			kvp, err := getTunIp(node)
+			if err != nil {
+				return []*model.KVPair{}, nil
+			}
+
+			kvps = append(kvps, kvp)
+		}
+	}
+
+	return kvps, nil
+}
+
+func getTunIp(n *v1.Node) (*model.KVPair, error) {
+	ip, _, err := net.ParseCIDR(n.Spec.PodCIDR)
+	if err != nil {
+		return nil, err
+	}
+	// We need to get the IP for the podCIDR and increment it to the
+	// first IP in the CIDR.
+	tunIp := ip.To4()
+	tunIp[3]++
+
+	kvp := &model.KVPair{
+		Key: model.HostConfigKey{
+			Hostname: n.Name,
+			Name: "IpInIpTunnelAddr",
+		},
+		Value: tunIp,
+	}
+
+	return kvp, nil
 }
--- a/lib/backend/k8s/k8s.go
+++ b/lib/backend/k8s/k8s.go
@@ -764,35 +764,37 @@ func (c *KubeClient) listHostConfig(l model.HostConfigListOptions) ([]*model.KVP
 	var kvps = []*model.KVPair{}
 
 	// Short circuit if they aren't asking for information we can provide.
-	if l.Name == "" || l.Name == "IpInIpTunnelAddr" {
-		// First see if we were handed a specific host, if not list all Nodes
-		if l.Hostname == "" {
-			nodes, err := c.clientSet.Nodes().List(v1.ListOptions{})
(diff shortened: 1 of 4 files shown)


lib/backend/k8s/k8s.go
```diff
@@ -739,9 +740,55 @@ func (c *KubeClient) deleteGlobalConfig(k *model.KVPair) error {
 }
 
 func (c *KubeClient) getHostConfig(k model.HostConfigKey) (*model.KVPair, error) {
+	if k.Name == "IpInIpTunnelAddr" {
+		n, err := c.clientSet.Nodes().Get(k.Hostname, metav1.GetOptions{})
+		if err != nil {
+			return nil, err
```
robbrockbank: I think we need to translate k82 to Calico errors

lib/backend/k8s/k8s.go
```diff
@@ -739,9 +740,55 @@ func (c *KubeClient) deleteGlobalConfig(k *model.KVPair) error {
...
+
+		return &model.KVPair{Key: k, Value: tunIp.String()}, nil
+	}
+
 	return nil, errors.ErrorResourceDoesNotExist{Identifier: k}
 }
 
 func (c *KubeClient) listHostConfig(l model.HostConfigListOptions) ([]*model.KVPair, error) {
+	if l.Name == "IpInIpTunnelAddr" {
+		n, err := c.clientSet.Nodes().Get(l.Hostname, metav1.GetOptions{})
```
robbrockbank: Shouldn't this do a List operation rather than a Get()

lib/backend/k8s/k8s.go
```diff
@@ -739,9 +740,55 @@ func (c *KubeClient) deleteGlobalConfig(k *model.KVPair) error {
...
+		ipNet.IP = tunIp
+
+		return &model.KVPair{Key: k, Value: tunIp.String()}, nil
+	}
+
 	return nil, errors.ErrorResourceDoesNotExist{Identifier: k}
 }
 
 func (c *KubeClient) listHostConfig(l model.HostConfigListOptions) ([]*model.KVPair, error) {
+	if l.Name == "IpInIpTunnelAddr" {
```
robbrockbank: IIRC the host config list options could have a blank name - which would mean enumerate over all hosts.  

I think this needs to be a List operation rather than a Get.

lib/backend/k8s/k8s.go
```diff
@@ -739,9 +740,55 @@ func (c *KubeClient) deleteGlobalConfig(k *model.KVPair) error {
...
+		n, err := c.clientSet.Nodes().Get(l.Hostname, metav1.GetOptions{})
+		if err != nil {
+			return nil, err
```
robbrockbank: If this returns no such resource type exception we should just return an empty list.

lib/backend/k8s/k8s.go
robbrockbank: It can also be called with a blank config name - which would mean iterate all configs.  So I think this also needs to handle l.Name == "" ....

which is what you are actually passing in from the syncer unless I'm mistaken.

robbrockbank:
I think this needs some UTs.

lib/backend/k8s/k8s.go
```diff
@@ -739,9 +740,55 @@ func (c *KubeClient) deleteGlobalConfig(k *model.KVPair) error {
...
+		}
+
+		ip, ipNet, err := net.ParseCIDR(n.Spec.PodCIDR)
```
robbrockbank: Feels like this chunk of code should be made into a common function `nodeToIPIPTunnelAddrConfig` or some such thing.

lib/backend/k8s/k8s.go
heschlie: In k8s I can't really pass it a hostname to look out for, so I could just do a list, and if we were handed a hostname I could filter it out of the list, or I could issue a get call if we have a hostname, and do a list if now.  I'm kinda leaning towards the 2nd option to make smaller API calls.

heschlie:
@robbrockbank This should be the complete set of changes, if it looks good I'll go ahead and rebase.

lib/backend/k8s/k8s.go
```diff
@@ -739,9 +741,85 @@ func (c *KubeClient) deleteGlobalConfig(k *model.KVPair) error {
...
+			Value: tunIp,
+		}
+
+		kvps = append(kvps, kvp)
+	}
+
+	return kvps, nil
 }
+
+func getTunIp(n *v1.Node) (string, error) {
```
robbrockbank: If this returned a pointer to the KVPair instead of the string I think that would simplify a chunk of the code above.

If we hit an error parsing the CIDR then just return a nil KVPair and log.  I think for the list case above - you probably just want to skip over nodes that don't have a valid CIDR rather than returning no results at all.

lib/backend/k8s/k8s.go
```diff
@@ -739,9 +741,85 @@ func (c *KubeClient) deleteGlobalConfig(k *model.KVPair) error {
...
+		return &model.KVPair{Key: k, Value: tunIp}, nil
+	}
+
 	return nil, errors.ErrorResourceDoesNotExist{Identifier: k}
 }
 
 func (c *KubeClient) listHostConfig(l model.HostConfigListOptions) ([]*model.KVPair, error) {
-	return []*model.KVPair{}, nil
+	var kvps = []*model.KVPair{}
+
```
robbrockbank: I assume if the config name is anything other than "IpInIpTunnelAddr" we should return no results?

heschlie: Or an empty string as it could be looking to list all info for all hosts or one host, I've now set it up to ignore anything other than `"" || "IpInIpTunnelAddr"`

lib/backend/k8s/k8s.go
```diff
@@ -739,9 +741,79 @@ func (c *KubeClient) deleteGlobalConfig(k *model.KVPair) error {
...
+	// Short circuit if they aren't asking for information we can provide.
+	if l.Name == "" || l.Name == "IpInIpTunnelAddr" {
```
robbrockbank: Minor nit:  If you checked for != "" && != "IpInIpTunnelAddr" you could just return an empty list and then the next block wouldn't need to be indented.

lib/backend/k8s/syncer.go
```diff
@@ -556,16 +571,21 @@ func (syn *kubeSyncer) parseNodeEvent(e watch.Event) *model.KVPair {
 	}
 
 	kvp, err := resources.K8sNodeToCalico(node)
-
 	if err != nil {
 		log.Panicf("%s", err)
 	}
 
+	kvpHostIp := &model.KVPair{
+		Key: model.HostConfigKey{Hostname: node.Name},
```
robbrockbank: Shouldn't this be HostIPKey?

lib/backend/k8s/k8s.go
```diff
@@ -739,9 +741,79 @@ func (c *KubeClient) deleteGlobalConfig(k *model.KVPair) error {
...
+
+			kvps = append(kvps, kvp)
+		}
+	}
+
+	return kvps, nil
 }
+
+func getTunIp(n *v1.Node) (*model.KVPair, error) {
+	ip, _, err := net.ParseCIDR(n.Spec.PodCIDR)
```
robbrockbank: As discussed - I think this should check if the CIDR exists and if so return no error and no KVPair - since it isn't really an error condition.  However, I think that means you'll also need to handle no KV returned in the above, which for a Get would correspond to a No such resource error.

lib/backend/k8s/k8s.go
```diff
@@ -739,9 +741,79 @@ func (c *KubeClient) deleteGlobalConfig(k *model.KVPair) error {
...
+	if err != nil {
+		return nil, err
```
robbrockbank: I think we just log here that the CIDR is invalid rather than needing to in the above branches.

lib/backend/k8s/syncer.go
heschlie: I think so, but the fact that it worked this way worries me a bit, going to rebuild felix/node and fire up another cluster to test

heschlie:
@robbrockbank Ready for another look.

robbrockbank:
LGTM.

PR #388, opened 2017-04-13
Add SRV discovery to etcd backend

jsok:
Adds a new JSON option `etcdDiscoverySrv` and env var `ETCD_DISCOVERY_SRV` to allow specifying a domain name with which the etcd client will perform DNS discovery. See https://coreos.com/etcd/docs/latest/op-guide/clustering.html#dns-discovery

Possibly related to an old issue https://github.com/projectcalico/calicoctl/issues/933.

### DIFF
--- a/lib/backend/etcd/etcd.go
+++ b/lib/backend/etcd/etcd.go
@@ -16,6 +16,7 @@ package etcd
 
 import (
 	goerrors "errors"
+	"fmt"
 	"strconv"
 	"strings"
 	"time"
@@ -46,8 +47,12 @@ type EtcdClient struct {
 	etcdKeysAPI etcd.KeysAPI
 }
 
-func NewEtcdClient(config *apiconfig.EtcdConfig) (*EtcdClient, error) {
-	// Determine the location from the authority or the endpoints.  The endpoints
+func getEtcdEndpoints(config *apiconfig.EtcdConfig) ([]string, error) {
+	if (config.EtcdAuthority != "" || config.EtcdEndpoints != "") && config.EtcdDiscoverySrv != "" {
+		return nil, goerrors.New("multiple discovery or bootstrap options specified, use either \"etcdEndpoints\" or \"etcdDiscoverySrv\"")
+	}
+
+	// Determine the location from the authority or the endpoints. The endpoints
 	// takes precedence if both are specified.
 	etcdLocation := []string{}
 	if config.EtcdAuthority != "" {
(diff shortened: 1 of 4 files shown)


djosborne:
Nice, this looks good, thanks for the contribution. Before this moves forward I want to make sure one thing is addressed - 

I'm basing this off of how etcd is run in DC/OS, but I think if etcd's address can only be looked up via SRV record we can assume it's slightly more volatile and has a higher chance than normal of changing IP _and port_. I want to be sure that if that were to happen, the way you've implemented this SRV discovery would allow calico to reconnect to the etcd cluster at it's new address. (Or at the very least we note that this behavior is not yet supported and leave it as future work).

@robbrockbank Any thoughts on Calico would handle a changing hostname for a regular DNS lookup today? I'm guessing oneshots like CNI would have no problem but I wonder how felix/confd (which keep persistent watches running) would handle it.

jsok:
The etcd client in calico only consults the SRV records on startup, after that any etcd member reconfigurations (orchestrated via the etcd API) would be broadcast across the whole etcd cluster. The etcd client in calico would therefore also receive the updated member list too

caseydavenport:
@jsok thanks for the PR and sorry for the delay on review!

One thing I think we'll need before merging are UTs for this PR.

caseydavenport:
@jsok is this something you're still working on?

jsok:
@caseydavenport yes this is still a useful piece of functionality. Happy to update/change based on feedback in order to make it merge-able.

jsok:
Rebased to accomodate some internal refactorings. I will also submit a complimentary docs PR soon.

lib/backend/etcd/etcd.go
```diff
@@ -46,8 +46,12 @@ type EtcdClient struct {
 	etcdKeysAPI etcd.KeysAPI
 }
 
-func NewEtcdClient(config *capi.EtcdConfig) (*EtcdClient, error) {
-	// Determine the location from the authority or the endpoints.  The endpoints
+func getEtcdEndpoints(config *capi.EtcdConfig) ([]string, error) {
+	if (config.EtcdAuthority != "" || config.EtcdEndpoints != "") && config.EtcdDiscoverySrv != "" {
+		return nil, goerrors.New("multiple discovery or bootstrap options specified, use either \"etcdAuthority\"/\"etcdEndpoints\" or \"etcdDiscoverySrv\"")
```
caseydavenport: Could you tweak this message so that it doesn't include `etcdAuthority`, since that's deprecated and not recommended anymore.

lib/backend/etcd/etcd.go
```diff
@@ -46,8 +46,12 @@ type EtcdClient struct {
 	etcdKeysAPI etcd.KeysAPI
 }
 
-func NewEtcdClient(config *capi.EtcdConfig) (*EtcdClient, error) {
-	// Determine the location from the authority or the endpoints.  The endpoints
+func getEtcdEndpoints(config *capi.EtcdConfig) ([]string, error) {
+	if (config.EtcdAuthority != "" || config.EtcdEndpoints != "") && config.EtcdDiscoverySrv != "" {
```
tmjd: I think this will always hit if EtcdDiscoverySrv is set because EtcdAuthority is defaulted.  Right?

lib/backend/etcd/etcd.go
jsok: Done

caseydavenport:
@jsok very sorry for the delay on this.

We've been focused on the changes for Calico v3.0, and I think we'll need to pick this discussion up again once that is released, which will be towards the end of the year.

caseydavenport:
This seems to have gotten rather out of date.

I'm going to close this PR for now.

PR #391, opened 2017-04-14
Changing stop targets to not error on non-existent items

insequent:
This is kind of a silly PR, and I totally understand if the Calico team is not interested. I was just getting annoyed with spam on the `stop-kubernetes-master` in particular. Especially when no volumes existed.

It'll still error on the explicit ones, but at least that's only 2 lines (one for docker, one for make)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,9 +9,9 @@ CALICO_BUILD?=calico/go-build
 PACKAGE_NAME?=projectcalico/libcalico-go
 LOCAL_USER_ID?=$(shell id -u $$USER)
 
-# Use this to populate the vendor directory after checking out the repository.
-# To update upstream dependencies, delete the glide.lock file first.
+## Use this to populate the vendor directory after checking out the repository.
 vendor: glide.yaml
+	# To update upstream dependencies, delete the glide.lock file first.
 	# To build without Docker just run "glide install -strip-vendor"
 	docker run --rm \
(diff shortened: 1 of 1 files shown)


caseydavenport:
@insequent thanks for opening, and sorry for the delay on review - I've been out for a few weeks.

Makefile
```diff
@@ -78,14 +78,14 @@ stop-kubernetes-master:
 	-docker rm -f calico-kubelet-master
 
 	# Remove any pods that the old kubelet may have started.
-	-docker rm -f $$(docker ps | grep k8s_ | awk '{print $$1}')
+	-docker ps -af name=k8s_ | awk 'NR < 2 {next}{print $$1}' | xargs -r docker rm -f
```
caseydavenport: Hm, I think the use of `xargs -r` breaks this on BSD / Mac.  See here: http://stackoverflow.com/questions/8803987/what-is-the-equivalent-to-xargs-r-under-osx

I think this needs to be tweaked so the tests can successfully run on a Mac.

insequent:
Thanks @caseydavenport. I hadn't tested on a Mac, but hopefully using the `while read` loop will work. Mind giving it a go? Also, I updated the other targets to have `##` comments, so `make help` is able to parse them correctly.

insequent:
Missed comments for `clean` and rebased with master.

caseydavenport:
OK - this worked for me on a Mac, and seems to be passing automated tests as well!

PR #393, opened 2017-04-17
Allow slash in certain names

### LINKED ISSUE #392: Allow "/" in Name fields (policy, profile etc.)
The current restriction is in place only because of path construction in etcd.  Names should be appropriately escaped for etcd so that we can use "/" - which is great for namespacing and something we already do for kdd.

With odd, it is not possible to use calicoctl to view or edit your policy due to the name restriction.

robbrockbank:
Fixes #392 

Rather than fix up every single resource that have have namespaced names (due to k8s object construction), I've focussed on the subset of objects that were affected by the original problem (you couldn't use calicoctl to view k8s KDD policy).  When we refactor libcalico-go it will be significantly easier to escape and unescape all name segments for etcd and other kv backends.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ bin
 *.coverprofile
 vendor
 .idea
+.go-pkg-cache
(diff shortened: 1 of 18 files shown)


tmjd:
LGTM

PR #394, opened 2017-04-17
Raise not supported errors in KDD for actions that are not supported.

robbrockbank:
Some KDD actions for certain resources types are not supported.  Rather than fail silently, KDD should return an error (otherwise operations through calicoctl appear to succeed).


heschlie:
LGTM

PR #396, opened 2017-04-18
Add ActiveInstanceID field under workloadEndpoint API

gunjan5:
Part of https://github.com/projectcalico/cni-plugin/issues/310
Required for https://github.com/projectcalico/cni-plugin/pull/312

### DIFF
--- a/lib/api/workloadendpoint.go
+++ b/lib/api/workloadendpoint.go
@@ -36,13 +36,23 @@ type WorkloadEndpointMetadata struct {
 	Name string `json:"name,omitempty" validate:"omitempty,namespacedname"`
 
 	// The name of the workload.
-	Workload string `json:"workload,omitempty" valid:"omitempty,namespacedname"`
+	Workload string `json:"workload,omitempty" validate:"omitempty,namespacedname"`
 
 	// The name of the orchestrator.
-	Orchestrator string `json:"orchestrator,omitempty" valid:"omitempty,namespacedname"`
+	Orchestrator string `json:"orchestrator,omitempty" validate:"omitempty,namespacedname"`
 
 	// The node name identifying the Calico node instance.
-	Node string `json:"node,omitempty" valid:"omitempty,name"`
+	Node string `json:"node,omitempty" validate:"omitempty,name"`
+
+	// ActiveInstanceID is an optional field that orchestrators may use to store additional
+	// information about the endpoint. The primary use case is to store a unique identifier
+	// for the active instance of a container. For example, with Calico CNI, a re-spawned
+	// container may use the same endpoint indexing (Node, Orchestrator, Workload, Endpoint)
+	// for the new container as for the old - the ActiveInstanceID is used to store an additional
+	// unique ID which the CNI plugin uses to determine whether the DEL operation needs to
+	// delete the Calico WorkloadEndpoint.
+	// This field is not an index field of the WorkloadEndpoint resource.
+	ActiveInstanceID string `json:activeInstanceID,omitempty" validate:"omitempty,name"`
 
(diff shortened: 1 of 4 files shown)


lib/api/workloadendpoint.go
```diff
@@ -36,13 +36,17 @@ type WorkloadEndpointMetadata struct {
 	Name string `json:"name,omitempty" validate:"omitempty,namespacedname"`
 
 	// The name of the workload.
-	Workload string `json:"workload,omitempty" valid:"omitempty,namespacedname"`
+	Workload string `json:"workload,omitempty" validate:"omitempty,namespacedname"`
+
+	// ActiveInstanceID contains an identifier for the container networked for
```
robbrockbank: Since this is optional and not actually a unique identifier, can you move this to be above the Labels field?

I don't like this wording, and it doesn't actually say what it does with it.  It should cover the following:

-  It isn't a unique identifier for the endpoint in that you cannot have multiple endpoints with the same Name, Workload, Orchestrator and Node and different ActiveInstanceIDs.
-  When you update/apply/create the endpoint, the endpoint uniquely identified by Node, Orchestrator, Workload and Name is updated to have the specified ActiveInstanceID.
-  When an ActiveInstanceID is store for the endpoint it is used to prevent deletion of the endpoint unless the delete request contains the same ActiveInstanceID  (+ rules about what happens when it isn't specified).

lib/backend/model/workloadendpoint.go
```diff
@@ -150,6 +150,7 @@ type WorkloadEndpoint struct {
 	// TODO: Validation for workload endpoint.
 	State       string            `json:"state"`
 	Name        string            `json:"name"`
+	ContainerID string            `json:"containerid"`
```
robbrockbank: Lets use one name or the other - it should be trivial to rename this in PyCharm.

lib/api/workloadendpoint.go
robbrockbank: Can we move this to be above the Label field.

I think the comment should:  Only mention the word container by way of an example, mention that it is not a unique identifier for the endpoint rather it is additional instance data that may be used by the orchestrator to provide additional consistency checks.  For example...

lib/api/workloadendpoint.go
```diff
@@ -36,13 +36,20 @@ type WorkloadEndpointMetadata struct {
...
 
 	// The name of the orchestrator.
-	Orchestrator string `json:"orchestrator,omitempty" valid:"omitempty,namespacedname"`
+	Orchestrator string `json:"orchestrator,omitempty" validate:"omitempty,namespacedname"`
 
 	// The node name identifying the Calico node instance.
-	Node string `json:"node,omitempty" valid:"omitempty,name"`
+	Node string `json:"node,omitempty" validate:"omitempty,name"`
+
+	// ActiveInstanceID contains an identifier for the WorkloadEndpoint, for example,
```
robbrockbank: Not sure I'd really understand what this is for from this comment.

robbrockbank: How about something like:

ActiveInstanceID is an optional field that orchestrators may use to store additional information about the endpoint. The primary use case is to store a unique identifier for the active instance of a container.  For example, with Calico CNI, a re-spawned container may use the same endpoint indexing (Node, Orchestrator, Workload, Endpoint) for the new container as for the old - the ActiveInstanceID is used to store an additional unique ID so that the CNI DEL operation can omit deleting the Calico WorkloadEndpoint data since it is updated for the new container instance. 

This field is not an index field of the WorkloadEndpoint resource.

gunjan5: ok done! PTAL

robbrockbank:
LGTM

PR #397, opened 2017-04-19
Add revision data for WorkloadEndpoint

robbrockbank:
Add to ObjectMetadata the revision info and plumb through for WEP
Add tests

### DIFF
--- a/lib/client/client.go
+++ b/lib/client/client.go
@@ -273,9 +273,12 @@ func (c *Client) delete(metadata unversioned.ResourceMetadata, helper conversion
 		return err
 	}
 
+	// Convert the Metadata to a Key and combine with the Metadata revision to create
+	// a KVPair for the delete operation.  At the moment only the WorkloadEndpoint Get
+	// operations fills in the revision information.
 	if k, err := helper.convertMetadataToKey(metadata); err != nil {
 		return err
-	} else if err := c.Backend.Delete(&model.KVPair{Key: k}); err != nil {
+	} else if err := c.Backend.Delete(&model.KVPair{Key: k, Revision: metadata.GetObjectMetadata().Revision}); err != nil {
 		return err
 	} else {
 		return nil
--- a/lib/client/workloadendpoint.go
+++ b/lib/client/workloadendpoint.go
@@ -166,6 +166,7 @@ func (w *workloadEndpoints) convertAPIToKVPair(a unversioned.Resource) (*model.K
 			IPv4Gateway: ah.Spec.IPv4Gateway,
 			IPv6Gateway: ah.Spec.IPv6Gateway,
 		},
+		Revision: ah.Metadata.Revision,
 	}
 
 	return &d, nil
@@ -210,6 +211,7 @@ func (w *workloadEndpoints) convertKVPairToAPI(d *model.KVPair) (unversioned.Res
 	}
 	ah.Spec.IPv4Gateway = bh.IPv4Gateway
 	ah.Spec.IPv6Gateway = bh.IPv6Gateway
+	ah.Metadata.Revision = d.Revision
 
 	return ah, nil
 }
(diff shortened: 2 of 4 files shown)


lib/client/workloadendpoint_e2e_test.go
```diff
@@ -184,8 +164,8 @@ var _ = testutils.E2eDatastoreDescribe("WorkloadEndpoint tests", testutils.Datas
 					"prod": "no",
 				}},
 			api.WorkloadEndpointMetadata{
-				Name:         "ep1/with_foo",
```
gunjan5: curious why you removed the `/` from the name?

robbrockbank: oh dammit - I actually copied this file across from my refactor branch.

I should really add that back in.  I will.

lib/client/workloadendpoint.go
```diff
@@ -63,7 +64,16 @@ func (w *workloadEndpoints) Apply(a *api.WorkloadEndpoint) (*api.WorkloadEndpoin
...
+	// Validate the supplied Metadata.
+	if err := validator.Validate(metadata); err != nil {
+		return err
+	}
+
+	k, err := w.convertMetadataToKey(metadata)
+	if err != nil {
+		return err
+	}
+	return w.c.Backend.Delete(&model.KVPair{Key: k, Revision: metadata.Revision})
```
gunjan5: why did you change the if/else chain here?Couldnt we just add Revision in the call to backend? 
```go
// Untyped get interface for deleting a single API object.  This is called from the typed
// interface.
func (c *Client) delete(metadata unversioned.ResourceMetadata, helper conversionHelper) error {
	// Validate the supplied Metadata.
	if err := validator.Validate(metadata); err != nil {
		return err
	}

	if k, err := helper.convertMetadataToKey(metadata); err != nil {
		return err
	} else if err := c.Backend.Delete(&model.KVPair{Key: k, Revision: metadata.Revision}); err != nil {
		return err
	} else {
		return nil
	}
}
```

gunjan5: ah ok nvm you wanted to keep this limited to WEP

gunjan5: mind sticking a comment so others don't trip when they modify `delete` in client.go and dont see it being reflected for WEP?

robbrockbank: I could add it generically I suppose.  It's not difficult to do.

robbrockbank: Actually - I think I will..  It will effectively be limited to WEP because no other resource type will fill in the Revision.

gunjan5: naw, lets keep the scope of this PR limited to WEP until your refactor is in :)

robbrockbank: it's actually a more compact code change i think.

gunjan5: im mostly worried about other things using this. There might be things we havent' considered that might break, for example something "working" even if it was not doing a proper CaS might break with this. Not sure

robbrockbank: If someone explicitly fills in the Revision then yes a Delete operation might fail - but in a way that is business as usual.

lib/client/client.go
```diff
@@ -273,9 +273,12 @@ func (c *Client) delete(metadata unversioned.ResourceMetadata, helper conversion
 		return err
 	}
 
+	// Convert the Metadata to a Key and combine with the Metadata revision to create
+	// a KVPair for the delete operation.  At the moment only the WorkloadEndpoint Get
+	// operations fills in the revision information.
 	if k, err := helper.convertMetadataToKey(metadata); err != nil {
 		return err
-	} else if err := c.Backend.Delete(&model.KVPair{Key: k}); err != nil {
+	} else if err := c.Backend.Delete(&model.KVPair{Key: k, Revision: metadata.GetObjectMetadata().Revision}); err != nil {
```
gunjan5: does this (`metadata.GetObjectMetadata().Revision` ) panic if `Revision` is `nil`?

robbrockbank: Nope.  The Revision has always been present in the backend.

PR #398, opened 2017-04-19
Don't use compat module for KDD datastore

### LINKED ISSUE #399: calico/node startup failing due to unsupported node operations with KDD
```
frank@tigera-fd-kubetest-master:~$ kubectl logs calico-node-kmdpr -n kube-system calico-node
time="2017-04-19T02:22:17Z" level=info msg="Loading config from environment" 
Checking datastore connection
time="2017-04-19T02:22:17Z" level=info msg="Datastore is ready" 
Datastore connection verified
time="2017-04-19T02:22:17Z" level=info msg="Unable to query node configuration" Name=tigera-fd-kubetest-01 error="operation Get is not supported on Node(name=tigera-fd-kubetest-01)" 
ERROR: Unable to access datastore to query node configuration
Terminating
Calico node failed to start
```

robbrockbank:
Fixes #399 

The handling of composite resource types in KDD was a bit wonky.  Basically for KDD we should be completely bypassing the compat module.

This PR removes the insertion of the compat module for KDD and removes the workaround code in compat.go to deal with KDD operations.

### DIFF
--- a/lib/backend/model/profile.go
+++ b/lib/backend/model/profile.go
@@ -168,13 +168,6 @@ type ProfileRules struct {
 	OutboundRules []Rule `json:"outbound_rules,omitempty" validate:"omitempty,dive"`
 }
 
-type client interface {
-	Create(object *KVPair) (*KVPair, error)
-	Update(object *KVPair) (*KVPair, error)
-	Apply(object *KVPair) (*KVPair, error)
-	Get(key Key) (*KVPair, error)
-}
-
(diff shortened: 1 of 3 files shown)


lib/backend/model/profile.go
```diff
@@ -168,13 +168,6 @@ type ProfileRules struct {
 	OutboundRules []Rule `json:"outbound_rules,omitempty" validate:"omitempty,dive"`
 }
 
-type client interface {
```
robbrockbank: This was not used anywhere.

PR #403, opened 2017-04-28
Relax label regex to just check chars, not construction

### LINKED ISSUE #404: Upgrading from older Felix deployment is rejecting endpoint config
Customer issue where older version of Felix allowed a more permissive regex for the label syntax.  

We need to relax the label syntax so that it only checks the allowed chars and does not attempt to check the construction (i.e. we don't need to check it follows a URL style syntax).

robbrockbank:
We have existing deployments that have label syntax that is more permissive than our current - this relaxes the  regex to allow these older deployments to operate.

Fixes #404

### DIFF
--- a/lib/selector/tokenizer/tokenizer.go
+++ b/lib/selector/tokenizer/tokenizer.go
@@ -55,7 +55,7 @@ type Token struct {
 
 const (
 	// LabelKeyMatcher is the base regex for a valid label key.
-	LabelKeyMatcher = `([a-zA-Z0-9_./-]{0,253}/)?[a-zA-Z0-9]([a-zA-Z0-9_.-]{0,61}[a-zA-Z0-9])?`
+	LabelKeyMatcher = `[a-zA-Z0-9_./-]{1,512}`
 	hasExpr         = `has\(\s*(` + LabelKeyMatcher + `)\s*\)`
 	allExpr         = `all\(\s*\)`
 	notInExpr       = `not\s*in\b`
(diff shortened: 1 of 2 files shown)


tmjd:
LGTM but I don't really know the origins of the validation so there may be someone else that should look at this.

lib/selector/tokenizer/tokenizer.go
```diff
@@ -55,7 +55,7 @@ type Token struct {
 
 const (
 	// LabelKeyMatcher is the base regex for a valid label key.
-	LabelKeyMatcher = `([a-zA-Z0-9_./-]{0,253}/)?[a-zA-Z0-9]([a-zA-Z0-9_.-]{0,61}[a-zA-Z0-9])?`
+	LabelKeyMatcher = `[a-zA-Z0-9_./-]{1,512}`
```
caseydavenport: We've had to do this a few times now, and I _think_ there is another place that needs to be updated IIRC.

caseydavenport: @heschlie did it last time!

robbrockbank: The validator (for the API) pulls in this value now which it originally never did - so I think we should be OK.

tmjd:
Info point.  With kubectl version 1.5.3 if I put in an invalid label name it returns `name part must match the regex ([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]`.

fasaxc:
@robbrockbank This shouldn't pose any problem in Felix.  We treat labels as opaque strings.  Good to keep them printable though, we don't want control chars in the logs.

I also checked the tokeniser logic and I don't think it'll cause any problems.  (It might if we add parens or other chars that are part of the selector syntax in other ways.)

PR #410, opened 2017-05-03
Option to disable host polling in KDD

caseydavenport:
No need to watch the k8s API for Nodes when we're not using that information.

This adds a config option in the syncer which can be set by clients that don't use node information.


lib/backend/k8s/k8s.go
```diff
@@ -68,6 +70,7 @@ type KubeConfig struct {
 	K8sCAFile                string `json:"k8sCAFile" envconfig:"K8S_CA_FILE" default:""`
 	K8sAPIToken              string `json:"k8sAPIToken" envconfig:"K8S_API_TOKEN" default:""`
 	K8sInsecureSkipTLSVerify bool   `json:"k8sInsecureSkipTLSVerify" envconfig:"K8S_INSECURE_SKIP_TLS_VERIFY" default:""`
+	K8sDisableHostPoll       bool
```
robbrockbank: K8sDisableNodePoll?  should it have json properties too?

robbrockbank:
Is there a corresponding docs change for this - I don't have a good feel for the impact on this - I assume not watching the nodes is only valid if the BGP config is never changed after start of day?

caseydavenport:
@robbrockbank I don't think there is a docs change.

PR #417, opened 2017-05-10
Add labels for IP pool

xychu:


caseydavenport:
@xychu thanks for the PR!

Could you add a bit of context to the description as to what this PR is for?

xychu:
@caseydavenport use case is something like: let admin to define IP pools with tenant labels, so different tenant users will be able to filter their own IP pools to use.

caseydavenport:
@xychu OK - I have a follow up question for you :)

Why would you like different tenants to use different IP Pools?

robbrockbank:
Hi @xychu 

The change in this PR only modifies the definition of the API structure, but does not plumb through the extra value to the backend datastore, nor adds any functionality that uses the label for selection.

The addition of labels and making them selectable on different resources is something we are looking at doing.  Would you be OK if I closed this PR and raised an Issue instead to track this.  It would be useful to add all the details about the use cases to that issue.

xychu:
@robbrockbank yeah, sure.
I've already noticed that it will not working by only adding the labels, just like profile. :-)

PR #422, opened 2017-05-18
k8s client-go update to 3.0

redbaron:


Makefile
```diff
@@ -15,11 +15,10 @@ vendor: glide.yaml
 	# To build without Docker just run "glide install -strip-vendor"
 	docker run --rm \
     -v $(CURDIR):/go/src/github.com/$(PACKAGE_NAME):rw \
-    -v $(HOME)/.glide:/home/user/.glide:rw \
```
tmjd: Why removing the caching capabilities of this?  Doesn't this greatly slow down the build?

glide.yaml
```diff
@@ -22,8 +22,10 @@ import:
 - package: gopkg.in/tchap/go-patricia.v2
   subpackages:
   - patricia
+- package: golang.org/x/text
+  version: 19e51611da83d6be54ddafce4a4af510cb3e9ea4
 - package: k8s.io/client-go
-  version: 243d8a9cb66a51ad8676157f79e71033b4014a2a
+  version: release-3.0
```
tmjd: This is currently a branch, preferably we vendor against tags or commit hashes.

run-uts
```diff
@@ -12,7 +12,7 @@ fi
 # Run tests in random order find tests recursively (-r).
 echo WHAT: $WHAT
 echo SKIP: $SKIP
-ginkgo -cover -r --skipPackage $SKIP $WHAT
+ginkgo -cover -r --skipPackage $SKIP $WHAT || { sleep 5; exit 1; }
```
tmjd: What is the purpose of this addition?

Makefile
redbaron: 2 reasons:
 - glide has a bug where  it doesn't update to latest versions of branches if "good enough" is found in cache. Therefore updates should happen with clean cache. Since there is no dedicated target for updates, I figures correctness is better than speed.
 - mounting home dirs doesn't work with Bash on Windows :)

run-uts
redbaron: Bash on Windows + Docker on Windows didn't produce any output at compilation error, possibly container was killed to soon to flush stderr or something like that. It was super confusing just to see blank screeen, fan spinning and then back to shell prompt with no indication on what went wrong. At some point I added sleep and it solved it

glide.yaml
redbaron: Exact versions are locked in `glide.lock`, unless it is deleted exact version will be downloaded during build. Usually when `glide.lock` is deleted is to update dependencies while staying within constraints of versions specified in `glide.yaml`. 

In case of client-go, given their branching policy, using a branch is a safe bet, especially given that many other deps are in `glide.yaml` are not locked (==locked to master)

caseydavenport:
@redbaron thanks for this!

@robbrockbank could you review?

Makefile
robbrockbank: I think a better approach here would be to have a separate `update-vendor` or similar make target rather than removing the cache from here - but that should be a separate PR.

Could you reinstate the cache and I'll raise a couple of issues covering:  update-vendor and getting this working in a windows environment.

robbrockbank:
This LGTM, but I'd quite like to reinstate the cache and remove any windows specific build changes - with the intention that i'll raise a couple of issues covering windows builds and vendor updates.

redbaron:
@robbrockbank updated

robbrockbank:
@redbaron Apologies for the delay in merging.  I wanted to give these changes a spin without the Makefile changes (i.e. so we'd be testing against a 1.4 Kubernetes), but I was hitting a bunch of test failures when I did that.  I'm still investigating.

robbrockbank:
Thanks for the PR @redbaron, LGTM.

Issue #423, opened 2017-05-19
IPAM doesn't go through the whole IPPools list if the first IPPool is disabled or has no IPs left


gunjan5:
## Expected Behavior

Calico IPAM should continue going through IPPools if the first one is not available

## Current Behavior

When you pass a list of IPPools to select IP address from and if the first IPPool is disabled or doesn't have any IPs available, Calico IPAM doesn't continue looking for an IP in any of the rest of the IPPools.

## Possible Solution

This seems to be the offending code: https://github.com/projectcalico/libcalico-go/blob/master/lib/client/ipam_block_reader_writer.go#L82-L89

## Steps to Reproduce (for bugs)

1. Configure CNI config to assign IPs from 2 IPPools:
```json
"ipam": {
    "type": "calico-ipam",
    "ipv4_pools": ["10.0.0.0/24", "20.0.0.0/24"]
},
```
2. Disable the first IPPool `10.0.0.0/24`
3. Start a pod
4. You'll get error: 
```json
{
    "code": 100,
    "msg": "The given pool (10.0.0.0/24) is disabled."
}
```
instead of IPAM picking an IP from the second IPPool

## Your Environment

* libcalico-go version: `v1.2.2`
* CNI version: `v1.8.3`

PR #424, opened 2017-05-19
bugfix: continue looking for IPs from rest of the IPPools if the first…

gunjan5:
… IPPool is disabled or not available

Fixes https://github.com/projectcalico/libcalico-go/issues/423

### DIFF
--- a/lib/client/ipam_block_reader_writer.go
+++ b/lib/client/ipam_block_reader_writer.go
@@ -16,7 +16,6 @@ package client
 
 import (
 	goerrors "errors"
-	"fmt"
 	"hash/fnv"
 	"math/big"
 	"math/rand"
@@ -73,37 +72,23 @@ func (rw blockReaderWriter) getAffineBlocks(host string, ver ipVersion, pools []
 }
 
 func (rw blockReaderWriter) claimNewAffineBlock(
-	host string, version ipVersion, givenPools []cnet.IPNet, config IPAMConfig) (*cnet.IPNet, error) {
+	host string, version ipVersion, requestedPools []cnet.IPNet, config IPAMConfig) (*cnet.IPNet, error) {
 
-	// If givenPools is not empty, use it.  Otherwise, default to
+	// If requestedPools is not empty, use it.  Otherwise, default to
 	// all configured pools.
 	pools := []cnet.IPNet{}
-	if len(givenPools) != 0 {
-		for _, pool := range givenPools {
-			// Validate the given pool is actually configured and matches the version.
-			if err := rw.checkPoolAvailability(pool); err != nil {
-				return nil, err
-			} else if version.Number != pool.Version() {
-				estr := fmt.Sprintf("The given pool (%s) does not match IP version %d", pool.String(), version.Number)
-				return nil, goerrors.New(estr)
-			}
-			pools = append(pools, pool)
-		}
-	} else {
-		// Default to all configured pools.
-		allPools, err := rw.client.IPPools().List(api.IPPoolMetadata{})
-		if err != nil {
-			log.Errorf("Error reading configured pools: %s", err)
-			return nil, err
-		}
 
-		// Grab all the IP networks in these pools.
-		for _, p := range allPools.Items {
-			// Don't include disabled pools or pools that don't match
-			// the requested IP version.
-			if !p.Spec.Disabled && version.Number == p.Metadata.CIDR.Version() {
-				pools = append(pools, p.Metadata.CIDR)
-			}
+	// Get all the configured pools.
+	allPools, err := rw.client.IPPools().List(api.IPPoolMetadata{})
+	if err != nil {
+		log.Errorf("Error reading configured pools: %s", err)
+		return nil, err
+	}
+
+	for _, p := range allPools.Items {
+		// Only include pools that are not disabled and are the correct version.
+		if !p.Spec.Disabled && version.Number == p.Metadata.CIDR.Version() && isPoolInRequestedPools(p.Metadata.CIDR, requestedPools) {
+			pools = append(pools, p.Metadata.CIDR)
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


lib/client/ipam_block_reader_writer.go
```diff
@@ -82,10 +82,9 @@ func (rw blockReaderWriter) claimNewAffineBlock(
...
 			// Validate the given pool is actually configured and matches the version.
 			if err := rw.checkPoolAvailability(pool); err != nil {
-				return nil, err
+				log.Errorf("IPPool not available: %s", err.Error())
 			} else if version.Number != pool.Version() {
-				estr := fmt.Sprintf("The given pool (%s) does not match IP version %d", pool.String(), version.Number)
-				return nil, goerrors.New(estr)
+				log.Errorf("The given pool (%s) does not match IP version %d", pool.String(), version.Number)
 			}
 			pools = append(pools, pool)
```
robbrockbank: Either this needs to be in an `else` branch, or you need to `continue` in the previous if clauses.

gunjan5: yes yes good point!

lib/client/ipam_block_reader_writer.go
```diff
@@ -82,10 +82,9 @@ func (rw blockReaderWriter) claimNewAffineBlock(
 		for _, pool := range givenPools {
 			// Validate the given pool is actually configured and matches the version.
 			if err := rw.checkPoolAvailability(pool); err != nil {
```
robbrockbank: checkPoolAvailability() performs a List of the pools which means it'll end up querying for each pool that is requested.   How easy would it be to make it query the pools once?

gunjan5: im going to refactor that code, not sure why it's done twice, we could just make one call and then check for the ones we care about

lib/client/ipam_block_reader_writer.go
```diff
@@ -78,30 +78,25 @@ func (rw blockReaderWriter) claimNewAffineBlock(
...
+		if len(givenPools) != 0 {
+			for _, pool := range givenPools {
+				// Check for the matching IP Pool and add if it's not disabled.
+				if reflect.DeepEqual(p.Metadata.CIDR, pool) && !p.Spec.Disabled {
+					pools = append(pools, pool)
+				}
+			}
+		} else {
+			if !p.Spec.Disabled {
 				pools = append(pools, p.Metadata.CIDR)
```
robbrockbank: You still need to check version is correct since List will return both versions.  I think this can still be simplified a bit...

When checking if the pool is one of the given pools you can just `continue` on the outer for loop using a labeled for loop.

And then after performing the given pools check you can check the IP version and the Disabled flag as a separate check (so you don't have to repeat yourself)

lib/client/ipam_block_reader_writer.go
```diff
@@ -139,6 +124,20 @@ func (rw blockReaderWriter) claimNewAffineBlock(
 	return nil, noFreeBlocksError("No Free Blocks")
 }
 
+// checkIPPoolCIDR checks if the IP Pool that is passed in belongs to the list of IP Pools
+// that should be used for assigning IPs from.
+func checkIPPoolCIDR(pool api.IPPool, givenPools []cnet.IPNet) bool {
```
gunjan5: how about renaming this to `checkIPPoolMembership` ?

robbrockbank: I was actually thinking that `givenPools` seems like an odd name (I realize this was pre-existing).  I wonder if we should rename to `requestedPools`  and then this method could be something like

isIPPoolInRequestedPools().

Or rather than passing in an IPPool() you could pass in the CIDR and then have a more generic method that checks if a CIDR is in a slice of CIDRs.  I think this is a nicer interface since it doesn't mix []IPNet and IPPool.

robbrockbank:
Looks great!  Think we need a couple of UTs to test the new branches - nothing major though.

A minor comment about method/variable naming, but could live without it.

caseydavenport:
@gunjan5 Are there corresponding test cases for the changes in this PR?

gunjan5:
tests added and renamed variables/function

robbrockbank:
It's failing semaphore at the moment

Also missing a couple of UTs to test the IPv6 pool being specified for IPv4 and the other way.

lib/client/ipam_block_reader_writer.go
```diff
@@ -139,6 +124,20 @@ func (rw blockReaderWriter) claimNewAffineBlock(
 	return nil, noFreeBlocksError("No Free Blocks")
 }
 
+// isIPPoolInRequestedPools checks if the IP Pool that is passed in belongs to the list of IP Pools
+// that should be used for assigning IPs from.
+func isIPPoolInRequestedPools(pool net.IPNet, requestedPools []cnet.IPNet) bool {
```
robbrockbank: This is a weird interface since you are comparing two different types.  It would make sense for both of these parameters to be cnet.IPNet - which would have simply required you to not deref the `CIDR.IPNet` above.  Then I think this method could be `isIPNetInIPNets` or similar since it is no longer strictly related to pools.

gunjan5: makes sense but I don't like the name `isIPNetInIPNets`, I think it's too generic and would make it hard to read. I think we should name it `isIPNetInRequestedPools`

robbrockbank: If you want IPNet to mean "IPPool" then better just left as it is, except maybe change Pools -> IPPools

robbrockbank: Should mention in the doc that requestedPools may be omitted which means allow all pools.

robbrockbank:
LGTM

lib/client/ipam_block_reader_writer.go
gunjan5: @robbrockbank we mention it here: http://docs.projectcalico.org/master/reference/cni-plugin/configuration#ipam 
```
By default, Calico IPAM will assign IP addresses from all the available IP Pools.

Optionally, the list of possible IPv4 and IPv6 pools can also be specified via the following properties:
```

Issue #425, opened 2017-05-25
Restrict IP Pool with an un-strict CIDR to be created


xychu:
## Expected Behavior

Once a IP Pool with an un-strict CIDR(e.g. 10.2.2.0/16) been created,
the `bird` will be not running and this IP Pool cannot be delete.

Maybe we should prevent such IP Pool to be created.

## Steps to Reproduce (for bugs)

1. create ippool with cidr: 10.2.2.0/16
2. try to delete it, failed with:
    
   ` DEBU[2017-05-25T15:11:47+08:00] Get Key: /calico/v1/ipam/v4/pool/10.2.0.0-16 ` 
   ` DEBU[2017-05-25T15:11:47+08:00] Key not found error ` 
    `ERRO[2017-05-25T15:11:47+08:00] [DELETE] /v2/ippools/10.2.2.0/16 GOT error: resource does not exist: IPPool(cidr=10.2.0.0/16) `

3. the `bird` failed to start, logs:

    `[root@sryslave1 calico]# tail -f -n 40 bird/current`                                                                                                                                                                                        
    `2017-05-25_06:51:05.90235 bird: /etc/calico/confd/config/bird_ipam.cfg, line 10: Invalid network prefix: 10.2.2.0/16.`                                                                                                                      
    `2017-05-25_06:51:06.90458 bird: /etc/calico/confd/config/bird_ipam.cfg, line 10: Invalid network prefix: 10.2.2.0/16.`

## Your Environment

* libcalico-go version: 12e511c
* Operating System and version:
* Link to your project (optional):

caseydavenport:
Looks to me like there are two bugs here:

### Not properly handling pools that aren't in CIDR notation

> /etc/calico/confd/config/bird_ipam.cfg, line 10: Invalid network prefix: 10.2.2.0/16.

### Not sending through the correct delete to etcd

> [DELETE] /v2/ippools/10.2.2.0/16 GOT error: resource does not exist: IPPool(cidr=10.2.0.0/16)

This seems to indicate that we're sending through a delete to `/v2/ippools/10.2.2.0/16` rather than `/v2/ippools/10.2.0.0-16` which is what we should be doing.



I think we should either:
- Normalize all IPPool requests so that everything is represented in correct CIDR format.
- OR, prevent people from creating Pools with invalid CIDRs.

[...]

PR #426, opened 2017-05-25
Add strict CIDR check in IP Pool validator

xychu:
## Description
PR for #425 

## Todos
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/lib/validator/validator.go
+++ b/lib/validator/validator.go
@@ -47,6 +47,7 @@ var (
 	reasonString        = "Reason: "
 	poolSmallIPv4       = "IP pool size is too small (min /26) for use with Calico IPAM"
 	poolSmallIPv6       = "IP pool size is too small (min /122) for use with Calico IPAM"
+	poolUnStictCIDR     = "IP pool CIDR is not strictly masked"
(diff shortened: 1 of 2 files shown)


lib/validator/validator.go
```diff
@@ -47,6 +47,7 @@ var (
 	reasonString        = "Reason: "
 	poolSmallIPv4       = "IP pool size is too small (min /26) for use with Calico IPAM"
 	poolSmallIPv6       = "IP pool size is too small (min /122) for use with Calico IPAM"
+	poolUnStictCIDR     = "IP pool CIDR is not strictly masked"
```
robbrockbank: UnStict -> Unstrict

lib/validator/validator.go
```diff
@@ -348,6 +349,16 @@ func validateIPPool(v *validator.Validate, structLevel *validator.StructLevel) {
 			}
 		}
 
+		// The Calico CIDR should be strictly masked
+		if !pool.Spec.Disabled {
```
robbrockbank: I don't think we should check the Disabled flag here - we should prevent configuration or disabled pools too.

xychu:
@robbrockbank  thanks for your review and I've updated the code.
Also add one test for this.

robbrockbank:
LGTM - than you @xychu for the contribution.

PR #428, opened 2017-05-26
Fix vendoring in Makefile

robbrockbank:
Add an update-vendor Make target and simplify and cleanup the vendoring and test targets


caseydavenport:
@robbrockbank what's the status / plan with this one?

robbrockbank:
Ah I think it needs some work.  I'll close for now.

PR #434, opened 2017-06-02
Add System Network Policy (Calico policy as a TPR)

robbrockbank:
## Description

This PR adds a new TPR type, the system-network-policy.

It allow the user to create SNP TPR resources in the kube-system namespace that Felix will interpret as system-wide Calico policy.

The format of the TPR Spec is identical to the  spec of the Calico Policy Resource.

## Notes

-  To avoid circular references it was necessary to move the etcd and kube config into the API - which I think is actually the correct place for it anyway.
-  Some of the profile->backend manipulation was moved out of client into it's own directory (converter)
-  There are UTs that test the SNP CRUD operations and Syncer notifications.
-  Whilst the CRUD code is implemented it is not exposed through the libcalico-go client API - policy resources can still only be queried for KDD.

## TODO
Would be nice to add some TPRs via kubectl and check that they are correctly interpreted.

### DIFF
--- a/lib/backend/k8s/k8s_fv_test.go
+++ b/lib/backend/k8s/k8s_fv_test.go
@@ -52,6 +52,19 @@ var (
 			},
 		},
 	}
+	calicoDisallowPolicyModel = model.Policy{
+		Order: &zeroOrder,
+		InboundRules: []model.Rule{
+			{
+				Action: "deny",
+			},
+		},
+		OutboundRules: []model.Rule{
+			{
+				Action: "deny",
+			},
+		},
+	}
 
 	// Use a back-off set of intervals for testing deletion of a namespace
 	// which can sometimes be slow.
@@ -190,17 +203,39 @@ func (c cb) ExpectDeleted(kvps []model.KVPair) {
 	}
 }
 
-// GetCheckEntryFunc returns a function that can be used to query an entry
-// in our state store.  It's useful for performing "Eventually" testing.
-func (c cb) GetCheckEntryFunc(key model.Key) func() interface{} {
+// GetSyncerValueFunc returns a function that can be used to query the value of
+// an entry in our syncer state store.  It's useful for performing "Eventually" testing.
+//
+// The returned function returns the cached entry or nil if the entry does not
+// exist in the cache.
+func (c cb) GetSyncerValueFunc(key model.Key) func() interface{} {
 	return func() interface{} {
-		log.Infof("Checking entry in cache: %s")
+		log.Infof("Checking entry in cache: %s", key)
 		c.Lock.Lock()
-		defer func() { c.Lock.Unlock() }()
+		defer func() {
+			c.Lock.Unlock()
+		}()
 		if entry, ok := c.State[key.String()]; ok {
-			return entry
+			return entry.Value
 		}
 		return nil
+	}
+}
+
+// GetSyncerValuePresentFunc returns a function that can be used to query whether an entry
+// is in our syncer state store.  It's useful for performing "Eventually" testing.
+//
+// When checking for presence use this function rather than GetSyncerValueFunc() because
+// the Value may itself by nil.
+//
+// The returned function returns true if the entry is present.
+func (c cb) GetSyncerValuePresentFunc(key model.Key) func() interface{} {
+	return func() interface{} {
+		log.Infof("Checking entry in cache: %s", key)
+		c.Lock.Lock()
+		defer func() { c.Lock.Unlock() }()
+		_, ok := c.State[key.String()]
+		return ok
 	}
 }
 
@@ -291,9 +326,9 @@ var _ = Describe("Test Syncer API for Kubernetes backend", func() {
 
 		By("Checking the correct entries are in our cache", func() {
 			expectedName := "ns.projectcalico.org/test-syncer-namespace-default-deny"
-			Eventually(cb.GetCheckEntryFunc(model.ProfileRulesKey{ProfileKey: model.ProfileKey{expectedName}})).ShouldNot(BeNil())
-			Eventually(cb.GetCheckEntryFunc(model.ProfileTagsKey{ProfileKey: model.ProfileKey{expectedName}})).ShouldNot(BeNil())
-			Eventually(cb.GetCheckEntryFunc(model.ProfileLabelsKey{ProfileKey: model.ProfileKey{expectedName}})).ShouldNot(BeNil())
+			Eventually(cb.GetSyncerValuePresentFunc(model.ProfileRulesKey{ProfileKey: model.ProfileKey{expectedName}})).Should(BeTrue())
+			Eventually(cb.GetSyncerValuePresentFunc(model.ProfileTagsKey{ProfileKey: model.ProfileKey{expectedName}})).Should(BeTrue())
+			Eventually(cb.GetSyncerValuePresentFunc(model.ProfileLabelsKey{ProfileKey: model.ProfileKey{expectedName}})).Should(BeTrue())
 		})
 
 		By("Deleting the namespace", func() {
@@ -303,9 +338,9 @@ var _ = Describe("Test Syncer API for Kubernetes backend", func() {
 
 		By("Checking the correct entries are no longer in our cache", func() {
 			expectedName := "ns.projectcalico.org/test-syncer-namespace-default-deny"
-			Eventually(cb.GetCheckEntryFunc(model.ProfileRulesKey{ProfileKey: model.ProfileKey{expectedName}}), slowCheck...).Should(BeNil())
-			Eventually(cb.GetCheckEntryFunc(model.ProfileTagsKey{ProfileKey: model.ProfileKey{expectedName}})).Should(BeNil())
-			Eventually(cb.GetCheckEntryFunc(model.ProfileLabelsKey{ProfileKey: model.ProfileKey{expectedName}})).Should(BeNil())
+			Eventually(cb.GetSyncerValuePresentFunc(model.ProfileRulesKey{ProfileKey: model.ProfileKey{expectedName}}), slowCheck...).Should(BeFalse())
+			Eventually(cb.GetSyncerValuePresentFunc(model.ProfileTagsKey{ProfileKey: model.ProfileKey{expectedName}})).Should(BeFalse())
+			Eventually(cb.GetSyncerValuePresentFunc(model.ProfileLabelsKey{ProfileKey: model.ProfileKey{expectedName}})).Should(BeFalse())
 		})
 	})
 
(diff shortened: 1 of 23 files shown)


lib/backend/k8s/resources/systemnetworkpolicies.go
```diff
@@ -0,0 +1,205 @@
...
+// through the libcalico-go API as standard Policy resources)
+func (c *snpClient) List(list model.ListInterface) ([]*model.KVPair, error) {
+	kvps := []*model.KVPair{}
+	l := list.(model.PolicyListOptions)
+
+	// If the Name is specified, k8s will return a single resource
+	// rather than a list, so handle this case separately, using our
+	// Get method to return the single result.
+	if l.Name != "" {
+		log.Info("Performing System Network Policy List with name")
```
caseydavenport: I think this should be a debug log.

lib/backend/k8s/resources/systemnetworkpolicies.go
```diff
@@ -0,0 +1,205 @@
...
+		if kvp, err := c.Get(model.PolicyKey{Name: l.Name}); err == nil {
+			kvps = append(kvps, kvp)
+		} else {
+			if !kerrors.IsNotFound(err) {
```
caseydavenport: I think the errors returned from `Get` are Calico errors, not Kubernetes errors.

lib/backend/k8s/resources/systemnetworkpolicies.go
```diff
@@ -0,0 +1,205 @@
...
+				return nil, K8sErrorToCalico(err, l)
+			}
+		}
+		return kvps, nil
+	}
+
+	// Since are not performing an exact Get, Kubernetes will return a list
```
caseydavenport: Missing `we`

lib/backend/k8s/resources/systemnetworkpolicies_conversion.go
```diff
@@ -0,0 +1,75 @@
...
+
+	return kvp
+}
+
+// SystemNetworkPolicyToThirdParty takes a Calico Policy and returns the equivalent
+// ThirdPartyResource representation.
+func SystemNetworkPolicyToThirdParty(kvp *model.KVPair) *thirdparty.SystemNetworkPolicy {
+	r, _ := converter.PolicyConverter{}.ConvertKVPairToAPI(kvp)
+
+	tprName := systemNetworkPolicyTprName(kvp.Key)
```
caseydavenport: I think it should be `systemNetworkPolicyTPRName`

lib/backend/k8s/resources/systemnetworkpolicies_conversion.go
```diff
@@ -0,0 +1,75 @@
...
+// systemNetworkPolicyTprName converts a Policy (specifically a System Network Policy)
+// name to a TPR name.
+func systemNetworkPolicyTprName(key model.Key) string {
+	// The name should be policed before we get here.
+	pk := key.(model.PolicyKey)
+	if !strings.HasPrefix(pk.Name, SystemNetworkPolicyNamePrefix) {
+		panic("System Network Policy name is not correctly namespaced")
+	}
+	// Trim the namespace and ensure lowercase.
+	return strings.ToLower(strings.TrimPrefix(pk.Name, SystemNetworkPolicyNamePrefix))
```
caseydavenport: Seems fine.  I wonder if we need to policy any special characters here?  

I guess users will still get an "Invalid name" type error out, so it's probably fine.

lib/backend/k8s/k8s_fv_test.go
```diff
@@ -386,6 +417,115 @@ var _ = Describe("Test Syncer API for Kubernetes backend", func() {
...
+			_, err := c.snpClient.Create(kvp1)
+			Expect(err).To(HaveOccurred())
+		})
+
+		By("Updating an existing System Network Policy", func() {
+			_, err := c.snpClient.Update(kvp1)
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		By("Applying a non-existent existing System Network Policy", func() {
```
robbrockbank: tee hee.. urm.   covfefe.

robbrockbank: fixed in next push.

lib/backend/k8s/k8s_fv_test.go
```diff
@@ -386,6 +425,115 @@ var _ = Describe("Test Syncer API for Kubernetes backend", func() {
...
+			_, err := c.snpClient.Apply(kvp2)
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		By("Updating the System Network Policy created by Apply", func() {
+			_, err := c.snpClient.Apply(kvp2)
```
robbrockbank: Done, in next push.

lib/backend/k8s/resources/systemnetworkpolicies.go
```diff
@@ -0,0 +1,205 @@
+// Copyright (c) 2016 Tigera, Inc. All rights reserved.
```
gunjan5: 2016?

lib/backend/k8s/thirdparty/systemnetworkpolicy.go
```diff
@@ -0,0 +1,73 @@
+package thirdparty
```
gunjan5: missing license :)

lib/backend/k8s/resources/systemnetworkpolicies.go
robbrockbank: in next push

lib/backend/k8s/resources/systemnetworkpolicies.go
robbrockbank: Correct - good spot!  And I've added another test in next push.

lib/backend/k8s/resources/systemnetworkpolicies.go
robbrockbank: k

lib/backend/k8s/resources/systemnetworkpolicies_conversion.go
```diff
@@ -0,0 +1,75 @@
...
+	return &tpr
+}
+
+// systemNetworkPolicyTprName converts a Policy (specifically a System Network Policy)
+// name to a TPR name.
+func systemNetworkPolicyTprName(key model.Key) string {
+	// The name should be policed before we get here.
+	pk := key.(model.PolicyKey)
+	if !strings.HasPrefix(pk.Name, SystemNetworkPolicyNamePrefix) {
+		panic("System Network Policy name is not correctly namespaced")
```
gunjan5: why panic instead of log.Fatal() or something else?

lib/backend/k8s/resources/systemnetworkpolicies_conversion.go
```diff
@@ -0,0 +1,75 @@
...
+	kvp.Revision = t.Metadata.ResourceVersion
+
+	return kvp
+}
+
+// SystemNetworkPolicyToThirdParty takes a Calico Policy and returns the equivalent
+// ThirdPartyResource representation.
+func SystemNetworkPolicyToThirdParty(kvp *model.KVPair) *thirdparty.SystemNetworkPolicy {
+	r, _ := converter.PolicyConverter{}.ConvertKVPairToAPI(kvp)
+
```
robbrockbank: These never threw errors, so removed from the definition.  The conversion will be having an overhaul if I ever manage to get my refactor in.

lib/backend/k8s/resources/systemnetworkpolicies_conversion.go
robbrockbank: k - although I had a feeling with golang people tend to use camel for everything... might be wrong there though.

lib/backend/k8s/resources/systemnetworkpolicies_conversion.go
robbrockbank: fine - i'll no op.

lib/backend/k8s/resources/systemnetworkpolicies_conversion.go
robbrockbank: yeah Fatal would be fine.  I'll change.

lib/backend/k8s/thirdparty/systemnetworkpolicy.go
robbrockbank: Yup - i'll go through and make sure copyright years are correct on files I've edited.

robbrockbank:
@caseydavenport @gunjan5 I believe I have address all of the review comments.

I have beefed up the FVs a little to fix issues with Namespace cleanup (we didn't previously wait for the Namespace to actually be deleted which meant the tests were running in unknown states).  I've improved that a bit.  I've also added value checking of the data in the syncer store so that I can actually test for updates of the System Network Profiles.

We probably should update our other tests to actually check update contents.

I would still like to add a test to run kubectl to create a TPR and check it updates our Syncer as expected -- and also behavior in error cases when the TPR data is badly formatted.  I'll do that tomorrow.

lib/backend/k8s/resources/systemnetworkpolicies_conversion.go
caseydavenport: Nah, it's not quite camel case.  If it's an abbreviation then it should all be capitalized.

e.g. `IPPools` vs `IpPools` and in this case `TPR` instead of `Tpr`

lib/backend/k8s/resources/systemnetworkpolicies_conversion.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+	kvp := converter.PolicyConverter{}.ConvertAPIToKVPair(r)
+	kvp.Revision = t.Metadata.ResourceVersion
+
+	return kvp
+}
+
+// SystemNetworkPolicyToThirdParty takes a Calico Policy and returns the equivalent
+// ThirdPartyResource representation.
+func SystemNetworkPolicyToThirdParty(kvp *model.KVPair) *thirdparty.SystemNetworkPolicy {
```
caseydavenport: Thinking about it now, I think these names are a little bit weird... it's not converting a SystemNetworkPolicy to a ThirdPartyResource, it's converting a Policy KVP to a SystemNetorkPolicy TPR.

So maybe these should be `PolicyToSystemNetworkPolicy` and `SystemNetworkPolicyToPolicy` or similar.

robbrockbank: Maybe, but inconsistent with IPPools, and IPPools can't really use that naming scheme.

TBH, part of the KDD routing PR will end up refactoring the TPR processing to define clear interfaces that need to be implemented, at which point the naming scheme matters less.

caseydavenport: > to define clear interfaces that need to be implemented, at which point the naming scheme matters less.

Yeah, that sounds really good.  Fine to keep as-is until then I guess.

PR #437, opened 2017-06-05
Restart single watches

neiljerram:
@caseydavenport This is where I've got to - as a basis for discussion whether it is worth keeping at all.

There are some changes here that I think turned out not to be required, so could be reverted:
- getting an initial ResourceVersion for GlobalConfig
- the 'open' arg, when reading from a watch channel; I thought it might be different from (obj != nil), but it appears always to be the same

There are other things that would need adding or finishing, for a complete PR:
- the `eventNeedsNewWatch` calls all need changing so that they do as the code now does for (obj == nil), but set `needResync`, as the code used to do, if the event indicates an error
- I think it would be good to add a fallback to the old behavior (full resync), if we restart a watch and then the next read gives (obj == nil) again - to make sure that we can't get into a tight loop there
- testing

Finally one query, in case you happen to know this.  According to ListOptions doc, I think we should be setting `Watch: true` in order to get changes since the specified version.  But we're not, and we are still only getting the new changes that we want.  So not sure why exactly why that is working.

### DIFF
--- a/lib/backend/k8s/syncer.go
+++ b/lib/backend/k8s/syncer.go
@@ -304,16 +304,16 @@ func (syn *kubeSyncer) readFromKubernetesAPI() {
 	needsResync := true
 
 	watchEventContinue := func(event watch.Event, key string) bool {
-		if event.Object == nil {
+		if event.Type == watch.Error {
+			log.Warnf("Event requires resync: %+v", event)
+			needsResync = true
+			return true
+		} else if event.Object == nil {
 			log.Warnf("Need new %v watch: %+v", key, event)
 			w := openWatchers[key]
 			log.WithField("watcher", w).Debug("Closing old watcher.")
 			w.Stop()
 			delete(openWatchers, key)
-			return true
-		} else if event.Type == watch.Error {
-			log.Warnf("Event requires resync: %+v", event)
-			needsResync = true
 			return true
 		}
 		return false
--- a/lib/backend/k8s/syncer.go
+++ b/lib/backend/k8s/syncer.go
@@ -178,6 +178,7 @@ func newSyncer(kubeAPI kubeAPI, converter converter, callbacks api.SyncerCallbac
 		tracker:         map[string]model.Key{},
 		disableNodePoll: disableNodePoll,
 		stopChan:        make(chan int),
+		openWatchers:    map[string]watch.Interface{},
 	}
 	return syn
 }
@@ -190,6 +191,7 @@ type kubeSyncer struct {
 	tracker         map[string]model.Key
 	disableNodePoll bool
 	stopChan        chan int
+	openWatchers    map[string]watch.Interface
 }
 
 // Holds resource version information.
@@ -291,33 +293,9 @@ func (syn *kubeSyncer) readFromKubernetesAPI() {
 	var event watch.Event
 	var kvp *model.KVPair
 	var opts metav1.ListOptions
-	var openWatchers = map[string]watch.Interface{}
-	closeWatchers := func() {
-		for _, w := range openWatchers {
-			log.WithField("watcher", w).Debug("Closing old watcher.")
-			w.Stop()
-		}
-		openWatchers = map[string]watch.Interface{}
-	}
 
 	// Always perform an initial snapshot.
 	needsResync := true
-
-	watchEventContinue := func(event watch.Event, key string) bool {
-		if event.Type == watch.Error {
-			log.Warnf("Event requires resync: %+v", event)
-			needsResync = true
-			return true
-		} else if event.Object == nil {
-			log.Warnf("Need new %v watch: %+v", key, event)
-			w := openWatchers[key]
-			log.WithField("watcher", w).Debug("Closing old watcher.")
-			w.Stop()
-			delete(openWatchers, key)
-			return true
-		}
-		return false
-	}
 
 	log.Info("Starting Kubernetes API read loop")
 	for {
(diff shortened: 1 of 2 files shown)


lib/backend/k8s/k8s.go
```diff
@@ -728,7 +729,7 @@ func (c *KubeClient) getGlobalConfig(k model.GlobalConfigKey) (*model.KVPair, er
 }
 
 // listGlobalConfig lists all global configs.
-func (c *KubeClient) listGlobalConfig(l model.GlobalConfigListOptions) ([]*model.KVPair, error) {
+func (c *KubeClient) listGlobalConfig(l model.GlobalConfigListOptions) ([]*model.KVPair, error, string) {
```
caseydavenport: Seems fine, but the error should be the last return value.

caseydavenport:
> Finally one query, in case you happen to know this. According to ListOptions doc, I think we should be setting Watch: true in order to get changes since the specified version. But we're not, and we are still only getting the new changes that we want. So not sure why exactly why that is working.

Weird, I wonder if that is a new change.  The watches definitely have been working as expected to-date.

lib/backend/k8s/k8s_fv_test.go
```diff
@@ -209,6 +209,10 @@ var _ = Describe("Test Syncer API for Kubernetes backend", func() {
 		go cb.ProcessUpdates()
 	})
 
+	FIt("should allow watches to timeout", func() {
```
caseydavenport: Or we can just remove this test and do unit testing or system testing of the timeout logic

neiljerram: Agreed, I was just using this for manual testing; will remove from the PR.

lib/backend/k8s/syncer.go
```diff
@@ -465,9 +502,8 @@ func (syn *kubeSyncer) performSnapshotDeletes(exists map[string]bool) {
 // a mapping of model.Key objects representing the objects which exist in the datastore, and
 // populates the provided resourceVersions with the latest k8s resource version
 // for each.
-func (syn *kubeSyncer) performSnapshot() ([]model.KVPair, map[string]bool, resourceVersions) {
+func (syn *kubeSyncer) performSnapshot(versions *resourceVersions) ([]model.KVPair, map[string]bool) {
```
neiljerram: It's because of variable declaration scope changes.

There is this decl at the top of readFromKubernetesAPI:
```
	latestVersions := resourceVersions{}
```

The old performSnapshot call,
```
			snap, existingKeys, latestVersions := syn.performSnapshot()
```
used to _shadow_ that top-level decl.  But it didn't matter, because the other places where we _used_ latestVersions were within this same scope (i.e. `if needsResync ...`).

But now the other using places (i.e. the watch setup code) are _outside_ this scope.  So it is important that we get rid of the shadowing and ensure that everyone is playing with the same, top-level, latestVersions.

lib/backend/k8s/k8s.go
neiljerram: I'm reverting this anyway - don't want to introduce any extra risk into this change.

neiljerram:
@caseydavenport @robbrockbank This is now rebased and I believe complete.  Looks like I still have a test failure to deal with - but you may like to give it another review.

lib/backend/k8s/syncer.go
```diff
@@ -280,18 +291,34 @@ func (syn *kubeSyncer) readFromKubernetesAPI() {
...
+	watchEventContinue := func(event watch.Event, key string) bool {
+		if event.Type == watch.Error {
+			log.Warnf("Event requires resync: %+v", event)
+			needsResync = true
+			return true
+		} else if event.Object == nil {
+			log.Warnf("Need new %v watch: %+v", key, event)
+			w := openWatchers[key]
+			log.WithField("watcher", w).Debug("Closing old watcher.")
+			w.Stop()
```
caseydavenport: Do we really only need to stop/restart the watcher in this case?

caseydavenport: Ah, all watchers are closed whenever we resync.  I think my general feeling of unease with this bit of code is because there's a lot of blurring of responsibilities / side-effects here.  For example, this function: sets a flag defined outside of its inner scope, returns a different flag, and sometimes closes a watcher but not always.

I'd rather have clearly defined lines in terms of responsibility: 

```
// Returns whether this event triggers a full resync.
func eventNeedsResync(e watch.Event) bool

// Returns whether this event requires a watch re-start, but
// not a full re-sync.
func eventRestartsWatch(e watch.Event) bool

// Closes a specific watcher
func closeWatcher(k key) 

// Closes all watchers (iterates over map and calls closeWatcher)
func closeAllWatchers()
```

neiljerram: Sure, I can rework it in this way - but it will result in a lot of repeated pattern code, which is really what I was trying to avoid.

caseydavenport: Yeah, I get that it removes some extra boilerplate code below, but I think it makes the code easier to read and follow.

robbrockbank:
LGTM.

I'm not sure, but I wonder if we might also need to store the entries on a per-resource type basis so that in the event a Watch fails we can perform a full re-sync of a single resource type (if say the revision version is too old for the API server to remember ... which in this case would trigger a full re-sync).

neiljerram:
@robbrockbank 
> I'm not sure, but I wonder if we might also need to store the entries on a per-resource type basis so that in the event a Watch fails we can perform a full re-sync of a single resource type (if say the revision version is too old for the API server to remember ... which in this case would trigger a full re-sync).

Yes, absolutely, that might be a useful further optimization (and one that Casey and I have discussed too). But in practice, don't you think it's likely now that your new syncer design will land soonish?  In that case, I guess that it would not make sense for me to think about adding that idea to _this_ design.  WDYT?

Issue #433, opened 2017-06-01
KDD may be resyncing too much


fasaxc:
It seems the k8s times out an inactive watch after about 30 minutes.  When one of KDD's watches times out, it does a full resync, which can be very expensive at high scale.  This introduces a lot of unnecessary latency if the watch that timed out could be restarted or only that resource resynced.

## Expected Behavior
If a watch for a low-traffic resource times out, reestablish it without doing full resync.

## Current Behavior
Full resync including expensive pod resync.

## Possible Solution
Run each watcher in own goroutine, reestablish watch without disturbing others?

robbrockbank:
@caseydavenport:  I've been thinking about this a little with the KDD routing stuff - since I wanted to use that opportunity to refactor the various different ways we do stuff.  This could relatively easily be designed in.

caseydavenport:
@robbrockbank I was planning on looking at this later today or tomorrow.  Maybe we could chat if you've got some time.

PR #443, opened 2017-06-06
Allow resyncs of individual resources

caseydavenport:
## Description

Fixes https://github.com/projectcalico/libcalico-go/issues/433 

Ideally we really should split out each resource into its own goroutine so that the resync of one resource doesn't block other watches, but this should at least substantially reduce the amount of time spent resyncing when a low-traffic watch times out.

## Todos
- [x] Update code comments 
- [x] Get tests passing

### DIFF
--- a/lib/backend/k8s/syncer.go
+++ b/lib/backend/k8s/syncer.go
@@ -171,10 +171,31 @@ func (k *realKubeAPI) getReadyStatus(key model.ReadyFlagKey) (*model.KVPair, err
 
 func newSyncer(kubeAPI kubeAPI, converter converter, callbacks api.SyncerCallbacks, disableNodePoll bool) *kubeSyncer {
 	syn := &kubeSyncer{
-		kubeAPI:         kubeAPI,
-		converter:       converter,
-		callbacks:       callbacks,
-		tracker:         map[string]model.Key{},
+		kubeAPI:   kubeAPI,
+		converter: converter,
+		callbacks: callbacks,
+		trackers: map[string]map[string]model.Key{
+			KEY_NS:  map[string]model.Key{},
+			KEY_PO:  map[string]model.Key{},
+			KEY_NP:  map[string]model.Key{},
+			KEY_SNP: map[string]model.Key{},
+			KEY_GC:  map[string]model.Key{},
+			KEY_HC:  map[string]model.Key{},
+			KEY_IP:  map[string]model.Key{},
+			KEY_NO:  map[string]model.Key{},
+			KEY_RS:  map[string]model.Key{},
+		},
+		needsResync: map[string]bool{
+			KEY_NS:  true,
+			KEY_PO:  true,
+			KEY_NP:  true,
+			KEY_SNP: true,
+			KEY_GC:  true,
+			KEY_HC:  true,
+			KEY_IP:  true,
+			KEY_NO:  true,
+			KEY_RS:  true,
+		},
 		disableNodePoll: disableNodePoll,
 		stopChan:        make(chan int),
 		openWatchers:    map[string]watch.Interface{},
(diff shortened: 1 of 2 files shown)


lib/backend/k8s/syncer.go
```diff
@@ -172,10 +172,31 @@ func (k *realKubeAPI) getReadyStatus(key model.ReadyFlagKey) (*model.KVPair, err
...
 	syn := &kubeSyncer{
-		kubeAPI:         kubeAPI,
-		converter:       converter,
-		callbacks:       callbacks,
-		tracker:         map[string]model.Key{},
+		kubeAPI:   kubeAPI,
+		converter: converter,
+		callbacks: callbacks,
+		trackers: map[string]map[string]model.Key{
+			KEY_NS:  map[string]model.Key{},
```
robbrockbank: Wouldn't it be cleaner to group together each set of per-resource data (which I think would mean the trackers, the resync flag, watcher and the current resource version?

caseydavenport: Yeah, I think it might be nice to have a struct which contains all the state for a given resource rather than a bunch of maps.

lib/backend/k8s/syncer.go
```diff
@@ -478,264 +508,323 @@ func (syn *kubeSyncer) readFromKubernetesAPI() {
...
+			// Ensure maps are initialized.
+			snap[KEY_NO] = []model.KVPair{}
+			keys[KEY_NO] = map[string]bool{}
 
 			versions.nodeVersion = noList.ListMeta.ResourceVersion
 			for _, no := range noList.Items {
 				node, err := resources.K8sNodeToCalico(&no)
 				if err != nil {
-					log.Panicf("%s", err)
+					log.Fatal("Error converting node: %s", err)
```
robbrockbank: WithError(err)

lib/backend/k8s/syncer.go
caseydavenport: I'm going to pass on this for this PR - it makes some of the little things that work nicely in this version a bit harder / more verbose to do.

lib/backend/k8s/syncer.go
```diff
@@ -171,10 +171,31 @@ func (k *realKubeAPI) getReadyStatus(key model.ReadyFlagKey) (*model.KVPair, err
...
+		needsResync: map[string]bool{
+			KEY_NS:  true,
+			KEY_PO:  true,
+			KEY_NP:  true,
+			KEY_SNP: true,
+			KEY_GC:  true,
+			KEY_HC:  true,
+			KEY_IP:  true,
+			KEY_NO:  true,
+			KEY_RS:  true,
```
neiljerram: I don't believe we ever use `needsResync[KEY_RS]`, so could delete that entry.

lib/backend/k8s/syncer.go
```diff
@@ -477,264 +507,322 @@ func (syn *kubeSyncer) readFromKubernetesAPI() {
...
-			syn.sendUpdates([]model.KVPair{*kvp})
+			syn.sendUpdates([]model.KVPair{*kvp}, KEY_IP)
 		case event = <-noChan:
 			log.Debugf("Incoming Node watch event. Type=%s", event.Type)
 			if syn.eventNeedsResync(event) {
-				needsResync = true
+				syn.needsResync[KEY_NO] = true
 				continue
 			} else if syn.eventRestartsWatch(event, KEY_NO) {
+				syn.needsResync[KEY_NO] = true
```
neiljerram: Why force this resync for Nodes, given that they are not TPR-backed?

(Although IIUC won't matter in practice, given how frequently we get Node updates.)

caseydavenport: Oh huh, good spot.  We probably don't need to do that (though I don't think it hurts).

PR #447, opened 2017-06-10
Generate events for host config when nodes change.

fasaxc:
## Description
The `IpInIpTunnelAddr` HostConfig actually comes from the Node; but the Syncer's Node watch didn't publish HostConfig events when new nodes show up.  We got away with that when we're not running Typha because Felix (and its Syncer) does a get at start of day and the pod CIDR doesn't tend to change once Felix is up.  Typha also does a get at start of day but the start of its day is  before the node appears!

## Todos
- [ ] Tests
- [x] Documentation

### DIFF
--- a/lib/backend/k8s/syncer.go
+++ b/lib/backend/k8s/syncer.go
@@ -320,7 +320,6 @@ func (syn *kubeSyncer) readFromKubernetesAPI() {
 	// Other watcher vars.
 	var nsChan, poChan, npChan, snpChan, gcChan, poolChan, noChan <-chan watch.Event
 	var event watch.Event
-	var kvp *model.KVPair
 	var opts metav1.ListOptions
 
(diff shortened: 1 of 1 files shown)


lib/backend/k8s/syncer.go
```diff
@@ -893,7 +899,7 @@ func (syn *kubeSyncer) parseNamespaceEvent(e watch.Event) []model.KVPair {
 	return []model.KVPair{*rules, *tags, *labels}
 }
 
-func (syn *kubeSyncer) parseNodeEvent(e watch.Event) *model.KVPair {
+func (syn *kubeSyncer) parseNodeEvent(e watch.Event) (kvpHostIp, kvpIPIPAddr *model.KVPair) {
```
caseydavenport: Can we remove the named returns? It's not something that shows up in the rest of the code and we've generally tried to avoid it.

PR #448, opened 2017-06-10
Fix panic caused by Node IP being passed through as (*net.IP)(nil).

fasaxc:
- Fix that our IP object couldn't be stringified if it was nil.
- Fix the node conversion to convert to a nil interface.

### DIFF
--- a/lib/backend/k8s/syncer.go
+++ b/lib/backend/k8s/syncer.go
@@ -912,8 +912,13 @@ func (syn *kubeSyncer) parseNodeEvent(e watch.Event) (*model.KVPair, *model.KVPa
 
 	kvpHostIp := &model.KVPair{
 		Key:      model.HostIPKey{Hostname: node.Name},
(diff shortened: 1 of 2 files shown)


caseydavenport:
Actually, there are a couple of failing tests.

fasaxc:
Ugh, it doesn't seem to be possible to fix this without a big rework.  If you add a `String()` method with a pointer receiver then you lose the automatic call through to the value-receiver version on the unnamed field that was being used by `fmt.Sprintf()`.  You can't have both because go doesn't support overloading and adding a value-receiver version doesn't fix the problem.

caseydavenport:
Given we merged https://github.com/projectcalico/libcalico-go/pull/449,  think we should close this.

Sounds like the String() stuff is a larger rework and we've solved the immediate problem at hand.

caseydavenport:
Added this guy: https://github.com/projectcalico/libcalico-go/issues/452

PR #453, opened 2017-06-15
[WiP] Add support for multiple CIDRS in the Net fields.

fasaxc:
@robbrockbank what do you think to this approach to adding multiple CIDRs into rules?  The good is that it's back-compatible; the bad is that the field is called `Net` but it can now be a string or a list.

### DIFF
--- a/lib/backend/model/rule.go
+++ b/lib/backend/model/rule.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2016 Tigera, Inc. All rights reserved.
+// Copyright (c) 2016-2017 Tigera, Inc. All rights reserved.
 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -19,6 +19,8 @@ import (
 	"strconv"
 	"strings"
 
+	"encoding/json"
+	"errors"
 	"github.com/projectcalico/libcalico-go/lib/net"
 	"github.com/projectcalico/libcalico-go/lib/numorstring"
 )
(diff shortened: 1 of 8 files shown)


lib/backend/model/rule.go
```diff
@@ -37,26 +39,62 @@ type Rule struct {
...
+			return err
+		}
+		*ns = append(*ns, &singleIP)
+		return nil
+	}
+	nsAsSlice := (*[]*net.IPNet)(ns)
+	return json.Unmarshal(b, &nsAsSlice)
+}
+
+func (ns IPNets) MarshalJSON() ([]byte, error) {
```
robbrockbank: In the YAML you'll get a single entry returned if configure a list with a single value in it.

I think we should be consistent about returning the same type that you configure (single value vs slice), so I think you'd need to store an explicit indication of whether it was a list or single entry rather that just using the list length.

robbrockbank: Also, I think we should define in the net package.

robbrockbank:
I'm ok with this barring comments inline, but worth getting other peoples opinions on this too.

Could we just flag this as deprecated and have a new field?  We could assert that only one or the other be used so the user cannot mix both formats, and then stop documenting the old one.

fasaxc:
Probably too horrible to have to stash a flag somewhere for 1-vs-list.  Why don't I just add a `Nets` list

caseydavenport:
@fasaxc @robbrockbank am I right in thinking this is superseded by https://github.com/projectcalico/libcalico-go/pull/470?

If so, let's close this and remove it from the v2.4.0 milestone.

robbrockbank:
@fasaxc Closing the issue - I think this is now obsolete.  Feel free to re-open if that is incorrect.

PR #454, opened 2017-06-15
Add new NetworkSet datatype.

fasaxc:
## Description
Add a "Network set" datatype to front-end and back-end models.  A network set is a labelled set of CIDRs, with the intention being that Felix would be updated to add the CIDRs from matching network sets into the IP sets that it builds and puts int he dataplane.

Why "Network set"?  We seem to fairly consistently use the work "net" for CIDRs in the datamodel and this is a set of networks.  (My first thought was "CIDR group" or similar but that'd be inconsistent.)

## Todos
- [x] Agree datamodel
- [x] Tests
- [x] [Resource documentation](https://github.com/projectcalico/calico/pull/862)
- [ ] Verify impact on RBAC with @shatrugna 
- [ ] Implement support in Felix

### DIFF
--- a/lib/api/networkset.go
+++ b/lib/api/networkset.go
@@ -54,7 +54,7 @@ type NetworkSetSpec struct {
 func NewNetworkSet() *NetworkSet {
 	return &NetworkSet{
 		TypeMetadata: unversioned.TypeMetadata{
-			Kind:       "NetworkSet",
+			Kind:       "networkSet",
 			APIVersion: unversioned.VersionCurrent,
 		},
 	}
--- a/lib/api/networkset.go
+++ b/lib/api/networkset.go
@@ -46,7 +46,7 @@ type NetworkSetMetadata struct {
 // NetworkSetSpec contains the specification for a NetworkSet resource.
 type NetworkSetSpec struct {
 	// The list of IP networks that belong to this set.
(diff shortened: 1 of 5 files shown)


lib/api/networkset.go
```diff
@@ -0,0 +1,80 @@
...
+import (
+	"github.com/projectcalico/libcalico-go/lib/api/unversioned"
+	"github.com/projectcalico/libcalico-go/lib/net"
+)
+
+// NetworkSet contains a set of arbitrary IP sub-networks/CIDRs that share labels to
+// allow rules to refer to them via selectors.  The labels of NetworkSets are in the same
+// "namespace" as those for endpoints.  NetworkSets are useful to refer to groups
+// of "external" non-Calico workloads/hosts.
+type NetworkSet struct {
```
caseydavenport: Just a thought about naming - maybe this should be something more generic like "SourceSet", etc. in case we want to add more types of things beyond "networks" down the road.

fasaxc: But it could be used in a destination rule.  `TargetSet`?  OTOH, maybe it's better to have more specific datatypes rather than something generic.  The idea is that you can identify a mix of things with labels but each think is strongly typed.

fasaxc: Just to close the loop, I think we agreed over slack that the current name was OK given that there's only one label namespace so, if we add new types of "target" later, they'll just union in with what we've got.

PR #455, opened 2017-06-15
Add common infrastructure for TPR resources and add BGP Peering config

robbrockbank:
This PR adds:
-  Generic TPR resource handling (when there is a 1:1 mapping between a Calico Resource and a TPR.
-  Global BGP Peers
-  It migrates the IPPools and System Network Policies to the generic TPR code
-  Add client and k8s tests

This should be a standalone PR - and the Per-Node BGP Peers can be added as a separate PR, so removing the [WIP] label from this PR.

Other possible change (not intending to make at this stage):
-  Move syncer code into the TPR handler and the ResourceHandler interface (so that it takes care of converting to the KVPairs automatically).
-  Register each of the TPRs or ResourceHandlers so that the CRUD operations can automatically fan out to the appropriate helper without having to have separate branches for each (basically have each ResourceHelper register as a CRUD helper).
-  Migrate Global config over to the TPR helper


lib/backend/k8s/k8s.go
```diff
@@ -59,9 +59,10 @@ type KubeClient struct {
 	converter converter
 
 	// Clients for interacting with Calico resources.
-	ipPoolClient api.Client
-	nodeClient   api.Client
-	snpClient    api.Client
+	globalBgpHandler resources.ResourceHandler
```
robbrockbank: I should probably change Handler->Client

lib/backend/k8s/resources/globalbgppeer.go
```diff
@@ -0,0 +1,153 @@
...
+		Spec: r.(*api.BGPPeer).Spec,
+	}
+	if kvp.Revision != nil {
+		tpr.Metadata.ResourceVersion = kvp.Revision.(string)
+	}
+	return &tpr, nil
+}
+
+// ResourceNameToIP converts a name used for a k8s resource to an IP address.
+func ResourceNameToIP(name string) *net.IP {
```
robbrockbank: Strictly speaking we could just get the IP address from the Spec, but what worries me here is that if the user decides to manage these via kubectl then we have no way of managing them via calicoctl if there is a mismatch between the name in the Metadata and the IP address in the Spec.  However, in hindsight I don't think this really helps.  I guess we could just ignore entries where they don't match and in that case the only way to manage those is via kubectl.

robbrockbank: I *think* I should remove this method and just get the IP from the Spec and then ignore entries where Name and IP don't match.  WDYT?

lib/backend/k8s/syncer.go
```diff
@@ -1020,7 +1004,10 @@ func (syn *kubeSyncer) parseSystemNetworkPolicyEvent(e watch.Event) *model.KVPai
 	}
 
 	// Convert the received SystemNetworkPolicy into a profile KVPair.
-	kvp := resources.ThirdPartyToSystemNetworkPolicy(np)
+	// TODO: If we can't convert the resource to a KVPair then we should
+	// TODO: at least attempt to convert to a Key so that we can delete the entry
+	// TODO: if was previously ok.
+	kvp, _ := resources.SystemNetworkPolicyConverter{}.ResourceToKVPair(np)
```
caseydavenport: Don't we need to handle this error?

lib/backend/k8s/resources/tpr.go
```diff
@@ -0,0 +1,269 @@
...
+// Interface required to satisfy use as a Kubernetes Third Party Resource List.
+type k8sList interface {
+	runtime.Object
+	metav1.ListMetaAccessor
+}
+
+// tprHandler implements the ResourceHandler interface and provides a generic
+// mechanism for a 1:1 mapping between a Calico Resource and an equivalent
+// Kubernetes Third Party Resource.
+type tprHandler struct {
```
robbrockbank: Yup that's one of the things I wanted to discuss w/ you when we do an interface overview.

lib/backend/k8s/resources/globalbgppeer.go
```diff
@@ -0,0 +1,153 @@
...
+func (_ GlobalBGPPeerConverter) ListInterfaceToKey(l model.ListInterface) model.Key {
+	pl := l.(model.GlobalBGPPeerListOptions)
+	if pl.PeerIP.IP != nil {
+		return model.GlobalBGPPeerKey{PeerIP: pl.PeerIP}
+	}
+	return nil
+}
+
+func (_ GlobalBGPPeerConverter) KeyToResourceName(k model.Key) (string, error) {
+	// The name should be policed before we get here.
```
bcreane: policed->validated?

lib/backend/k8s/resources/handler.go
```diff
@@ -0,0 +1,78 @@
...
+type ResourceConverter interface {
+	// ListInterfaceToKey converts a ListInterface to a Key if the
+	// ListInterface specifies a specific instance, otherwise returns nil.
+	ListInterfaceToKey(model.ListInterface) model.Key
+
+	// Convert the Key to the Resource name.
+	KeyToResourceName(model.Key) (string, error)
+
+	// Convert the Resource to a KVPair.
+	ResourceToKVPair(k8sResource) (*model.KVPair, error)
```
robbrockbank: How about KVPairToK8sResource and K8sResourceToKVPair?

lib/backend/k8s/resources/tpr.go
```diff
@@ -0,0 +1,269 @@
...
+		"resource":     c.resource,
+	}).Info("Get TPR by name")
+	tpr := reflect.New(c.resourceType).Interface().(k8sResource)
+	err = c.tprClient.Get().
+		Resource(c.resource).
+		Namespace("kube-system").
+		Name(name).
+		Do().Into(tpr)
+	if err != nil {
+		log.WithError(err).WithField("Resource", c.resource).Info("Error getting TPR")
```
robbrockbank: The latter log should have been removed.

lib/backend/k8s/resources/globalbgppeer.go
caseydavenport: Ok, so maybe it's trying to say "This method expects that the name has been validated by the caller", but this method doesn't take a name, it takes a Key and _returns_ a name, right?

So is it really "This method expects that the provided Key has been validated by the caller"?

lib/backend/k8s/resources/globalbgppeer.go
caseydavenport: So, the benefits of pulling it from the spec:
- We don't need to the reverse conversion
- Folks can create TPRs with actual names, so long as a valid IP is present in the spec.

The benefits of pulling it from the key:
- It's symmetrical
- Don't need to think as much about edge cases 

Personally, I think I'd lean towards pulling it from the metadata, but I could be missing some context.

lib/backend/k8s/syncer.go
```diff
@@ -1047,20 +1015,55 @@ func (syn *kubeSyncer) parseGlobalConfigEvent(e watch.Event) *model.KVPair {
...
+	if err == nil {
+		// For deletes, we need to nil out the Value part of the KVPair
+		if e.Type == watch.Deleted {
+			kvp.Value = nil
+		}
+		return kvp
 	}
-	return kvp
+
+	// Error converting resource.  Attempt to determine the Key and treat as
```
caseydavenport: Huh, not sure I understand this logic (I don't think it existed before).  

Why do we want to handle this as a delete?

lib/client/bgppeer_e2e_test.go
```diff
@@ -49,114 +30,75 @@ import (
 	"github.com/projectcalico/libcalico-go/lib/testutils"
 )
 
+// Perform CRUD operations on BGP Peer Resources
+
```
caseydavenport: Extra newline.

lib/client/bgppeer_e2e_test.go
```diff
@@ -245,3 +187,109 @@ var _ = testutils.E2eDatastoreDescribe("BGPPeer tests", testutils.DatastoreEtcdV
...
+			api.BGPPeerMetadata{
+				Scope:  scope.Scope("global"),
+				PeerIP: net.MustParseIP("20.0.0.1"),
+			},
+			api.BGPPeerSpec{
+				ASNumber: numorstring.ASNumber(6512),
+			},
+			api.BGPPeerSpec{}),
+	)
+
```
caseydavenport: Extra newline

lib/converter/policy.go
```diff
@@ -25,19 +25,22 @@ import (
...
+	return k, nil
 }
 
 // ConvertAPIToKVPair converts an API Policy structure to a KVPair containing a
 // backend Policy and PolicyKey.
-func (p PolicyConverter) ConvertAPIToKVPair(a unversioned.Resource) *model.KVPair {
+func (p PolicyConverter) ConvertAPIToKVPair(a unversioned.Resource) (*model.KVPair, error) {
 	ap := a.(api.Policy)
-	k := p.ConvertMetadataToKey(ap.Metadata)
+	k, err := p.ConvertMetadataToKey(ap.Metadata)
```
caseydavenport: When do we ever expect to get an error?  Looks like ConvertMetataToKey was updated to include an error that is always `nil`, and then we handle it here?

lib/backend/k8s/k8s_fv_test.go
```diff
@@ -625,6 +625,123 @@ var _ = Describe("Test Syncer API for Kubernetes backend", func() {
...
+			_, err := c.Create(kvp1)
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		By("Attempting to recreate an existing Global BPGP Peer", func() {
+			_, err := c.Create(kvp1)
+			Expect(err).To(HaveOccurred())
+		})
+
+		By("Updating an existing Global BPGP Peer", func() {
```
robbrockbank: Bordering on Perfection Gateway Protocol?

lib/backend/k8s/resources/customresource.go
```diff
@@ -0,0 +1,345 @@
...
+	name            string
+	resource        string
+	description     string
+	k8sResourceType reflect.Type
+	k8sListType     reflect.Type
+	converter       CustomK8sResourceConverter
+}
+
+// newCustomK8sResourceClient creates a new customK8sResourceClient.
+func newCustomK8sResourceClient(c *kubernetes.Clientset, r *rest.RESTClient,
```
robbrockbank: Actually we could just drop the constructors since it just maps directly through to creating the struct.... so having an additional struct with the same data wouldn't make any sense here.

lib/backend/k8s/k8s.go
```diff
@@ -418,9 +431,11 @@ func (c *KubeClient) Get(k model.Key) (*model.KVPair, error) {
 	case model.ReadyFlagKey:
 		return c.getReadyStatus(k.(model.ReadyFlagKey))
 	case model.IPPoolKey:
-		return c.ipPoolClient.Get(k.(model.IPPoolKey))
+		return c.ipPoolClient.Get(k)
 	case model.NodeKey:
 		return c.nodeClient.Get(k.(model.NodeKey))
```
robbrockbank: I agree :-)

lib/converter/policy.go
robbrockbank: Converting from Metadata to Key should always succeed tbh.  It would be nice to just get rid of that error from all of the interfaces in both client and k8s.

lib/backend/k8s/syncer.go
robbrockbank: So the scenario this would really address is someone modifying the TPR using kubectl and creating something that did not parse correctly.

The question is what should we do?  For calico/nodes that are just starting they won't see the entry in the datastore.  For existing calico/nodes they'll still be using the previously valid instance of the resource.  A restart of a calico/node will have the effect of deleting the entry - so it feels like the syncer should therefore treat this as a delete.

WDYT?

lib/backend/k8s/k8s.go
robbrockbank: If you like I can update the Global config as part of this PR - that way all TPR based stuff is using this approach - and the only difference then are non-TPR ones.

lib/backend/k8s/k8s.go
```diff
@@ -198,6 +200,9 @@ func (c *KubeClient) createThirdPartyResources() error {
 	if err := c.snpClient.EnsureInitialized(); err != nil {
 		return err
 	}
+	if err := c.globalBgpClient.EnsureInitialized(); err != nil {
```
robbrockbank: Easy enough to do.  I'll do it now.

lib/converter/policy.go
robbrockbank: The reason I add it here though was to be consistent with the peer converter.

lib/backend/k8s/k8s.go
```diff
@@ -173,33 +175,46 @@ func (c *KubeClient) ensureThirdPartyResources() error {
...
+	go func() { done <- c.ipPoolClient.EnsureInitialized() }()
+	go func() { done <- c.snpClient.EnsureInitialized() }()
+	go func() { done <- c.globalBgpClient.EnsureInitialized() }()
+
+	// Wait for all 4 registrations to complete and keep track of the last
+	// error to return.
+	var lastErr error
+	for i := 0; i < 4; i++ {
+		if err := <- done; err != nil {
+			lastErr = err
```
caseydavenport: We should at least log out the other errors we're hitting here so that they're not completely lost.

lib/backend/k8s/k8s.go
```diff
@@ -173,33 +175,46 @@ func (c *KubeClient) ensureThirdPartyResources() error {
...
+	}()
+
+	go func() { done <- c.ipPoolClient.EnsureInitialized() }()
+	go func() { done <- c.snpClient.EnsureInitialized() }()
+	go func() { done <- c.globalBgpClient.EnsureInitialized() }()
+
+	// Wait for all 4 registrations to complete and keep track of the last
+	// error to return.
+	var lastErr error
+	for i := 0; i < 4; i++ {
```
caseydavenport: It might be better to use a WaitGroup here: https://golang.org/pkg/sync/#WaitGroup

Then, we can wait until each goroutine is done and we won't need to keep this loop count in sync with the number of goroutines that we launch.

e.g.

```go
wg.Wait()
close(done)
for err := <- done; err != nil {
  ...
}
```

robbrockbank: I did look at that, but you still need to update the count of the WaitGroup so that it waits for the correct number.

Longer term, I'd like these things to be registered and therefore the counting would be automated and not an extra bit of code that needs adding.  But I can use a WG if it's irking you.

lib/backend/k8s/k8s.go
robbrockbank: Yup yup

lib/backend/k8s/resources/globalbgppeer.go
robbrockbank: My personal pref is to pull from the Metadata.

lib/backend/k8s/k8s.go
caseydavenport: I'd rather that be a separate PR - it can come in right after this one though!

lib/backend/k8s/k8s.go
caseydavenport: Yeah, but the updating the count is easy and self-contained in each registration.

e.g.

```go
go func() { 
    wg.Add(1)
    done <- c.ipPoolClient.EnsureInitialized()
    wg.Done()
}()
```

lib/backend/k8s/syncer.go
caseydavenport: Ok, gotcha.  I think this is fine then.

lib/backend/k8s/syncer.go
```diff
@@ -1047,20 +1015,55 @@ func (syn *kubeSyncer) parseGlobalConfigEvent(e watch.Event) *model.KVPair {
...
+	if err == nil {
+		log.WithField("Key", key).WithError(err).Error("Error in resource data, treating as deleted")
+		return &model.KVPair{
+			Key: key,
+		}
+	}
+
+	// Could not determine the Key from the resource name - all we can do is
+	// ignore this event.
+	logContext.WithError(err).Error("Error in resource data, ignoring event")
```
caseydavenport: I think we should at least log out the resource type here, and maybe "Malformed resource" or "Failed to parse {resource type}" is a more clear wording?

lib/backend/k8s/k8s.go
robbrockbank: I think you need to do the count outside the go routine no?  otherwise you'll end up with a potential timing window for count 0.

So, it's not quite as self contained...  but yes I get it.

robbrockbank: As an aside, and for the config rework:

Once the config is migrated over to use the generic code, I'd like to move the retry processing to within each go routine - so only the ones that fail actually need to retry ... as it stands if one fails we retry all of them.

Once we have that then the behavior should be to exit as soon as we hit any error.

robbrockbank: ^^ the above doesn't change the usefulness of the wait group.

caseydavenport: All SGTM

lib/backend/k8s/syncer.go
robbrockbank: We do log out the resource type.  Have a look at logContext.  It's a trick Shaun uses a lot to re-use the log construction when we want the same parameters in multiple logs.  It's a minor performance improvement and it tidies up the log code a little.

I'll change the log wording though if you like.

caseydavenport: oops, missed that.

robbrockbank: and it looks like i'm using log when i should use logContext... (further up)

robbrockbank: urm... but also i missed a log where is should be logContext... So guilty as charged (just on the wrong log)

PR #456, opened 2017-06-20
validate all the requested pools

gunjan5:
The bug is IPAM doesn't throw an error if an IP Pool is requested but it doesn't exist, it will just silently use the other valid pools  
- this was introduced in this PR https://github.com/projectcalico/libcalico-go/pull/424 specifically this line https://github.com/projectcalico/libcalico-go/pull/424/files#diff-6dcc4bd87aacb5d0037a762fbc4dae72L302
- [x] add tests

This should also fix the CNI test failures

### DIFF
--- a/lib/client/ipam_block_reader_writer.go
+++ b/lib/client/ipam_block_reader_writer.go
@@ -22,6 +22,8 @@ import (
 	"net"
 	"reflect"
 
+	"fmt"
+
 	log "github.com/Sirupsen/logrus"
 	"github.com/projectcalico/libcalico-go/lib/api"
 	"github.com/projectcalico/libcalico-go/lib/backend/model"
@@ -71,8 +73,7 @@ func (rw blockReaderWriter) getAffineBlocks(host string, ver ipVersion, pools []
 	return ids, nil
 }
 
-func (rw blockReaderWriter) claimNewAffineBlock(
-	host string, version ipVersion, requestedPools []cnet.IPNet, config IPAMConfig) (*cnet.IPNet, error) {
+func (rw blockReaderWriter) claimNewAffineBlock(host string, version ipVersion, requestedPools []cnet.IPNet, config IPAMConfig) (*cnet.IPNet, error) {
 
 	// If requestedPools is not empty, use it.  Otherwise, default to
 	// all configured pools.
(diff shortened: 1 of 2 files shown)


gunjan5:
@caseydavenport I think this code can be simplified but would require more refactoring. 
Let me know if you think there's an easier way.

lib/client/ipam_block_reader_writer.go
```diff
@@ -92,6 +93,17 @@ func (rw blockReaderWriter) claimNewAffineBlock(
 		}
 	}
 
+reqPools:
```
caseydavenport: Alternatively something like this might work? Haven't tried it though.

```go
// Build a map so we can lookup existing pools.
pm := map[string]bool{}
for _, ap := range allPools.Items {
  pm[ap.Metadata.CIDR.String()] = true
}

// Make sure each requested pool exists.
for _, rp := range requestedPools {
  if _, ok := pm[rp.String()]; !ok {
    // The requested pool doesn't exist.
    return nil, fmt.Errorf("The given pool doesn't exist")
  }
}
```

gunjan5: Good point :) but I wonder if creating a map (memory) and doing `.String()` (cpu) costs us more than just n^2 for a small n

gunjan5:
seems like the map approach also works, but I'm still not sure if it's buying us anything (or costing more) in terms of efficiency https://github.com/projectcalico/libcalico-go/pull/456#discussion_r123115789

lib/client/ipam_test.go
```diff
@@ -431,6 +432,24 @@ var _ = testutils.E2eDatastoreDescribe("IPAM tests", testutils.DatastoreEtcdV2,
...
+		// This should return an error.
+		Context("AutoAssign an IP from two pools - one valid one doesn't exist", func() {
+			args := client.AutoAssignArgs{
+				Num4:      1,
+				Num6:      0,
+				Hostname:  host,
+				IPv4Pools: []cnet.IPNet{pool1, pool5_doesnot_exist},
+			}
+
+			v4, _, error := ic.AutoAssign(args)
```
caseydavenport: Also, we shouldn't use `error` as the variable name - that's a built-in type.

gunjan5: my b - totally missed it :man_facepalming:

caseydavenport:
LGTM - once the tests pass feel free to merge!

PR #459, opened 2017-06-22
Align BGP Peer model with other per-host and global config types

robbrockbank:
This PR aligns the BGP Peer backend model with other models that support per-node and global varieties (i.e. by having separate model types for each).

This was something that I wanted to change previously and ultimately limits the potential processing the backend has to implement.

Although not strictly required, it does align more nicely with our felix and bgp config model *and* for k8s simplifies the processing since they are handled differently, and for etcd allows a less permissive root to be used for the enumeration.  I think as we move to a more k8s idiomatic indexing, the per-Node model will effectively just be phased out, or converted to a different type (using a node selector)...

This also splits out the API->KVPair conversion so that k8s may use it when we construct a k8s resource that uses the same `api.BGPPeerSpec` (as we did previously for Network Policies)

### DIFF
--- a/lib/backend/k8s/k8s_fv_test.go
+++ b/lib/backend/k8s/k8s_fv_test.go
@@ -747,7 +747,7 @@ var _ = Describe("Test Syncer API for Kubernetes backend", func() {
 	}()
 
 	It("should not error on unsupported List() calls", func() {
-		objs, err := c.List(model.BGPPeerListOptions{})
+		objs, err := c.List(model.NodeBGPPeerListOptions{})
 		Expect(err).NotTo(HaveOccurred())
 		Expect(len(objs)).To(Equal(0))
 	})
--- a/lib/backend/model/bgppeer.go
+++ b/lib/backend/model/bgppeer.go
@@ -24,7 +24,6 @@ import (
 	"github.com/projectcalico/libcalico-go/lib/errors"
 	"github.com/projectcalico/libcalico-go/lib/net"
 	"github.com/projectcalico/libcalico-go/lib/numorstring"
-	"github.com/projectcalico/libcalico-go/lib/scope"
 )
 
 var (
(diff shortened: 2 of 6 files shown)


lib/backend/model/bgppeer.go
```diff
@@ -33,111 +33,65 @@ var (
 	typeBGPPeer        = reflect.TypeOf(BGPPeer{})
 )
 
-type BGPPeerKey struct {
-	Scope    scope.Scope `json:"-" validate:"omitempty"`
-	Hostname string      `json:"-" validate:"omitempty"`
-	PeerIP   net.IP      `json:"-" validate:"required"`
+type HostBGPPeerKey struct {
```
caseydavenport: Why don't we call this `NodeBGPPeerKey` and maybe eventually even `Nodename`?

lib/converter/bgppeer.go
```diff
@@ -0,0 +1,89 @@
...
+
+import (
+	"github.com/projectcalico/libcalico-go/lib/api"
+	"github.com/projectcalico/libcalico-go/lib/api/unversioned"
+	"github.com/projectcalico/libcalico-go/lib/backend/model"
+	"github.com/projectcalico/libcalico-go/lib/errors"
+	"github.com/projectcalico/libcalico-go/lib/scope"
+)
+
+// PolicyConverter implements a set of functions used for converting between
```
robbrockbank: Thie should be BGPPeerConverter in the comments.  I'll update.

lib/backend/model/bgppeer.go
```diff
@@ -151,7 +105,70 @@ func (options BGPPeerListOptions) KeyFromDefaultPath(path string) Key {
...
+func (key GlobalBGPPeerKey) valueType() reflect.Type {
+	return typeBGPPeer
+}
+
+func (key GlobalBGPPeerKey) String() string {
+	return fmt.Sprintf("BGPPeer(global, ip=%s)", key.PeerIP)
+}
+
+type GlobalBGPPeerListOptions struct {
+	Scope    scope.Scope `json:"-" validate:"omitempty"`
```
caseydavenport: I think we can remove Scope and Hostname from this now.

lib/backend/model/bgppeer.go
```diff
@@ -151,7 +105,70 @@ func (options BGPPeerListOptions) KeyFromDefaultPath(path string) Key {
...
+	}
+}
+
+func (options GlobalBGPPeerListOptions) KeyFromDefaultPath(path string) Key {
+	log.Debugf("Get BGPPeer key from %s", path)
+	peerIP := net.IP{}
+	ekeyb := []byte(path)
+
+	if r := matchGlobalBGPPeer.FindAllSubmatch(ekeyb, -1); len(r) == 1 {
+		_ = peerIP.UnmarshalText(r[0][1])
```
caseydavenport: We should probably handle this error, even just to log it out.

lib/client/bgppeer.go
```diff
@@ -73,67 +76,62 @@ func (h *bgpPeers) Get(metadata api.BGPPeerMetadata) (*api.BGPPeer, error) {
 // that match the Metadata (wildcarding missing fields).
 func (h *bgpPeers) List(metadata api.BGPPeerMetadata) (*api.BGPPeerList, error) {
 	l := api.NewBGPPeerList()
-	err := h.c.list(metadata, h, l)
-	return l, err
+
+	// Global and host peers are listed separately.  Work out which we need
+	// to list.
+	listGlobal := (metadata.Scope == scope.Global || metadata.Scope == scope.Undefined) && metadata.Node == ""
```
caseydavenport: Shouldn't we return an error for `metadata.Scope == scope.Global && metadata.Node != ""`?

caseydavenport: I think for clarity this should really be:
```
metadata.Scope == scope.Global || (metadata.Scope == scope.Undefined && metadata.Node == "")
```

right?

lib/backend/model/bgppeer.go
robbrockbank: We can - I kept it as Host to be inline with other ones - but it was kinda bugging me.

robbrockbank:
Added more tests to check additional validation.

The converter code is tested through the client tests - but not directly in the converter module.

lib/backend/model/bgppeer.go
robbrockbank: Updated.

lib/backend/model/bgppeer.go
robbrockbank: Oops.  Yup.

lib/backend/model/bgppeer.go
robbrockbank: Handled *and* logged.

lib/client/bgppeer.go
robbrockbank: Done.  Validation is handled by the validator code.

robbrockbank: Done - validation handled in the validator.

lib/backend/k8s/k8s_fv_test.go
```diff
@@ -747,7 +747,7 @@ var _ = Describe("Test Syncer API for Kubernetes backend", func() {
 	}()
 
 	It("should not error on unsupported List() calls", func() {
-		objs, err := c.List(model.BGPPeerListOptions{})
+		objs, err := c.List(model.NodeBGPPeerListOptions{})
```
caseydavenport: We might want to change this to something else given we're going to be supporting this soon!

Not sure if you want to do it in this PR, or if you'll remember in your other ones, but it might be that this test passes even though the resource is actually supported given the assertions it's making.

robbrockbank: I'll do that as part of the next PR.  It hadn't slipped my notice that this would soon fail.

robbrockbank:
Squashed.  Will merge as soon as tests pass.

lib/backend/k8s/k8s_fv_test.go
robbrockbank: Ah I see what you mean.  I'll do this first thing as part of the PR.  On my today list.

PR #460, opened 2017-06-26
Rework Global Felix config to use new custom resource infrastructure.

robbrockbank:
Rework the existing Global Felix config to use the new generic custom resource infrastructure.

This PR sneaks in another two changes (I can remove if you'd prefer, but it's a small change):
-  The BGP Peer PR didn't register the struct types - once you do that it's necessary to use camel-case convention for the structs, so GlobalBGPPeer -> GlobalBgpPeer.
-  Standardize on file/variable format


lib/backend/k8s/k8s_fv_test.go
```diff
@@ -943,6 +943,7 @@ var _ = Describe("Test Syncer API for Kubernetes backend", func() {
 		})
 
 		By("applying a new object", func() {
+			gc.Revision = nil // Revision should not be specified when creating.
```
caseydavenport: It's only test code, but I prefer to keep comments above the things they're referring to than inline.

lib/backend/k8s/resources/globalconfig.go
```diff
@@ -0,0 +1,95 @@
...
+	return strings.ToLower(k.(model.GlobalConfigKey).Name), nil
+}
+
+func (_ GlobalConfigConverter) NameToKey(name string) (model.Key, error) {
+	return model.GlobalConfigKey{Name: name}, nil
+}
+
+func (c GlobalConfigConverter) ToKVPair(r CustomK8sResource) (*model.KVPair, error) {
+	// Since we are using the Calico API Spec definition to store the Calico
+	// GlobalConfig, use the client conversion helper to convert between KV and API.
```
caseydavenport: I don't think this comment applies to this code.

lib/backend/k8s/resources/globalconfig.go
```diff
@@ -0,0 +1,95 @@
...
+		Key: model.GlobalConfigKey{
+			Name: t.Spec.Name,
+		},
+		Value:    t.Spec.Value,
+		Revision: t.Metadata.ResourceVersion,
+	}, nil
+}
+
+func (c GlobalConfigConverter) FromKVPair(kvp *model.KVPair) (CustomK8sResource, error) {
+	name, _ := c.KeyToName(kvp.Key)
```
caseydavenport: We should handle the error here.

lib/backend/k8s/k8s_fv_test.go
robbrockbank: kk

lib/backend/k8s/resources/globalconfig.go
robbrockbank: you are quite correct.  removed.

lib/backend/k8s/resources/globalconfig.go
robbrockbank: okey dokey.

robbrockbank:
@caseydavenport changes applied.

lib/backend/k8s/resources/globalconfig.go
```diff
@@ -0,0 +1,95 @@
...
+		return model.GlobalConfigKey{Name: pl.Name}
+	}
+	return nil
+}
+
+func (_ GlobalConfigConverter) KeyToName(k model.Key) (string, error) {
+	return strings.ToLower(k.(model.GlobalConfigKey).Name), nil
+}
+
+func (_ GlobalConfigConverter) NameToKey(name string) (model.Key, error) {
```
caseydavenport: So, because we lose the capitalization information in this method I think we probably don't want / need to implement this?


Basically, the lack of symmetry: `IpInIpEnabled -> ipinipenabled` != `ipipenabled -> ipipenabled`

It doesn't look like it's used in this file, and the only other place I see is here: https://github.com/projectcalico/libcalico-go/blob/cf00825e59c15c61741d13bbe10e5491997daf60/lib/backend/k8s/syncer.go#L1057

Might it be better to have this method return an error? I think the way it is might cause unexpected behavior since

lib/backend/k8s/resources/globalconfig_test.go
```diff
@@ -0,0 +1,105 @@
...
+	key1 := model.GlobalConfigKey{
+		Name: "AbCd",
+	}
+	name1 := "abcd"
+
+	// Compatible set of key and name (Used for Name to Key conversion)
+	key2 := model.GlobalConfigKey{
+		Name: "foobaz",
+	}
+	name2 := "foobaz"
```
caseydavenport: I think this isn't really the right test - if we wanted to support this it would be `foobaz -> FooBaz`.

Related to my other comment.

lib/backend/k8s/resources/globalconfig.go
robbrockbank: Yes that is true - and IIRC I only had this method because you asked me to add it for symmetry to the interface :-D

Yeah I was toying with just removing it from the interface.  We only really need the Key to Name function for the internals.

That said, the asymmetry in the name is interesting and it raises the issue of what happens in etcd if someone uses two different capitalizations of the same config name - which one wins?

robbrockbank: Oooh, actually, the NameToKey is useful for the Syncer (and is used there), so I guess we just want to make this a no-op and perhaps raise an error for this particular NameToKey() method.

PR #461, opened 2017-06-26
Bring KDD Node client inline with other resource clients

robbrockbank:
Minor change to:
-  Update the Node client to implement the K8sResourceClient interface
-  Tweaks to the List node processing to handle qualified List.

### DIFF
--- a/lib/backend/k8s/k8s.go
+++ b/lib/backend/k8s/k8s.go
@@ -62,7 +62,7 @@ type KubeClient struct {
 	globalConfigClient resources.K8sResourceClient
 	ipPoolClient       resources.K8sResourceClient
 	snpClient          resources.K8sResourceClient
-	nodeClient         api.Client
+	nodeClient         resources.K8sResourceClient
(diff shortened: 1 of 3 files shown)


lib/backend/k8s/resources/globalconfig.go
```diff
@@ -0,0 +1,95 @@
...
+func (_ GlobalConfigConverter) KeyToName(k model.Key) (string, error) {
+	return strings.ToLower(k.(model.GlobalConfigKey).Name), nil
+}
+
+func (_ GlobalConfigConverter) NameToKey(name string) (model.Key, error) {
+	return model.GlobalConfigKey{Name: name}, nil
+}
+
+func (c GlobalConfigConverter) ToKVPair(r CustomK8sResource) (*model.KVPair, error) {
+	// Since we are using the Calico API Spec definition to store the Calico
```
caseydavenport: This comment doesn't need to be here for this one :)

PR #462, opened 2017-06-27
Move set and health packages here to avoid Felix/Typha duplication

neiljerram:
## Description
'set' and 'health' are library packages used by both Felix and Typha.  By moving them here we can eliminate duplicate code in the Felix and Typha repos - and anyway they probably should be here as this is _lib_ calico-go.

### DIFF
--- a/lib/health/health.go
+++ b/lib/health/health.go
@@ -1,0 +1,171 @@
+// Copyright (c) 2017 Tigera, Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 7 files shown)


lib/health/health.go
```diff
@@ -0,0 +1,171 @@
...
+	state := NewHealthState()
+
+	go MonitorHealth(state, neededForReady, neededForLive, c)
+
+	log.WithField("port", port).Info("Starting health endpoints")
+	http.HandleFunc("/readiness", func(rsp http.ResponseWriter, req *http.Request) {
+		log.Debug("GET /readiness")
+		status := STATUS_BAD
+		if state.Ready() {
+			log.Debug("Felix is ready")
```
caseydavenport: We shouldn't have Felix specific logging in libcalico.

caseydavenport:
I need to think about whether or not this belongs in libcalico-go some more.  It's nice to make things common, but also libcalico shouldn't be the go-to place to put any lines of code that are shared across >1 Calico components.

neiljerram:
For the record, we've agreed that libcalico-go is a reasonable home for common packages such as these.

I've just updated this PR so as to have the latest health package code, as reviewed and revised at https://github.com/projectcalico/felix/pull/1489.  Next steps:
- change Felix to use this instead of its built-in set and health packages
- finish Typha health work to use this and to implement correct liveness and readiness criteria
- review, test and merge everything.

PR #463, opened 2017-06-28
"loose" pin logrus

gunjan5:
pinning logrus to the same version semantic to avoid build issues when libcalico is pulled into CNI

```
[WARN]  --> Unable to find semantic version for constraint github.com/Sirupsen/logrus ^1.0.0, ^0.10.0
[WARN]  Unable to set version on github.com/Sirupsen/logrus to ^1.0.0, ^0.10.0. Err: Unable to update checked out version

[INFO]  Combining github.com/Sirupsen/logrus semantic version constraints ^1.0.0, ^0.10.0, ^0.10.0 and ^0.10.0
[WARN]  --> Unable to find semantic version for constraint github.com/Sirupsen/logrus ^1.0.0, ^0.10.0, ^0.10.0
[WARN]  Unable to set version on github.com/Sirupsen/logrus to ^1.0.0, ^0.10.0, ^0.10.0. Err: Unable to update checked out version
[INFO]  --> Fetching updates for k8s.io/apimachinery
[INFO]  Combining github.com/Sirupsen/logrus semantic version constraints ^1.0.0, ^0.10.0, ^0.10.0, ^0.10.0 and ^0.10.0
[WARN]  --> Unable to find semantic version for constraint github.com/Sirupsen/logrus ^1.0.0, ^0.10.0, ^0.10.0, ^0.10.0
[WARN]  Unable to set version on github.com/Sirupsen/logrus to ^1.0.0, ^0.10.0, ^0.10.0, ^0.10.0. Err: Unable to update checked out version
```

CC @lwr20

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -6,14 +6,13 @@ owners:
   email: rob@tigera.io
 import:
 - package: github.com/Sirupsen/logrus
-  version: ^0.10.0
+  version: ^1.0.0
 - package: github.com/coreos/etcd
   subpackages:
   - client
   - pkg/transport
 - package: github.com/projectcalico/go-yaml-wrapper
 - package: github.com/kelseyhightower/envconfig
-- package: github.com/onsi/ginkgo
 - package: github.com/satori/go.uuid
 - package: golang.org/x/net
   subpackages:
@@ -45,3 +44,4 @@ import:
   - tools/clientcmd
 testImport:
 - package: github.com/onsi/gomega
+- package: github.com/onsi/ginkgo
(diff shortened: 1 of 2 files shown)


caseydavenport:
@gunjan5 looks like you've got a failing test.

glide.yaml
```diff
@@ -6,14 +6,13 @@ owners:
   email: rob@tigera.io
 import:
 - package: github.com/Sirupsen/logrus
-  version: ^0.10.0
```
fasaxc: How about `>=0.10.0`?  `^1.0.0` conflicts with downstream projects that pin to `^0.10.0`

caseydavenport: Unless 1.0.0 has complicated breaking changes, feels like we should just upstream our downstream projects so they pin to ^1.0.0, yeah?

PR #464, opened 2017-06-28
Add per-node config infrastructure and per-node BGP peering

robbrockbank:
Should be ready for initial review.

This PR adds generic per-node configuration (so it *could* if necessary be used for per-node Felix config and per-node BGP config).  Longer term the per-node stuff should disappear as we take a more selector based approach to configuration - but having the generic code should make it easier if we need to add the additional per-node config as an interim measure.

This also adds per-node BGP Peering and includes e2e tests that test against etcdv2 and kdd.

Note that due to the node being an oft-updated resource, it is necessary to retry the update if there is a conflict.  I added a generic retryWrapper to handle this.  Any resource where we are tacking Calico data onto another Kubernetes resource could benefit from being wrapped.

### DIFF
--- a/lib/backend/k8s/k8s_fv_test.go
+++ b/lib/backend/k8s/k8s_fv_test.go
@@ -495,20 +495,20 @@ var _ = Describe("Test Syncer API for Kubernetes backend", func() {
 		// to fan-out Policy CRUD operations to the appropriate KDD client
 		// based on the prefix).
 		kvp1Name := "snp.projectcalico.org/my-test-snp"
-		kvp1 := &model.KVPair{
+		kvp1a := &model.KVPair{
 			Key:   model.PolicyKey{Name: kvp1Name},
 			Value: &calicoAllowPolicyModel,
 		}
-		kvp1_2 := &model.KVPair{
+		kvp1b := &model.KVPair{
 			Key:   model.PolicyKey{Name: kvp1Name},
 			Value: &calicoDisallowPolicyModel,
 		}
 		kvp2Name := "snp.projectcalico.org/my-test-snp2"
-		kvp2 := &model.KVPair{
+		kvp2a := &model.KVPair{
 			Key:   model.PolicyKey{Name: kvp2Name},
 			Value: &calicoAllowPolicyModel,
 		}
-		kvp2_2 := &model.KVPair{
+		kvp2b := &model.KVPair{
 			Key:   model.PolicyKey{Name: kvp2Name},
 			Value: &calicoDisallowPolicyModel,
 		}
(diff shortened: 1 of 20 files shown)


lib/backend/k8s/resources/retrywrapper.go
```diff
@@ -0,0 +1,131 @@
...
+	// Excessive retries.  Return the last error.
+	log.WithField("Key", object.Key).Warning("Failed to create object: too many update conflicts.")
+	return nil, err
+}
+
+func (r *retryWrapper) Update(object *model.KVPair) (*model.KVPair, error) {
+	var kvp *model.KVPair
+	var err error
+	for i := 0; i < maxActionRetries; i++ {
+		if kvp, err = r.client.Update(object); err == nil {
```
robbrockbank: Yup this wrapper is wrapped around the outside of the BGPPeer Client.  So if that client filters up a "please retry this" error, this wrapper will re-invoke the Update on the BGPPeer Client (which will perform the Get/Modify/Update).

lib/client/bgppeer_e2e_test.go
```diff
@@ -186,109 +189,3 @@ var _ = testutils.E2eDatastoreDescribe("BGPPeer tests", testutils.DatastoreEtcdV
 		})
 	})
 })
-
-// Perform CRUD operations on Global BGP Peer Resources
-var _ = testutils.E2eDatastoreDescribe("Global BGPPeer tests", testutils.DatastoreAll, func(config api.CalicoAPIConfig) {
```
robbrockbank: The guys ^^ should be sufficient.  Previously I couldn't flag the guys ^^ to be tested against KDD because of the lack of Per-Node BGP support.  Now that I can, I can get rid of these guys down below which were basically a copy of the guys above (but without any Per-node BGP Peers).

lib/backend/k8s/resources/customnoderesource.go
```diff
@@ -0,0 +1,491 @@
...
+		"Key":      kvp.Key,
+		"Value":    kvp.Value,
+		"Resource": c.resource,
+	})
+	logContext.Debug("Create per-Node resource")
+
+	// Get the names and the latest Node settings associated with the Key.
+	_, resName, node, err := c.getNamesAndNodeFromKey(kvp.Key)
+	if err != nil {
+		logContext.WithError(err).Info("Error getting current settings when creating resource.")
```
robbrockbank: I don't think it should be an Error.  Perhaps a Warning though - it doesn't necessarily mean anything bad is happening with our code, just that the upstream API is rejecting the Get Node query.

lib/backend/k8s/resources/customnoderesource.go
```diff
@@ -0,0 +1,491 @@
...
+		return nil, err
+	}
+
+	// Extract the resource from the annotations.  It should exist.
+	kvps, err := c.extractFromAnnotations(node, resName)
+	if err != nil {
+		logContext.WithError(err).Error("Error extracting annotation when updating resource")
+		return nil, err
+	}
+	if len(kvps) != 1 {
```
robbrockbank: Nope, and it'd be an error if we did.

lib/backend/k8s/resources/customnoderesource.go
```diff
@@ -0,0 +1,491 @@
...
+		return nil, err
+	}
+
+	// If the resourceName has been specified, then update the map to only
+	// include that entry.
+	if reqName != "" {
+		logContext.Debug("Restrict results to only include requested resource name")
+		val, ok := raw[reqName]
+		raw = make(map[string]string, 0)
+		if ok {
```
robbrockbank: No, this is used for Listing too.

caseydavenport: Ah, ok.  I see now.

lib/backend/k8s/resources/customnoderesource.go
```diff
@@ -0,0 +1,491 @@
...
+		}
+		nodes = nodeList.Items
+		rev = nodeList.GetResourceVersion()
+	}
+
+	// Loop through each of the nodes and extract the required data.
+	for _, node := range nodes {
+		nodeKVPs, err := c.extractFromAnnotations(&node, resName)
+		if err != nil {
+			logContext.WithField("NodeName", node.GetName()).WithError(err).Warning("Error listing resources for Node")
```
robbrockbank: I think we should always return what we can (otherwise stuffing up a single entry in a Node annotation could potentially take down all resources and render them useless.

My take would be, for all operations:
-  If the node name is specified and the node annotation cannot be parsed at all due to corrupted data:  return an error 
-  If the resource name is explicitly specified, and that specific entry within the annotation is corrupted:  return an error
-  If the resource name is not specified (which will only be the case for List) and some of the entries are corrupt (but not all of them) then return what entries within the annotation that we can.  This would minimize the impact of someone having attempted to hand edit (badly) an annotation.

robbrockbank: These are all *unexpected* error cases - but I think it's nice to be able to handle as gracefully as we can these cases.

robbrockbank:
@caseydavenport :  Feedback applied (plus a tad more).

I tweaked the interface for the custom-node-resource helper ... so that it the responsibility of the derived resource to know how to marshal and unmarshal between the annotations and the resource.

Main reason for doing this is that we may (and I think probably do) need to support per-node felix and bgp config.  I suspect in this case we might just want annotation entries such as:

felix.projectcalico.org/<config>=<value>

and 

bgp.projectcalico.org/<config>=<value>

rather than having a single entry that stores a dict of all the config values.

Hmmm, I wonder if we do that for the peers:

bgppeer.projectcalico.org/peer1=<marshalled value1>
bgppeer.projectcalico.org/peer2=<marshalled value2>
bgppeer.projectcalico.org/peer3=<marshalled value3>

lib/backend/k8s/resources/nodebgppeer_test.go
```diff
@@ -0,0 +1,131 @@
...
+		Expect(err).To(BeNil())
+		Expect(node).To(Equal("nodeY"))
+		Expect(name).To(Equal("aa-ff--0012"))
+	})
+
+	It("should convert a valid node name and resource name to a Key (IPv4)", func() {
+		key, err := converter.NodeAndNameToKey("nodeA", "1-2-3-4")
+		Expect(err).To(BeNil())
+		Expect(key).To(Equal(model.NodeBGPPeerKey{
+			Nodename: "nodeS",
```
caseydavenport: huh, shouldn't this be `nodeA`?  Why are the tests passing?

lib/backend/k8s/resources/retrywrapper.go
```diff
@@ -0,0 +1,163 @@
...
+
+func (r *retryWrapper) Create(object *model.KVPair) (*model.KVPair, error) {
+	var kvp *model.KVPair
+	var err error
+	for i := 0; i < maxActionRetries; i++ {
+		if kvp, err = r.client.Create(object); err == nil {
+			// No error, exit returning the KVPair.
+			return kvp, nil
+		} else {
+			return nil, err
```
caseydavenport: Something's not clicking here for me... isn't this just going to return after one retry whether or not the Create succeeds / fails?

caseydavenport: Ah, I think we need to check if it's a retry error (maybe Create never returns a retry error?)

lib/backend/k8s/resources/customnoderesource.go
```diff
@@ -0,0 +1,436 @@
...
+// Action strings - used for context logging.
+type action string
+
+const (
+	actApply  action = "Apply"
+	actCreate action = "Create"
+	actUpdate action = "Update"
+)
+
+// K8sResourceConverter defines an interface to map between the model and the
```
caseydavenport: stale comment - K8sResourceConverter

lib/backend/k8s/resources/customnoderesource.go
```diff
@@ -0,0 +1,436 @@
...
+	ListInterfaceToNodeAndName(model.ListInterface) (string, string, error)
+
+	// KeyToNodeAndName converts the Key to the node name and resource name.
+	KeyToNodeAndName(model.Key) (string, string, error)
+
+	// NodeAndNameToKey converts the Node name and resource name to a Key.
+	NodeAndNameToKey(string, string) (model.Key, error)
+}
+
+// customK8sNodeResourceClient implements the K8sResourceClient interface and provides a generic
```
caseydavenport: wrong capitalization in the comment here (and `NewCustomK8sNodeResourceClientConfig`)

lib/backend/k8s/resources/customnoderesource.go
```diff
@@ -0,0 +1,436 @@
...
+	logContext.Debug("Update per-Node resource")
+
+	// Get the resource name and the latest Node settings associated with the Key.
+	resName, node, err := c.getNameAndNodeFromKey(kvp.Key)
+	if err != nil {
+		logContext.WithError(err).Error("Failed to create resource")
+		return nil, err
+	}
+
+	// There should be no existing entry for a Create.
```
caseydavenport: comment is wrong - needs to be _updated_.

lib/backend/k8s/resources/customnoderesource.go
```diff
@@ -0,0 +1,436 @@
...
+func (c *customK8sNodeResourceClient) Update(kvp *model.KVPair) (*model.KVPair, error) {
+	logContext := log.WithFields(log.Fields{
+		"Key":       kvp.Key,
+		"Value":     kvp.Value,
+		"Resource":  c.ResourceType,
+		"Namespace": c.Namespace,
+	})
+	logContext.Debug("Update per-Node resource")
+
+	// Get the resource name and the latest Node settings associated with the Key.
```
caseydavenport: What are "Node settings"?  Does this mean "Node resource"?

lib/backend/k8s/resources/customnoderesource.go
```diff
@@ -0,0 +1,436 @@
...
+	// Get the name and the latest Node settings associated with the Key, and
+	// convert the name to the annotation key.
+	resName, node, err := c.getNameAndNodeFromKey(kvp.Key)
+	if err != nil {
+		logContext.WithError(err).Error("Failed to delete resource")
+		return err
+	}
+	ak := c.nameToAnnotationKey(resName)
+
+	// There should be no existing entry for a Create.
```
caseydavenport: stale comment

lib/backend/k8s/resources/customnoderesource.go
```diff
@@ -0,0 +1,436 @@
...
+			newLogContext.WithError(err).Warning("Return no results for resource list: node does not exist")
+			return kvps, "", nil
+		}
+		nodes = append(nodes, *node)
+		rev = node.GetResourceVersion()
+	} else {
+		nodeList, err := c.ClientSet.Nodes().List(metav1.ListOptions{})
+		if err != nil {
+			logContext.WithError(err).Info("Failed to list resources: unable to list Nodes")
+			err = K8sErrorToCalico(err, nodeName)
```
caseydavenport: Could we combine this line into the return below?

lib/backend/k8s/resources/customnoderesource.go
```diff
@@ -0,0 +1,436 @@
...
+
+	return kvps, rev, nil
+}
+
+func (c *customK8sNodeResourceClient) EnsureInitialized() error {
+	return nil
+}
+
+// getNameAndNodeFromKey extracts the resource name from the Key
+// and gets the current Node resource config from the Kubernetes API.
```
caseydavenport: "Node resource config" I think is just a complicated way of saying "Node" :)

lib/backend/k8s/resources/customnoderesource.go
```diff
@@ -0,0 +1,436 @@
...
+// Returns: the Resource name, The Node resource.
```
caseydavenport: weird capitalization.

lib/backend/k8s/k8s_fv_test.go
```diff
@@ -742,6 +742,163 @@ var _ = Describe("Test Syncer API for Kubernetes backend", func() {
 		})
 	})
 
+	It("should handle a CRUD of Node BGP Peer", func() {
+		var kvp1, kvp1_2, kvp2, kvp2_2 *model.KVPair
```
robbrockbank: okay okay - i've changed to 1a,1b and 2a,2b.

lib/backend/k8s/resources/retrywrapper.go
robbrockbank: Ah - yes we aren't checking the error type, so it'll be retrying all error types (which is a bit sad - although wouldn't actually do anything bad I don't think).

lib/backend/k8s/resources/nodebgppeer_test.go
robbrockbank: Well that's just embarrassing - no test suite defined, so not actually running any of these tests.  Will fix (and fix up broken tests).

robbrockbank:
@caseydavenport hopefully last iteration

lib/backend/k8s/resources/customresource.go
```diff
@@ -103,9 +103,12 @@ func (c *customK8sResourceClient) Create(kvp *model.KVPair) (*model.KVPair, erro
 		return nil, K8sErrorToCalico(err, kvp.Key)
 	}
 
-	// Update the revision information from the response.
-	kvp.Revision = resOut.GetObjectMeta().GetResourceVersion()
-	return kvp, nil
+	// Return the Key and Value with updated Revision information.
```
robbrockbank: Well we don't really need it although it makes the testing easier.   Reasoning is, if you call

kvp_out, err := client.Create(kvp_in)

You wouldn't expect kvp_in to change, and therefore you might wish to use it again.  But as it stands we'll update kvp_in and just return it as kvp_out.

Anyhow, I've removed as it wasn't needed for this PR.

caseydavenport:
@robbrockbank feel free to merge after squashing :crossed_fingers:

PR #467, opened 2017-06-30
Fixed bug preventing specific profiles from being accessed from namespaces

mgleung:
## Description
Fixed a bug where the calico prefix on namespaces was being used to query for profiles (which was invalid). Also added a debug statement to display errors if there are issues with accessing profile information from the KDD.

Also fixed tests which were expecting to use etcd notation for namespaces against the kube client

Fixes https://github.com/projectcalico/calicoctl/issues/1658

## Tests
Set up Calico using the kubernetes datastore.
Ran ./calicoctl -l debug get profile ns.projectcalico.org/<name> and verified profile information was returned

### DIFF
--- a/lib/backend/k8s/k8s.go
+++ b/lib/backend/k8s/k8s.go
@@ -471,6 +471,7 @@ func (c *KubeClient) listProfiles(l model.ProfileListOptions) ([]*model.KVPair, 
 	if l.Name != "" {
 		kvp, err := c.getProfile(model.ProfileKey{Name: l.Name})
 		if err != nil {
+			log.WithError(err).Debug("Error retrieving profile")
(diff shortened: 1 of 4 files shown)


lib/backend/k8s/k8s.go
```diff
@@ -471,6 +471,7 @@ func (c *KubeClient) listProfiles(l model.ProfileListOptions) ([]*model.KVPair,
 	if l.Name != "" {
 		kvp, err := c.getProfile(model.ProfileKey{Name: l.Name})
 		if err != nil {
+			log.Debugf("Error retrieving profile: %s", err)
```
caseydavenport: I think we generally prefer to use `log.WithError(err).Debug(...)` for this sort of thing.

PR #468, opened 2017-07-03
Add per-node and global BGP config to KDD

robbrockbank:
Adds per-node and global BGP config to KDD.   Most importantly the global BGP config option to enabled/disable the node mesh is present.

This PR includes some changes to the client and etcd code too:
Reason:
-  etcd exposes the BGP config parameters in an odd way:
  -  capitalization differs from the Felix "standard"
  -  the node-2-node mesh is a json blob rather than a single boolean value.

Changes:
-  Hide the etcd oddities in the compat driver - so on the KVPair interface we'll just define camelcase options AsNumber NodeMeshEnabled (this is what Felix uses, should we? ... or should I change AsNumber to ASNumber?) parameters.
-  compat module will convert these to the original as_num and node_mesh.
-  In addition, the NodeMeshEnabled will be exposed in the KVPair as a boolean string - in compat we'll convert this to the json blob.

This allows the KVPair interface to expose sensible config names and values with consistent format as used by Felix.  I think this is *only* used by etcd so I'm wondering if we should just be more explicit about that and do all conversion in that module.   

Converting KVPairs / Keys in the compat module is a bit of a PITA because error responses from the etcd driver will have the modified Key as the Identifier -- so that means we have to jump through hoops to adjust the Identifier in the error back to the original Key.  Nothing terrible, but a side effect of adding the identifier inforamation in the backend when it should really be handed by the client.  Anyhoo.

### DIFF
--- a/lib/backend/compat/compat.go
+++ b/lib/backend/compat/compat.go
@@ -15,9 +15,11 @@
 package compat
 
 import (
+	"encoding/json"
 	goerrors "errors"
 
 	log "github.com/Sirupsen/logrus"
+
 	"github.com/projectcalico/libcalico-go/lib/backend/api"
 	"github.com/projectcalico/libcalico-go/lib/backend/model"
 	"github.com/projectcalico/libcalico-go/lib/errors"
@@ -46,7 +48,7 @@ func (c *ModelAdaptor) EnsureCalicoNodeInitialized(node string) error {
 // Create an entry in the datastore.  This errors if the entry already exists.
 func (c *ModelAdaptor) Create(d *model.KVPair) (*model.KVPair, error) {
 	var err error
-	switch d.Key.(type) {
+	switch k := d.Key.(type) {
 	case model.ProfileKey:
 		t, l, r := ToTagsLabelsRules(d)
 		if t, err = c.client.Create(t); err != nil {
@@ -76,6 +78,14 @@ func (c *ModelAdaptor) Create(d *model.KVPair) (*model.KVPair, error) {
 		b, err := c.client.Create(d)
 		if err != nil {
 			return nil, err
+		}
+		d.Revision = b.Revision
+		return d, nil
+	case model.GlobalBGPConfigKey:
+		nd := toDatastoreGlobalBGPConfig(*d)
+		b, err := c.client.Create(nd)
+		if err != nil {
+			return nil, errors.UpdateErrorIdentifier(err, k)
 		}
 		d.Revision = b.Revision
 		return d, nil
@@ -121,6 +131,14 @@ func (c *ModelAdaptor) Update(d *model.KVPair) (*model.KVPair, error) {
 		}
 		d.Revision = b.Revision
 		return d, nil
+	case model.GlobalBGPConfigKey:
+		nd := toDatastoreGlobalBGPConfig(*d)
+		b, err := c.client.Update(nd)
+		if err != nil {
+			return nil, errors.UpdateErrorIdentifier(err, d.Key)
+		}
+		d.Revision = b.Revision
+		return d, nil
 	default:
 		return c.client.Update(d)
 	}
@@ -138,7 +156,7 @@ func (c *ModelAdaptor) Apply(d *model.KVPair) (*model.KVPair, error) {
 		} else if _, err := c.client.Apply(l); err != nil {
 			return nil, err
 		} else if _, err := c.client.Apply(r); err != nil {
-			return nil, err
+			return nil, errors.UpdateErrorIdentifier(err, d.Key)
 		} else {
 			d.Revision = t.Revision
 			return d, nil
@@ -163,6 +181,14 @@ func (c *ModelAdaptor) Apply(d *model.KVPair) (*model.KVPair, error) {
 		}
 		d.Revision = b.Revision
 		return d, nil
+	case model.GlobalBGPConfigKey:
+		nd := toDatastoreGlobalBGPConfig(*d)
+		b, err := c.client.Apply(nd)
+		if err != nil {
+			return nil, errors.UpdateErrorIdentifier(err, d.Key)
+		}
+		d.Revision = b.Revision
+		return d, nil
 	default:
 		return c.client.Apply(d)
 	}
@@ -181,6 +207,10 @@ func (c *ModelAdaptor) Delete(d *model.KVPair) error {
 			return err
 		}
 		return nil
+	case model.GlobalBGPConfigKey:
+		nd := toDatastoreGlobalBGPConfig(*d)
+		err := c.client.Delete(nd)
+		return errors.UpdateErrorIdentifier(err, d.Key)
 	default:
 		return c.client.Delete(d)
 	}
@@ -195,6 +225,13 @@ func (c *ModelAdaptor) Get(k model.Key) (*model.KVPair, error) {
 		return c.getNode(kt)
 	case model.BlockKey:
 		return c.getBlock(k)
+	case model.GlobalBGPConfigKey:
+		nk := toDatastoreGlobalBGPConfigKey(kt)
+		if kvp, err := c.client.Get(nk); err != nil {
+			return nil, errors.UpdateErrorIdentifier(err, k)
+		} else {
+			return fromDatastoreGlobalBGPConfig(*kvp), nil
+		}
 	default:
 		return c.client.Get(k)
 	}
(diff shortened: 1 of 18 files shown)


lib/backend/compat/compat.go
```diff
@@ -617,36 +664,134 @@ func toNodeDeleteComponents(d *model.KVPair) (primary *model.KVPair, optional []
...
+	}
+	return key
+}
+
+// toDatastoreGlobalBGPConfig modifies the Global BGP Config KVPair to the format required in the
+// datastore (for back-compatibility).
+func toDatastoreGlobalBGPConfig(d model.KVPair) *model.KVPair {
+	// Copy the KVPair, so we aren't modifying the original.
+	k := d.Key.(model.GlobalBGPConfigKey)
+	d.Key = toDatastoreGlobalBGPConfigKey(k)
```
caseydavenport: Hm, so we're copying it so that toDatastoreGlobalBGPConfigKey doesn't modify the original, but then we're overwriting the original in the same line?

Are we just keeping this so that we can use `k.Name` below? If so, why don't we either:
- Switch on the "datastore" representation of the name, OR
- Just save off the same and use that in the switch below.

Or, am I misunderstanding?

lib/backend/k8s/k8s.go
```diff
@@ -397,15 +415,19 @@ func (c *KubeClient) Delete(d *model.KVPair) error {
 	log.Debugf("Performing 'Delete' for %+v", d)
 	switch d.Key.(type) {
 	case model.GlobalConfigKey:
-		return c.globalConfigClient.Delete(d)
+		return c.globalFelixConfigClient.Delete(d)
```
caseydavenport: Hm, so this is nice because it makes it clear that it's for felix, however the model is called `GlobalConfig`, so it might be nice to keep this in-sync with what the model is called.

WDYT?

lib/backend/k8s/resources/globalbgpconfig.go
```diff
@@ -0,0 +1,97 @@
...
+	GlobalBgpConfigTPRName      = "global-bgp-config.projectcalico.org"
+)
+
+func NewGlobalBGPConfigClient(c *kubernetes.Clientset, r *rest.RESTClient) K8sResourceClient {
+	return &customK8sResourceClient{
+		clientSet:       c,
+		restClient:      r,
+		name:            GlobalBgpConfigTPRName,
+		resource:        GlobalBgpConfigResourceName,
+		description:     "Calico Global Configuration",
```
caseydavenport: Calico Global BGP Configuration?

lib/backend/k8s/resources/globalbgpconfig.go
```diff
@@ -0,0 +1,97 @@
...
+		converter:       GlobalBgpConfigConverter{},
+	}
+}
+
+// GlobalBgpConfigConverter implements the K8sResourceConverter interface.
+type GlobalBgpConfigConverter struct {
+}
+
+func (_ GlobalBgpConfigConverter) ListInterfaceToKey(l model.ListInterface) model.Key {
+	pl := l.(model.GlobalBGPConfigListOptions)
```
caseydavenport: What is `pl` short for?  Could we just do `name := l.(model.GlobalBGPConfigListOptions).Name`?

lib/backend/model/bgpconfig.go
```diff
@@ -17,13 +17,18 @@ package model
...
+
+	log "github.com/Sirupsen/logrus"
 
 	"github.com/projectcalico/libcalico-go/lib/errors"
 )
 
 var (
-	typeGlobalBGPConfig = rawStringType
-	typeHostBGPConfig   = rawStringType
+	matchGlobalBGPConfig = regexp.MustCompile("^/?calico/bgp/v1/global/(.+)$")
```
caseydavenport: Hm, what are the changes to this file for?

lib/client/config_e2e_test.go
```diff
@@ -89,6 +89,89 @@ var _ = testutils.E2eDatastoreDescribe("with config option API tests", testutils
 		Expect(cs).To(Equal(client.ConfigLocationGlobal))
 	})
 
+	It("should handle node IP in IP tunnel address", func() {
```
caseydavenport: Right, so this fails with KDD because KDD gets the tunnel address from the Node.PodCIDR field, and we needed to move it to an etcd-only set of tests, yeah?  Just making sure I understand.

lib/client/config_e2e_test.go
```diff
@@ -89,6 +89,89 @@ var _ = testutils.E2eDatastoreDescribe("with config option API tests", testutils
...
+		By("checking address set to nil")
+		err = config.SetNodeIPIPTunnelAddress("node1", nil)
+		Expect(err).NotTo(HaveOccurred())
+
+		ip, err = config.GetNodeIPIPTunnelAddress("node1")
+		Expect(err).NotTo(HaveOccurred())
+		Expect(ip).To(BeNil())
+	})
+
+	It("should handle per-node felix config", func() {
```
caseydavenport: This should work now in KDD right?

lib/errors/errors.go
```diff
@@ -126,3 +126,31 @@ type ErrorResourceUpdateConflict struct {
...
 }
+
+// UpdateErrorIdentifier modifies the supplied error to use the new resource
+// identifier.
+func UpdateErrorIdentifier(err error, id interface{}) error {
+	if err == nil {
+		return nil
+	}
+
+	switch e := err.(type) {
```
caseydavenport: Hm, why do we need to create a new var `e` and then re-assign to `err`?

Couldn't we just do `err.Identifier= id` for each of these?

lib/backend/compat/compat.go
robbrockbank: Just keeping it to switch on Name.  We can combine the two lines above and then switch on the DS name.  No issues doing that (in fact I think that's what I originally had).

robbrockbank: Oh the Copy KVPair comment is bogus.  It's already copied by virtue of us not passing in a pointer.

lib/backend/k8s/k8s.go
robbrockbank: Yeah don't mind either way.  Makes it a little less verbose and easy to rename.

lib/backend/model/bgpconfig.go
robbrockbank: We didn't previously support list operations for config.  I was adding that in, although for the life of me I can't remember why.

robbrockbank: Oh I think it was for "no surprises" later on -- was thinking ahead about moving away from confd and using a common set of library function for both backends to configure bird.  Strictly not required for this PR though.

lib/client/config_e2e_test.go
robbrockbank: Yup - and that is why I didn't just add node-specific Felix config in this PR.  It's not required for KDD routing, and needs a little more thought regarding interactions with this field.

lib/client/config_e2e_test.go
robbrockbank: Nope - as with the above, I haven't added per-node Felix config to KDD.  It isn't required for KDD routing.

I think we should add it soon, and then we should be feature complete (in terms of what can be configured through calicoctl).

lib/errors/errors.go
robbrockbank: ha... yes, oops.

robbrockbank: No we cannot.

It's a golang thang: `e := err.(type)` dereferences the interface "err" thus creating a copy of the exception struct.  So you are actually modifying a copy of the exception that is passed in.

lib/backend/k8s/k8s.go
```diff
@@ -183,13 +188,14 @@ func (c *KubeClient) createThirdPartyResources() error {
 	done := make(chan error)
 	go func() { done <- c.ipPoolClient.EnsureInitialized() }()
 	go func() { done <- c.snpClient.EnsureInitialized() }()
-	go func() { done <- c.globalBgpClient.EnsureInitialized() }()
+	go func() { done <- c.globalBgpPeerClient.EnsureInitialized() }()
 	go func() { done <- c.globalConfigClient.EnsureInitialized() }()
+	go func() { done <- c.globalBgpConfigClient.EnsureInitialized() }()
 
 	// Wait for all 4 registrations to complete and keep track of the last
```
caseydavenport: Comment is out-of-date

PR #469, opened 2017-07-05
Expose per-node listing as external interfaces

robbrockbank:
Add the ability to query the per-node resources from the k8s Node config.

Whilst generic, the primary focus of this is for simplifying the processing in confd.

### DIFF
--- a/lib/backend/k8s/resources/client.go
+++ b/lib/backend/k8s/resources/client.go
@@ -16,6 +16,8 @@ package resources
 
 import (
 	"github.com/projectcalico/libcalico-go/lib/backend/model"
+
+	apiv1 "k8s.io/client-go/pkg/api/v1"
 )
 
 // K8sResourceClient is the interface to the k8s datastore for CRUD operations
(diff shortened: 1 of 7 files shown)


lib/backend/k8s/resources/customnoderesource.go
```diff
@@ -433,3 +445,11 @@ func (c *customK8sNodeResourceClient) extractResourcesFromAnnotation(node *apiv1
 
 	return kvps, nil
 }
+
+// ExtractResourcesFromNode returns the resources stored in the Node configuration.
+//
+// This convenience method is not guaranteed to be part of the libcalico-go API
```
caseydavenport: nit: the wording of this seems to imply guarantees that I don't think we necessarily make about other methods in libcalico-go :)

Might be better worded as: "This convenience method is expected to be removed in a future libcalico-go release"

lib/backend/k8s/resources/client.go
```diff
@@ -63,3 +65,11 @@ type K8sResourceClient interface {
 	// any ready to be used.
 	EnsureInitialized() error
 }
+
+// K8sNodeResourceClient extends the K8sResourceClient to add a helper method to
+// extract resources from the supplied K8s Node.  This additional convenience method
```
caseydavenport: s/method/interface (and see below comment)

PR #470, opened 2017-07-06
Add a Nets field, deprecate the Net field.

fasaxc:
## Description
Alternative approach to allowing multiple nets in a rule.  

- Deprecate the net field, add a new nets field.  
- Validate against specifying both.
- On read, convert net to nets.

## Todos
- [x] Tests
- [x] [Documentation](https://github.com/projectcalico/libcalico-go/pull/470)

## Release Note

```release-note
Add new Source.Nets and Destination.Nets fields (and their negated couterparts) 
to rules, allowing multiple CIDRs to be matched in a single rule.  The Source.Net 
and Destination.Net fields are now deprecated; when reading back data that 
contains a Net field, it will be converted to a single-entry Nets field.  Felix (and 
Typha, if in use) should be upgraded before using the new Nets fields in a rule.
```

### DIFF
--- a/lib/api/rule.go
+++ b/lib/api/rule.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2016 Tigera, Inc. All rights reserved.
+// Copyright (c) 2016-2017 Tigera, Inc. All rights reserved.
 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -75,7 +75,7 @@ type ICMPFields struct {
 // to a particular entity (that is either the source or destination).
 //
 // A source EntityRule matches the source endpoint and originating traffic.
-// A desination EntityRule matches the destination endpoint and terminating traffic.
+// A destination EntityRule matches the destination endpoint and terminating traffic.
 type EntityRule struct {
 	// Tag is an optional field that restricts the rule to only apply to traffic that
 	// originates from (or terminates at) endpoints that have profiles with the given tag
@@ -84,7 +84,12 @@ type EntityRule struct {
 
 	// Net is an optional field that restricts the rule to only apply to traffic that
 	// originates from (or terminates at) IP addresses in the given subnet.
+	// Deprecated: superseded by the Nets field.
 	Net *net.IPNet `json:"net,omitempty" validate:"omitempty"`
+
+	// Nets is an optional field that restricts the rule to only apply to traffic that
+	// originates from (or terminates at) IP addresses in any of the given subnets.
+	Nets []*net.IPNet `json:"nets,omitempty" validate:"omitempty"`
 
 	// Selector is an optional field that contains a selector expression (see Policy for
 	// sample syntax).  Only traffic that originates from (terminates at) endpoints matching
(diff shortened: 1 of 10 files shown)


fasaxc:
@robbrockbank think I prefer this.  WDYT?

lib/api/rule.go
```diff
@@ -128,3 +139,28 @@ type EntityRule struct {
 	// Protocol match in the Rule to be set to "tcp" or "udp".
 	NotPorts []numorstring.Port `json:"notPorts,omitempty" validate:"omitempty,dive"`
 }
+
+func combineNets(n *net.IPNet, nets []*net.IPNet) []*net.IPNet {
```
caseydavenport: Is there a requirement to support both `nets` and `net` on the same object?

Maybe we could simplify this by just validating that only one of `net` and `nets` is provided - might also be useful as a way to get users to drop using the deprecated field (they _need_ to stop using it in order to use the new feature) so that there's less pain when that field is dropped some day in the future.

robbrockbank: Yup I agree - I think I'd suggested similar on the other PR :-)

fasaxc: @robbrockbank That was the approach I took, the validation function for the front-end does do that so a normal user shouldn't be able to hit this.  I thought it was better to have felix tolerate having both in the backend though; no point dropping a policy on the floor if it has a fairly sensible interpretation.

lib/api/rule.go
```diff
@@ -128,3 +139,28 @@ type EntityRule struct {
...
+		return []*net.IPNet{n}
+	}
+	combined := make([]*net.IPNet, len(nets)+1)
+	copy(combined, nets)
+	combined[len(combined)-1] = n
+	return combined
+}
+
+// GetNets returns either r.Nets or a slice containing r.Net.  It is useful for unifying the
+// two representations.  If both values are specified, it joins them together.
```
robbrockbank: I don't think the "If both values are specified" sentence should be there since it shouldn't be possible to have that and it sounds like we are encouraging the use of both.

lib/backend/model/rule.go
```diff
@@ -87,83 +128,93 @@ func (r Rule) String() string {
...
+		}
+		if r.DstSelector != "" {
+			toParts = append(toParts, "selector", fmt.Sprintf("%#v", r.DstSelector))
+		}
+		dstNets := r.AllDstNets()
+		if len(dstNets) != 0 {
+			toParts = append(toParts, "cidr", joinNets(dstNets))
+		}
+		if len(r.NotDstPorts) > 0 {
+			NotDstPorts := make([]string, len(r.NotDstPorts))
```
robbrockbank: NotDstPorts -> notDstPorts

lib/validator/validator.go
```diff
@@ -406,25 +407,49 @@ func validateRule(v *validator.Validate, structLevel *validator.StructLevel) {
 		}
 	}
 
-	// Check for mismatch in IP versions
-	if rule.Source.Net != nil && rule.IPVersion != nil {
-		if rule.Source.Net.Version() != *(rule.IPVersion) {
-			structLevel.ReportError(reflect.ValueOf(rule.Source.Net), "Source.Net",
-				"", reason("rule contains an IP version that does not match src CIDR version"))
-		}
+	// Check that only one of the net or nets fields is specified.
```
robbrockbank: Given our desire to ween people off the singular, I wonder if we should be a little more strict and reject if there is a mix of any of the Net and Nets fields (e.g.  Source.Net  && Dest.NotNets)

WDYT?

robbrockbank:
A couple of minor comments, and branch needs updating.

Otherwise LGTM.

lib/api/rule.go
```diff
@@ -115,9 +120,15 @@ type EntityRule struct {
 	// NotTag is the negated version of the Tag field.
 	NotTag string `json:"notTag,omitempty" validate:"omitempty,tag"`
 
-	// NotNet is the negated version of the Net field.
+	// NotNet is an optional field that restricts the rule to only apply to traffic that
+	// does not originate from (or terminate at) an IP address in the given subnet.
+	// Deprecated: superseded by NotNets.
```
caseydavenport: Spurred by the conversation in slack - is there a good place to add a warning log when users use `Nets`, so we can indicate that it's deprecated?

fasaxc:
@caseydavenport How about this: https://github.com/projectcalico/libcalico-go/pull/470/commits/0886ade6a62d269a5c831c38d7b43883e993f9e6

caseydavenport:
@fasaxc that LGTM - sorry for the last minute comment :p

PR #471, opened 2017-07-06
Support annotations on Policy objects

caseydavenport:
## Description

Enhance Policy objects so that they support annotations.  This will be useful for clients who want to track extra metadata about a given resource, or e.g. keep track of resources they specifically have created.

Once this is approved, I intend to do the same for Profiles.

## Todos
- [x] Tests
- [x] Documentation

### DIFF
--- a/lib/api/policy.go
+++ b/lib/api/policy.go
@@ -48,6 +48,9 @@ type PolicyMetadata struct {
 
 	// The name of the selector-based security policy.
 	Name string `json:"name,omitempty" validate:"omitempty,namespacedname"`
+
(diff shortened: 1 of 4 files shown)


caseydavenport:
Still want to do a manual test of this, but @robbrockbank might want to take a first look.

robbrockbank:
LGTM.   Should we just add these to all our consumer facing resources?

caseydavenport:
@robbrockbank I have a particular use-case in mind for policy/profiles, but I would like to see it added to other things down the road.

I'd really like to see us update to use a common metadata struct, but that's a longer discussion :)

Issue #65, opened 2016-08-18
Only validate the fields required for a particular request type (e.g. create, delete, get)


robbrockbank:
Currently we have a lot of validation fields marked as "omitempty" when we really should be validating them

I think we need to do a few things:
-  For create, update, apply validate the Spec (and honor required indicator)
-  For create, update, apply, delete validate the Metadata (And honor the required indicator)

otherwise
-  Validate Spec and/or Metadata but allow fields to be empty (in which case they won't be validated)

caseydavenport:
I'm not sure what actually needs to be done for this task - is this just adding more fields to `validator.go`?

robbrockbank:
At the moment we always validate the full contents of the YAML data.    This means, at the moment, you can't just feed in a list of YAML with empty specs to delete a set of objects (you really only need the metadata) -- because the spec data is validated.

When deleting/getting/listing it is not necessary to validate the Spec data since that is not actually used in the request.

We should limit exactly what is validated based on the type of request.

I'm considering this low priority though and not required for GA.

PR #472, opened 2017-07-12
Added generic validations to ensure the existence of lookup fields

mgleung:
## Description
Validations for missing fields were complete on etcd but incomplete for KDD.  Wrote generic validations for Key fields that are used for looking up resources that will run no matter what datastore is used.

Addresses #65 .  Validations seem to have already been removed on spec data where it is unnecessary so those changes are not covered in this PR.

## Testing
- Wrote/Ran Unit Tests
- Ran manual tests against KDD and observed the correct errors were thrown

### DIFF
--- a/lib/api/unversioned/types.go
+++ b/lib/api/unversioned/types.go
@@ -19,6 +19,12 @@ type Resource interface {
 	GetTypeMetadata() TypeMetadata
 }
 
+// All singular resources (all resources not including lists) implement the ResourceObject interface
+type ResourceObject interface {
+	Resource
+	GetResourceMetadata() ResourceMetadata
+}
+
 // Define available versions.
 var (
 	// `apiVersion` in the config yaml files
--- a/lib/validator/common.go
+++ b/lib/validator/common.go
@@ -1,0 +1,82 @@
+// Copyright (c) 2016-2017 Tigera, Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package validator
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/projectcalico/libcalico-go/lib/api"
+	"github.com/projectcalico/libcalico-go/lib/api/unversioned"
+	"github.com/projectcalico/libcalico-go/lib/errors"
+	"github.com/projectcalico/libcalico-go/lib/scope"
+)
+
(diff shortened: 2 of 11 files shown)


lib/api/unversioned/types.go
```diff
@@ -17,6 +17,7 @@ package unversioned
 // All resources (and resource lists) implement the Resource interface.
 type Resource interface {
 	GetTypeMetadata() TypeMetadata
+	GetResourceMetadata() ResourceMetadata
```
robbrockbank: The ResourceMetadata interface calls the equivalent method: "GetObjectMetadata()", so we should use the same name here - it's returning the same information.

lib/validator/common.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/projectcalico/libcalico-go/lib/api/unversioned"
+	"github.com/projectcalico/libcalico-go/lib/errors"
+	"github.com/projectcalico/libcalico-go/lib/scope"
+)
+
+// ValidateKeyFieldsExist is used to validate the Key fields in a resource
+// to ensure that all necessary fields are present.
+// This is split from the validator so it can be used conditionally
+// depending on the command.
+func ValidateKeyFieldsExist(resource unversioned.Resource) error {
```
robbrockbank: Not sure I get the naming differences here between the method that takes the resource and the one that takes the metadata.

I'd imagine we just want the method below and scrap this one.  It's up to the calling code to get the Metadata first to pass in.

lib/validator/common.go
```diff
@@ -0,0 +1,88 @@
...
+	return ValidateMetadataFieldsExist(resource.GetResourceMetadata())
+}
+
+// ValidateMetadataFieldsExist is used to validate the Resource Metadata to ensure
+// that all necessary fields are present.
+// This is split from the validator so it can be used conditionally
+// depending on the command.
+func ValidateMetadataFieldsExist(rm unversioned.ResourceMetadata) error {
```
robbrockbank: The name suggests this is validating all metadata fields exist - which isn't true, so name needs to be tweaked.  Perhaps ValidateMetadataIDsAssigned()?

lib/validator/common.go
```diff
@@ -0,0 +1,88 @@
...
+	switch metadata := rm.(type) {
+	case api.BGPPeerMetadata:
+		if metadata.PeerIP.IP == nil {
+			return errors.ErrorInsufficientIdentifiers{Name: "peerIP"}
+		}
+		if metadata.Scope != scope.Global && metadata.Node == "" {
```
robbrockbank: Also need to check that Scope is not "unassigned".

lib/validator/common.go
```diff
@@ -0,0 +1,88 @@
...
+			return errors.ErrorInsufficientIdentifiers{Name: "orchestrator"}
+		}
+		if metadata.Workload == "" {
+			return errors.ErrorInsufficientIdentifiers{Name: "workload"}
+		}
+		if metadata.Name == "" {
+			return errors.ErrorInsufficientIdentifiers{Name: "name"}
+		}
+	default:
+		return fmt.Errorf("Unexpected resource metadata: %s", metadata)
```
robbrockbank: I think this should be Fatal - it indicates some code that hasn't but should have been written.

PR #474, opened 2017-07-13
Moving over logutils from Felix

mgleung:
## Description
Moving logutils functions out of Felix and into libcalico-go.  This will allow other repos to use the same log formatting that Felix does and will make logs more consistent.

## Todos
- [x] Testing

## Release note
```release-note
None required
```

### DIFF
--- a/lib/logutils/logutils.go
+++ b/lib/logutils/logutils.go
@@ -1,0 +1,436 @@
+// Copyright (c) 2016-2017 Tigera, Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package logutils
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"os"
+	"path"
+	"runtime"
+	"sort"
+	"strings"
+	"sync"
+	"time"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+// FilterLevels returns all the logrus.Level values <= maxLevel.
+func FilterLevels(maxLevel log.Level) []log.Level {
+	levels := []log.Level{}
+	for _, l := range log.AllLevels {
+		if l <= maxLevel {
+			levels = append(levels, l)
+		}
+	}
+	return levels
+}
+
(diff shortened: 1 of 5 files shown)


lib/logutils/logutils.go
```diff
@@ -0,0 +1,430 @@
...
+	levels := []log.Level{}
+	for _, l := range log.AllLevels {
+		if l <= maxLevel {
+			levels = append(levels, l)
+		}
+	}
+	return levels
+}
+
+// Formatter is our custom log formatter, which mimics the style used by the Python version of
```
fasaxc: This comment is probably less helpful than when I wrote it since there's no Python version of Felix any more.  How about adding a bit of design exposition so that people following know why it's done the way it's done:
```
Formatter is a custom log formatter designed to balance ease of machine processing with human readability.  It includes:

    - A sortable millisecond timestamp, which makes it easier to scan and correlate logs.
    - The level, up front, to aid visual scanning.
    - The PID of the process, to make it easier to spot log discontinuities (If you're looking at two disjoint chunks of log, were they written by the same process?  Was there a restart in-between?)
    - The file name and line number, as essential context.
    - The message!
    - Log fields appended in sorted order.

... example ...

Note: there are tools in Calico's CI pipeline that reply on being able to parse (at lest the prefix section of) this log format in order to detect errors.  If you're tempted to change it, make sure you track down those tools and adjust them accordingly!
```

mgleung: Makes sense.  I'll make those changes.  Thanks for writing up the descriptive comment for me

lib/logutils/logutils.go
```diff
@@ -0,0 +1,436 @@
...
+	levels := []log.Level{}
+	for _, l := range log.AllLevels {
+		if l <= maxLevel {
+			levels = append(levels, l)
+		}
+	}
+	return levels
+}
+
+// Formatter is our custom log formatter designed to balance ease of machien processing
```
fasaxc: Typo: `s/machien/machine/`

lib/logutils/logutils.go
```diff
@@ -0,0 +1,436 @@
...
+// with human readability.  Logs include:
+//    - A sortable millisecond timestamp, for scanning and correlating logs
+//    - The log level, near the beginning of the line, to aid in visual scanning
+//    - The PID of the process to make it easier to spot log discontinuities (If
+//      you are looking at two disjoint chunks of log, were they written by the
+//      same process?  Was there a restart in-between?
```
fasaxc: Typo: unclosed paren.

lib/logutils/logutils_test.go
```diff
@@ -0,0 +1,413 @@
...
+	"strings"
+	"time"
+
+	log "github.com/Sirupsen/logrus"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/ginkgo/extensions/table"
+	. "github.com/onsi/gomega"
+	"github.com/prometheus/client_golang/prometheus"
+
+	"sync"
```
fasaxc: Nit: this should be in the first block of imports.

PR #481, opened 2017-07-24
Add ID fields to Resource Objects for easy lookup and validations

mgleung:
## Description
Adds in an ID field for every Resource Object and validations on top of those ID fields to make it easier to validate that important fields were not changed.

Will be of help to finishing https://github.com/projectcalico/calicoctl/pull/1564

## Todos
- [x] Tests
- [x] Documentation (none required)
- [x] Release note (none required)

## Release Note
```release-note
None required
```

### DIFF
--- a/lib/api/bgppeer.go
+++ b/lib/api/bgppeer.go
@@ -15,6 +15,8 @@
 package api
 
 import (
+	"fmt"
+
 	"github.com/projectcalico/libcalico-go/lib/api/unversioned"
 	"github.com/projectcalico/libcalico-go/lib/net"
 	"github.com/projectcalico/libcalico-go/lib/numorstring"
@@ -35,6 +37,15 @@ type BGPPeer struct {
 
 func (t BGPPeer) GetResourceMetadata() unversioned.ResourceMetadata {
 	return t.Metadata
+}
+
+// String() returns the human-readable string representation of a BGPPeer instance
+// which is defined by its PeerIP and Scope.
+func (t BGPPeer) String() string {
+	if t.Metadata.Scope == scope.Node && t.Metadata.Node == "" {
+		return fmt.Sprintf("BGPPeer(PeerIP=%s, Scope=%s)", t.Metadata.PeerIP.IP.String(), t.Metadata.Scope)
+	}
+	return fmt.Sprintf("BGPPeer(PeerIP=%s, Scope=%s, Node=%s)", t.Metadata.PeerIP.IP.String(), t.Metadata.Scope, t.Metadata.Node)
 }
 
 // BGPPeerMetadata contains the metadata for a BGPPeer resource.
--- a/lib/api/unversioned/types.go
+++ b/lib/api/unversioned/types.go
@@ -19,10 +19,16 @@ type Resource interface {
 	GetTypeMetadata() TypeMetadata
 }
 
-// All singular resources (all resources not including lists) implement the ResourceObject interface
+// All singular resources (all resources not including lists) implement the ResourceObject interface.
 type ResourceObject interface {
 	Resource
+
+	// GetResourceMetadata returns the ResourceMetadata for each Resource Object.
 	GetResourceMetadata() ResourceMetadata
+
+	// String returns a human-readable string representation of a ResourceObject which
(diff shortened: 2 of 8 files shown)


lib/api/bgppeer.go
```diff
@@ -37,6 +37,10 @@ func (t BGPPeer) GetResourceMetadata() unversioned.ResourceMetadata {
 	return t.Metadata
 }
 
+func (t BGPPeer) GetResourceID() string {
```
robbrockbank: I think you're better off having this method return a user friendly name that identifies this resource.  That way it could be returned to the user as well.  e.g. "BGP Peer (PeerIP=xxx, Scope=Global)"

lib/validator/common.go
```diff
@@ -80,3 +82,39 @@ func ValidateMetadataIDsAssigned(rm unversioned.ResourceMetadata) error {
 
 	return nil
 }
+
+// ValidateIDFieldsSame is used to validate that the Resource Objects
```
robbrockbank: I think this is only useful for calicoctl, so I'd not include this in libcalico-go.

lib/validator/common.go
```diff
@@ -80,3 +82,39 @@ func ValidateMetadataIDsAssigned(rm unversioned.ResourceMetadata) error {
...
+		}
+	}
+
+	return nil
+}
+
+// GetIDFields is a helper function that returns the field labels
+// on the metadata objects that are ID fields excluding Labels,
+// Annotations, and Tags
+func GetIDFields(rm unversioned.ResourceMetadata) []string {
```
robbrockbank: I don't like this.  I think if you really want to get a slice of IDs as strings, then the getResourceID() should probably have been getResourceIDs() returning a slice of strings. 

*Personally* I don't think that's necessary - if you have a method that returns a nice user friendly name of the resource then you can craft some text that says  Resource X should be Resource Y.

mgleung: I see.  My thought was to basically write a function that would work for all the resources without needing to create more individual methods on each that would be need to be maintained should something change.  To that end, adding a new tag type would probably be best as far as that line of thinking goes.  You're right though; all of this is probably unnecessary if I can get a user friendly name working so then I can parse those values.  `WorkloadEndpoint` might have a little bit of a long string name though....

lib/api/bgppeer.go
```diff
@@ -37,6 +39,10 @@ func (t BGPPeer) GetResourceMetadata() unversioned.ResourceMetadata {
 	return t.Metadata
 }
 
+func (t BGPPeer) String() string {
+	return fmt.Sprintf("BGPPeer(IP=%s, Scope=%s, Node=%s)", t.Metadata.PeerIP.IP.String(), t.Metadata.Scope, t.Metadata.Node)
```
robbrockbank: IP->PeerIP

lib/api/unversioned/types.go
```diff
@@ -23,6 +23,7 @@ type Resource interface {
 type ResourceObject interface {
 	Resource
 	GetResourceMetadata() ResourceMetadata
+	String() string
```
robbrockbank: Given the very specific behavior of this String() method, I think it is worth adding a doc comment.

lib/api/workloadendpoint.go
```diff
@@ -31,6 +31,10 @@ func (t WorkloadEndpoint) GetResourceMetadata() unversioned.ResourceMetadata {
 	return t.Metadata
 }
 
+func (t WorkloadEndpoint) String() string {
+	return fmt.Sprintf("WorkloadEndpoint(Node=%s, Orchestrator=%s, Workload=%s, Name=%s, ActiveInstanceID=%s)", t.Metadata.Node, t.Metadata.Orchestrator, t.Metadata.Workload, t.Metadata.Name, t.Metadata.ActiveInstanceID)
```
robbrockbank: I think we can remove the Active Instance ID, or perhaps only include it if it is set to something - it's used under very specific circumstances.

lib/api/bgppeer.go
robbrockbank: Not too fussed about this, but perhaps if scope is node and node=="" we could omit node from the name.

Issue #485, opened 2017-07-28
KDD doesn’t pass node IP information to Felix during resync


bcreane:
Specific to KDD, when calico-node is restarted, we see a data plane connectivity issue that lasts 5-12 seconds. This issue is seen in both v2.3 and v2.4.

## Expected Behavior
Restarting calico-node should not result in data plane disruption.

## Current Behavior
Running intra-node iperf session, I observe a 5-12 second break in connectivity.

## Steps to Reproduce (for bugs)

1. deploy k8s cluster with KDD backend
2. spin up iperf on two different nodes: `kubectl run iperf --image=phlak/iperf --replicas=2`
3. run iperf: `kubectl exec -it iperf-X -n default sh`, then: `# iperf -c 192.168.X.X -t 3600 -i 1`
4. kill one of the calico-nodes where iperf is running: `kubectl delete pod -n kube-system calico-node-XX`

See https://github.com/projectcalico/calicoctl/issues/1419 for related debugging info.

## Your Environment

* libcalico-go version:
* Operating System and version:
* Link to your project (optional):

caseydavenport:
Tracked this down to the KDD syncer not sending through Node IPs to Felix during its initial snapshot - [see here](https://github.com/projectcalico/libcalico-go/blob/master/lib/backend/k8s/syncer.go#L781).

Felix thus removes the node IPs from the "all-ips" ipset and re-populates them as it receives subsequent watch updates.

bcreane:
A DROP rule was getting hit during calico-node restart, observed via (on each node):
    `sudo watch -d -c -n 0.1 "iptables-save -c -t filter | grep DROP"`

Subsequently watched the associated ipset change for a brief moment (on each node):
    `sudo watch -d -n 0.1 "ipset list cali4-all-hosts"`

PR #486, opened 2017-08-04
Treat initial and subsequent resync events identically for node ip

bcreane:
## Description
Refactor parsing the KDD node syncer event so that common code between a new sync and an update sync event are treated the same. Prior to this fix the new sync event path sent incorrectly formatted node event to clients (e.g. felix). 

See https://github.com/projectcalico/libcalico-go/issues/485 for details.

Note: a unit test is in the works.

## Todos
- [x] Tests - manually tested the scenario outlined in the bug, and also confirmed that a hand run of a unit test also sees the correct syncer behavior.
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/backend/k8s/syncer.go
+++ b/lib/backend/k8s/syncer.go
@@ -794,14 +794,15 @@ func (syn *kubeSyncer) performSnapshot(versions *resourceVersions) (map[string][
 
 			versions.nodeVersion = noList.ListMeta.ResourceVersion
 			for _, no := range noList.Items {
-				node, err := resources.K8sNodeToCalico(&no)
-				if err != nil {
-					log.WithError(err).Fatal("Error converting node")
-				}
-				if node != nil {
-					snap[KEY_NO] = append(snap[KEY_NO], *node)
-					keys[KEY_NO][node.Key.String()] = true
-				}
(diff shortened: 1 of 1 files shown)


lib/backend/k8s/syncer.go
```diff
@@ -884,12 +885,7 @@ func (syn *kubeSyncer) parseNamespaceEvent(e watch.Event) []model.KVPair {
 	return []model.KVPair{*rules, *tags, *labels}
 }
 
-func (syn *kubeSyncer) parseNodeEvent(e watch.Event) (*model.KVPair, *model.KVPair) {
-	node, ok := e.Object.(*k8sapi.Node)
-	if !ok {
-		log.Panicf("Invalid node event. Type: %s, Object: %+v", e.Type, e.Object)
-	}
-
+func splitNodeEvent(node *k8sapi.Node) (*model.KVPair, *model.KVPair) {
```
caseydavenport: minor naming nit: I'd just call this `splitNode` since it's not actually taking an `Event` struct as an input, rather a Node.

caseydavenport:
OK LGTM - test's in a separate PR as discussed.

PR #489, opened 2017-08-05
Now specify UpdateType for ExpectExists helper function & add new test for Node sync

bcreane:
## Description
This request adds two commits:
 1. Modify ExpectExists() to accept an Update which consists of both a KVPair as well as an UpdateType. ExpectExists now requires that the update type match as well as that the KVPair is present. All existing unit tests pass consistently (on my laptop).
2. Add another expected HostConfigKey that should be updated at initial sync time. This is a test to cover https://github.com/projectcalico/libcalico-go/issues/485.

Note this additional test verifies the fix associated with above issue - the test fails when the fix is removed.

## Todos
- [x] Tests - all unit tests are passing

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/backend/k8s/k8s_fv_test.go
+++ b/lib/backend/k8s/k8s_fv_test.go
@@ -146,21 +146,21 @@ func (c cb) ProcessUpdates() {
 	}
 }
 
-func (c cb) ExpectExists(kvps []model.KVPair) {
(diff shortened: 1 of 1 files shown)


caseydavenport:
Ok, this was merged as part of https://github.com/projectcalico/libcalico-go/pull/490

PR #491, opened 2017-08-08
Migrate from TPR to CRD for the Kubernetes backend

gunjan5:
## Description
 Move to CustomResourceDefination from ThirdPartyResource for KDD backend

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Migrate from TPR to CRD for the Kubernetes backend
```


CustomResourceDefinition.yaml
```diff
@@ -0,0 +1,69 @@
+kind: List
```
caseydavenport: This is just for the tests, right?

I think this should be in a subdirectory so it's out of the way.

e.g. `test/crds.yaml`

(and a test/README.md)

lib/backend/k8s/custom/globalbgpconfig.go
```diff
@@ -21,23 +21,24 @@ import (
...
 	metav1.TypeMeta `json:",inline"`
-	Metadata        metav1.ObjectMeta `json:"metadata"`
+	Metadata        metav1.ObjectMeta   `json:"metadata"`
+	Spec            GlobalBgpConfigSpec `json:"spec"`
+}
 
-	Spec GlobalBgpConfigSpec `json:"spec"`
+type GlobalBgpConfigSpec struct {
+	// The reason we have Name field in Spec is because k8s metadata
+	// name field makes the string lowercase, so Name field in Spec is
```
caseydavenport: rather than "make the string lowercase", "requires the string to be lowercase" is more accurate :)

lib/backend/k8s/resources/customresource.go
```diff
@@ -301,26 +295,9 @@ func (c *customK8sResourceClient) List(list model.ListInterface) ([]*model.KVPai
 	return kvps, reslOut.GetListMeta().GetResourceVersion(), nil
 }
 
-// EnsureInitialized performs client initialization required for a specific
-// Custom K8s Resource type.
+// EnsureInitialized is no-op since CRD should be initialized
+// by applying the manifest.
```
caseydavenport: what manifest?  Someone reading the code might not know.

I'd just say "initialized in advance"

CustomResourceDefinition.yaml
gunjan5: yup, just for the tests. Will move it to tests/

lib/backend/k8s/resources/customresource.go
caseydavenport: also, nits: 
- "is _a_ no-op" not "is no-op"
- "the CRD" not "CRD"

lib/backend/k8s/resources/customresource.go
```diff
@@ -301,26 +295,9 @@ func (c *customK8sResourceClient) List(list model.ListInterface) ([]*model.KVPai
...
-	}
-	_, err := c.clientSet.Extensions().ThirdPartyResources().Create(&res)
-	if err != nil {
-		// Don't care if it already exists.
-		if !kerrors.IsAlreadyExists(err) {
-			logContext.WithError(err).Info("Error initializing ThirdParyResource")
-			return K8sErrorToCalico(err, res)
-		}
-	}
+	// no-op, should be created by applying the manifest.
```
caseydavenport: You said this two lines earlier, so can delete this one.

lib/backend/k8s/resources/ippool.go
```diff
@@ -69,41 +69,56 @@ func (_ IPPoolConverter) NameToKey(name string) (model.Key, error) {
...
+		} else {
+			ipipInterface = ""
+		}
 	}
+
 	return &model.KVPair{
-		Key:      model.IPPoolKey{CIDR: v.CIDR},
-		Value:    &v,
+		Key: model.IPPoolKey{CIDR: *cidr},
+		Value: &model.IPPool{
```
caseydavenport: Why don't we use the conversion methods that already exist for this?

See how it's done for example in BGP peers: https://github.com/projectcalico/libcalico-go/blob/master/lib/backend/k8s/resources/globalbgppeer.go#L99

gunjan5: we don't have any common converter methods for IPPool, I can take the one in `client/` and move it to `converter/` and use it here

lib/backend/k8s/custom/systemnetworkpolicy.go
```diff
@@ -22,7 +22,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
-// SystemNetworkPolicy is the ThirdPartyResource definition of a Calico Policy resource in
+// SystemNetworkPolicy is the CustomResourceDefinition of a Calico Policy resource in
 // the Kubernetes API.
 type SystemNetworkPolicy struct {
```
caseydavenport: I believe we want to change this to `GlobalNetworkPolicy` instead.

lib/backend/k8s/k8s_fv_test.go
```diff
@@ -1196,7 +1189,7 @@ var _ = Describe("Test Syncer API for Kubernetes backend", func() {
 					CIDR:          *cidr,
 					IPIPInterface: "tunl0",
 					Masquerade:    true,
-					IPAM:          true,
```
caseydavenport: Why this change?

lib/backend/k8s/resources/customresource.go
```diff
@@ -117,32 +115,60 @@ func (c *customK8sResourceClient) Update(kvp *model.KVPair) (*model.KVPair, erro
...
+		// Send the update request using the name.
+		name := resIn.GetObjectMeta().GetName()
+		logContext = logContext.WithField("Name", name)
+		logContext.Debug("Update resource by name")
+		updateError = c.restClient.Put().
+			Resource(c.resource).
+			Body(resIn).
+			Name(name).
+			Do().Into(resOut)
+		if updateError == nil {
```
caseydavenport: This logic is a bit different from what was here before.

If `providedRV != ""`, then we don't want to retry on failure - we should just return right away.

We should only retry when:
- No Revision was provided on the KVP, AND
- The error we got back was due to a CAS error.

caseydavenport: Haven't looked elsewhere yet, but make sure we add a UT for this if there isn't one already.

lib/backend/k8s/resources/customresource.go
```diff
@@ -155,19 +181,23 @@ func (c *customK8sResourceClient) Apply(kvp *model.KVPair) (*model.KVPair, error
...
 	logContext.Debug("Apply custom Kubernetes resource")
 
-	// Attempt and Update and roll back to a Create if the resource does
-	// not exist.  We only log debug here since the Update and Create will
-	// also log.
-	updated, err := c.Update(kvp)
+	// Attempt to Create and do an Update if the resource already exists.
+	// We only log debug here since the Create and Update will also log.
+	// Can't set Revision while creating a resource.
+	updated, err := c.Create(&model.KVPair{
```
caseydavenport: why do we need this change?

Does "Create" fail if a revision was provided?  If so, what code is providing a revision?

Perhaps we want to skip "Create" if the revision is set?

lib/backend/k8s/resources/globalbgpconfig.go
```diff
@@ -28,19 +28,19 @@ import (
...
 	return &customK8sResourceClient{
 		clientSet:       c,
 		restClient:      r,
-		name:            GlobalBgpConfigTPRName,
+		name:            GlobalBgpConfigCRDName,
 		resource:        GlobalBgpConfigResourceName,
 		description:     "Calico Global BGP Configuration",
-		k8sResourceType: reflect.TypeOf(thirdparty.GlobalBgpConfig{}),
-		k8sListType:     reflect.TypeOf(thirdparty.GlobalBgpConfigList{}),
+		k8sResourceType: reflect.TypeOf(custom.GlobalBgpConfig{}),
```
caseydavenport: Should this be `GlobalBGPConfig` now?

lib/backend/k8s/resources/globalbgpconfig.go
```diff
@@ -80,17 +80,19 @@ func (c GlobalBgpConfigConverter) FromKVPair(kvp *model.KVPair) (CustomK8sResour
 	if err != nil {
 		return nil, err
 	}
-	tpr := thirdparty.GlobalBgpConfig{
-		Metadata: metav1.ObjectMeta{
-			Name: name,
+	crd := custom.GlobalBgpConfig{
+		TypeMeta: metav1.TypeMeta{
```
caseydavenport: I don't fully understand why we needed to add TypeMeta for GlobalBgpConfig but not for other resources.  This needs more investigation.

lib/client/ippool.go
```diff
@@ -132,70 +133,21 @@ func (h *ipPools) convertMetadataToListInterface(m unversioned.ResourceMetadata)
 // convertMetadataToKey converts an IPPoolMetadata to an IPPoolKey
 // This is part of the conversionHelper interface.
 func (h *ipPools) convertMetadataToKey(m unversioned.ResourceMetadata) (model.Key, error) {
```
caseydavenport: `(h *ipPools)` - any idea why this is called `h`?

:p

gunjan5: i believe historically it was called a "helper" (not a fan of that name), and they's why the `h` has been carried with it

lib/backend/k8s/k8s_fv_test.go
gunjan5: because you can't have both `IPAM` and `Disabled` `true` at the same time. They're `!` of each other in the conversion code.

lib/backend/k8s/resources/customresource.go
gunjan5: Create fails if revision is provided. I haven't spotted a code that provides revision for create (but nothing for update either). 
 > Perhaps we want to skip "Create" if the revision is set?

hmm, that sounds good in theory, but I wonder if there's any client that's setting something as the revision even for create and doing this will fail (but it's the right thing)

lib/backend/k8s/k8s_fv_test.go
```diff
@@ -173,7 +173,7 @@ func (c cb) ExpectExists(updates []api.Update) {
 		})
 
 		// Expect the key to have existed.
-		Expect(matches).To(Equal(true), fmt.Sprintf("Expected update not found: %s", update.Key))
+		// Expect(matches).To(Equal(true), fmt.Sprintf("Expected update not found: %s", update.Key))
```
caseydavenport: Make sure you undo this change.

gunjan5: yup, that's why the commit message is "remove me" 😄

lib/backend/k8s/resources/globalnetworkpolicies.go
```diff
@@ -30,61 +30,61 @@ import (
 )
 
 const (
-	SystemNetworkPolicyResourceName = "systemnetworkpolicies"
-	SystemNetworkPolicyTPRName      = "system-network-policy.alpha.projectcalico.org"
-	SystemNetworkPolicyNamePrefix   = "snp.projectcalico.org/"
+	GlobalNetworkPolicyResourceName = "GlobalNetworkPolicies"
+	GlobalNetworkPolicyCRDName      = "globalnetworkpolicies.crd.projectcalico.org"
+	GlobalNetworkPolicyNamePrefix   = "crd.projectcalico.org/"
```
caseydavenport: I think what we want to do is remove the `GlobalNetworkPolicyNamePrefix` altogether.

Here's what I'm thinking so far, and have discussed a bit with @robbrockbank 

A NetworkPolicy with name "foo" should show up like this in `calicoctl get policies`:
- `knp.default.foo`

A GlobalNetworkPolicy with name "foo" should show up like this in `calicoctl get policies`:
- `foo`

Makefile
```diff
@@ -69,14 +69,33 @@ run-kubernetes-master: stop-kubernetes-master
...
 		/hyperkube controller-manager \
                         --master=127.0.0.1:8080 \
                         --min-resync-period=3m \
                         --allocate-node-cidrs=true \
                         --cluster-cidr=10.10.0.0/16 \
                         --v=5
 
+	# Create CustomResourceDefinition (CRD) for Calico resources
+	# from the manifest crds.yaml 
+	docker run \
```
caseydavenport: We should simplify this by mounting in the manifests directory into the `st-apiserver` container and just calling out to `kubectl` from within.

This is nice because it means fewer images to pull when running the tests.

lib/backend/k8s/custom/globalbgppeer.go
```diff
@@ -12,76 +12,85 @@
...
 	metav1.TypeMeta `json:",inline"`
 	Metadata        metav1.ObjectMeta `json:"metadata"`
-	Spec            api.BGPPeerSpec   `json:"spec"`
+	Spec            BGPPeerSpec       `json:"spec"`
 }
 
-// GlobalBgpPeerList is a list of Calico Global BGP Peer resources.
-type GlobalBgpPeerList struct {
+type BGPPeerSpec struct {
+	Spec   api.BGPPeerSpec
```
caseydavenport: I think you need this to be an anonymous, inline struct.  You don't want a `Spec.Spec`

lib/backend/k8s/custom/ippool.go
```diff
@@ -12,83 +12,81 @@
...
+// IPPool is the CustomResourceDefinition definition of an IPPool in the Kubernetes API.
+type IPPool struct {
 	metav1.TypeMeta `json:",inline"`
 	Metadata        metav1.ObjectMeta `json:"metadata"`
+	Spec            IPPoolSpec        `json:"spec"`
+}
 
-	Spec IpPoolSpec `json:"spec"`
+type IPPoolSpec struct {
+	Spec api.IPPoolSpec
```
caseydavenport: Same here - this should be anonymous.

lib/backend/k8s/k8s.go
```diff
@@ -325,8 +251,8 @@ func (c *KubeClient) Syncer(callbacks api.SyncerCallbacks) api.Syncer {
 func (c *KubeClient) Create(d *model.KVPair) (*model.KVPair, error) {
 	log.Debugf("Performing 'Create' for %+v", d)
 	switch d.Key.(type) {
-	case model.GlobalConfigKey:
-		return c.globalConfigClient.Create(d)
+	case model.GlobalFelixConfigKey:
```
caseydavenport: oops, I don't think we want to rename the `model.GlobalConfig` stuff.  That's a breaking change that requires corresponding Felix changes.

lib/backend/k8s/k8s.go
```diff
@@ -704,7 +630,7 @@ func (c *KubeClient) getPolicy(k model.PolicyKey) (*model.KVPair, error) {
 	}
 
 	// Check to see if this is backed by a NetworkPolicy or a Namespace.
```
caseydavenport: I know this isn't yours, but this comment is wrong now.

lib/backend/k8s/conversion.go
```diff
@@ -68,13 +68,13 @@ func (c converter) parsePolicyNameNamespace(name string) (string, error) {
...
-	// Policies backed by NetworkPolicies have form "np.projectcalico.org/<ns_name>.<np_name>"
-	if !strings.HasPrefix(name, "np.projectcalico.org/") {
+	// Policies backed by NetworkPolicies have form "knp.default.<ns_name>.<np_name>"
+	if !strings.HasPrefix(name, "knp.default.") {
 		// This is not backed by a Kubernetes NetworkPolicy.
 		return "", "", fmt.Errorf("Policy %s not backed by a NetworkPolicy", name)
 	}
 
-	splits := strings.SplitN(strings.TrimPrefix(name, "np.projectcalico.org/"), ".", 2)
+	splits := strings.SplitN(strings.TrimPrefix(name, "knp.default."), ".", 2)
```
caseydavenport: Hm, this logic is actually already wrong.  If there's a namespace or a NetworkPolicy with a `.` in it, this will break.

Some examples:
- namespace="casey.davenport" -> `casey.davenport.mypolicy`
- policy="my.policy" -> `default.my.policy`

We should add UTs for these and then fix up the code, but let's do it in a separate PR immediately after this one since it's not a regression caused by this PR.

lib/backend/k8s/k8s_fv_test.go
```diff
@@ -516,140 +491,143 @@ var _ = Describe("Test Syncer API for Kubernetes backend", func() {
...
-		// based on the prefix).
-		kvp1Name := "snp.projectcalico.org/my-test-snp"
+	It("should handle a CRUD of Global Network Policy", func() {
+
+		kvp1Name := "my-test-gnp"
 		kvp1a := &model.KVPair{
 			Key:   model.PolicyKey{Name: kvp1Name},
 			Value: &calicoAllowPolicyModel,
 		}
-		kvp1b := &model.KVPair{
```
caseydavenport: Why did we need to move these around?

lib/backend/k8s/resources/customresource.go
```diff
@@ -117,32 +115,68 @@ func (c *customK8sResourceClient) Update(kvp *model.KVPair) (*model.KVPair, erro
...
+		} else {
+			// We don't retry for any other errors or if the revision number is provided.
+			break
+		}
 	}
 
-	// Update the revision information from the response.
-	kvp.Revision = resOut.GetObjectMeta().GetResourceVersion()
-	return kvp, nil
+	// Failed to update the resource after 5 retries.
```
caseydavenport: This comment isn't correct - it's either after 5 retries, or if an RV was passed.

lib/backend/k8s/resources/globalbgpconfig.go
```diff
@@ -75,22 +75,24 @@ func (c GlobalBgpConfigConverter) ToKVPair(r CustomK8sResource) (*model.KVPair,
...
+func (c GlobalBGPConfigConverter) FromKVPair(kvp *model.KVPair) (CustomK8sResource, error) {
 	name, err := c.KeyToName(kvp.Key)
 	if err != nil {
 		return nil, err
 	}
-	tpr := thirdparty.GlobalBgpConfig{
-		Metadata: metav1.ObjectMeta{
-			Name: name,
+	crd := custom.GlobalBGPConfig{
+		TypeMeta: metav1.TypeMeta{
```
caseydavenport: I don't think we ever answered why this is needed, @gunjan5.  Any thoughts?

lib/backend/k8s/resources/globalbgppeer_test.go
```diff
@@ -123,9 +118,4 @@ var _ = Describe("Global BGP conversion methods", func() {
 		Expect(kvp.Value).To(BeAssignableToTypeOf(&model.BGPPeer{}))
 		Expect(kvp.Value).To(Equal(kvp1.Value))
 	})
-
-	It("should fail to convert an invalid Kuberenetes resource", func() {
```
caseydavenport: Why did you delete this test case?

lib/backend/k8s/resources/globalnetworkpolicies_test.go
```diff
@@ -155,18 +152,13 @@ var _ = Describe("System Network Policies conversion methods", func() {
 		Expect(k).To(Equal(key2))
 	})
 
-	It("should fail to convert an invalid Key to a resource name", func() {
```
caseydavenport: And this one?

lib/backend/k8s/k8s.go
gunjan5: oh yes, find/replace mistake, ill fix it

lib/backend/k8s/custom/globalbgppeer.go
gunjan5: but then in the conversion code, we will have to assign individual fields from api Spec when we do ToKVPair, which will need an update every time we change the spec. https://github.com/projectcalico/libcalico-go/pull/491/files#diff-dd846d0ae3114b5d3ffe01b897c4e3c4R94

lib/backend/k8s/k8s_fv_test.go
gunjan5: because we use that one for an Update operation which needs a revision number which we only get _after_ the previous Create operation, so I moved it after the Create to insert the Revision, and also because we should keep the variables close to where they're used

lib/backend/k8s/resources/globalbgpconfig.go
gunjan5: everything looks identical, I will play around with it a bit to see

lib/backend/k8s/resources/globalbgppeer_test.go
gunjan5: because it was essentially testing `ResourceNameToIP(t.Metadata.Name)`  in the conversion code https://github.com/projectcalico/libcalico-go/pull/491/files#diff-dd846d0ae3114b5d3ffe01b897c4e3c4L86 which is not needed now since we include the peerIP in the Spec

lib/backend/k8s/resources/globalnetworkpolicies_test.go
gunjan5: same as above, we added CIDR to the Spec so we don't need the conversion this test was testing

lib/backend/k8s/resources/ippool_test.go
```diff
@@ -134,14 +123,4 @@ var _ = Describe("IP Pool conversion methods", func() {
 		Expect(kvp.Value).To(BeAssignableToTypeOf(&model.IPPool{}))
 		Expect(kvp.Value).To(Equal(kvp1.Value))
 	})
-
```
gunjan5: same as above, we added CIDR to the Spec so we don't need the conversion this test was testing

lib/backend/k8s/resources/names.go
```diff
@@ -28,8 +28,29 @@ import (
 
 // IPToResourceName converts an IP address to a name used for a k8s resource.
 func IPToResourceName(ip net.IP) string {
```
caseydavenport: I think there are some corresponding changes / cleanup we can do here: https://github.com/projectcalico/libcalico-go/pull/491/files#diff-711892c034cb78da74e3f90f690c86b0R113

lib/backend/k8s/resources/ippool_test.go
caseydavenport: I'm not sure that's true.  We added it to the spec, but we're not changing the logic around the name at all.

lib/backend/k8s/k8s_fv_test.go
caseydavenport: > or an Update operation which needs a revision number 

I don't think that's true, now that Update does a Get first.  Unless there's another reason I think we should keep these tests as they were, or at least remove the revision # so that we're not fundamentally changing the tests.

We can add another test that sets the revision on Update.

lib/backend/k8s/custom/globalbgppeer.go
```diff
@@ -12,76 +12,85 @@
...
 	metav1.TypeMeta `json:",inline"`
 	Metadata        metav1.ObjectMeta `json:"metadata"`
-	Spec            api.BGPPeerSpec   `json:"spec"`
+	Spec            BGPPeerSpec       `json:"spec"`
 }
 
-// GlobalBgpPeerList is a list of Calico Global BGP Peer resources.
-type GlobalBgpPeerList struct {
+type BGPPeerSpec struct {
+	BGPPeerSpec api.BGPPeerSpec
```
caseydavenport: I think this is still not right.

It should be:

```go
type BGPPeerSpec struct {
    api.BGPPeerSpec
    Scope  scope.Scope
    Node   string
    PeerIP net.IP
}
```

robbrockbank: The Scope field is not required in the Spec - it was required in the Metadata to allow us to do queries for "get me the node peers".

Makefile
caseydavenport: @gunjan5 you missed this one.  Can you open a second PR to do this ASAP so we don't block this PR?

gunjan5: yeah will do

lib/backend/k8s/k8s_fv_test.go
```diff
@@ -546,110 +518,112 @@ var _ = Describe("Test Syncer API for Kubernetes backend", func() {
...
-		By("Checking cache does not have System Network Policy entries", func() {
+		By("Checking cache does not have Global Network Policy entries", func() {
 			Eventually(cb.GetSyncerValuePresentFunc(kvp1a.Key)).Should(BeFalse())
 			Eventually(cb.GetSyncerValuePresentFunc(kvp2a.Key)).Should(BeFalse())
 		})
 
-		By("Creating a System Network Policy", func() {
-			_, err := c.snpClient.Create(kvp1a)
+		By("Creating a Global Network Policy", func() {
+			var err error
```
caseydavenport: Don't think we need this.

lib/backend/k8s/k8s_fv_test.go
```diff
@@ -546,110 +518,112 @@ var _ = Describe("Test Syncer API for Kubernetes backend", func() {
...
-		By("Checking cache has correct System Network Policy entries", func() {
+		By("Checking cache has correct Global Network Policy entries", func() {
 			Eventually(cb.GetSyncerValueFunc(kvp1a.Key)).Should(Equal(kvp1b.Value))
 			Eventually(cb.GetSyncerValuePresentFunc(kvp2a.Key)).Should(BeFalse())
 		})
 
-		By("Applying a non-existent System Network Policy", func() {
-			_, err := c.snpClient.Apply(kvp2a)
+		By("Applying a non-existent Global Network Policy", func() {
+			var err error
```
caseydavenport: Don't need this `var err error`

lib/backend/k8s/k8s_fv_test.go
```diff
@@ -700,7 +677,8 @@ var _ = Describe("Test Syncer API for Kubernetes backend", func() {
 		}()
 
 		By("Creating a Global BGP Peer", func() {
-			_, err := c.Create(kvp1a)
+			var err error
```
caseydavenport: Don't need this

lib/backend/k8s/k8s_fv_test.go
```diff
@@ -747,10 +726,17 @@ var _ = Describe("Test Syncer API for Kubernetes backend", func() {
 			kvps, err := c.List(model.GlobalBGPPeerListOptions{})
 			Expect(err).ToNot(HaveOccurred())
 			Expect(kvps).To(HaveLen(2))
-			Expect(kvps[0].Key).To(Equal(kvp1b.Key))
-			Expect(kvps[0].Value).To(Equal(kvp1b.Value))
-			Expect(kvps[1].Key).To(Equal(kvp2b.Key))
-			Expect(kvps[01].Value).To(Equal(kvp2b.Value))
+			keys := []model.Key{}
+			vals := []interface{}{}
+			for _, k := range kvps {
```
caseydavenport: Would be better if we used a map so we can assert that a specific value exists for a given key.

lib/backend/k8s/resources/globalbgpconfig.go
caseydavenport: @gunjan5 ping

lib/backend/k8s/conversion.go
robbrockbank: can't we just use `/` as the separator for now since that is not a valid char for the namespace.

caseydavenport: Don't think that fixes the problem I'm describing above at least.

PR #493, opened 2017-08-09
Stop running an entire master for KDD FV tests

caseydavenport:
## Description

Kubernetes hyperkube 1.7.0 no longer ships with the scripts necessary to run the full dockerized single-node cluster like we were doing before, so this lets us update our FVs to run against a k8s 1.7+ api.

We didn't actually need to run a full controller-manager, kubelet, etc. anyway so this is a nice simplification, and will mean that these tests can run on a mac once more :)

This also spotted what I think is a bug in the KDD where it doesn't handle an empty Annotations on the Node properly. 

## Todos
- [x] Tests
- [x] Documentation
- [x] Release note

## Release Note

```release-note
None required
```


lib/backend/k8s/resources/node_conversion.go
```diff
@@ -70,6 +70,11 @@ func mergeCalicoK8sNode(calicoNode *model.Node, k8sNode *kapiv1.Node) (*kapiv1.N
 	if calicoNode.BGPIPv4Net != nil {
 		subnet, _ := calicoNode.BGPIPv4Net.Mask.Size()
 		ipCidr := fmt.Sprintf("%s/%d", calicoNode.BGPIPv4Addr.String(), subnet)
+
+		// If the Annotations map is nil, initialize it.
```
caseydavenport: This turned out to be causing a panic in the new tests.

I had thought this would always be initialized, but seems like perhaps not.

lib/backend/k8s/k8s_fv_test.go
```diff
@@ -277,9 +277,38 @@ var _ = Describe("Test Syncer API for Kubernetes backend", func() {
...
+				},
+			},
+		}
+
+		// Delete the node to ensure a clean start.
+		c.clientSet.Nodes().Delete(n.Name, &metav1.DeleteOptions{})
+
+		// Re-create the node.
+		_, err := c.clientSet.Nodes().Create(&n)
+		if err != nil {
```
gunjan5: not a big deal, but why not assert with Gomega like `Expect(err).NotTo(HaveOccurred(), "Failed to create node in k8s API for test")`?

PR #494, opened 2017-08-10
Add Types field to Policy

neiljerram:
## Description
This is the libcalico-go part of an enhancement to allow a Calico Policy to specify explicitly whether it applies on ingress or on egress, or both.
 
## Todos
- [x] Tests
- [x] Documentation - in https://github.com/projectcalico/calico/pull/1016
- [x] Release note - in https://github.com/projectcalico/calico/pull/1016

### DIFF
--- a/lib/api/policy.go
+++ b/lib/api/policy.go
@@ -118,8 +118,20 @@ type PolicySpec struct {
 	// PreDNAT indicates to apply the rules in this policy before any DNAT.
 	PreDNAT bool `json:"preDNAT,omitempty"`
 
-	// Types indicates whether this policy applies to ingress, or to egress, or to both.  An
-	// empty or nil value here indicates both ingress and egress.
+	// Types indicates whether this policy applies to ingress, or to egress, or to both.  When
+	// not explicitly specified (and so the value on creation is empty or nil), Calico defaults
+	// Types according to what IngressRules and EgressRules are present in the policy.  The
+	// default is:
+	//
+	// - [ PolicyTypeIngress ], if there are no EgressRules (including the case where there are
+	//   also no IngressRules)
+	//
+	// - [ PolicyTypeEgress ], if there are EgressRules but no IngressRules
+	//
+	// - [ PolicyTypeIngress, PolicyTypeEgress ], if there are both IngressRules and EgressRules.
+	//
+	// When the policy is read back again, Types will always be one of these values, never empty
+	// or nil.
 	Types []PolicyType `json:"types,omitempty" validate:"omitempty,dive,policytype"`
 }
 
--- a/lib/api/policy.go
+++ b/lib/api/policy.go
@@ -117,6 +117,10 @@ type PolicySpec struct {
 
 	// PreDNAT indicates to apply the rules in this policy before any DNAT.
 	PreDNAT bool `json:"preDNAT,omitempty"`
+
+	// Types indicates whether this policy applies to ingress, or to egress, or to both.  An
+	// empty or nil value here indicates both ingress and egress.
+	Types []string `json:"types,omitempty" validate:"omitempty,dive,policytype"`
 }
 
 // NewPolicy creates a new (zeroed) Policy struct with the TypeMetadata initialised to the current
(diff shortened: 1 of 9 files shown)


lib/api/policy.go
```diff
@@ -117,6 +117,10 @@ type PolicySpec struct {
 
 	// PreDNAT indicates to apply the rules in this policy before any DNAT.
 	PreDNAT bool `json:"preDNAT,omitempty"`
+
+	// Types indicates whether this policy applies to ingress, or to egress, or to both.  An
+	// empty or nil value here indicates both ingress and egress.
+	Types []string `json:"types,omitempty" validate:"omitempty,dive,policytype"`
```
caseydavenport: Might be nice if instead of a string we used a new type for this, so calling code could make use of the constants.  I think this is fairly standard practice for fields like this.

e.g.

```go
type PolicyType string

const (
    PolicyTypeIngress PolicyType = "ingress"
    PolicyTypeEgress PolicyType = "egress"
)
```

lib/validator/validator.go
```diff
@@ -519,4 +532,13 @@ func validatePolicySpec(v *validator.Validate, structLevel *validator.StructLeve
 		structLevel.ReportError(reflect.ValueOf(m.EgressRules),
 			"PolicySpec.EgressRules", "", reason("PreDNAT PolicySpec cannot have any EgressRules"))
 	}
+
```
caseydavenport: We probably also want the ability to validate the entire list of policy types beyond just the individual value validation done above.

i.e. I think this should be an error:

```
types:
- egress
- egress
```

neiljerram:
@bcreane @caseydavenport I believe I'm fully done now on this libcalico-go part of the Types work, including all the testing that I had in mind for this repo.  @caseydavenport Please could you re-review?  @bcreane Please could you check that this is everything you're expecting me to hand over to you, as far as libcalico-go is concerned?

bcreane:
Hi @neiljerram  - yes, this looks like everything we've discussed for libcalico-go, so thumbs up. @caseydavenport is out of the office but may find time to review.

lib/converter/policy.go
```diff
@@ -73,6 +99,15 @@ func (p PolicyConverter) ConvertKVPairToAPI(d *model.KVPair) (unversioned.Resour
 	ap.Spec.Selector = bp.Selector
 	ap.Spec.DoNotTrack = bp.DoNotTrack
 	ap.Spec.PreDNAT = bp.PreDNAT
+	ap.Spec.Types = nil
+
+	// Note: avoid accidentally converting from nil to [].
+	if bp.Types != nil {
```
caseydavenport: Does this mean there's a case where `ap.Spec.Types` will be `nil` on read?

I think we probably want to be populating the `Types` field with the correct values if they're not there when someone does a "get", since we're enforcing the presence of this field on "create" (via defaults and validation).

I was imagining this would be done in the etcd/kdd backends so that the defaulting would also apply to the Syncer.

lib/api/policy.go
```diff
@@ -117,8 +117,20 @@ type PolicySpec struct {
 
 	// PreDNAT indicates to apply the rules in this policy before any DNAT.
 	PreDNAT bool `json:"preDNAT,omitempty"`
+
+	// Types indicates whether this policy applies to ingress, or to egress, or to both.  An
+	// empty or nil value here indicates both ingress and egress.
```
neiljerram: Yes, sorry, a hangover from what I wrote first - will fix.

run-uts
```diff
@@ -9,22 +9,39 @@ else
     SKIP="vendor,$SKIP"
 fi
 
+echo "Removing old coverprofiles..."
+find . -name "*.coverprofile" -type f -delete
+
+echo "Calculating packages to cover..."
+go_dirs=$(find -type f -name '*.go' | \
+	      grep -vE '/vendor/|\./proto/|.glide|/k8sfv/' | \
```
fasaxc: I think this (welcome) fix-up belongs in its own PR.

Aren't these exclusions felix-specific?

neiljerram: Yes, indeed - sorry for conflating these.  I'll split these to another PR.

Yes, I guess that .proto and k8sfv are Felix-specific, so will remove those.

PR #497, opened 2017-08-14
Tests and extra parent paths for deletion

### LINKED ISSUE #370: Deleting a node resource does not delete all node specific directories
The calico/bgp/v1/host/<hostname> directory persists after deleting the node <hostname> using calicoctl.

There doesn't appear to be any side effect other than leaking directories (not a common operation though).

heschlie:
## Description
Fixes #370 
I've added some tests for Node and WEPs to verify they have cleaned up after themselves.

The extra parent dirs will error (though it doesn't get propegated up the chain) if they still have something in them.  This means that for the Node struct the final key to be removed will also remove its parent
directory.

## Todos
- [x] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
Calico will ensure empty dirs under its control for Node and Workload Endpoints are removed once they are not needed.
```

### DIFF
--- a/lib/backend/model/bgpconfig.go
+++ b/lib/backend/model/bgpconfig.go
@@ -112,7 +112,7 @@ func (key NodeBGPConfigKey) defaultDeletePath() (string, error) {
 }
 
 func (key NodeBGPConfigKey) defaultDeleteParentPaths() ([]string, error) {
-	return nil, nil
+	return []string{fmt.Sprintf("/calico/bgp/v1/host/%s", key.Nodename)}, nil
 }
 
 func (key NodeBGPConfigKey) valueType() reflect.Type {
--- a/lib/backend/model/workloadendpoint.go
+++ b/lib/backend/model/workloadendpoint.go
@@ -71,7 +71,9 @@ func (key WorkloadEndpointKey) defaultDeleteParentPaths() ([]string, error) {
 	workload := fmt.Sprintf("/calico/v1/host/%s/workload/%s/%s",
 		key.Hostname, escapeName(key.OrchestratorID), escapeName(key.WorkloadID))
 	endpoints := workload + "/endpoint"
-	return []string{endpoints, workload}, nil
+	orchestrator := fmt.Sprintf("/calico/v1/host/%s/workload/%s",
+		key.Hostname, escapeName(key.OrchestratorID))
+	return []string{endpoints, workload, orchestrator}, nil
 }
 
 func (key WorkloadEndpointKey) valueType() reflect.Type {
--- a/lib/client/node_e2e_test.go
+++ b/lib/client/node_e2e_test.go
@@ -40,6 +40,8 @@ import (
 	"github.com/projectcalico/libcalico-go/lib/net"
 	"github.com/projectcalico/libcalico-go/lib/numorstring"
 	"github.com/projectcalico/libcalico-go/lib/testutils"
+	"os/exec"
+	"fmt"
 )
 
 var _ = testutils.E2eDatastoreDescribe("Node tests", testutils.DatastoreEtcdV2, func(config api.CalicoAPIConfig) {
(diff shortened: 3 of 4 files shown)


lib/client/node_e2e_test.go
```diff
@@ -141,6 +143,15 @@ var _ = testutils.E2eDatastoreDescribe("Node tests", testutils.DatastoreEtcdV2,
 			_, err = c.Nodes().Get(meta1)
 			Expect(err.Error()).To(Equal(errors.New("resource does not exist: Node(name=node1)").Error()))
 
+			// Node1 directory in etcd should be removed.
```
robbrockbank: Isn't this test covered by the next test?

lib/client/node_e2e_test.go
```diff
@@ -141,6 +143,15 @@ var _ = testutils.E2eDatastoreDescribe("Node tests", testutils.DatastoreEtcdV2,
...
+			out, _ := exec.Command("curl", "http://127.0.0.1:2379/v2/keys/calico/v1/host/node1").Output()
+			Expect(string(out)).To(ContainSubstring("Key not found"))
+
+			// Node1 should not exist in etcd anymore.
+			dump, _ := exec.Command("curl", "http://127.0.0.1:2379/v2/keys/calico/?recursive=true").Output()
+			fmt.Println(string(dump))
+			Expect(string(dump)).NotTo(ContainSubstring("node1"))
+
 			// Delete node2 should not error.
```
robbrockbank: I think I'd rather see a separate test to explicitly test this stuff.  I think it better to implement a test that does the following:
-  Creates a node
-  Allocates some IP addresses for that node
-  Creates a WorkloadEndpoint on that node
-  Creates a BGPPeer on that node 
-  Deletes the node
-  Verifies all node configuration is removed from etcd.

lib/backend/model/bgpconfig.go
```diff
@@ -112,7 +112,7 @@ func (key NodeBGPConfigKey) defaultDeletePath() (string, error) {
 }
 
 func (key NodeBGPConfigKey) defaultDeleteParentPaths() ([]string, error) {
-	return nil, nil
+	return []string{fmt.Sprintf("/calico/bgp/v1/host/%s", key.Nodename)}, nil
```
robbrockbank: This isn't the correct fix.  You really need to have code in the node handling to explicitly delete the BGP peer node sub-tree.  Have a look in node.go in the client directory.  As part of the delete process we also delete the IPAM node sub-tree - I think we want something similar for the bgp sub tree, so perhaps define a BGP Node model that is just used for deletion purposes.

lib/client/node_e2e_test.go
robbrockbank: I'm in two minds as to whether you need any of the workload endpoint changes.  If you do beef up the test then a valid test would be:
-  Create two endpoints under the same workload
-  Delete one.
-  Check the workload dir still exists
-  Delete the other
-  Check the workload dir no longer exists.

lib/client/workloadendpoint_e2e_test.go
```diff
@@ -148,6 +156,15 @@ var _ = testutils.E2eDatastoreDescribe("WorkloadEndpoint tests", testutils.Datas
...
 			Expect(err).NotTo(HaveOccurred())
 			Expect(len(workloadEndpointList.Items)).To(Equal(0))
+
+			// Now that all workloads have been removed there should no longer be any
+			// directories under '/calico/v1/host/$node/workload/'.
+			key = fmt.Sprintf("/calico/v1/host/%s/workload/", meta1.Node)
+			out, _ = exec.Command("curl", "http://127.0.0.1:2379/v2/keys"+key+"?recursive=true").Output()
+			Expect(string(out)).NotTo(ContainSubstring(meta1.Orchestrator))
+			key = fmt.Sprintf("/calico/v1/host/%s/workload/", meta2.Node)
+			out, _ = exec.Command("curl", "http://127.0.0.1:2379/v2/keys"+key+"?recursive=true").Output()
```
robbrockbank: you should move this "curl and check for a string" in to a test utility.

lib/backend/model/workloadendpoint.go
```diff
@@ -71,7 +71,9 @@ func (key WorkloadEndpointKey) defaultDeleteParentPaths() ([]string, error) {
 	workload := fmt.Sprintf("/calico/v1/host/%s/workload/%s/%s",
 		key.Hostname, escapeName(key.OrchestratorID), escapeName(key.WorkloadID))
 	endpoints := workload + "/endpoint"
-	return []string{endpoints, workload}, nil
+	orchestrator := fmt.Sprintf("/calico/v1/host/%s/workload/%s",
+		key.Hostname, escapeName(key.OrchestratorID))
+	return []string{endpoints, workload, orchestrator}, nil
```
heschlie: I may have misunderstood the issue, but this was quick to implement so I figured why not, now you've given me a why not :) will remove

lib/backend/model/bgpconfig.go
heschlie: Will have a look, may harass you tomorrow for some clarity if need be.

lib/client/node_e2e_test.go
heschlie: Cool I'll write something up for it, and yea it sounds like we should ditch all of the WEP changes.

heschlie:
@robbrockbank based on your feedback I'm going to close this PR, it seems like I'll be ditching most of what is here anyway, will open a new PR based on the feedback here.

PR #500, opened 2017-08-23
Ensuring calico-node does not leak dirs into etcd

### LINKED ISSUE #370: Deleting a node resource does not delete all node specific directories

heschlie:
## Description
Fixes #370 

I've added a test to ensure we're not leaking directories into etcd when we delete a node.  I've also fixed a directory leak we were aware of by adding a new model similar to how we do with the IPAM data.

## Todos
- [x] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
Calico Node will now ensure it is not leaking directories into etcd.
```

### DIFF
--- a/lib/client/node.go
+++ b/lib/client/node.go
@@ -115,6 +115,16 @@ func (h *nodes) Delete(metadata api.NodeMetadata) error {
 		}
 	}
 
+	// Remove BGP Node directory
+	log.Debug("Removing BGP Node data")
+	err = h.RemoveBGPNode(metadata.Name)
+	if err != nil {
+		log.Debug("Error removing BGP Node data: %v", err)
+		if _, ok := err.(errors.ErrorResourceDoesNotExist); ok {
+			return err
+		}
+	}
+
 	// Finally remove the node.
 	return h.c.delete(metadata, h)
 }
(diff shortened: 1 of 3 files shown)


heschlie:
@caseydavenport @fasaxc Would one of you have time to give this a review?

heschlie:
@robbrockbank ping

robbrockbank:
NodeHost is an odd name - we use Node and Host to mean the same thing (Host was the old name, which we changed to Node).  I think this should be BGPNode.

Other than that, looks fine.

heschlie:
@robbrockbank This seems more like what you had in mind? May I go ahead and squash and merge?

lib/client/node.go
```diff
@@ -228,3 +238,18 @@ func (h *nodes) convertKVPairToAPI(d *model.KVPair) (unversioned.Resource, error
 
 	return apiNode, nil
 }
+
+// RemoveNodeHost removes all Node specific data from the datastore.
+func (h *nodes) RemoveNodeHost(host string) error {
```
caseydavenport: This is still a bit strangely named - shouldn't this be `RemoveBGPNode` now?

lib/client/node.go
```diff
@@ -115,6 +115,16 @@ func (h *nodes) Delete(metadata api.NodeMetadata) error {
 		}
 	}
 
+	// Remove Node host directory
+	log.Debug("Removing Node host data")
```
caseydavenport: same with this log - I think it means to be "Removing Node BGP configuration" or similar.

lib/client/node.go
```diff
@@ -115,6 +115,16 @@ func (h *nodes) Delete(metadata api.NodeMetadata) error {
...
+	err = h.RemoveNodeHost(metadata.Name)
+	if err != nil {
+		log.Debug("Error removing host data: %v", err)
```
caseydavenport: and this one

heschlie:
@caseydavenport Whoops, ok I've updated `node.go` to match.

@robbrockbank @caseydavenport am I ok to squash? I would guess we don't want to merge until after the 2.6 release?

robbrockbank:
lgtm

PR #502, opened 2017-08-29
Change KDD:NetPol to be ingress-only, delete Outbound allow-all

bcreane:
## Description

When converting from a k8s:NetworkPolicy to the Calico model, no longer append
an allow-all OutboundRule; rather, specify the policy is Types:ingress.

The PR includes changes to KDD backend and associated UTs:
- all UTs pass
- Related to Neil's Types PR: https://github.com/projectcalico/libcalico-go/pull/494

## Todos
- [X] Tests
- [X] Release note

## Release Note

```release-note
When converting from a Kubernetes:NetworkPolicy to a Calico:Policy, in place of appending an allow-all egress rule, now specify this policy is ingress-only.

As a result, Kubernetes:NetworkPolicies allow subsequent policies to match an explicit egress rule.
```

### DIFF
--- a/lib/backend/k8s/conversion.go
+++ b/lib/backend/k8s/conversion.go
@@ -219,10 +219,10 @@ func (c converter) networkPolicyToPolicy(np *extensions.NetworkPolicy) (*model.K
 			Name: policyName,
 		},
 		Value: &model.Policy{
-			Order:         &order,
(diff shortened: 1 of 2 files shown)


redbaron:
Is it going to work with 1.7 Kube clusters where no egress in  NetworkPolicy exist?

neiljerram:
@redbaron Yes; the effect, both before and after this change, is that there is no restriction on pod egress traffic (i.e. traffic leaving a pod).

PR #503, opened 2017-08-31
Policy conversion code made public

svInfra17:
## Description
To avoid duplication of policy conversion code in k8s-policy, made Kubernetes network policy to calico policy conversion API public so that k8s-policy can simply import backend package and reuse existing code.
-  Tests ran: `make test-containerized` 

## Todos
- [x] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/backend/k8s/conversion.go
+++ b/lib/backend/k8s/conversion.go
@@ -198,8 +198,8 @@ func (c converter) podToWorkloadEndpoint(pod *kapiv1.Pod) (*model.KVPair, error)
 	return &kvp, nil
 }
 
-// networkPolicyToPolicy converts a k8s NetworkPolicy to a model.KVPair.
-func (c converter) networkPolicyToPolicy(np *extensions.NetworkPolicy) (*model.KVPair, error) {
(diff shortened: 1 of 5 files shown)


lib/backend/k8s/conversion.go
```diff
@@ -198,8 +198,8 @@ func (c converter) podToWorkloadEndpoint(pod *kapiv1.Pod) (*model.KVPair, error)
 	return &kvp, nil
 }
 
-// networkPolicyToPolicy converts a k8s NetworkPolicy to a model.KVPair.
-func (c converter) networkPolicyToPolicy(np *extensions.NetworkPolicy) (*model.KVPair, error) {
+// NetworkPolicyToPolicy converts a k8s NetworkPolicy to a model.KVPair.
```
caseydavenport: I think you'll need to make the converter struct that this function is bound to public as well.

See here: https://github.com/projectcalico/libcalico-go/pull/503/files#diff-4b89792d3cf3070e2152b6ec11c1c517R39

When you do it could you add a TODO comment up there which says "TODO: make this private and expose a public conversion interface instead."

PR #504, opened 2017-08-31
Add named ports to datamodel and KDD

fasaxc:
## Description

This PR adds support for named ports to the calico datamodel. 

## Todos
- [x] Validation
- [x] Tests
  - [x] conversion of pods
  - [x] conversion of policies
  - [x] validation
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/backend/k8s/conversion.go
+++ b/lib/backend/k8s/conversion.go
@@ -186,9 +186,9 @@ func (c Converter) podToWorkloadEndpoint(pod *kapiv1.Pod) (*model.KVPair, error)
 				var modelProto numorstring.Protocol
 				switch containerPort.Protocol {
 				case kapiv1.ProtocolUDP:
-					numorstring.ProtocolFromString("udp")
+					modelProto = numorstring.ProtocolFromString("udp")
 				case kapiv1.ProtocolTCP, kapiv1.Protocol("") /* K8s default is TCP. */ :
-					numorstring.ProtocolFromString("tcp")
+					modelProto = numorstring.ProtocolFromString("tcp")
 				default:
 					log.WithFields(log.Fields{
 						"protocol": containerPort.Protocol,
@@ -197,10 +197,7 @@ func (c Converter) podToWorkloadEndpoint(pod *kapiv1.Pod) (*model.KVPair, error)
 					}).Debug("Ignoring named port with unknown protocol")
 					continue
 				}
-				modelProto = numorstring.ProtocolFromString("tcp")
-				if containerPort.Protocol == kapiv1.ProtocolUDP {
-					modelProto = numorstring.ProtocolFromString("udp")
-				}
+
 				endpointPorts = append(endpointPorts, model.EndpointPort{
 					Name:     containerPort.Name,
 					Protocol: modelProto,
@@ -353,7 +350,7 @@ func (c Converter) k8sIngressRuleToCalico(r extensions.NetworkPolicyIngressRule,
 		ports = []*extensions.NetworkPolicyPort{nil}
 	}
 
-	// Combine destintations with sources to generate rules.
+	// Combine destinations with sources to generate rules.
 	// TODO: This currently creates a lot of rules by making every combination of from / ports
 	// into a rule.  We can combine these so that we don't need as many rules!
 	for _, port := range ports {
--- a/lib/backend/k8s/conversion.go
+++ b/lib/backend/k8s/conversion.go
@@ -178,6 +178,38 @@ func (c Converter) podToWorkloadEndpoint(pod *kapiv1.Pod) (*model.KVPair, error)
 	}
 	labels["calico/k8s_ns"] = pod.ObjectMeta.Namespace
 
+	// Map any named ports through.
+	var endpointPorts []model.EndpointPort
+	for _, container := range pod.Spec.Containers {
+		for _, containerPort := range container.Ports {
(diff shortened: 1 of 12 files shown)


lib/numorstring/numorstring_test.go
```diff
@@ -71,12 +71,14 @@ func init() {
 		Entry("should accept 65535 port as int", "65535", portType, numorstring.SinglePort(65535)),
 		Entry("should accept 0:65535 port range as string", "\"0:65535\"", portType, portFromRange(0, 65535)),
 		Entry("should accept 1:10 port range as string", "\"1:10\"", portType, portFromRange(1, 10)),
+		Entry("should accept foo-bar port range as named port", "\"foo-bar\"", portType, numorstring.NamedPort("foo-bar")),
```
neiljerram: Why does this say "port range"?  It's just a single named port, isn't it?

lib/numorstring/port.go
```diff
@@ -18,52 +18,79 @@ import (
...
 		msg := fmt.Sprintf("minimum port number (%d) is greater than maximum port number (%d) in port range", minPort, maxPort)
 		return port, errors.New(msg)
 	}
 	return port, nil
 }
 
+var (
+	allDigits = regexp.MustCompile(`^\d+$`)
+	portRange = regexp.MustCompile(`^(\d+):(\d+)$`)
+	nameRegex = regexp.MustCompile("^[a-zA-Z0-9_.-]{1,128}$")
```
neiljerram: Is there no restriction such as the first character being alpha?  As this stands, I believe a number would also match nameRegex.

lib/validator/validator.go
```diff
@@ -112,6 +112,10 @@ func init() {
 	registerStructValidator(validateICMPFields, api.ICMPFields{})
 	registerStructValidator(validateRule, api.Rule{})
 	registerStructValidator(validateBackendRule, model.Rule{})
+	registerStructValidator(validateBackendEndpointPort, model.EndpointPort{})
+	registerStructValidator(validateEndpointPort, api.EndpointPort{})
+	registerStructValidator(validateWorkloadEndpoint, model.WorkloadEndpoint{})
```
neiljerram: insert "Backend" in the name here, for consistency

lib/validator/validator.go
```diff
@@ -112,6 +112,10 @@ func init() {
...
+	registerStructValidator(validateHostEndpoint, model.HostEndpoint{})
```
neiljerram: here too

lib/validator/validator.go
neiljerram: Also, when do these backend model validators come into play?  (I guess they do, since there are already others above - but don't immediately see the scenario.)

lib/backend/k8s/conversion.go
```diff
@@ -178,6 +178,38 @@ func (c Converter) podToWorkloadEndpoint(pod *kapiv1.Pod) (*model.KVPair, error)
 	}
 	labels["calico/k8s_ns"] = pod.ObjectMeta.Namespace
 
+	// Map any named ports through.
+	var endpointPorts []model.EndpointPort
+	for _, container := range pod.Spec.Containers {
+		for _, containerPort := range container.Ports {
+			if containerPort.Name != "" && containerPort.ContainerPort != 0 {
+				var modelProto numorstring.Protocol
+				switch containerPort.Protocol {
```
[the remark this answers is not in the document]
fasaxc: Ugh, that's just a mistake a bit of code rework that I missed.

lib/numorstring/port.go
fasaxc: I think that'd be sensible if we'd done it on day one but now it'd be a breaking change in the existing places we use names.

lib/validator/validator.go
fasaxc: Two places: 

- Felix uses them before feeding models into the calculation graph. (So they prevent someone writing rubbish to etcd directly.)
- I think calicoctl applies both front and backend validation (but the errors are cleaner with the former).

caseydavenport:
Is there a release note somewhere for this feature?

fasaxc:
@caseydavenport There's one in Felix. https://github.com/projectcalico/felix/pull/1534.

lib/numorstring/port.go
```diff
@@ -18,52 +18,73 @@ import (
...
 }
 
 // SinglePort creates a Port struct representing a single port.
 func SinglePort(port uint16) Port {
-	return Port{port, port}
+	return Port{MinPort: port, MaxPort: port}
+}
+
+func NamedPort(name string) Port {
+	return Port{PortName: name}
```
bcreane: are there any invalid characters we should reject? e.g. weird whitespace, non-printable chars, etc.

lib/validator/validator.go
```diff
@@ -520,6 +554,66 @@ func validateBGPPeerMeta(v *validator.Validate, structLevel *validator.StructLev
 	}
 }
 
+func validateBackendEndpointPort(v *validator.Validate, structLevel *validator.StructLevel) {
```
bcreane: All this boilerplate (also the code above which checks for duplicate named ports for both `WorkloadEndpointSpec` and `HostEndpointSpec`) ... I'll let more experienced golang devs comment on this, but it does look verbose.

PR #507, opened 2017-09-09
Fix how we generate inbound rules

caseydavenport:
## Description

This brings the rules generated through KDD so they match the nil vs empty semantics which is present when using etcd and certain lists are not present in the datastore.

This is important for letting the policy controller import KDD conversion code to translate policies to be programmed into etcd.


bcreane:
Lgtm!

PR #511, opened 2017-09-11
[WIP] Define v2 client API

robbrockbank:
WIP definition of the v2 client API.  Put up for initial review of the structure definitions and the client API.

Things to review:
-  apiv2
-  clientv2
-  options
-  apiconfig (move the client config out into a separate dir)... perhaps this should be clientconfig rather than apiconfig?

Issues to resolve:
-  Should the API methods take a context rather than specifying a list/watch timeout? 
-  Any suggestions for directory structure/naming.  At the moment I *think* the api and client will be removed and so apiv2->api and clientv2->client (i.e. how we currently have it).  But we can also take this opportunity to restructure further.   Main thing to consider though is minimal changes to felix/typha and so perhaps we want to keep the model definitions and the backend client api paths unchanged.

Main client interface defined in /clientv2/interface.go


lib/apiconfig/load.go
```diff
@@ -0,0 +1,77 @@
+package apiconfig
```
robbrockbank: Moved this over from the client and api.

PR #513, opened 2017-09-13
Make WorkloadEndpoint conversion code public

caseydavenport:
## Description

Make WEP conversion code public so that it can be used by the policy controller.

Necessary for making named ports available to the policy controller.

### DIFF
--- a/lib/converter/workloadendpoint.go
+++ b/lib/converter/workloadendpoint.go
@@ -1,0 +1,138 @@
+// Copyright (c) 2016-2017 Tigera, Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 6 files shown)


lib/converter/workloadendpoint.go
```diff
@@ -0,0 +1,138 @@
...
+
+	ah := api.NewWorkloadEndpoint()
+	ah.Metadata.Node = bk.Hostname
+	ah.Metadata.Orchestrator = bk.OrchestratorID
+	ah.Metadata.Workload = bk.WorkloadID
+	ah.Metadata.Name = bk.EndpointID
+	ah.Metadata.Labels = bh.Labels
+	ah.Spec.InterfaceName = bh.Name
+	ah.Metadata.ActiveInstanceID = bh.ActiveInstanceID
+	ah.Spec.MAC = bh.Mac
```
gunjan5: kek

lib/converter/workloadendpoint.go
```diff
@@ -0,0 +1,138 @@
+// Copyright (c) 2016 Tigera, Inc. All rights reserved.
```
gunjan5: 2017 :)

PR #515, opened 2017-09-14
Named ports + FV fixes

fasaxc:
## Description
This PR restores the named ports feature, which was backed out under #519. 

I found some bugs while FV testing the feature, this PR also closes those up.

## Todos
- [x] Tests
- [x] Documentation
- [x] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/api/hostendpoint.go
+++ b/lib/api/hostendpoint.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2016 Tigera, Inc. All rights reserved.
+// Copyright (c) 2016-2017 Tigera, Inc. All rights reserved.
 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -77,6 +77,9 @@ type HostEndpointSpec struct {
 	// profile is applied in the order that they appear in this list.  Profile rules are applied
 	// after the selector-based security policy.
 	Profiles []string `json:"profiles,omitempty" validate:"omitempty,dive,namespacedname"`
+
+	// Ports contains the endpoint's named ports, which may be referenced in security policy rules.
+	Ports []EndpointPort `json:"ports,omitempty" validate:"omitempty,dive"`
 }
 
 // NewHostEndpoint creates a new (zeroed) HostEndpoint struct with the TypeMetadata initialised to the current
--- a/lib/api/workloadendpoint.go
+++ b/lib/api/workloadendpoint.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2016 Tigera, Inc. All rights reserved.
+// Copyright (c) 2016-2017 Tigera, Inc. All rights reserved.
 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -19,6 +19,7 @@ import (
 
 	"github.com/projectcalico/libcalico-go/lib/api/unversioned"
 	"github.com/projectcalico/libcalico-go/lib/net"
+	"github.com/projectcalico/libcalico-go/lib/numorstring"
 )
 
 type WorkloadEndpoint struct {
@@ -104,6 +105,9 @@ type WorkloadEndpointSpec struct {
 
 	// MAC is the MAC address of the endpoint interface.
 	MAC *net.MAC `json:"mac,omitempty" validate:"omitempty,mac"`
+
+	// Ports contains the endpoint's named ports, which may be referenced in security policy rules.
+	Ports []EndpointPort `json:"ports,omitempty" validate:"omitempty,dive"`
 }
 
 // IPNat contains a single NAT mapping for a WorkloadEndpoint resource.
@@ -114,6 +118,12 @@ type IPNAT struct {
 
 	// The external IP address.
 	ExternalIP net.IP `json:"externalIP"`
+}
+
+type EndpointPort struct {
+	Name     string               `json:"name" validate:"name"`
+	Protocol numorstring.Protocol `json:"protocol"`
+	Port     uint16               `json:"port" validate:"gt=0"`
 }
 
 // String returns a friendly form of an IPNAT.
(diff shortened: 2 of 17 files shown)


caseydavenport:
Looks like this is missing this change? https://github.com/projectcalico/libcalico-go/pull/519/files#diff-4b89792d3cf3070e2152b6ec11c1c517L170

fasaxc:
@caseydavenport Not sure I understand, that change is on master; it reverts the named ports feature, which went in too soon.  This change follows on from that, reinstating those changes and then applying some fixes on top.

fasaxc:
@caseydavenport I chatted to Erik, we think that there was a github funny or I forgot to push something before I went away because we both see the code there now but he swears it wasn't showing up before.

caseydavenport:
@fasaxc huh yeah, looks like it's there now. There was definitely code missing from this PR when I looked last week but it's here now!

lib/api/workloadendpoint.go
```diff
@@ -104,6 +105,9 @@ type WorkloadEndpointSpec struct {
 
 	// MAC is the MAC address of the endpoint interface.
 	MAC *net.MAC `json:"mac,omitempty" validate:"omitempty,mac"`
+
+	// Ports contains the endpoint's named ports, which may be referenced in security policy rules.
+	Ports []EndpointPort `json:"ports,omitempty" validate:"omitempty,dive"`
```
neiljerram: (trailing dive here too)

lib/backend/k8s/conversion.go
```diff
@@ -167,6 +167,35 @@ func (c Converter) PodToWorkloadEndpoint(pod *kapiv1.Pod) (*model.KVPair, error)
 	}
 	labels["calico/k8s_ns"] = pod.ObjectMeta.Namespace
 
+	// Map any named ports through.
+	var endpointPorts []model.EndpointPort
+	for _, container := range pod.Spec.Containers {
+		for _, containerPort := range container.Ports {
```
neiljerram: We seem to be losing that - in the k8s data structures - a named port belongs to a *particular* container within the pod.  Does that matter?

lib/backend/k8s/conversion_test.go
```diff
@@ -245,12 +299,20 @@ var _ = Describe("Test NetworkPolicy conversion", func() {
...
+				Action:      "allow",
+				Protocol:    &protoTCP, // Defaulted to TCP.
+				SrcSelector: "calico/k8s_ns == 'default' && k == 'v' && k2 == 'v2'",
+				DstPorts:    []numorstring.Port{numorstring.SinglePort(80)},
+			},
+			model.Rule{
+				Action:      "allow",
+				Protocol:    &protoTCP, // Defaulted to TCP.
+				SrcSelector: "calico/k8s_ns == 'default' && k == 'v' && k2 == 'v2'",
+				DstPorts:    []numorstring.Port{numorstring.NamedPort("foo")},
```
neiljerram: Could we have instead just one unified model.Rule?
```
model.Rule{
	Action:      "allow",
	Protocol:    &protoTCP, // Defaulted to TCP.
	SrcSelector: "calico/k8s_ns == 'default' && k == 'v' && k2 == 'v2'",
	DstPorts:    []numorstring.Port{numorstring.SinglePort(80), numorstring.NamedPort("foo")},
}
```
Is there a good reason why the converter does not do that?

lib/backend/model/rule.go
```diff
@@ -40,23 +40,23 @@ type Rule struct {
...
 	NotSrcNets     []*net.IPNet       `json:"!src_nets,omitempty" validate:"omitempty"`
 	NotSrcSelector string             `json:"!src_selector,omitempty" validate:"omitempty,selector"`
-	NotSrcPorts    []numorstring.Port `json:"!src_ports,omitempty" validate:"omitempty"`
+	NotSrcPorts    []numorstring.Port `json:"!src_ports,omitempty" validate:"omitempty,dive"`
 	NotDstTag      string             `json:"!dst_tag,omitempty" validate:"omitempty"`
 	NotDstSelector string             `json:"!dst_selector,omitempty" validate:"omitempty,selector"`
 	NotDstNet      *net.IPNet         `json:"!dst_net,omitempty" validate:"omitempty"`
 	NotDstNets     []*net.IPNet       `json:"!dst_nets,omitempty" validate:"omitempty"`
-	NotDstPorts    []numorstring.Port `json:"!dst_ports,omitempty" validate:"omitempty"`
+	NotDstPorts    []numorstring.Port `json:"!dst_ports,omitempty" validate:"omitempty,dive"`
```
neiljerram: (more trailing dives above)

lib/api/hostendpoint.go
```diff
@@ -77,6 +77,9 @@ type HostEndpointSpec struct {
 	// profile is applied in the order that they appear in this list.  Profile rules are applied
 	// after the selector-based security policy.
 	Profiles []string `json:"profiles,omitempty" validate:"omitempty,dive,namespacedname"`
+
+	// Ports contains the endpoint's named ports, which may be referenced in security policy rules.
+	Ports []EndpointPort `json:"ports,omitempty" validate:"omitempty,dive"`
```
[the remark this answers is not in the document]
fasaxc: That just tells the validator to examine each EndpointPort in the slice too.  It knows how to do that because we register a validator for EndpointPort too.

lib/backend/k8s/conversion.go
fasaxc: I don't think so, the containers share a network namespace and k8s enforces uniqueness across a whole pod.  I think the association with containers in the k8s model is fairly natural because ports belong to the particular process and hence the particular container but, since we don't have the concept of a sub-workload in our datamodel, I don't think we lose anything.

lib/backend/k8s/conversion_test.go
fasaxc: Yes, we could make that optimisation but it's non-trivial due to the need to split out TCP/UDP ports into separate rules.

lib/api/hostendpoint.go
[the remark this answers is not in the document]
fasaxc: For context: initially, I failed to add the dive and the validator ignored the Ports validation, so my new UTs failed.

PR #517, opened 2017-09-15
[WIP] First iteration of v2 client supporting etcdv3 CRUD and IPAM 

robbrockbank:
This is the first iteration of the v2 client from which we should create an integration branch for the v2 client work.

THIS SHOULD NOT BE MERGED INTO MASTER

The main things to review are:
-  The API itself.  This was in another PR, but it has probably changed a little bit during implementation.
-  The plumbing through of the v2 client to the only (currently) supported backend: etcdv3
-  The etcdv3 backend
-  Minor tweaks to the IPAM module to remove it from the client and to remove the v1/v2 client use of the IPPool enumeration - instead I've added an interface to the IPAM module that the consumer must provide that is used to enumerate IPPools - this way it's not tied to either the v1 or v2 client.  I did this primarily to remove the hard-linking with the client and to make ipam a standalone unit.

Note that the backend client still uses a KVPair style API.  Reason for this is:
-  We already had a roughly working etcdv3 backend implemented to use this
-  We still need to support IPAM which is making use of this particular interface
-  It was generally a lot less effort to change.
-  Felix still talks the KVPairs and we want to minimize the churn in Felix and Typha.

For the new resources, we have a single model.ResourceKey to represent all new resource types.  The value for a new resource is the resource itself - so no mapping is required between the data exposed over the API and the data stored in etcd.

For now, I have kept the old `client` and `api` directories, and added the new client and definitions in `clientv2` and `apiv2` directories.  I do not expect these to be the final resting places, and we should take this opportunity to decide where these should live.

I would recommend reviewing the following:
-  apiv2:  new resource definitions
   -  note conversion of net.IP and net.IPNet fields to be simple strings
   -  note removal of deprecated Net and NotNet fields in the rule
   -  camel case "one of" values, e.g. cross-subnet -> CrossSubnet
   -  removed `pip.enabled` field
-  clientv2:  the new client interface
   -  especially the untyped.go module which is the meat of the mapping between the typed interface and the untyped backend interface.
-  ipam
   - most of this code is the same, but you might want to look at the code that uses the `PoolAccessorInterface` since this is new and differs from the old code where it would call directly back into the client to query the pools.  I wanted to avoid a circular dependency on the client, so added this interface.  In hindsight, I'm not sure why I bothered - although I do like not having all the IPAM code in the client directory.
-  etcdv3
   -  this is all new code, mostly based on the open etcdv3 PR, but it does have modifications.
   - DO NOT REVIEW THE SYNCER.  This is untested and there is a sprint item to get the syncer working.
-  apiconfig
   -  This is the config used to instantiate the client.  I moved it out of api.  The LoadFrom methods I moved from the client. 
   -  This isn't new code, but the location is new.

One thing I'm not sure about is how I've handled the namespaced resources.  So to create the NetworkPolicies interface you pass in the namespace, but then the namespace is also part of the resource definition which feels like duplication.  WDPT?  Should I just add the namespace to the Delete/Get/List calls and remove from the client NetworkPolicies() accessor method?

Most of the other changes are to keep things compiling ok.

```release-note
None required
```


lib/apiv2/node.go
```diff
@@ -0,0 +1,105 @@
...
+	metav1.TypeMeta `json:",inline"`
+	// Standard object's metadata.
+	Metadata metav1.ObjectMeta `json:"metadata,omitempty"`
+	// Specification of the Node.
+	Spec NodeSpec `json:"spec,omitempty"`
+}
+
+// NodeSpec contains the specification for a Node resource.
+type NodeSpec struct {
+	// BGP configuration for this node.  If this omitted, the Calico node
```
caseydavenport: Wait, is this comment correct?

lib/apiv2/node.go
```diff
@@ -0,0 +1,105 @@
...
+	// will be run in policy-only mode.
+	BGP *NodeBGPSpec `json:"bgp,omitempty" validate:"omitempty"`
+}
+
+// NodeSpec contains the specification for the Node BGP configuration.
+type NodeBGPSpec struct {
+	// The AS Number of the node.  If this is not specified, the global
+	// default value will be used.
+	ASNumber *numorstring.ASNumber `json:"asNumber,omitempty"`
```
caseydavenport: Are we sticking with our custom `numorstring` package or should we be switching to just use the upstream `intstr` package from k8s?

lib/apiv2/node.go
```diff
@@ -0,0 +1,105 @@
...
+	// IPv4Address is the IPv4 address and network of this node.  At least
+	// one of the IPv4 and IPv6 addresses should be specified.
+	IPv4Address string `json:"ipv4Address,omitempty" validate:"omitempty,ipv4"`
```
caseydavenport: nit: there should be newlines in here somewhere.

lib/apiv2/policy.go
```diff
@@ -0,0 +1,234 @@
...
+// 	- Fail to match the packet; in which case the packet proceeds to the next policy. If there
+// 	  are no more policies then the packet is dropped.
+//
+// Calico implements the security policy for each endpoint individually and only the policies that
+// have matching selectors are implemented. This ensures that the number of rules that actually need
+// to be inserted into the kernel is proportional to the number of local endpoints rather than the
+// total amount of policy.
+//
+// GlobalNetworkPolicy is globally-scoped (i.e. not Namespaced).
+type GlobalNetworkPolicy struct {
```
caseydavenport: Bit odd that all the other resources are in other files and these two share one.

lib/apiv2/policy.go
```diff
@@ -0,0 +1,234 @@
...
+	Spec PolicySpec `json:"spec,omitempty"`
+}
+
+// NetworkPolicy is the Namespaced-equivalent of the GlobalNetworkPolicy.
+type NetworkPolicy struct {
+	metav1.TypeMeta `json:",inline"`
+	// Standard object's metadata.
+	Metadata metav1.ObjectMeta `json:"metadata,omitempty"`
+	// Specification of the Policy.
+	Spec PolicySpec `json:"spec,omitempty"`
```
caseydavenport: Does it make sense to share the `PolicySpec` type here? The spec is going to behave differently based on which object it's in.

Maybe it's pragmatic to do this for now.

lib/apiv2/profile.go
```diff
@@ -0,0 +1,97 @@
...
+type Profile struct {
+	metav1.TypeMeta `json:",inline"`
+	// Standard object's metadata.
+	Metadata metav1.ObjectMeta `json:"metadata,omitempty"`
+	// Specification of the Profile.
+	Spec ProfileSpec `json:"spec,omitempty"`
+}
+
+// ProfileSpec contains the specification for a security Profile resource.
+type ProfileSpec struct {
```
caseydavenport: Now that I think about it, we should probably have a `LabelsToApply` spec field here in order to be specific and to model kubernetes behavior.

Having the `Metadata.Labels` field be the ones that are inherited feels wrong and scary.

lib/apiv2/node.go
robbrockbank: For a lot of these definitions there seem to be a mixed approach for whether fields are separated with newlines.  I see this a lot in the k8s resource definitions too.  I'm not sure which I prefer tbh.

lib/apiv2/policy.go
robbrockbank: yup - i'll split this out.

lib/apiv2/policy.go
robbrockbank: but the actual definition is identical no?

caseydavenport: yep.

lib/apiv2/profile.go
robbrockbank: Sounds good.

robbrockbank:
Closing in preference to this PR:

https://github.com/projectcalico/libcalico-go/pull/518

PR #518, opened 2017-09-16
Initial checkin for etcdv3 client and IPAM

robbrockbank:
This is the first iteration of the v2 client from which we should create an integration branch for the v2 client work.

THIS SHOULD NOT BE MERGED INTO MASTER.  NOTE THAT THIS REVIEW IS AGAINST THE v3.0-integration BRANCH AND CAN THEREFORE BE MERGED.

The main things to review are:
-  The API itself.  This was in another PR, but it has probably changed a little bit during implementation.
-  The plumbing through of the v2 client to the only (currently) supported backend: etcdv3
-  The etcdv3 backend
-  [on hold - no need to review this] Minor tweaks to the IPAM module to remove it from the client and to remove the v1/v2 client use of the IPPool enumeration - instead I've added an interface to the IPAM module that the consumer must provide that is used to enumerate IPPools - this way it's not tied to either the v1 or v2 client.  I did this primarily to remove the hard-linking with the client and to make ipam a standalone unit.

Note that the backend client still uses a KVPair style API.  Reason for this is:
-  We already had a roughly working etcdv3 backend implemented to use this
-  We still need to support IPAM which is making use of this particular interface
-  It was generally a lot less effort to change.
-  Felix still talks the KVPairs and we want to minimize the churn in Felix and Typha.

For the new resources, we have a single model.ResourceKey to represent all new resource types.  The value for a new resource is the resource itself - so no mapping is required between the data exposed over the API and the data stored in etcd.

For now, I have kept the old `client` and `api` directories, and added the new client and definitions in `clientv2` and `apiv2` directories.  I do not expect these to be the final resting places, and we should take this opportunity to decide where these should live.

I would recommend reviewing the following:
-  apiv2:  new resource definitions
   -  note conversion of net.IP and net.IPNet fields to be simple strings
   -  note removal of deprecated Net and NotNet fields in the rule
   -  camel case "one of" values, e.g. cross-subnet -> CrossSubnet
   -  removed `ipip.enabled` field
   -  added a LabelsToApply field to the profile definition because using the Metadata labels for that purpose doesn't really make sense.
-  clientv2:  the new client interface
   -  especially the untyped.go module which is the meat of the mapping between the typed interface and the untyped backend interface.
   -  I've only done the plumbing for BGPPeer and NetworkPolicy - once we are happy with those, I will finish off the remaining resource types (it's pretty simple boiler plate code)
-  ipam
   - don't review this just yet.  there is a separate backlog item to sort out IPAM.  The current code does work and tests run successfully, but we may completely change how we do IPAM in v2.  TBD.
 -  etcdv3
    -  this is all new code, mostly based on the open etcdv3 PR, but it does have modifications.
    - DO NOT REVIEW THE SYNCER.  This is untested and there is a backlog item to get the syncer working.
-  apiconfig
   -  This is the config used to instantiate the client.  I moved it out of api.  The LoadFrom methods I moved from the client. 
   -  This isn't new code, but the location is new.

One thing I'm not sure about is how I've handled the namespaced resources.  So to create the NetworkPolicies interface you pass in the namespace, but then the namespace is also part of the resource definition which feels like duplication.  WDPT?  Should I just add the namespace to the Delete/Get/List calls and remove from the client NetworkPolicies() accessor method?

Most of the other changes are to keep things compiling ok.


lib/apiv2/policy.go
```diff
@@ -0,0 +1,195 @@
...
+	// NotPorts is the negated version of the Ports field.
+	// Since only some protocols have ports, if any ports are specified it requires the
+	// Protocol match in the Rule to be set to "tcp" or "udp".
+	NotPorts []numorstring.Port `json:"notPorts,omitempty" validate:"omitempty,dive"`
+}
+
+type Action string
+
+const (
+	Allow Mode = "Allow"
```
gunjan5: why is this `Mode` here? Shouldn't it be `PolicyType`?

robbrockbank: That should be `Action` not `Mode`.  I will fix.That should be `Action` not `Mode`.  I will fix.

lib/backend/k8s/conversion.go
```diff
@@ -83,10 +83,10 @@ func (c Converter) parseProfileName(profileName string) (string, error) {
 	return strings.TrimPrefix(profileName, "k8s_ns."), nil
 }
 
-// NamespaceToProfile converts a Namespace to a Calico Profile.  The Profile stores
+// namespaceToProfile converts a Namespace to a Calico Profile.  The Profile stores
```
gunjan5: I think it needs to be exported for the Golang policy controller

lib/apiv2/ippool.go
```diff
@@ -0,0 +1,92 @@
...
+type IPIPConfiguration struct {
+	// The IPIP mode.  This can be one of "off", "always" or "cross-subnet".  A mode
+	// of "always" will also use IPIP tunneling for routing to destination IP
+	// addresses within this pool.  A mode of "cross-subnet" will only use IPIP
+	// tunneling when the destination node is on a different subnet to the
+	// originating node.  The default value (if not specified) is "always".
+	Mode Mode `json:"mode,omitempty" validate:"omitempty,ipipmode"`
+}
+
+type Mode string
```
gunjan5: can we call this `IPIPMode` or something a bit more descriptive? this is under the `api` package so, an exported field called `Mode` under `api` package sounds too vague

lib/apiv2/policy.go
gunjan5: yes yes I meant Action

lib/backend/k8s/conversion.go
robbrockbank: hmm.  i'll try and rebase - this shouldn't have changed.

robbrockbank: grrr - rebase didn't fix.  ok  i'll fix up by hand.

lib/apiconfig/load.go
```diff
@@ -0,0 +1,77 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		c, err := LoadClientConfigFromBytes(b)
+		if err != nil {
+			return nil, fmt.Errorf("syntax error in %s: %s", filename, err)
+		}
+		return c, nil
+	} else {
```
gunjan5: done need this `else`

lib/apiv2/doc.go
```diff
@@ -0,0 +1,32 @@
...
+/*
+Package api implements the struct definitions used on the northbound client API.
+
+In particular this includes:
+	-  The various resource types as exposed on the Northbound API.  The valid resource
+	   types are:
+		-  BGPPeer
+		-  HostEndpoint
+		-  Policy
+		-  Pool
```
gunjan5: IPPool?

lib/apiv2/doc.go
```diff
@@ -0,0 +1,32 @@
...
+		-  Profile
```
gunjan5: Isn't this list missing some types? WEP, etc?

lib/backend/client.go
```diff
@@ -18,32 +18,32 @@ import (
 	"errors"
 	"fmt"
 
-	"github.com/projectcalico/libcalico-go/lib/api"
-	bapi "github.com/projectcalico/libcalico-go/lib/backend/api"
-	"github.com/projectcalico/libcalico-go/lib/backend/compat"
-	"github.com/projectcalico/libcalico-go/lib/backend/etcd"
-	"github.com/projectcalico/libcalico-go/lib/backend/k8s"
+	"github.com/projectcalico/libcalico-go/lib/apiconfig"
+	"github.com/projectcalico/libcalico-go/lib/backend/api"
```
gunjan5: I liked the old approach of importing backend/api as `bapi` to be clear it's backend API

lib/backend/etcdv3/etcdv3.go
```diff
@@ -0,0 +1,417 @@
...
+		return nil, errors.ErrorDatastoreError{Err: err}
+	}
+
+	// Etcd V3 does not return a error when compare condition fails we must verify the
+	// response Succeeded field instead.  If the compare did not succeed then check for
+	// a successful get to return either an UpdateConflict or a ResourceDoesNotExist error.
+	if !txnResp.Succeeded {
+		getResp := (*etcdv3.GetResponse)(txnResp.Responses[0].GetResponseRange())
+		if len(getResp.Kvs) == 0 {
+			return nil, errors.ErrorResourceDoesNotExist{Identifier: d.Key}
```
gunjan5: I think we should add some logs in this block to make it easier to troubleshoot

lib/backend/model/resource.go
```diff
@@ -0,0 +1,161 @@
...
+var (
+	matchGlobalResource     = regexp.MustCompile("^/calico/resources/v2/([^/]+)/([^/]+)$")
+	matchNamespacedResource = regexp.MustCompile("^/calico/resources/v2/([^/]+)/([^/]+)/([^/]+)$")
+	kindToType              = map[string]reflect.Type{
+		strings.ToLower(apiv2.KindBGPPeer):          reflect.TypeOf(apiv2.BGPPeer{}),
+		strings.ToLower(apiv2.KindIPPool):           reflect.TypeOf(apiv2.IPPool{}),
+		strings.ToLower(apiv2.KindNode):             reflect.TypeOf(apiv2.Node{}),
+		strings.ToLower(apiv2.KindHostEndpoint):     reflect.TypeOf(apiv2.HostEndpoint{}),
+		strings.ToLower(apiv2.KindNetworkPolicy):    reflect.TypeOf(apiv2.NetworkPolicy{}),
+		strings.ToLower(apiv2.KindNetworkPolicy):    reflect.TypeOf(apiv2.NetworkPolicy{}),
```
gunjan5: shouldnt this be `kindGlobalNetworkPolicy`?

lib/backend/model/resource.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+func (key ResourceKey) defaultPath() (string, error) {
+	return key.defaultDeletePath()
+}
+
+func (key ResourceKey) defaultDeletePath() (string, error) {
+	if namespace.IsNamespaced(key.Kind) {
+		return fmt.Sprintf("/calico/resources/v2/%s/%s/%s", strings.ToLower(key.Kind), key.Namespace, key.Name), nil
+	} else {
```
gunjan5: this `else` can go

lib/backend/model/resource.go
```diff
@@ -0,0 +1,161 @@
...
+	if t == nil {
+		panic("Unexpected resource kind: " + key.Kind)
+	}
+	return t
+}
+
+func (key ResourceKey) String() string {
+	if namespace.IsNamespaced(key.Kind) {
+		return fmt.Sprintf("%s(%s/%s)", key.Kind, key.Namespace, key.Name)
+	} else {
```
gunjan5: same here

lib/backend/model/resource.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+func (key ResourceKey) defaultDeleteParentPaths() ([]string, error) {
+	return nil, nil
+}
+
+func (key ResourceKey) valueType() reflect.Type {
+	t := kindToType[strings.ToLower(key.Kind)]
+	if t == nil {
+		panic("Unexpected resource kind: " + key.Kind)
```
gunjan5: how about a `log.Fatalf` or something? panics are not userfriendly

lib/backend/model/resource.go
```diff
@@ -0,0 +1,161 @@
...
+		if len(options.Namespace) != 0 && namespace != options.Namespace {
+			log.Debugf("Didn't match namespace %s != %s", options.Namespace, namespace)
+			return nil
+		}
+		if len(options.Name) != 0 && name != options.Name {
+			log.Debugf("Didn't match name %s != %s", options.Name, name)
+			return nil
+		}
+		return ResourceKey{Kind: options.Kind, Namespace: namespace, Name: name}
+	} else {
```
gunjan5: ~else~ :)

lib/clientv2/client.go
```diff
@@ -0,0 +1,164 @@
...
+	"github.com/satori/go.uuid"
+	log "github.com/sirupsen/logrus"
+
+	"github.com/projectcalico/libcalico-go/lib/apiconfig"
+	"github.com/projectcalico/libcalico-go/lib/backend"
+	bapi "github.com/projectcalico/libcalico-go/lib/backend/api"
+	"github.com/projectcalico/libcalico-go/lib/backend/model"
+	"github.com/projectcalico/libcalico-go/lib/errors"
+	"github.com/projectcalico/libcalico-go/lib/ipam"
+	"github.com/projectcalico/libcalico-go/lib/net"
```
gunjan5: silly question, now that we're using k8s style `string` type IP, why do we still need our own `net` package?

lib/client/client.go
```diff
@@ -101,8 +104,28 @@ func (c *Client) BGPPeers() BGPPeerInterface {
...
-	return newIPAM(c)
+func (c *Client) IPAM() ipam.Interface {
+	return ipam.NewIPAM(c.Backend, poolAccessor{})
+}
+
+type poolAccessor struct {
+	client *Client
+}
+
+func (p poolAccessor) GetEnabledPools(ipVersion int) ([]net.IPNet, error) {
```
gunjan5: can you clarify why is this in v1 client?

lib/clientv2/client.go
```diff
@@ -0,0 +1,164 @@
...
+// client contains
+type client struct {
+	// The backend client is currently public to allow access to datastore
+	// specific functions that are used by calico/node.  This is a temporary
+	// measure and users of the client API should not assume that the backend
+	// will be available in the future.
+	Backend bapi.Client
+
+	// The untyped client used internally
+	untyped untyped
```
gunjan5: not too concerned but I think we should rename this to something else

lib/clientv2/client.go
```diff
@@ -0,0 +1,164 @@
...
+	// irrespective of how Calico is deployed.
+	kv := &model.KVPair{
+		Key:   model.GlobalConfigKey{Name: "ClusterGUID"},
+		Value: fmt.Sprintf("%v", hex.EncodeToString(uuid.NewV4().Bytes())),
+	}
+	if _, err := c.Backend.Create(kv); err == nil {
+		log.WithField("ClusterGUID", kv.Value).Info("Assigned cluster GUID")
+	} else {
+		if _, ok := err.(errors.ErrorResourceAlreadyExists); !ok {
+			log.WithError(err).WithField("ClusterGUID", kv.Value).Warn("Failed to assign cluster GUID")
```
gunjan5: wouldn't this throw an error if calico-node simply reboots and calls `EnsureInitialized` to check the connection because ClusterGUID was already `Create`d from the first bootup?

lib/clientv2/doc.go
```diff
@@ -0,0 +1,37 @@
...
+definitions for each resource type are defined in the following package:
+	github.com/projectcalico/libcalico-go/lib/api
+
+The client has a number of methods that return interfaces for managing:
+    -  BGP Peer resources
+	-  Policy resources
+	-  IP Pool resources
+	-  Host endpoint resources
+	-  Workload endpoint resources
+	-  Profile resources
```
gunjan5: missing GNP?

lib/clientv2/interface.go
```diff
@@ -0,0 +1,45 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package clientv2
+
+import "github.com/projectcalico/libcalico-go/lib/ipam"
+
+type Interface interface {
```
gunjan5: an interface called `Interface`? How about `ClientInterface` or something similar? Same for the filename

lib/namespace/resource.go
```diff
@@ -0,0 +1,12 @@
+package namespace
```
gunjan5: nit: a bunch of files in this package are missing the copyright notice

gunjan5:
left some comments/questions/suggestions, still haven't gotten around reviewing the etcdv3 backend and the tests

lib/apiconfig/load.go
robbrockbank: kk

lib/apiv2/doc.go
robbrockbank: yup.

lib/apiv2/doc.go
robbrockbank: yup and needs GNP and NP

lib/backend/client.go
robbrockbank: yup fair enough.

lib/backend/model/resource.go
robbrockbank: Yup!

lib/backend/model/resource.go
robbrockbank: Sure - although this is really to catch errors in the code and isn't something a user should ever hit when using the client.

lib/backend/model/resource.go
robbrockbank: kkkkkk

lib/clientv2/client.go
robbrockbank: We still have other places that use it - for example the IPAM module and the v1 models. 

In this particular case because it's used by the old IPAM module.

lib/clientv2/client.go
```diff
@@ -0,0 +1,164 @@
...
+	"github.com/projectcalico/libcalico-go/lib/backend"
+	bapi "github.com/projectcalico/libcalico-go/lib/backend/api"
+	"github.com/projectcalico/libcalico-go/lib/backend/model"
+	"github.com/projectcalico/libcalico-go/lib/errors"
+	"github.com/projectcalico/libcalico-go/lib/ipam"
+	"github.com/projectcalico/libcalico-go/lib/net"
+	"github.com/projectcalico/libcalico-go/lib/options"
+)
+
+// client contains
```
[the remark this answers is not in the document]
robbrockbank: kk

lib/clientv2/client.go
robbrockbank: Ok, I've made the naming resource-centric rather than "untyped"

lib/clientv2/client.go
robbrockbank: eh?  this doesn't return an error if the ClusterGUID was already created.

lib/client/client.go
robbrockbank: for compilation purposes.  also it allowed me to test a subset of the changes using etcdv2.  I am expecting the v1 client to be removed eventually, so I wouldn't dwell on this too much at this point.

lib/clientv2/interface.go
robbrockbank: actually this is common practice in k8s and I quite like it.  The Interface represents the primary interface for the package.  Sub-interfaces do need additional naming, but having the main entry point as just Interface is quite nice.  Thus a caller to the code uses client.Interface.

client.ClientInterface is generally discouraged due to repetition.

lib/namespace/resource.go
robbrockbank: there is only one file in this package.  You mean another package?

lib/clientv2/client.go
gunjan5: oooo nvm, it's `!ok` case. Ignore that.

lib/namespace/resource.go
gunjan5: oo, I mean teh options package

lib/backend/etcdv3/etcdv3.go
```diff
@@ -0,0 +1,462 @@
...
+	).Commit()
+	if err != nil {
+		logCxt.WithError(err).Warning("Create failed")
+		return nil, errors.ErrorDatastoreError{Err: err}
+	}
+
+	if !txnResp.Succeeded {
+		// The resource must already exist.  Extract the current value and
+		// return that if possible.
+		logCxt.Info("Create transaction failed due to resource already existing")
```
gunjan5: is this always the case when `txnResp.Scceeded` = `false`?

lib/backend/etcdv3/etcdv3.go
```diff
@@ -0,0 +1,462 @@
...
+		return nil, err
+	}
+
+	return &EtcdV3Client{etcdClient: client}, nil
+}
+
+// Create an entry in the datastore.  If the entry already exists, this will return
+// an ErrorResourceAlreadyExists error and the current entry.
+func (c *EtcdV3Client) Create(d *model.KVPair) (*model.KVPair, error) {
+	logCxt := log.WithFields(log.Fields{"model-key": d.Key, "value": d.Value, "ttl": d.TTL, "rev": d.Revision})
```
gunjan5: will there be a `Revision` for `Create`?

robbrockbank: There should not be - but tracing it out doesn't seem like a terrible thing to do.

lib/backend/etcdv3/etcdv3.go
robbrockbank: I believe the Succeeded is a response to the If() portion of the transaction, and so false would be because the resource already exists.  However, since you are the primary reviewer for this it might be good to dive into the etcdv3 calls to check my assumptions.

lib/backend/etcdv3/etcdv3.go
```diff
@@ -0,0 +1,462 @@
...
+func (c *EtcdV3Client) Apply(d *model.KVPair) (*model.KVPair, error) {
+	logCxt := log.WithFields(log.Fields{"key": d.Key, "value": d.Value, "ttl": d.TTL, "rev": d.Revision})
+	logCxt.Debug("Processing Apply request")
+	key, value, err := getKeyValueStrings(d)
+	if err != nil {
+		return nil, err
+	}
+
+	logCxt.Debug("Performing etcdv3 Put for Apply request")
+	resp, err := c.etcdClient.Put(context.Background(), key, value)
```
gunjan5: for `Apply` we usually call `Update` & `Create`, why haven't we done the same here?

gunjan5: another reason being `apply` = `create` + `update` so we want to use the same options we used in create and update, and if we change one of those we want `apply` to change with it

lib/backend/etcdv3/etcdv3.go
```diff
@@ -0,0 +1,462 @@
...
+
+	if !txnResp.Succeeded {
+		logCxt.Info("Delete transaction failed due to resource update conflict")
+		return errors.ErrorResourceUpdateConflict{Identifier: k}
+	}
+
+	delResp := txnResp.Responses[0].GetResponseDeleteRange()
+	if delResp.Deleted == 0 {
+		logCxt.Info("Delete transaction failed due resource not existing")
+		return errors.ErrorResourceDoesNotExist{Identifier: k}
```
gunjan5: so does this mean, when the resource doesn't exist and you try to delete it, you still get `txnResp.Succeeded` = `true`??

lib/backend/etcdv3/etcdv3.go
robbrockbank: I am planning to remove the Apply  and just support Create/Update.

For calicoctl, Apply will need additional logic to ensure we don't lose the read-only fields in the Metadata.

lib/backend/etcdv3/etcdv3.go
robbrockbank: Yes I believe that is the case, and the UTs would suggest that it correct too.  Checking the number of deleted entries is what indicates whether the entry existed in the first place.

Again - it would make sense for you to dive into the etcdv3 calls too to make sure I am not misusing them.

robbrockbank: I'm going to add a commit to beef up the e2e UT/FVs to check the actual error contents.

lib/backend/api/api.go
```diff
@@ -106,10 +104,11 @@ type Client interface {
...
 
-	// Perform any "backdoor" initialization required by the components
-	// used in calico/node.  This is a temporary mechanism and will be
-	// removed.
-	EnsureCalicoNodeInitialized(node string) error
+	// Clean removes Calico data from the backend datastore.  Used for test purposes.
+	Clean() error
+
+	// Close the client.
+	//Close()
```
gunjan5: why is this removed? there would be goroutine leaks if we don't close the client. I don't see `clientv3.Close()` being called manually either anywhere in the code

lib/backend/etcdv3/etcdv3.go
```diff
@@ -0,0 +1,465 @@
...
+
+	putOpts, err := c.getTTLOption(d)
+	if err != nil {
+		return nil, err
+	}
+
+	// Checking for 0 version of the key, which means it doesn't exists yet,
+	// and if it does, get the current value.
+	logCxt.Debug("Performing etcdv3 transaction for Create request")
+	txnResp, err := c.etcdClient.Txn(context.Background()).If(
```
gunjan5: should we have some sort of client timeout for the context (different from the `DialTimeout`)? 
```go
ctx, cancel := context.WithTimeout(context.Background(), timeout)
txnResp, err := c.etcdClient.Txn(ctx).If(
...
```

lib/backend/etcdv3/etcdv3.go
gunjan5: There are these server error codes the etcdv3 client can return https://github.com/coreos/etcd/blob/master/etcdserver/api/v3rpc/rpctypes/error.go#L23

See example in https://godoc.org/github.com/coreos/etcd/clientv3
```go
resp, err := kvc.Put(ctx, "", "")
if err != nil {
	if err == context.Canceled {
		// ctx is canceled by another routine
	} else if err == context.DeadlineExceeded {
		// ctx is attached with a deadline and it exceeded
	} else if verr, ok := err.(*v3rpc.ErrEmptyKey); ok {
		// process (verr.Errors)
	} else {
		// bad cluster endpoints, which are not etcd servers
	}
}
```

lib/backend/etcdv3/etcdv3.go
```diff
@@ -0,0 +1,465 @@
...
+		clientv3.Compare(clientv3.Version(key), "=", 0),
+	).Then(
+		clientv3.OpPut(key, value, putOpts...),
+	).Else(
+		clientv3.OpGet(key),
+	).Commit()
```
gunjan5: so interestingly, this is how they're doing it in k8s API server: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/storage/etcd3/store.go#L175-L179

robbrockbank: yes - i think i stole the idea from there.

lib/backend/etcdv3/etcdv3.go
robbrockbank: yes we should.  can you ignore that for now though - I was planning on handling the watch/list timeouts as part of another task.

lib/clientv2/resources.go
```diff
@@ -0,0 +1,281 @@
...
+	return nil, nil
+}
+
+// resourceToKVPair converts the resource to a KVPair that can be consumed by the
+// backend datastore client.
+func (c *resources) resourceToKVPair(opts options.SetOptions, kind string, in resource) *model.KVPair {
+	// Prepare the resource to remove non-persisted fields.
+	rv := in.GetObjectMeta().GetResourceVersion()
+	in.GetObjectMeta().SetResourceVersion("")
+	in.GetObjectMeta().SetSelfLink("")
```
gunjan5: what about other things in the ObjectMetadata like `creationTimestamp`, and `UID`?

lib/clientv2/resources.go
```diff
@@ -0,0 +1,281 @@
...
+// resource.
+func (c *resources) kvPairToResource(kvp *model.KVPair) resource {
+	// Extract the resource from the returned value - the backend will already have
+	// decoded it.
+	out := kvp.Value.(resource)
+
+	// Remove the SelfLink which Calico does not use, and set the ResourceVersion from the
+	// value returned from the backend datastore.
+	out.GetObjectMeta().SetSelfLink("")
+	out.GetObjectMeta().SetResourceVersion(kvp.Revision)
```
gunjan5: do we want to expose `creationTimestamp` to the clients? Not too worried about Get calls listing it but not sure if we want to let clients set it

lib/backend/etcdv3/etcdv3.go
gunjan5: ok as long as you're keeping track of those :)

lib/backend/api/api.go
robbrockbank: Actually it was never removed, because it was never implemented in the first place for v1.  I added it here, but I haven't implemented it yet since it isn't required for CRUD, and possibly not even for MVP.  I don't think we have anything that needs to close an existing client.

lib/clientv2/resources.go
robbrockbank: those things are persisted in the datastore so should not be removed.

lib/clientv2/resources.go
robbrockbank: Why?  It's a valid field in the Metadata.  Why should it not be exposed - who uses it otherwise?

robbrockbank: The resource version is special because the source of truth is the datastore itself.  It provides the revision information and it is not stored in the resource data.

The self link is also a special field because it is filled in by the AAPIS and doesn't mean anything when just using the libcalico-go API.  We shouldn't persist it.  If it turns out the AAPIS needs it to be persisted we'll revisit, but I don't believe it does and I don't want to store data that we don't need to persist.

lib/backend/model/resource.go
```diff
@@ -0,0 +1,159 @@
...
+
+	log "github.com/sirupsen/logrus"
+
+	"github.com/projectcalico/libcalico-go/lib/apiv2"
+	"github.com/projectcalico/libcalico-go/lib/namespace"
+)
+
+var (
+	matchGlobalResource     = regexp.MustCompile("^/calico/resources/v2/([^/]+)/([^/]+)$")
+	matchNamespacedResource = regexp.MustCompile("^/calico/resources/v2/([^/]+)/([^/]+)/([^/]+)$")
```
matthewdupre: We don't use these for validation right?

lib/clientv2/networkpolicy.go
```diff
@@ -0,0 +1,90 @@
...
+	res := &apiv2.NetworkPolicyList{}
+	if err := r.client.resources.List(opts, apiv2.KindNetworkPolicy, apiv2.KindNetworkPolicyList, r.namespace, AllNames, res); err != nil {
+		return nil, err
+	}
+	return res, nil
+}
+
+// Watch returns a watch.Interface that watches the NetworkPolicies that match the
+// supplied options.
+func (r networkPolicies) Watch(opts options.ListOptions) (watch.Interface, error) {
```
shatrugna: The Watch API needs to account for watching at an individual resource level granularity.
Ref: Storage Watch API "Watch(ctx context.Context, key string, resourceVersion string, p storage.SelectionPredicate)" . where key can reference an individual object or a collection.

lib/apiconfig/load.go
```diff
@@ -0,0 +1,76 @@
...
+
+	"github.com/kelseyhightower/envconfig"
+	yaml "github.com/projectcalico/go-yaml-wrapper"
+	"github.com/projectcalico/libcalico-go/lib/apiv2"
+	log "github.com/sirupsen/logrus"
+)
+
+// LoadClientConfig loads the ClientConfig from the specified file (if specified)
+// or from environment variables (if the file is not specified).
+func LoadClientConfig(filename string) (*CalicoAPIConfig, error) {
```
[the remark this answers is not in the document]
robbrockbank: This is not a new thing.  If you think the config loading should be handled differently then best to either raise a prioritized backlog item or add an issue.

lib/backend/k8s/k8s.go
```diff
@@ -159,8 +159,23 @@ func (c *KubeClient) EnsureInitialized() error {
 	return nil
 }
 
-func (c *KubeClient) EnsureCalicoNodeInitialized(node string) error {
-	log.WithField("Node", node).Info("Ensuring node is initialized")
+func (c *KubeClient) Clean() error {
```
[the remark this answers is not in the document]
robbrockbank: Matt - this is k8s code.  As per the PR notes - don't review the k8s stuff or the watcher stuff in this PR.  This PR only covered CRUD or etcdv3.

lib/backend/model/resource.go
robbrockbank: Validation of what?

We use these to covert from a etcdv3 key back to a resource.

lib/backend/etcdv3/etcdv3.go
robbrockbank: Actually this was lifted from the etcdv3 PR which I've modified a little.  So no from the k8s code.

So it does need a decent review.  That said, G5 has already given it a fairly in-depth review.

lib/backend/etcdv3/etcdv3.go
```diff
@@ -0,0 +1,465 @@
...
+		logCxt.Info("Create transaction failed due to resource already existing")
+		var existing *model.KVPair
+		getResp := (*clientv3.GetResponse)(txnResp.Responses[0].GetResponseRange())
+		if len(getResp.Kvs) != 0 {
+			if v, err := model.ParseValue(d.Key, getResp.Kvs[0].Value); err == nil {
+				logCxt.Debug("Parsed existing entry to return in response")
+				existing = &model.KVPair{
+					Key:      d.Key,
+					Value:    v,
+					Revision: strconv.FormatInt(getResp.Kvs[0].ModRevision, 10),
```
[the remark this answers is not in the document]
robbrockbank: What do you mean?

lib/clientv2/networkpolicy.go
```diff
@@ -0,0 +1,90 @@
...
+func (r networkPolicies) Update(res *apiv2.NetworkPolicy, opts options.SetOptions) (*apiv2.NetworkPolicy, error) {
+	out, err := r.client.resources.Update(opts, apiv2.KindNetworkPolicy, r.namespace, res)
+	if out != nil {
+		return out.(*apiv2.NetworkPolicy), err
+	}
+	return nil, err
+}
+
+// Delete takes name of the NetworkPolicy and deletes it. Returns an error if one occurs.
+func (r networkPolicies) Delete(name string, opts options.DeleteOptions) error {
```
shatrugna: Delete should return the value of the 'name' object it deleted.

lib/clientv2/networkpolicy.go
```diff
@@ -0,0 +1,90 @@
...
+	err := r.client.resources.Delete(opts, apiv2.KindNetworkPolicy, r.namespace, name)
+	return err
+}
+
+// Get takes name of the NetworkPolicy, and returns the corresponding NetworkPolicy object,
+// and an error if there is any.
+func (r networkPolicies) Get(name string, opts options.GetOptions) (*apiv2.NetworkPolicy, error) {
```
shatrugna: Get doesn't seem to be setting object.TypeMeta.Kind and object.TypeMeta.APIVersion. It seems to be set fine though when Listing. Can you confirm?

lib/clientv2/networkpolicy.go
shatrugna: For lack of a better place leaving the comment here. Wondering if you could just use the Types from upstream:
```
const (
	Added    EventType = "ADDED"
	Modified EventType = "MODIFIED"
	Deleted  EventType = "DELETED"
	Error    EventType = "ERROR"

	DefaultChanSize int32 = 100
)
```

lib/clientv2/networkpolicy.go
shatrugna: Delete could take an optional parameter, UID, for conditional deletion.

PR #521, opened 2017-09-20
Add support for k8s egress rules and CIDR blocks

bcreane:
## Description
Add support for k8s egress rules and CIDR blocks during NetworkPolicy -> Calico policy conversion. Add several UTs to validate correct conversion. FV, if it's necessary, will be a subsequent PR.

## Todos
- [X] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
Add support for converting K8s NetworkPolicy Egress rules and IPBlock selectors to Calico policy.
```


lib/backend/k8s/conversion_test.go
```diff
@@ -665,6 +673,166 @@ var _ = Describe("Test NetworkPolicy conversion", func() {
...
+		np := extensions.NetworkPolicy{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "testPolicy",
+				Namespace: "default",
+			},
+			Spec: extensions.NetworkPolicySpec{
+				PodSelector: metav1.LabelSelector{},
+				Egress: []extensions.NetworkPolicyEgressRule{
+					{
+						To: []extensions.NetworkPolicyPeer{
```
caseydavenport: I think we should have a `Ports` section in this policy, or better yet another test which validates an egress rule with both ports + peers.

bcreane: added a separate test

PR #523, opened 2017-09-20
Completes support for Calico policy when using KDD client

doublek:
## Description
Adds support for creating, updating and deleting Calico policy when using `KubeClient`.

Tested by running through the [Advanced Policy Demo](https://docs.projectcalico.org/v2.5/getting-started/kubernetes/tutorials/advanced-policy) and confirmed that it works as expected when using KDD backend.

## Todos
- [x] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/backend/k8s/k8s.go
+++ b/lib/backend/k8s/k8s.go
@@ -257,6 +257,8 @@ func (c *KubeClient) Syncer(callbacks api.SyncerCallbacks) api.Syncer {
 func (c *KubeClient) Create(d *model.KVPair) (*model.KVPair, error) {
 	log.Debugf("Performing 'Create' for %+v", d)
 	switch d.Key.(type) {
+	case model.PolicyKey:
+		return c.gnpClient.Create(d)
 	case model.GlobalConfigKey:
 		return c.globalFelixConfigClient.Create(d)
 	case model.IPPoolKey:
@@ -285,6 +287,8 @@ func (c *KubeClient) Create(d *model.KVPair) (*model.KVPair, error) {
 func (c *KubeClient) Update(d *model.KVPair) (*model.KVPair, error) {
 	log.Debugf("Performing 'Update' for %+v", d)
 	switch d.Key.(type) {
+	case model.PolicyKey:
+		return c.gnpClient.Update(d)
 	case model.GlobalConfigKey:
 		return c.globalFelixConfigClient.Update(d)
 	case model.IPPoolKey:
@@ -315,6 +319,8 @@ func (c *KubeClient) Apply(d *model.KVPair) (*model.KVPair, error) {
 	switch d.Key.(type) {
 	case model.WorkloadEndpointKey:
 		return c.applyWorkloadEndpoint(d)
+	case model.PolicyKey:
+		return c.gnpClient.Apply(d)
 	case model.GlobalConfigKey:
 		return c.globalFelixConfigClient.Apply(d)
 	case model.IPPoolKey:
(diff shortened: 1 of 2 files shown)


lib/backend/k8s/k8s.go
```diff
@@ -662,6 +671,40 @@ func (c *KubeClient) getPolicy(k model.PolicyKey) (*model.KVPair, error) {
...
+// deletePolicy deletes a Policy. It takes care of appropriate conversions whether
+// the Policy is backed by a NetworkPolicy or a Global Network Policy.
+func (c *KubeClient) deletePolicy(d *model.KVPair) error {
+	k, ok := d.Key.(model.PolicyKey)
+	if !ok {
+		return goerrors.New("Incorrect key")
+	} else if k.Name == "" {
+		return goerrors.New("Missing policy name")
+	}
+	if strings.HasPrefix(k.Name, "knp.default.") {
```
caseydavenport: AFAIK we don't need to implement delete for knp.default policies, unless I'm forgetting something.

These should pretty much be exclusively managed through the k8s API, so we should probably just return an "operation not supported" error.

Ultimately I think `calicoctl` should prevent people from deleting / modifying `knp.default` policies anyway for both etcd and kdd.

Discussed with @lxpollitt and he agrees.

lib/backend/k8s/k8s.go
```diff
@@ -662,6 +671,40 @@ func (c *KubeClient) getPolicy(k model.PolicyKey) (*model.KVPair, error) {
...
 }
 
+// deletePolicy deletes a Policy. It takes care of appropriate conversions whether
+// the Policy is backed by a NetworkPolicy or a Global Network Policy.
+func (c *KubeClient) deletePolicy(d *model.KVPair) error {
+	k, ok := d.Key.(model.PolicyKey)
+	if !ok {
+		return goerrors.New("Incorrect key")
+	} else if k.Name == "" {
+		return goerrors.New("Missing policy name")
```
caseydavenport: Probably fine to have this check, but I _think_ this will be validated before it gets to us so we won't have to worry about this case.

doublek:
@caseydavenport I've made the requested changes. PTAL...

lib/backend/k8s/k8s.go
```diff
@@ -344,10 +350,13 @@ func (c *KubeClient) Apply(d *model.KVPair) (*model.KVPair, error) {
 	}
 }
 
-// Delete an entry in the datastore. This is a no-op when using the k8s backend.
+// Delete an entry in the datastore. Returns an error if the entry does not exist or if
+// deleteion is performed on an unsupported resource.
```
caseydavenport: typo: deleteion

caseydavenport:
@doublek one more thing - we need a way to indicate to users that they can't create policies that start with `knp.default` otherwise our logic here doesn't work.

That logic should probably be in `calicoctl` since the policy controller _does_ need to be able to create policies with those names.  Are you planning to do that?

We could also add a check in the KDD backend code.  WDYT?

doublek:
@caseydavenport Good point.

If the policy controller is using libcalico-go's client (which it probably does), then this PR, with the check for `knp.default` in the `Delete` method, will break policy controllers ability to delete policies that begin with `knp.default`. In that case, I think we should:

1. remove the check for `knp.default` in the `Delete` method 
2. enforce this restriction in `calicoctl` for Create, Update, and Delete calls.

PR #524, opened 2017-09-21
Make validation of policy types more permissive.

fasaxc:
## Description

Allow ingress/egress rules to be specified even if the corresponding Type isn't enabled.

This allows for smoother upgrades since we can have policy that works for old and new Felix versions.

## Todos
- [x] Tests
- [x] Documentation (no change needed, text didn't cover this case)
- [x] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/validator/validator.go
+++ b/lib/validator/validator.go
@@ -552,21 +552,4 @@ func validatePolicySpec(v *validator.Validate, structLevel *validator.StructLeve
 			mp[t] = true
 		}
 	}
-
-	// When Types is explicitly specified:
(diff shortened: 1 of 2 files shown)


neiljerram:
Code lgtm, but I would prefer a couple of improvements to the commit message:

- 'corresponsing' should be 'corresponding'

- It could be read as saying that ingress/egress rules will be *honoured* even when the corresponding Types value is absent.  So perhaps: "Allow a Policy to specify ingress/egress rules even if the corresponding Type isn't enabled."

PR #527, opened 2017-09-22
Add context to client interface

robbrockbank:
This is built on the Watcher support, so should only be reviewed once #522 is merged.

Add context to all of the v2 client calls and remove the timeout options from the List/Watch processing.

This will be required by the AAPIS integration.


lib/clientv2/bgppeer_e2e_test.go
```diff
@@ -58,7 +60,7 @@ var _ = testutils.E2eDatastoreDescribe("BGPPeer tests", testutils.DatastoreAll,
 			be.Clean()
 
 			By("Updating the BGPPeer before it is created")
-			res, outError := c.BGPPeers().Update(&apiv2.BGPPeer{
+			res, outError := c.BGPPeers().Update(context.Background(), &apiv2.BGPPeer{
```
gunjan5: can't you create a test level shared context var `ctx := context.Background()` and reuse it in the tests? especially since we're not canceling any of the contexts in any of the tests

robbrockbank: You can, but the Background context is static anyways

robbrockbank: If you really want this changed - I'll fix in #528 if that's ok with you because I've added a bunch of tests using the same test template - so I'll have to fix that up anyways ...

PR #530, opened 2017-09-23
IPAM updates for v2 client interface and etcdv3

robbrockbank:
@gunjan5   THIS PR MUST BE SQUASHED BEFORE MERGING

If you have time this sprint, do you think you could review the changes I made to the IPAM module for the v2 client.  Could you review these files plus the poolAccessor in the clientv2/client.go file in the current v3.0-integration branch.

So here's the thing.  The code is already checked into the v3.0-integration branch but we shelved reviewing the IPAM code because we expected that we might change the code significantly.  Current thinking is that we now will not - so I would like to get this reviewed and the task closed down (which will allow us to basically complete the CNI integration task).

Now because the code is already checked in and I didn't do a clever git move of the files I couldn't easily work out how to get a sensible PR to review.  So I've done this.  This PR has two commits.  One to back out the changes and the second to put them back in.  Could you just review the SECOND commit.

Most notable changes are:
-  I moved the IPAM into a separate package (it's not clear from the review, but it used to be in the lib/client package)
-  Removed the code which directly accessed the Calico client.  I did this for a couple of reasons:
   -  Moving IPAM into the separate package means you'd end up with circular references since client references IPAM.
   -  I'm still not sure what we need to do about upgrade/backlevel support so I wanted to abstract out the Calico client from the IPAM module.
-  Fixed up the tests since they were performing a lot of setup in a different thread to the actual tests which was causing weird issues for me.  Hopefully these are actually fixed and correct now.
-  Moved the IPAM interface definition from the ipam.go into a separate interface.go file (which seems to be what the cool kids are doing).

On the second point, I removed the direct calls to the Calico client by abstracting out the function that queries the IP Pools.  Hence I added an interface to return the pool CIDRs.

### DIFF
--- a/lib/ipam/ipam.go
+++ b/lib/ipam/ipam.go
@@ -15,13 +15,12 @@
 package ipam
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"os"
 
 	log "github.com/sirupsen/logrus"
-
-	"context"
 
 	bapi "github.com/projectcalico/libcalico-go/lib/backend/api"
 	"github.com/projectcalico/libcalico-go/lib/backend/model"
@@ -36,16 +35,16 @@ const (
 	ipamKeyErrRetries = 3
 )
 
-// NewIPAM returns a new ipamClient, which implements the IPAMInterface.
+// NewIPAMClient returns a new ipamClient, which implements Interface.
 // Consumers of the Calico API should not create this directly, but should
-// access IPAM through the main client IPAM accessor.
-func NewIPAM(client bapi.Client, pools PoolAccessorInterface) *ipams {
-	return &ipams{client: client, pools: pools,
+// access IPAM through the main client IPAM accessor (e.g. clientv2.IPAM())
+func NewIPAMClient(client bapi.Client, pools PoolAccessorInterface) Interface {
+	return &ipamClient{client: client, pools: pools,
 		blockReaderWriter: blockReaderWriter{client: client, pools: pools}}
 }
 
-// ipamClient implements the IPAMInterface
-type ipams struct {
+// ipamClient implements Interface
+type ipamClient struct {
 	client            bapi.Client
 	pools             PoolAccessorInterface
 	blockReaderWriter blockReaderWriter
@@ -54,7 +53,7 @@ type ipams struct {
 // AutoAssign automatically assigns one or more IP addresses as specified by the
 // provided AutoAssignArgs.  AutoAssign returns the list of the assigned IPv4 addresses,
 // and the list of the assigned IPv6 addresses.
-func (c ipams) AutoAssign(args AutoAssignArgs) ([]net.IP, []net.IP, error) {
+func (c ipamClient) AutoAssign(args AutoAssignArgs) ([]net.IP, []net.IP, error) {
 	// Determine the hostname to use - prefer the provided hostname if
 	// non-nil, otherwise use the hostname reported by os.
 	hostname := decideHostname(args.Hostname)
@@ -96,7 +95,7 @@ func (c ipams) AutoAssign(args AutoAssignArgs) ([]net.IP, []net.IP, error) {
 	return v4list, v6list, nil
 }
 
-func (c ipams) autoAssign(num int, handleID *string, attrs map[string]string, pools []net.IPNet, version ipVersion, host string) ([]net.IP, error) {
+func (c ipamClient) autoAssign(num int, handleID *string, attrs map[string]string, pools []net.IPNet, version ipVersion, host string) ([]net.IP, error) {
 
 	// Start by trying to assign from one of the host-affine blocks.  We
 	// always do strict checking at this stage, so it doesn't matter whether
@@ -159,7 +158,7 @@ func (c ipams) autoAssign(num int, handleID *string, attrs map[string]string, po
 		}
 
 		if retries == 0 {
-			return nil, errors.New("Max retries hit")
+			return nil, errors.New("Max retries hit - excessive concurrent IPAM requests")
 		}
 	}
 
@@ -224,7 +223,7 @@ func (c ipams) autoAssign(num int, handleID *string, attrs map[string]string, po
 // in order to satisfy the assignment.  An error will be returned if the IP address
 // is already assigned, or if StrictAffinity is enabled and the address is within
 // a block that does not have affinity for the given host.
-func (c ipams) AssignIP(args AssignIPArgs) error {
+func (c ipamClient) AssignIP(args AssignIPArgs) error {
 	hostname := decideHostname(args.Hostname)
 	log.Infof("Assigning IP %s to host: %s", args.IP, hostname)
 
@@ -292,12 +291,12 @@ func (c ipams) AssignIP(args AssignIPArgs) error {
 		}
 		return nil
 	}
-	return errors.New("Max retries hit")
+	return errors.New("Max retries hit - excessive concurrent IPAM requests")
 }
 
 // ReleaseIPs releases any of the given IP addresses that are currently assigned,
 // so that they are available to be used in another assignment.
-func (c ipams) ReleaseIPs(ips []net.IP) ([]net.IP, error) {
+func (c ipamClient) ReleaseIPs(ips []net.IP) ([]net.IP, error) {
 	log.Infof("Releasing IP addresses: %v", ips)
 	unallocated := []net.IP{}
 
@@ -330,7 +329,7 @@ func (c ipams) ReleaseIPs(ips []net.IP) ([]net.IP, error) {
 	return unallocated, nil
 }
 
-func (c ipams) releaseIPsFromBlock(ips []net.IP, blockCIDR net.IPNet) ([]net.IP, error) {
+func (c ipamClient) releaseIPsFromBlock(ips []net.IP, blockCIDR net.IPNet) ([]net.IP, error) {
 	for i := 0; i < ipamEtcdRetries; i++ {
 		obj, err := c.client.Get(context.Background(), model.BlockKey{CIDR: blockCIDR}, "")
 		if err != nil {
@@ -389,10 +388,10 @@ func (c ipams) releaseIPsFromBlock(ips []net.IP, blockCIDR net.IPNet) ([]net.IP,
 		}
 		return unallocated, nil
 	}
-	return nil, errors.New("Max retries hit")
+	return nil, errors.New("Max retries hit - excessive concurrent IPAM requests")
 }
 
-func (c ipams) assignFromExistingBlock(
+func (c ipamClient) assignFromExistingBlock(
 	blockCIDR net.IPNet, num int, handleID *string, attrs map[string]string, host string, affCheck bool) ([]net.IP, error) {
 	// Limit number of retries.
 	var ips []net.IP
@@ -444,7 +443,7 @@ func (c ipams) assignFromExistingBlock(
 // pool.  Returns a list of blocks that were claimed, as well as a
 // list of blocks that were claimed by another host.
 // If an empty string is passed as the host, then the value of os.Hostname is used.
-func (c ipams) ClaimAffinity(cidr net.IPNet, host string) ([]net.IPNet, []net.IPNet, error) {
+func (c ipamClient) ClaimAffinity(cidr net.IPNet, host string) ([]net.IPNet, []net.IPNet, error) {
 	// Validate that the given CIDR is at least as big as a block.
 	if !largerThanOrEqualToBlock(cidr) {
 		estr := fmt.Sprintf("The requested CIDR (%s) is smaller than the minimum.", cidr.String())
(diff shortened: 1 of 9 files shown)


lib/ipam/ipam.go
```diff
@@ -33,80 +35,18 @@ const (
...
-	// RemoveIPAMHost does not release any IP addresses claimed on the given host.
-	// If an empty string is passed as the host then the value returned by os.Hostname is used.
-	RemoveIPAMHost(host string) error
-}
-
-// newIPAM returns a new ipamClient, which implements the IPAMInterface
-func newIPAM(c *Client) *ipams {
-	return &ipams{c, blockReaderWriter{c}}
+// NewIPAM returns a new ipamClient, which implements the IPAMInterface.
+// Consumers of the Calico API should not create this directly, but should
```
gunjan5: > Consumers of the Calico API should not create this directly, but should access IPAM through the main client IPAM accessor.

Then who is this for?

robbrockbank: the v2 client

lib/ipam/ipam.go
```diff
@@ -33,80 +35,18 @@ const (
...
+// access IPAM through the main client IPAM accessor.
+func NewIPAM(client bapi.Client, pools PoolAccessorInterface) *ipams {
+	return &ipams{client: client, pools: pools,
+		blockReaderWriter: blockReaderWriter{client: client, pools: pools}}
 }
 
 // ipamClient implements the IPAMInterface
 type ipams struct {
```
gunjan5: can we rename this to something else? also the comment seems wrong

lib/ipam/ipam.go
```diff
@@ -218,7 +158,7 @@ func (c ipams) autoAssign(num int, handleID *string, attrs map[string]string, po
 		}
 
 		if retries == 0 {
-			return nil, goerrors.New("Max retries hit")
+			return nil, errors.New("Max retries hit")
```
gunjan5: can we add some more context into this error? Max retries of what

lib/ipam/ipam.go
```diff
@@ -296,21 +228,21 @@ func (c ipams) AssignIP(args AssignIPArgs) error {
...
 	if !c.blockReaderWriter.withinConfiguredPools(args.IP) {
-		return goerrors.New("The provided IP address is not in a configured pool\n")
+		return errors.New("The provided IP address is not in a configured pool\n")
 	}
 
 	blockCIDR := getBlockCIDRForAddress(args.IP)
 	log.Debugf("IP %s is in block '%s'", args.IP.String(), blockCIDR.String())
 	for i := 0; i < ipamEtcdRetries; i++ {
-		obj, err := c.client.Backend.Get(model.BlockKey{blockCIDR})
+		obj, err := c.client.Get(context.Background(), model.BlockKey{blockCIDR}, "")
```
gunjan5: should we expose the context to the client? so when a client calls `AssignIP` it can pass a context and cancel it if needed? WDYT?

lib/ipam/ipam.go
```diff
@@ -349,7 +281,7 @@ func (c ipams) AssignIP(args AssignIPArgs) error {
 		// Update the block using the original KVPair to do a CAS.  No need to
 		// update the Value since we have been manipulating the Value pointed to
 		// in the KVPair.
-		_, err = c.client.Backend.Update(obj)
+		_, err = c.client.Update(context.Background(), obj)
```
gunjan5: also we should probably have one context in this function even if we don't expose it to the caller of `Autoassign`

lib/ipam/ipam.go
robbrockbank: we can - but it's old code, it's an unusual error condition, it's in multiple places and I'm not sure it gains us anything useful.

lib/ipam/ipam.go
```diff
@@ -359,7 +291,7 @@ func (c ipams) AssignIP(args AssignIPArgs) error {
 		}
 		return nil
 	}
-	return goerrors.New("Max retries hit")
+	return errors.New("Max retries hit")
```
gunjan5: needs more context

lib/ipam/ipam.go
robbrockbank: We could - not sure anyone will use it though.

lib/ipam/ipam.go
robbrockbank: what do you mean?

lib/ipam/ipam.go
robbrockbank: why?

robbrockbank: see comment above.

lib/ipam/ipam.go
robbrockbank: Background() returns a fixed context, so I'm not sure it makes any difference does it?

lib/ipam/ipam.go
robbrockbank: i've beefed up the comment.

lib/ipam/ipam.go
robbrockbank: renamed to ipamClient as per the comment.

lib/ipam/ipam.go
```diff
@@ -923,20 +851,20 @@ func (c ipams) SetIPAMConfig(cfg IPAMConfig) error {
 	}
 
 	if !cfg.StrictAffinity && !cfg.AutoAllocateBlocks {
-		return goerrors.New("Cannot disable 'StrictAffinity' and 'AutoAllocateBlocks' at the same time")
+		return errors.New("Cannot disable 'StrictAffinity' and 'AutoAllocateBlocks' at the same time")
```
gunjan5: side question, are we planning to expose these config options?

robbrockbank: nothing in plan to at the moment.  it may be that we decide later on to revisit the IPAM architecture prior to GA - but nothing at the moment.

lib/ipam/ipam_block_reader_writer.go
```diff
@@ -75,41 +76,39 @@ func (rw blockReaderWriter) getAffineBlocks(host string, ver ipVersion, pools []
...
 func (rw blockReaderWriter) claimNewAffineBlock(host string, version ipVersion, requestedPools []cnet.IPNet, config IPAMConfig) (*cnet.IPNet, error) {
 
-	// If requestedPools is not empty, use it.  Otherwise, default to
-	// all configured pools.
+	// If requestedPools is not empty, use it.  Otherwise, default to all configured pools.
 	pools := []cnet.IPNet{}
 
 	// Get all the configured pools.
-	allPools, err := rw.client.IPPools().List(api.IPPoolMetadata{})
+	allPools, err := rw.pools.GetEnabledPools(version.Number)
```
gunjan5: shouldn't be called `allPools` anymore

lib/ipam/ipam_block_reader_writer.go
```diff
@@ -145,8 +144,10 @@ func isPoolInRequestedPools(pool cnet.IPNet, requestedPools []cnet.IPNet) bool {
 	if len(requestedPools) == 0 {
 		return true
 	}
+	// Compare the requested pools against the actual pool CIDR.  Note that we don't use deep equals
+	// because golang interchangeably seems to use 4-byte and 16-byte representations of IPv4 addresses.
```
gunjan5: this seems like a useful thing to know. Do you have a link or an issue you want to add to the comment so we have a reference for future?

lib/ipam/ipam_block_reader_writer.go
```diff
@@ -276,29 +273,27 @@ func (rw blockReaderWriter) releaseBlockAffinity(host string, blockCIDR cnet.IPN
...
-	return goerrors.New("Max retries hit")
+	return errors.New("Max retries hit")
 }
 
 // withinConfiguredPools returns true if the given IP is within a configured
 // Calico pool, and false otherwise.
 func (rw blockReaderWriter) withinConfiguredPools(ip cnet.IP) bool {
-	allPools, _ := rw.client.IPPools().List(api.IPPoolMetadata{})
-	for _, p := range allPools.Items {
+	allPools, _ := rw.pools.GetEnabledPools(ip.Version())
```
gunjan5: same here, it's not not allPools anymore

lib/ipam/ipam_block_reader_writer.go
robbrockbank: a golang issue?  nope - just something I've spotted in the past.  It's a bit yucky.

lib/ipam/ipam_test.go
```diff
@@ -12,88 +12,62 @@
...
 	cnet "github.com/projectcalico/libcalico-go/lib/net"
 	"github.com/projectcalico/libcalico-go/lib/testutils"
 )
 
+// Implement an IP pools accessor for the IPAM client.
+type ipPoolAccessor struct {
+	pools map[string]bool
+}
+
+func (i *ipPoolAccessor) GetEnabledPools(ipVersion int) ([]cnet.IPNet, error) {
```
gunjan5: why is this utility function in a test file?

lib/ipam/ipam_test.go
```diff
@@ -12,88 +12,62 @@
...
+	sorted := []string{}
+	// Get a sorted list of enabled pool CIDR strings.
+	for p, e := range i.pools {
+		if e {
+			sorted = append(sorted, p)
+		}
+	}
+	sort.Strings(sorted)
```
gunjan5: not sure I understand the need for sorting here. Can you explain?

lib/ipam/ipam_test.go
robbrockbank: It's not a utility function.  It's a mock version of an IP Pool Accessor.

lib/ipam/ipam_test.go
robbrockbank: Just to mimic what etcd will do and to minimize the changes required for the UTs (since they made the assumption that the pools were ordered).

robbrockbank: Oh yes - and to allow the tests to be deterministic.

robbrockbank: I'll add a comment.

lib/ipam/ipam_block_reader_writer.go
robbrockbank: kk, tweaking.

lib/ipam/ipam_block_reader_writer.go
robbrockbank: yup - updated.

lib/ipam/ipam_test.go
robbrockbank: added a comment to that effect.

lib/ipam/ipam_test.go
```diff
@@ -103,364 +77,323 @@ type testArgsClaimAff struct {
...
-		Context("AutoAssign 1 IP without specifying a pool", func() {
-			args := client.AutoAssignArgs{
-				Num4:     1,
-				Num6:     0,
-				Hostname: host,
-			}
+			It("should auto-assign from the only available pool", func() {
+				bc.Clean()
+				deleteAllPools()
+				applyPool("10.0.0.0/24", true)
```
gunjan5: how about having global strings `pool1Str := "10.0.0.0/24"` and one for pool2 so it's easy to correlate when reading the tests what's what, and it's also less error-prone

lib/ipam/ipam_test.go
```diff
@@ -103,364 +77,323 @@ type testArgsClaimAff struct {
...
-			})
+			It("should assign from a new pool for a new host (old pool is removed)", func() {
+				deleteAllPools()
+				applyPool("20.0.0.0/24", true)
+
+				p, _ := ipPools.GetEnabledPools(4)
+				Expect(len(p)).To(Equal(1))
+				Expect(p[0].String()).To(Equal(pool2.String()))
+				p, _ = ipPools.GetEnabledPools(6)
+				Expect(len(p)).To(Equal(0))
```
gunjan5: fyi: gomega also has a `BeZero()` function. Not saying you need to change this

lib/ipam/ipam_test.go
```diff
@@ -514,15 +461,53 @@ var _ = testutils.E2eDatastoreDescribe("IPAM tests", testutils.DatastoreEtcdV2,
...
+
+				// Re-initialize it to an empty slice to flush out any IP if passed in by mistake.
+				inIPs = []cnet.IP{}
+
+				inIPs = append(inIPs, cnet.IP{assignIP})
+
+			}
+
+			if autoAssignNumIPv4 != 0 {
+				assignedIPv4, _, _ := ic.AutoAssign(AutoAssignArgs{
```
gunjan5: why is the err ignored here?

robbrockbank: dunno - I just moved the code from a standalone function to inside here since it was only called from this one test.

I daresay we should assert the other things though, I'll add.

lib/ipam/ipam_test.go
robbrockbank: yeah that might be nice - but this is old code and I don't want the scope of this task to keep creeping out.

lib/ipam/ipam_test.go
robbrockbank: good to know.

gunjan5:
LGTM, S&M

PR #531, opened 2017-09-24
k8s conversion code handles 1.7 clusters

caseydavenport:
## Description

Fixes our conversion logic in KDD such that it works on clusters running k8s <=v1.7.0.

In this scenario, the API doesn't return a `policyTypes` for any NetworkPolicies and we need to thus infer that the policy should only affect ingress traffic since k8s pre-1.8 (and pre-policyTypes) didn't support egress policies.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/backend/k8s/conversion_test.go
+++ b/lib/backend/k8s/conversion_test.go
@@ -846,6 +846,421 @@ var _ = Describe("Test NetworkPolicy conversion", func() {
 	})
 })
 
+// This suite of tests is useful for ensuring we continue to support kubernetes apiserver
+// versions <= 1.7.x, and can be removed when that is no longer required.
+var _ = Describe("Test NetworkPolicy conversion (k8s <= 1.7, no policyTypes)", func() {
+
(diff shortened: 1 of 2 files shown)


lib/backend/k8s/conversion_test.go
```diff
@@ -846,6 +846,421 @@ var _ = Describe("Test NetworkPolicy conversion", func() {
 	})
 })
 
+// This suite of tests is useful for ensuring we continue to support kubernetes apiserver
+// versions <= 1.7.x, and can be removed when that is no longer required.
+var _ = Describe("Test NetworkPolicy conversion (k8s <= 1.7, no policyTypes)", func() {
```
caseydavenport: This is a copy of the NetworkPolicy tests from above, but each with the `policyTypes` section removed to make it look like it would appear on a k8s 1.7 cluster.

The assertions made in each test are the same.

PR #534, opened 2017-09-25
Fix and simplify watcher code

robbrockbank:
The watcher code was panicking occasionally due to sending events down the results channel after the channel was closed.

I've restructured the code to remove the separate error channel, and to basically have a single go routine for handling the watch.


lib/clientv2/prefixdeletewatch_e2e_test.go
```diff
@@ -1,84 +0,0 @@
-// Copyright (c) 2016-2017 Tigera, Inc. All rights reserved.
```
robbrockbank: This file is now watch_e2e_test.go.  The first test is tiger same prefix delete test, but the second test is a multi-threaded events/watch/cancel test.

lib/backend/etcdv3/watcher.go
```diff
@@ -109,36 +73,18 @@ func (wc *watcher) ResultChan() <-chan api.WatchEvent {
...
-	// event type to WatchAdded.
-	for _, kv := range list.KVPairs {
-		log.Info("Sending create events for each existing entry")
-		wc.sendEvent(&api.WatchEvent{
-			Type: api.WatchAdded,
-			New:  kv,
-		})
-	}
-	return nil
+// HasTerminated implements the api.WatchInterface.
```
gunjan5: HasTerminated itself doesn’t implement the WatchInterface

lib/backend/etcdv3/watcher.go
```diff
@@ -109,36 +73,18 @@ func (wc *watcher) ResultChan() <-chan api.WatchEvent {
...
+func (wc *watcher) HasTerminated() bool {
+	terminated := wc.terminated
```
gunjan5: cant you just do `return wc.terminated`?

lib/backend/etcdv3/watcher.go
```diff
@@ -45,10 +45,9 @@ func (c *etcdV3Client) Watch(cxt context.Context, l model.ListInterface, revisio
 		list:       l,
 		initialRev: rev,
 		resultChan: make(chan api.WatchEvent, outgoingBufSize),
-		errChan:    make(chan error, 1),
 	}
 	wc.ctx, wc.cancel = context.WithCancel(cxt)
-	go wc.run()
+	go wc.watchLoop()
```
gunjan5: rename the `outgoingBufSize` to something more relevant now that we don't have an incoming one. (adding the comment here because can't comment on the code that doesn't show up in the diff view)

lib/clientv2/watch_e2e_test.go
```diff
@@ -0,0 +1,182 @@
...
+				})
+			}
+
+			By("Listing all the resources")
+			outList, outError := c.BGPPeers().List(ctx, options.ListOptions{})
+			Expect(outError).NotTo(HaveOccurred())
+			Expect(outList.Items).To(HaveLen(numEvents))
+
+			By("Creating a watcher, watching the current resource version")
+			w, _ := c.BGPPeers().Watch(ctx, options.ListOptions{ResourceVersion: outList.ResourceVersion})
```
gunjan5: why are we ignoring this error? we should assert the error here

lib/clientv2/watch_e2e_test.go
```diff
@@ -0,0 +1,182 @@
...
+
+			By("Stopping the watcher and checking for full termination")
+			testWatcher.Stop()
+			Eventually(w.(*watcher).hasTerminated).Should(BeTrue())
+		})
+	})
+
+	Describe("Test constant stream of events whilst closing watcher", func() {
+		It("should handle gracefully closing watchers while events are occurring", func() {
+			c, err := New(config)
```
gunjan5: we could possibly put the client setup code in a `BeforeEach` block so you don't have to repeat it

lib/clientv2/watch_e2e_test.go
```diff
@@ -0,0 +1,182 @@
...
+				defer wg.Done()
+				for {
+					if _, err := c.BGPPeers().Create(ctx, &apiv2.BGPPeer{
+						ObjectMeta: metav1.ObjectMeta{Name: "name1"},
+						Spec: apiv2.BGPPeerSpec{
+							PeerIP:   "1.2.3.4",
+							ASNumber: numorstring.ASNumber(12345),
+						},
+					}, options.SetOptions{}); err != nil {
+						panic(err)
```
gunjan5: make this an assertion

lib/clientv2/watch_e2e_test.go
```diff
@@ -0,0 +1,182 @@
...
+					}
+					time.Sleep(5 * time.Millisecond)
+					if err := c.BGPPeers().Delete(ctx, "name1", options.DeleteOptions{}); err != nil {
+						panic(err)
```
gunjan5: same

lib/clientv2/watch_e2e_test.go
```diff
@@ -0,0 +1,182 @@
...
+					}
+					time.Sleep(5 * time.Millisecond)
```
gunjan5: is 5ms enough to check `finishctx.Err()`

lib/clientv2/watch_e2e_test.go
```diff
@@ -0,0 +1,182 @@
...
+
+					if finishctx.Err() != nil {
+						break
```
gunjan5: should we log before breaking out if err != nil?

lib/clientv2/watch_e2e_test.go
```diff
@@ -0,0 +1,182 @@
...
+				watchers[i], err = c.BGPPeers().Watch(ctx, options.ListOptions{})
+				Expect(err).NotTo(HaveOccurred())
+				defer watchers[i].Stop()
+
+				wg.Add(1)
+				go func(w watch.Interface) {
+					defer wg.Done()
+					for e := range w.ResultChan() {
+						if e.Type == watch.Deleted {
+							if e.Previous == nil {
```
gunjan5: Since this is a test a better thing to write would be `Expect(e.Previous).NotTo(BeNil(), "no previous value on deleted event")`, same for the following assertions

lib/backend/etcdv3/watcher.go
```diff
@@ -166,35 +112,82 @@ func (wc *watcher) watchLoop(watchClosedCh chan struct{}) {
...
+			Type: api.WatchAdded,
+			New:  kv,
+		})
+	}
+	return nil
+}
+
+// terminateWatcher terminates the resources associated with the watcher.
+func (wc *watcher) terminateWatcher() {
+	log.Debug("Terminate watcher")
```
gunjan5: shouldn't this be "terminating"?

lib/backend/etcdv3/watcher.go
```diff
@@ -166,35 +112,82 @@ func (wc *watcher) watchLoop(watchClosedCh chan struct{}) {
...
+	wc.terminated = true
 }
 
-// sendError sends an error down the errChan or waits for done notification.
+// sendError packages up the error as an event and sends it in the results channel.
 func (wc *watcher) sendError(err error) {
-	select {
-	case wc.errChan <- err:
-	case <-wc.ctx.Done():
+	// Skip cancelled errors propagating up from etcd.
```
gunjan5: can you add some more explanation in the comment here?

lib/backend/etcdv3/watcher.go
```diff
@@ -166,35 +112,82 @@ func (wc *watcher) watchLoop(watchClosedCh chan struct{}) {
...
+	if err == context.Canceled {
+		return
```
gunjan5: I personally prefer writing `return nil` to be more explicit

lib/backend/etcdv3/watcher.go
robbrockbank: ha!  yes.  I think I previously had a log in there too, and I want to make sure that the log and the return value were in-sync.

robbrockbank: I'm gonna change this to use an atomic operator.

lib/backend/etcdv3/watcher.go
robbrockbank: are you sure?  This method has no return type, so I don't think it is valid to return nil.

lib/clientv2/watch_e2e_test.go
robbrockbank: Not sure I follow.  The 5ms delay is purely so this doesn't tight-loop doing add/delete events.

lib/clientv2/watch_e2e_test.go
robbrockbank: We could - probably need to add logs elsewhere too then.

lib/clientv2/watch_e2e_test.go
robbrockbank: Hmm, I'll try, but I had a feeling that wouldn't work within a goroutine.

robbrockbank: Oh, it seems to work.  I'll update the ones above too then.

lib/clientv2/watch_e2e_test.go
robbrockbank: yeah we could, but meh.

lib/backend/etcdv3/watcher.go
gunjan5: o nvm I thought it was returning error

PR #537, opened 2017-09-26
Fix namespace handling and other minor cosmetic changes

robbrockbank:


robbrockbank:
Functional changes are in `lib/clientv2/resources.go`
The rest is just variable names and code comments

lib/clientv2/resources.go
```diff
@@ -256,27 +256,23 @@ func (c *resources) kvPairToResource(kvp *model.KVPair) resource {
...
+			in.GetObjectMeta().SetNamespace(defaultNamespace)
+		} else if in.GetObjectMeta().GetNamespace() == "" {
+			in.GetObjectMeta().SetNamespace(ns)
+		} else if in.GetObjectMeta().GetNamespace() != ns {
 			return cerrors.ErrorValidation{
 				ErroredFields: []cerrors.ErroredField{{
 					Name:   "Metadata.Namespace",
-					Reason: "Namespace should be specified",
+					Reason: "Namespace does not match client namespace",
 					Value:  in.GetObjectMeta().GetNamespace(),
```
gunjan5: is there a way to specify the client namespace in the error value? like maybe `fmt.Sprintf("%s != %s",  in.GetObjectMeta().GetNamespace(), ns)`?

robbrockbank: Hmmm, it's possible, but only by including the name in the error which is probably a move away from how the validation code is likely to change.

I've added sub task EPB-443 to the validation story to pick this up when we handle validation.

PR #542, opened 2017-09-26
Code markups to support client requirements

robbrockbank:
This consists of:
-  Removing the handling of the default namespace - this should be handled by the client
-  Removing the namespace from the client constructor and instead adding to:  the Get/Delete methods on namespaced resources, and the ListWatchOptions struct.
-  Returning the deleted resource as part of the Delete request
-  Allowing List and Watch to list/watch a specific resource.

### DIFF
--- a/lib/apiv2/workloadendpoint.go
+++ b/lib/apiv2/workloadendpoint.go
@@ -42,7 +42,9 @@ type WorkloadEndpointSpec struct {
 	// The container ID.
 	ContainerID string `json:"containerID,omitempty" validate:"omitempty,name"`
 	// The Pod name.
-	Pod string `json:"podID,omitempty" validate:"omitempty,name"`
+	Pod string `json:"pod,omitempty" validate:"omitempty,name"`
+	// The Endpoint name.
+	Endpoint string `json:"endpoint,omitempty" validate:"omitempty,name"`
 	// IPNetworks is a list of subnets allocated to this endpoint. IP packets will only be
 	// allowed to leave this interface if they come from an address in one of these subnets.
 	// Currently only /32 for IPv4 and /128 for IPv6 networks are supported.
--- a/lib/backend/api/api.go
+++ b/lib/backend/api/api.go
@@ -85,7 +85,7 @@ type Client interface {
 	// also be removed when deleting the objects that implicitly created it.
 	// For example, deleting the last WorkloadEndpoint in a Workload will
 	// also remove the Workload.
-	Delete(ctx context.Context, key model.Key, revision string) error
+	Delete(ctx context.Context, key model.Key, revision string) (*model.KVPair, error)
 
 	// Get returns the object identified by the given key as a KVPair with
 	// revision information.
--- a/lib/backend/compat/compat.go
+++ b/lib/backend/compat/compat.go
@@ -197,22 +197,22 @@ func (c *ModelAdaptor) Apply(d *model.KVPair) (*model.KVPair, error) {
 }
 
 // Delete an entry in the datastore.  This errors if the entry does not exists.
-func (c *ModelAdaptor) Delete(ctx context.Context, k model.Key, rev string) error {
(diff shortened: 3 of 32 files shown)


lib/backend/api/api.go
```diff
@@ -85,7 +85,7 @@ type Client interface {
 	// also be removed when deleting the objects that implicitly created it.
 	// For example, deleting the last WorkloadEndpoint in a Workload will
 	// also remove the Workload.
-	Delete(ctx context.Context, key model.Key, revision string) error
+	Delete(ctx context.Context, key model.Key, revision string) (*model.KVPair, error)
```
gunjan5: I'm guessing this is because of the API server?

lib/backend/compat/compat.go
```diff
@@ -197,22 +197,22 @@ func (c *ModelAdaptor) Apply(d *model.KVPair) (*model.KVPair, error) {
...
-			return err
+			return nil, err
 		}
-		if err = c.client.Delete(ctx, p.Key, rev); err != nil {
-			return err
+		if _, err = c.client.Delete(ctx, p.Key, rev); err != nil {
+			return nil, err
 		}
-		return nil
+		return nil, nil
```
gunjan5: shouldn't you be returning `KVPair` and `nil` here?

lib/clientv2/networkpolicy.go
```diff
@@ -51,23 +50,26 @@ func (r networkPolicies) Create(ctx context.Context, res *apiv2.NetworkPolicy, o
...
 		return out.(*apiv2.NetworkPolicy), err
 	}
 	return nil, err
 }
 
 // Delete takes name of the NetworkPolicy and deletes it. Returns an error if one occurs.
-func (r networkPolicies) Delete(ctx context.Context, name string, opts options.DeleteOptions) error {
-	err := r.client.resources.Delete(ctx, opts, apiv2.KindNetworkPolicy, r.namespace, name)
-	return err
+func (r networkPolicies) Delete(ctx context.Context, namespace, name string, opts options.DeleteOptions) (*apiv2.NetworkPolicy, error) {
```
gunjan5: thinking out loud, what if we just have the clients make sure namespace is set in the resource before they make a Get or Delete call, so we can just extract the namespace from ` r.namespace`? This way we don't have to change the function signature for namespaced resources

robbrockbank: there isn't a resource on a Get or Delete call.

lib/apiv2/workloadendpoint.go
```diff
@@ -42,7 +42,9 @@ type WorkloadEndpointSpec struct {
 	// The container ID.
 	ContainerID string `json:"containerID,omitempty" validate:"omitempty,name"`
 	// The Pod name.
-	Pod string `json:"podID,omitempty" validate:"omitempty,name"`
+	Pod string `json:"pod,omitempty" validate:"omitempty,name"`
+	// The Endpoint name.
```
gunjan5: can you add one more sentence in the comment explaining this is the container side of the veth pair interface?

robbrockbank: it doesn't have to be.

robbrockbank: in libnetwork it's an arbitrary ID.

lib/backend/compat/compat.go
robbrockbank: yeah - but we aren't using the compat module anymore.   I'm keeping it around in case I copy some of the code, at the moment I just want it to compile.

lib/backend/api/api.go
robbrockbank: Yup.

lib/clientv2/networkpolicy.go
gunjan5: Doh, yes

gunjan5:
Ok lgtm, feel free to merge so I can pull it in and update my PR

lib/apiv2/workloadendpoint.go
gunjan5: K

PR #544, opened 2017-09-28
WorkloadEndpoint name construction helper

robbrockbank:
This will be useful for the CNI plugin integration and other integrations in the future.

This contains methods to construct the WEP name from a set of Identifiers, and to compare a WEP name against the set of identifiers (which will be required if you are doing a prefix list).

### DIFF
--- a/lib/names/workloadendpoint.go
+++ b/lib/names/workloadendpoint.go
@@ -1,0 +1,224 @@
+// Copyright (c) 2017 Tigera, Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package names
+
+import (
+	"strings"
+
+	cerrors "github.com/projectcalico/libcalico-go/lib/errors"
+)
+
+// WorkloadEndpointIdentifiers is a collection of identifiers that are used to uniquely
+// identify a WorkloadEndpoint resource.  Since a resource is identified by a single
+// name field, Calico requires the name to be constructed in a very specific format.
+// The format is dependent on the Orchestrator type:
+// -  k8s:  <node>-k8s-<pod>-<endpoint>
+// -  cni:  <node>-cni-<containerID>-<endpoint>
+// -  libnetwork:  <node>-libnetwork-libnetwork-<endpoint>
+// -  (other):  <node>-<orchestrator>-<workload>-<endpoint>
+//
+// Each parameter cannot start or end with a dash (-), and dashes within the parameter
+// will be escaped to a double-dash (--) in the constructed name.
+//
+// List queries allow for prefix lists (for non-KDD), the client should verify that
+// then items returned in the list match the supplied identifiers using the
+// NameMatches() method.  This is necessary because a prefix match may return endpoints
+// that do not exactly match the required identifiers.  For example , suppose you are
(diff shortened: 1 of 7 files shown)


lib/names/workloadendpoint_test.go
```diff
@@ -0,0 +1,200 @@
...
+package names_test
+
+import (
+	. "github.com/onsi/ginkgo/extensions/table"
+	. "github.com/onsi/gomega"
+	cerrors "github.com/projectcalico/libcalico-go/lib/errors"
+	"github.com/projectcalico/libcalico-go/lib/names"
+)
+
+func init() {
```
gunjan5: why is this test in an `init()` function?

lib/names/workloadendpoint_test.go
```diff
@@ -0,0 +1,200 @@
...
+	DescribeTable("WorkloadEndpoint name construction, fully qualified names",
+		func(ids names.WorkloadEndpointIdentifiers, name string, erroredField string) {
```
gunjan5: nit: for table driven test, it's nice to have expected output variables named something like `expectedName` so it's easy to understand when someone is reading the test

lib/names/workloadendpoint.go
```diff
@@ -0,0 +1,217 @@
...
+
+	// Node and Orchestrator are *always* required.
+	if len(node.value) == 0 {
+		return nil, cerrors.ErrorValidation{
+			ErroredFields: []cerrors.ErroredField{
+				{Name: node.field, Reason: "field should be assigned"},
+			},
+		}
+	}
+	if len(orch.value) == 0 {
```
gunjan5: shouldn't we also be making sure that the `endpoint` is specified since that's also required for all the orchestrators?

lib/names/workloadendpoint_test.go
```diff
@@ -0,0 +1,200 @@
...
+			Node:         "node-1",
+			Orchestrator: "k8s",
+			Pod:          "pod-name",
+			Endpoint:     "eth0",
+		}, "node--1-k8s-pod--name-eth", false, ""),
+		Entry("CNI endpoint missing container ID and matching name", names.WorkloadEndpointIdentifiers{
+			Node:         "node-1",
+			Orchestrator: "cni",
+			Endpoint:     "eth0",
+		}, "node--1-cni-orchestratorid-eth0", true, ""),
```
gunjan5: not sure I understand where this `orchestratorid` string comes from? How is that being placed in place of `ContainerID`?

lib/names/workloadendpoint.go
robbrockbank: nope - it's the last part in the name construction, so for prefixes it won't be specified.  For non-prefixes it'll be caught when we construct the name from the segments, because that does a final validation of all parts.

lib/names/workloadendpoint_test.go
robbrockbank: hmm, i think i've seen that as an alternative to  _ = Describe()

robbrockbank: but actually it just makes everything more indented.  I'll change

lib/names/workloadendpoint_test.go
robbrockbank: fixed.

lib/names/workloadendpoint_test.go
robbrockbank: the key thing is that the IDs don't include the ContainerID - so it can be regarded as a wildcard match.  So the name does match that.

robbrockbank: workflow is:

-  create a WEPIdentifiers with the subset of ids
-  construct the name prefix
-  do a prefix list
-  parse the items in the list and check the name matches the IDs

this final step is necessary because it's possible to get a prefix match that doesn't actually match the required identifiers.  For example (and this would be a real scenario for CNI), suppose you are querying endpoints with node=node1, orch=k8s, pod=pod and endpoints is wild carded.

The name prefix would be   `node1-k8s-pod-`

but that could also return endpoints with a pod call "pod-1", because the name of that endpoints would be `node1-k8s-pod--1-eth0` which matches the prefix.

robbrockbank: i've beefed up the comment for the WorkloadEndpointIdentifiers doc.

PR #550, opened 2017-10-03
Add ability to List using a name prefix.

robbrockbank:
PR adds the ability to perform a List using a name prefix when using etcdv3.

Required for the ability to perform WEP lists per workload.

### DIFF
--- a/lib/backend/model/keys.go
+++ b/lib/backend/model/keys.go
@@ -173,6 +173,14 @@ func ListOptionsIsFullyQualified(listOptions ListInterface) bool {
 	return listOptions.KeyFromDefaultPath(listOptions.defaultPathRoot()) != nil
 }
 
+// IsListOptionsLastSegmentPrefix returns true if the final segment of the default path
+// root is a name prefix rather than the full name.
+func IsListOptionsLastSegmentPrefix(listOptions ListInterface) bool {
+	// Only supported for ResourceListOptions.
+	rl, ok := listOptions.(ResourceListOptions)
+	return ok && rl.IsLastSegmentIsPrefix()
+}
+
 // KeyFromDefaultPath parses the default path representation of a key into one
 // of our <Type>Key structs.  Returns nil if the string doesn't match one of
 // our key types.
--- a/lib/backend/model/resource.go
+++ b/lib/backend/model/resource.go
@@ -87,6 +87,17 @@ type ResourceListOptions struct {
 	Namespace string
 	// The resource kind.
 	Kind string
+	// Whether the name is prefix rather than the full name.
+	Prefix bool
+}
+
+// If the Kind, Namespace and Name are specified, but the Name is a prefix then the
+// last segment of this path is a prefix.
+func (options ResourceListOptions) IsLastSegmentIsPrefix() bool {
+	return len(options.Kind) != 0 &&
+		(len(options.Namespace) != 0 || !namespace.IsNamespaced(options.Kind)) &&
+		len(options.Name) != 0 &&
+		options.Prefix
 }
 
 func (options ResourceListOptions) KeyFromDefaultPath(path string) Key {
@@ -115,9 +126,14 @@ func (options ResourceListOptions) KeyFromDefaultPath(path string) Key {
 			log.Debugf("Didn't match namespace %s != %s", options.Namespace, namespace)
 			return nil
 		}
-		if len(options.Name) != 0 && name != options.Name {
-			log.Debugf("Didn't match name %s != %s", options.Name, name)
-			return nil
+		if len(options.Name) != 0 {
+			if options.Prefix && !strings.HasPrefix(name, options.Name) {
+				log.Debugf("Didn't match name prefix %s != prefix(%s)", options.Name, name)
+				return nil
+			} else if !options.Prefix && name != options.Name {
+				log.Debugf("Didn't match name %s != %s", options.Name, name)
+				return nil
+			}
 		}
 		return ResourceKey{Kind: options.Kind, Namespace: namespace, Name: name}
 	}
@@ -134,9 +150,14 @@ func (options ResourceListOptions) KeyFromDefaultPath(path string) Key {
 		log.Debugf("Didn't match name %s != %s", options.Kind, kind)
 		return nil
 	}
-	if len(options.Name) != 0 && name != options.Name {
-		log.Debugf("Didn't match name %s != %s", options.Name, name)
-		return nil
+	if len(options.Name) != 0 {
+		if options.Prefix && !strings.HasPrefix(name, options.Name) {
+			log.Debugf("Didn't match name prefix %s != prefix(%s)", options.Name, name)
+			return nil
+		} else if !options.Prefix && name != options.Name {
+			log.Debugf("Didn't match name %s != %s", options.Name, name)
+			return nil
+		}
 	}
 	return ResourceKey{Kind: options.Kind, Name: name}
 }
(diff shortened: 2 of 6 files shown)


lib/clientv2/workloadendpoint_e2e_test.go
```diff
@@ -38,21 +38,34 @@ var _ = testutils.E2eDatastoreDescribe("WorkloadEndpoint tests", testutils.Datas
 	ctx := context.Background()
 	namespace1 := "namespace-1"
 	namespace2 := "namespace-2"
-	name1 := "wep-1"
-	name2 := "wep-2"
-	spec1 := apiv2.WorkloadEndpointSpec{
+	name1 := "node--1-k8s-abcdef-eth0"
```
robbrockbank: btw, I changed these as a future fix for when we add validation - since the name and the spec are inextricably linked I couldn't just swap the same Spec between different resource instances.  didn't need to be done now, but there you go.

lib/clientv2/workloadendpoint_e2e_test.go
```diff
@@ -397,4 +410,106 @@ var _ = testutils.E2eDatastoreDescribe("WorkloadEndpoint tests", testutils.Datas
...
+			outList, err := c.WorkloadEndpoints().List(ctx, options.ListOptions{Namespace: "namespace1", Name: "node--1-k8s-pod-eth0"})
+			Expect(err).NotTo(HaveOccurred())
+			Expect(outList.Items).To(HaveLen(1))
+			testutils.ExpectResource(&outList.Items[0], apiv2.KindWorkloadEndpoint, "namespace1", "node--1-k8s-pod-eth0", outRes1.Spec)
+
+			By("Doing a short prefix get to retrieve both workload endpoints in namespace1")
+			outList, err = c.WorkloadEndpoints().List(ctx, options.ListOptions{Namespace: "namespace1", Name: "node--1-k8s-pod-", Prefix: true})
+			Expect(err).NotTo(HaveOccurred())
+			Expect(outList.Items).To(HaveLen(2))
+			testutils.ExpectResource(&outList.Items[0], apiv2.KindWorkloadEndpoint, "namespace1", "node--1-k8s-pod--1-eth0", outRes2.Spec)
```
gunjan5: is it always guaranteed to get the last one we put first in the list? Will this also be true for KDD?

lib/clientv2/workloadendpoint_e2e_test.go
```diff
@@ -397,4 +410,106 @@ var _ = testutils.E2eDatastoreDescribe("WorkloadEndpoint tests", testutils.Datas
...
+
+			By("Doing a short prefix get with wildcarded namespace to retrieve all workload endpoints")
+			outList, err = c.WorkloadEndpoints().List(ctx, options.ListOptions{Name: "node--1-k8s-pod-", Prefix: true})
+			Expect(err).NotTo(HaveOccurred())
+			Expect(outList.Items).To(HaveLen(3))
+			testutils.ExpectResource(&outList.Items[0], apiv2.KindWorkloadEndpoint, "namespace1", "node--1-k8s-pod--1-eth0", outRes2.Spec)
+			testutils.ExpectResource(&outList.Items[1], apiv2.KindWorkloadEndpoint, "namespace1", "node--1-k8s-pod-eth0", outRes1.Spec)
+			testutils.ExpectResource(&outList.Items[2], apiv2.KindWorkloadEndpoint, "namespace2", "node--1-k8s-pod--2-eth0", outRes3.Spec)
+
+			By("Doing a long prefix get with wildcarded names to retrieve the workload endpoint in namespace2")
```
gunjan5: not sure what this test is buying us. Not too fussed about it tho

lib/clientv2/workloadendpoint_e2e_test.go
```diff
@@ -397,4 +410,106 @@ var _ = testutils.E2eDatastoreDescribe("WorkloadEndpoint tests", testutils.Datas
...
+			outList, err = c.WorkloadEndpoints().List(ctx, options.ListOptions{Name: "node--1-k8s-pod--2", Prefix: true})
+			Expect(err).NotTo(HaveOccurred())
+			Expect(outList.Items).To(HaveLen(1))
+			testutils.ExpectResource(&outList.Items[0], apiv2.KindWorkloadEndpoint, "namespace2", "node--1-k8s-pod--2-eth0", outRes3.Spec)
+
+			By("Deleting all endpoints")
+			_, err = c.WorkloadEndpoints().Delete(ctx, outRes1.Namespace, outRes1.Name, options.DeleteOptions{})
```
gunjan5: Can't we use the result of a wildcard list to loop through and delete the WEPs?

lib/backend/model/resource.go
```diff
@@ -87,6 +87,17 @@ type ResourceListOptions struct {
...
 	Kind string
+	// Whether the name is prefix rather than the full name.
+	Prefix bool
+}
+
+// If the Kind, Namespace and Name are specified, but the Name is a prefix then the
+// last segment of this path is a prefix.
+func (options ResourceListOptions) ListOptionsLastSegmentIsPrefix() bool {
+	return len(options.Kind) != 0 &&
+		(len(options.Namespace) != 0 || !namespace.IsNamespaced(options.Kind)) &&
```
gunjan5: I wonder if we can add some validation (doesn't have to be now), to make sure this is only for WEPs, NPs and GNP kinds? or is that too much? WDYT?

lib/backend/model/keys.go
```diff
@@ -173,6 +173,14 @@ func ListOptionsIsFullyQualified(listOptions ListInterface) bool {
 	return listOptions.KeyFromDefaultPath(listOptions.defaultPathRoot()) != nil
 }
 
+// ListOptionsLastSegmentIsPrefix returns true if the final segment of the default path
+// root is a name prefix rather than the full name.
+func ListOptionsLastSegmentIsPrefix(listOptions ListInterface) bool {
```
gunjan5: don't care too much about it but, I personally prefer func/methods that check something and return a bool names something like `IsLastSegmentPrefixListOptions`.

lib/clientv2/workloadendpoint_e2e_test.go
robbrockbank: These test won't be valid for KDD since WEPs aren't createable.  I'll update to be etcdv3 only/

lib/backend/model/resource.go
robbrockbank: For etcdv3 this function will be completely generic, so we don't need to limit for certain resource types.

lib/clientv2/workloadendpoint_e2e_test.go
robbrockbank: it's just testing that the filtering is working correctly when namespace is wild carded.

lib/clientv2/workloadendpoint_e2e_test.go
robbrockbank: We could - but I've kinda done it now and it doesn't buy us anything extra.  I'll leave unless you feel strongly.

lib/backend/model/resource.go
robbrockbank: Actually thinking about it, for O.S. in KDD, only WEP prefix is supported.  I'll update the comment in the ListOptions.

lib/clientv2/workloadendpoint_e2e_test.go
robbrockbank: The results will be returned by etcdv3 in alphanumeric order.

PR #551, opened 2017-10-04
Implemented interface definitions for BGP Config resource

mgleung:
## Description
Adding the BGPConfiguration resource as well as the functionality and tests for CRUD operations.

## Todos
- [x] Tests
- [x] Documentation (none required here)
- [x] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/apiv2/bgpconfig.go
+++ b/lib/apiv2/bgpconfig.go
@@ -1,0 +1,73 @@
+// Copyright (c) 2017 Tigera, Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package apiv2
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	//"k8s.io/apimachinery/pkg/runtime/schema"
+
+	"github.com/projectcalico/libcalico-go/lib/numorstring"
+)
+
+const (
+	KindBGPConfiguration     = "BGPConfiguration"
+	KindBGPConfigurationList = "BGPConfigurationList"
+)
+
+// BGPConfiguration contains the configuration for any BGP routing.
+type BGPConfiguration struct {
+	metav1.TypeMeta `json:",inline"`
+	// Standard object's metadata.
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+	// Specification of the BGPConfiguration.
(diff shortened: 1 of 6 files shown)


lib/apiv2/bgpconfig.go
```diff
@@ -0,0 +1,73 @@
...
+	// Standard object's metadata.
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+	// Specification of the BGPConfiguration.
+	Spec BGPConfigurationSpec `json:"spec,omitempty"`
+}
+
+// TODO: Add validation on LogSeverityScreen for valid values
+// BGPConfigurationSpec contains the values of the BGP configuration.
+type BGPConfigurationSpec struct {
+	LogSeverityScreen     string                `json:"log_severity_screen,omitempty" validate:"omitempty"`
```
robbrockbank: following k8s / go standards we're not using _ delimiters - use camel case, with lower case first letter lowercase.

so, `log_severity_screen` -> `logSeverityScreen`

lib/clientv2/bgpconfig.go
```diff
@@ -0,0 +1,107 @@
...
+// bgpConfigurations implements BGPConfigurationInterface
+type bgpConfigurations struct {
+	client client
+}
+
+// Create takes the representation of a BGpConfiguration and creates it.
+// Returns the stored representation of the BGPConfiguration, and an error
+// if there is any.
+func (r bgpConfigurations) Create(ctx context.Context, res *apiv2.BGPConfiguration, opts options.SetOptions) (*apiv2.BGPConfiguration, error) {
+	if res.ObjectMeta.GetName() != "default" && (res.Spec.NodeToNodeMeshEnabled != nil || res.Spec.DefaultNodeASNumber != nil) {
```
robbrockbank: I realize this is a simple bit of code, but it is duplicated for Create and Update.  Could you pull out into a separate validation fn?

lib/clientv2/bgpconfig.go
```diff
@@ -0,0 +1,107 @@
...
+		return nil, errors.New("Cannot set spec fields node_to_node_mesh_enabled and default_node_as_number on a non default BGP Configuration.")
```
robbrockbank: You should return a validation error (defined in the errors package).

lib/clientv2/bgpconfig.go
```diff
@@ -0,0 +1,145 @@
...
+					Reason: "Cannot set defaultNodeASNumber on a non default BGP Configuration.",
+				})
+				err = tempErr
+			} else {
+				err = cerrors.ErrorValidation{
+					ErroredFields: []cerrors.ErroredField{{
+						Name:   "BGPConfiguration.Spec.DefaultNodeASNumber",
+						Reason: "Cannot set defaultNodeASNumber on a non default BGP Configuration.",
+					}},
+				}
```
robbrockbank: So I think this workflow would have been a little easier if you'd done something like:

```
erroredFields := []errors.ErroredField{}
if meshSet {
    erroredFields = append(erroredFields, errors.ErroredField{meshField})
}
if asNumberSet {
    erroredFields = append(erroredFields, errors.ErroredField{asNumberField})
}
if len(erroredFields) > 0 {
    return ErrorValidation{ErroredFields: erroredFields}
}
```

PR #552, opened 2017-10-05
[WIP] Add ability to sync using multiple watchers

robbrockbank:
PLEASE DON'T REVIEW JUST YET.

This adds a generic "watcherSyncer" (not sure I like that name... open to suggestions) that implements the Syncer interface required by Felix, syncing the data from multiple arbitrary Watchers.

I'm going to use this for the etcdv3 syncer, and once we have Watch support covering all of the necessary K8s resources, we could use this to replace the K8s syncer code too.

For Felix, the current plan is to sync v1 data.  This generic helper class allows the user to specify an arbitrary converter interface to convert between the raw set of updates and the ones that we actually send to the consumer.

### DIFF
--- a/lib/backend/watchersyncer/watchercache.go
+++ b/lib/backend/watchersyncer/watchercache.go
@@ -1,0 +1,285 @@
+// Copyright (c) 2017 Tigera, Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package watchersyncer
+
+import (
+	"context"
+	"time"
+
+	"github.com/sirupsen/logrus"
+
+	"github.com/projectcalico/libcalico-go/lib/backend/api"
+	"github.com/projectcalico/libcalico-go/lib/backend/model"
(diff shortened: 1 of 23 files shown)


lib/backend/watchersyncer/watchercache.go
```diff
@@ -0,0 +1,285 @@
...
+}
+
+// Create a new watcherCache.
+// Note that the results channel is untyped.  The watcherSyncer expects one of the following
+// types:
+// -  An error
+// -  An api.Update
+// -  A api.SyncStatus (only for the very first InSync notification)
+func newWatcherCache(client api.Client, resourceType ResourceType, results chan<- interface{}) *watcherCache {
+	wc := &watcherCache{
```
gunjan5: cant you just return this instead of storing it then return?

robbrockbank: yup

lib/backend/watchersyncer/watchercache.go
```diff
@@ -0,0 +1,285 @@
...
+	}
+	return wc
+}
+
+// run creates the watcher and loops indefinitely reading from the watcher.
+func (wc *watcherCache) run() {
+	wc.createWatcher()
+	for {
+		select {
+		case event := <-wc.watch.ResultChan():
```
gunjan5: I think you should be able to simplify:
```go
for{
  select{
      case event := <-wc.watch.ResultChan():
          switch event.Type {
               api.WatchSynced:
 ...
``` 
to 
```go
for{
  switch <-wc.watch.ResultChan().Type {
       case api.WatchSynced:
...
```
because `select` `case` is usually if you're waiting on multiple channels, in this case we just have one channel we're waiting on. I tried this on a simple program, and it seemed to work.

robbrockbank: I don't think that's quite right, but you are correct that I could drop the select.

I think it would be more like the following, because I need access to event, not just event.Type:

```
event := <=wc.watch.ResultChan()
switch event.Type {
...
```

robbrockbank:
Replaced by #560.

PR #553, opened 2017-10-05
Implemented interface definitions for the Felix Configuration resource

mgleung:
## Description
Adding the FelixConfiguration resource as well as the functionality and tests for CRUD operations.

## Todos
- [x] Tests
- [x] Documentation (not needed here)
- [x] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/apiv2/felixconfig.go
+++ b/lib/apiv2/felixconfig.go
@@ -1,0 +1,219 @@
+// Copyright (c) 2017 Tigera, Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package apiv2
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	//"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+const (
+	KindFelixConfiguration     = "FelixConfiguration"
+	KindFelixConfigurationList = "FelixConfigurationList"
+)
+
+// Felix Configuration contains the configuration for Felix.
+type FelixConfiguration struct {
+	metav1.TypeMeta `json:",inline"`
+	// Standard object's metadata.
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+	// Specification of the FelixConfiguration.
+	Spec FelixConfigurationSpec `json:"spec,omitempty"`
+}
+
+// TODO: Add validation on LogSeverityScreen, LogSeveritySys, LogSeverityFile, DatastoreType,
+// DefaultEndpointToHostAction, IptablesFilterAllowAction, IptablesMangleAllowAction,
+// ChainInsertModefor valid values.
+// FelixConfigurationSpec contains the values of the Felix configuration.
+type FelixConfigurationSpec struct {
+	UseInternalDataplaneDriver *bool  `json:"useInternalDataplaneDriver,omitempty"`
+	DataplaneDriver            string `json:"dataplaneDriver,omitempty"`
+
+	Ipv6Support    *bool `json:"ipv6Support,omitempty"`
+	IgnoreLooseRPF *bool `json:"ignoreLooseRPF,omitempty"`
+
+	// RouterefreshInterval is the period, in seconds, at which Felix re-checks the routes
+	// in the dataplane to ensure that no other process has accidentally broken Calico’s rules.
+	// Set to 0 to disable route refresh. [Default: 90]
+	RouteRefreshIntervalSecs *int `json:"routeRefreshIntervalSecs,omitempty" configname:"RouteRefreshInterval"`
+	// IptablesRefreshInterval is the period, in seconds, at which Felix re-checks the IP sets
+	// in the dataplane to ensure that no other process has accidentally broken Calico’s rules.
+	// Set to 0 to disable IP sets refresh. Note: the default for this value is lower than the
+	// other refresh intervals as a workaround for a Linux kernel bug that was fixed in kernel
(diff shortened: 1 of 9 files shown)


lib/apiv2/felixconfig.go
```diff
@@ -0,0 +1,152 @@
...
+	PrometheusMetricsEnabled        *bool `json:"prometheusMetricsEnabled,omitempty"`
+	PrometheusMetricsPort           *int  `json:"prometheusMetricsPort,omitempty"`
+	PrometheusGoMetricsEnabled      *bool `json:"prometheusGoMetricsEnabled,omitempty"`
+	PrometheusProcessMetricsEnabled *bool `json:"prometheusProcessMetricsEnabled,omitempty"`
+
+	FailsafeInboundHostPorts  *[]ProtoPort `json:"failsafeInboundHostPorts,omitempty"`
+	FailsafeOutboundHostPorts *[]ProtoPort `json:"failsafeOutboundHostPorts,omitempty"`
+
+	UsageReportingEnabled *bool `json:"usageReportingEnabled,omitempty"`
+	// TODO: Check if CalicoVersion is important
```
robbrockbank: Good spot - nope that needs to be moved into ClusterInfo

lib/apiv2/felixconfig.go
```diff
@@ -0,0 +1,152 @@
...
+// DefaultEndpointToHostAction, IptablesFilterAllowAction, IptablesMangleAllowAction,
+// ChainInsertModefor valid values.
+// FelixConfigurationSpec contains the values of the Felix configuration.
+type FelixConfigurationSpec struct {
+	UseInternalDataplaneDriver *bool  `json:"useInternalDataplaneDriver,omitempty"`
+	DataplaneDriver            string `json:"dataplaneDriver,omitempty"`
+
+	DatastoreType string `json:"datastoreType,omitempty"`
+
+	TyphaAddr           string         `json:"typhaAddr,omitempty"`
```
robbrockbank: Ah, yes we should remove the Type settings too.

lib/apiv2/felixconfig.go
```diff
@@ -0,0 +1,152 @@
...
+
+// TODO: Add validation on LogSeverityScreen, LogSeveritySys, LogSeverityFile, DatastoreType,
+// DefaultEndpointToHostAction, IptablesFilterAllowAction, IptablesMangleAllowAction,
+// ChainInsertModefor valid values.
+// FelixConfigurationSpec contains the values of the Felix configuration.
+type FelixConfigurationSpec struct {
+	UseInternalDataplaneDriver *bool  `json:"useInternalDataplaneDriver,omitempty"`
+	DataplaneDriver            string `json:"dataplaneDriver,omitempty"`
+
+	DatastoreType string `json:"datastoreType,omitempty"`
```
robbrockbank: DatastoreType can be removed.

lib/apiv2/felixconfig.go
```diff
@@ -0,0 +1,152 @@
...
+	TyphaAddr           string         `json:"typhaAddr,omitempty"`
+	TyphaK8sServiceName string         `json:"typhaK8sServiceName,omitempty"`
+	TyphaK8sNamespace   string         `json:"typhaK8sNamespace,omitempty"`
+	TyphaReadTimeout    *time.Duration `json:"typhaReadTimeout"`
+	TyphaWriteTimeout   *time.Duration `json:"typhaWriteTimeout"`
+
+	Ipv6Support    *bool `json:"ipv6Support,omitempty"`
+	IgnoreLooseRPF *bool `json:"ignoreLooseRPF,omitempty"`
+
+	RouteRefreshInterval               *time.Duration `json:"routeRefreshInterval,omitempty"`
```
robbrockbank: Sorry to expand the scope here, but could you do the following for all of the duration/interval fields:
-  Convert from *time.Duration to *int
-  Make sure each field has the correct Suffix (which I think will be "Secs", but could you cross reference with: https://docs.projectcalico.org/v2.6/reference/felix/configuration to make sure).
-  For every fields where you are modifying the field name from the original, could you add a configname tag that specifies the original field name.   e.g

```
	RouteRefreshIntervalSecs           *int `json:"routeRefreshInterval,omitempty" configname:"RouteRefreshInterval"`
```

lib/apiv2/felixconfig.go
```diff
@@ -0,0 +1,152 @@
...
+	// TODO: Check if CalicoVersion is important
+	CalicoVersion string `json:"calicoVersion,omitempty"`
+
+	DebugMemoryProfilePath          string         `json:"debugMemoryProfilePath,omitempty"`
+	DebugDisableLogDropping         *bool          `json:"debugDisableLogDropping,omitempty"`
+	DebugSimulateCalcGraphHangAfter *time.Duration `json:"debugSimulateCalcGraphHangAfter,omitempty"`
+	DebugSimulateDataplaneHangAfter *time.Duration `json:"debugSimulateDataplaneHangafter,omitempty"`
+}
+
+// Define the ProtoPort here. Cannot use the equivalent ProtoPort in felix since
```
robbrockbank: Don't think you need the circular dependencies comment - generally it doesn't make sense for anything in libcalico-go to import things from felix because felix should be importing from here.

lib/apiv2/felixconfig.go
```diff
@@ -0,0 +1,219 @@
...
+	UseInternalDataplaneDriver *bool  `json:"useInternalDataplaneDriver,omitempty"`
+	DataplaneDriver            string `json:"dataplaneDriver,omitempty"`
+
+	Ipv6Support    *bool `json:"ipv6Support,omitempty"`
+	IgnoreLooseRPF *bool `json:"ignoreLooseRPF,omitempty"`
+
+	// RouterefreshInterval is the period, in seconds, at which Felix re-checks the routes
+	// in the dataplane to ensure that no other process has accidentally broken Calico’s rules.
+	// Set to 0 to disable route refresh. [Default: 90]
+	RouteRefreshIntervalSecs *int `json:"routeRefreshIntervalSecs,omitempty" configname:"routeRefreshInterval"`
```
robbrockbank: The configname should have capital r:  "RouteRefreshInterval"

robbrockbank: sim. for other fields

mgleung: oops my bad.  Didn't catch that detail.

robbrockbank:
LGTM once tests pass I'll merge.

PR #558, opened 2017-10-06
Implemented interface definitions for Cluster Information resource

mgleung:
## Description
Adding the ClusterInformation resource as well as the functionality and tests for CRUD operations.

## Todos
- [x] Tests
- [x] Documentation (not needed here now)
- [x] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/apiv2/clusterinfo.go
+++ b/lib/apiv2/clusterinfo.go
@@ -1,0 +1,74 @@
+// Copyright (c) 2017 Tigera, Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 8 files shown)


lib/apiv2/clusterinfo.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+// ClusterInformationSpec contains the values of describing the cluster.
+type ClusterInformationSpec struct {
+	// ClusterGUID is the GUID of the cluster
+	ClusterGUID string `json:"clusterGUID,omitempty" validate:"omitempty"`
+	// ClusterType describes the type of the cluster
+	ClusterType string `json:"clusterType,omitempty" validate:"omitempty"`
+	// The version of the cluster
+	Version string `json:"version,omitempty" validate:"omitempty"`
```
robbrockbank: I think Version and CalicoVersion are the same thing really.  We can drop the Version one.

mgleung: Oh oops thought that it was the Cluster version.  I'll remove it

PR #560, opened 2017-10-08
Add a generic syncer built upon an arbitrary set of watchers

robbrockbank:
This is a generic syncer implementation built against the backend API (it would be easy enough to translate to be against the main client watcher API, but for the time being while we're converting to v1 models I want to keep this in the backend).

I'm intending to to use this syncer to replace the existing confd processing.  Since KDD won't support Watchers for the initial releases this Watcher-Syncer is designed to drop in to a polling mode with a 5s poll interval - i.e. basically the same behavior for KDD as we have at the moment.

### DIFF
--- a/lib/backend/watchersyncer/watchercache.go
+++ b/lib/backend/watchersyncer/watchercache.go
@@ -1,0 +1,312 @@
+// Copyright (c) 2017 Tigera, Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package watchersyncer
+
+import (
+	"context"
+	"time"
+
+	"github.com/sirupsen/logrus"
+
+	"github.com/projectcalico/libcalico-go/lib/backend/api"
+	"github.com/projectcalico/libcalico-go/lib/backend/model"
+	cerrors "github.com/projectcalico/libcalico-go/lib/errors"
+)
+
+// The watcherCache provides watcher/syncer support for a single key type in the
+// backend.  These results are sent to the main WatcherSyncer on a buffered "results"
+// channel.  To ensure the order of events is received correctly by the main WatcherSyncer,
+// we send all notification types in this channel.  Note that because of this the results
+// channel is untyped - however the watcherSyncer only expects one of the following
+// types:
+// -  An error
+// -  An api.Update
+// -  A api.SyncStatus (only for the very first InSync notification)
+type watcherCache struct {
+	logger       *logrus.Entry
+	client       api.Client
+	watch        api.WatchInterface
+	resources    map[string]cacheEntry
+	oldResources map[string]cacheEntry
+	results      chan<- interface{}
+	hasSynced    bool
+	resourceType ResourceType
+}
+
+var (
+	ListRetryInterval = 1000 * time.Millisecond
+	WatchPollInterval = 5000 * time.Millisecond
+)
+
+// cacheEntry is an entry in our cache.  It groups the a key with the last known
+// revision that we processed.  We store the revision so that we can determine
+// if an entry has been updated (adn therefore whether we need to send an update
+// event in the syncer callback).
+type cacheEntry struct {
+	revision string
+	key      model.Key
+}
+
+// Create a new watcherCache.
+func newWatcherCache(client api.Client, resourceType ResourceType, results chan<- interface{}) *watcherCache {
+	return &watcherCache{
+		logger:       logrus.WithField("ListRoot", model.ListOptionsToDefaultPathRoot(resourceType.ListInterface)),
+		client:       client,
+		resourceType: resourceType,
+		results:      results,
+		resources:    make(map[string]cacheEntry, 0),
+	}
+}
+
+// run creates the watcher and loops indefinitely reading from the watcher.
+func (wc *watcherCache) run() {
+	wc.logger.Debug("Watcher cache starting, start initial sync processing")
+	wc.resyncAndCreateWatcher()
+
+	wc.logger.Debug("Starting main event processing loop")
+	for {
+		rc := wc.watch.ResultChan()
+		wc.logger.WithField("RC", rc).Debug("Reading event from results channel")
+		event := <-rc
+		switch event.Type {
+		case api.WatchAdded, api.WatchModified:
+			kvp := event.New
+			wc.handleWatchListEvent(kvp)
+		case api.WatchDeleted:
+			// Nil out the value to indicate a delete.
+			kvp := event.Old
+			kvp.Value = nil
+			wc.handleWatchListEvent(kvp)
+		case api.WatchError:
+			// Handle a WatchError.  First determine if the error type indicates that the
+			// watch has closed, and if so we'll need to resync and create a new watcher.
+			wc.results <- event.Error
+			if _, ok := event.Error.(cerrors.ErrorWatchTerminated); ok {
+				wc.logger.Info("Received watch terminated error - recreate watcher")
+				wc.resyncAndCreateWatcher()
+			}
+		default:
+			// Unknown event type - not much we can do other than log.
+			wc.logger.WithField("EventType", event.Type).Info("Unknown event type received from the datastore")
+		}
+	}
+}
+
+// resyncAndCreateWatcher loops performing resync processing until it successfully
+// completes a resync and starts a watcher.
+func (wc *watcherCache) resyncAndCreateWatcher() {
+	// Make sure any previous watcher is stopped.
+	wc.logger.Info("Starting watch sync/resync processing")
+	if wc.watch != nil {
+		wc.logger.Info("Stopping previous watcher")
+		wc.watch.Stop()
+		wc.watch = nil
+	}
+
+	for {
+		// Start the resync.  This processing loops until we create the watcher.  If the
+		// watcher continuously fails then this loop effectively becomes a polling based
+		// syncer.
+		wc.logger.Debug("Starting main resync loop")
+
+		// Notify the converter that we are resyncing.
+		if wc.resourceType.UpdateProcessor != nil {
+			wc.logger.Debug("Trigger converter resync notification")
+			wc.resourceType.UpdateProcessor.OnSyncerStarting()
+		}
+
+		// Start the sync by Listing the current resources.
+		l, err := wc.client.List(context.Background(), wc.resourceType.ListInterface, "")
+		if err != nil {
+			// Failed to perform the list.  Pause briefly (so we don't tight loop) and retry.
+			wc.logger.WithError(err).Debug("Failed to perform list of current data during resync")
+			time.Sleep(ListRetryInterval)
+			continue
+		}
+
+		// Move the current resources over to the oldResources
+		wc.oldResources = wc.resources
+		wc.resources = make(map[string]cacheEntry, 0)
+
+		// Send updates for each of the resources we listed - this will revalidate entries in
+		// the oldResources map.
+		for _, kvp := range l.KVPairs {
+			wc.handleWatchListEvent(kvp)
+		}
+
+		// We've listed the current settings.  Complete the sync by notifying the main WatcherSyncer
+		// go routine (if we haven't already) and by sending deletes for the old resources that were
+		// not acknowledged by the List.  The oldResources will be empty after this call.
+		wc.finishResync()
+
+		// And now start watching from the revision returned by the List.
+		w, err := wc.client.Watch(context.Background(), wc.resourceType.ListInterface, l.Revision)
+		if err != nil {
+			// Failed to create the watcher - we'll need to retry.  Sleep so that we don't
+			// tight loop.  Since we have just performed a list, we need a slightly longer
+			// delay to avoid overloading the datastore.  If the watcher keeps failing then
+			// we are effectively operating in a polling mode, so the interval should be a
+			// sensible polling interval.
+			wc.logger.WithError(err).Debug("Failed to create watcher")
+			time.Sleep(WatchPollInterval)
+			continue
+		}
+
+		// Store the watcher and exit back to the main event loop.
+		wc.logger.Debug("Resync completed, now watching for change events")
+		wc.watch = w
+		return
+	}
+}
+
+// finishResync handles processing to finish synchronization.
+// If this watcher has never been synced then notify the main watcherSyncer that we've synced.
+// We may also need to send deleted messages for old resources that were not validated in the
+// resync (i.e. they must have since been deleted).
+func (wc *watcherCache) finishResync() {
+	// If this is our first synced event then send a synced notification.  The main
+	// watcherSyncer code will send a Synced event when it has received synced events from
+	// each cache.
+	if !wc.hasSynced {
+		wc.logger.Info("Sending synced update")
+		wc.results <- api.InSync
+		wc.hasSynced = true
+	}
+
+	// If the watcher failed at any time, we end up recreating a watcher and storing off
+	// the current known resources for revalidation.  Now that we have finished the sync,
+	// any of the remaining resources that were not accounted for must have been deleted
+	// and we need to send deleted events for them.
+	numOldResources := len(wc.oldResources)
+	if numOldResources > 0 {
+		wc.logger.WithField("Num", numOldResources).Debug("Sending resync deletes")
+		updates := make([]api.Update, 0, len(wc.oldResources))
+		for _, r := range wc.oldResources {
+			updates = append(updates, api.Update{
+				UpdateType: api.UpdateTypeKVDeleted,
+				KVPair: model.KVPair{
+					Key: r.key,
+				},
+			})
+		}
+		wc.results <- updates
+	}
+	wc.oldResources = nil
+}
+
+// handleWatchListEvent handles a watch event converting it if required and passing to
+// handleConvertedWatchEvent to send the appropriate update types.
+func (wc *watcherCache) handleWatchListEvent(kvp *model.KVPair) {
+	if wc.resourceType.UpdateProcessor == nil {
+		// No update processor - handle immediately.
+		wc.handleConvertedWatchEvent(kvp)
+		return
+	}
+
+	// We have an update processor so use that to convert the event data.
+	kvps, err := wc.resourceType.UpdateProcessor.Process(kvp)
+	for _, kvp := range kvps {
+		wc.handleConvertedWatchEvent(kvp)
+	}
+
+	// If we hit a conversion error, notify the main syncer.
+	if err != nil {
+		wc.results <- err
+	}
+}
+
+// handleConvertedWatchEvent handles a converted watch event fanning out
+// to the add/mod or delete processing as necessary.
+func (wc *watcherCache) handleConvertedWatchEvent(kvp *model.KVPair) {
+	if kvp.Value == nil {
+		wc.handleDeletedUpdate(kvp.Key)
+	} else {
+		wc.handleAddedOrModifiedUpdate(kvp)
+	}
+}
+
+// handleAddedOrModifiedUpdate handles a single Added or Modified update request.
+// Whether we send an Added or Modified depends on whether we have already sent
+// an added notification for this resource.
+func (wc *watcherCache) handleAddedOrModifiedUpdate(kvp *model.KVPair) {
+	thisKey := kvp.Key
+	thisKeyString := thisKey.String()
+	thisRevision := kvp.Revision
+	wc.markAsValid(thisKeyString)
+
+	// If the resource is already in our map, then this is a modified event.  Check the
+	// revision to see if we actually need to send an update.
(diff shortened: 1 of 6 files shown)


lib/backend/watchersyncer/watchersyncer.go
```diff
@@ -0,0 +1,200 @@
...
+	// since the WatcherSyncer maintains it's own cache and will swallow duplicates.
+	// A KVPair with a nil value indicates a delete.  A non nil value indicates an add/modified.
+	// The processor may respond with any number of adds or deletes.
+	Process(*model.KVPair) ([]*model.KVPair, error)
+
+	// SyncStarting is called when syncer is starting a full sync for the associated resource
+	// type.  That means it is first going to list current resources and then watch for any updates.
+	// If the processor maintains a private internal cache, then the cache should be cleared at
+	// this point since the cache will be re-populated from the sync.
+	SyncStarting()
```
caseydavenport: another naming nit - what about `OnSyncerStart` to better indicate this is something called when the syncer starts

lib/backend/watchersyncer/watchersyncer.go
```diff
@@ -0,0 +1,200 @@
...
+}
+
+// New creates a new multiple Watcher-backed api.Syncer.
+func New(client api.Client, resourceTypes []ResourceType, callbacks api.SyncerCallbacks) api.Syncer {
+	rs := &watcherSyncer{
+		watcherCaches: make([]*watcherCache, len(resourceTypes)),
+		results:       make(chan interface{}, 2000),
```
caseydavenport: What's this magic number of 2000? Do we need a buffered channel here? I would have thought we shouldn't limit the length of this channel, or at least make it configurable.

lib/backend/watchersyncer/watchercache.go
```diff
@@ -0,0 +1,296 @@
...
+
+// Create a new watcherCache.
+// Note that the results channel is untyped.  The watcherSyncer expects one of the following
+// types:
+// -  An error
+// -  An api.Update
+// -  A api.SyncStatus (only for the very first InSync notification)
+func newWatcherCache(client api.Client, resourceType ResourceType, results chan<- interface{}) *watcherCache {
+	return &watcherCache{
+		logger:       logrus.WithField("ListRoot", model.ListOptionsToDefaultPathRoot(resourceType.ListInterface)),
```
caseydavenport: Wouldn't it be nicer to have this context logger log out the resource type rather than the etcd path?

lib/backend/watchersyncer/watchersyncer.go
```diff
@@ -0,0 +1,200 @@
...
+// to syncer updates.
+func (rs *watcherSyncer) run() {
+	log.Info("Running - sending status update")
+	rs.sendStatusUpdate(api.WaitForDatastore)
+	log.Info("Starting watcher caches")
+	for _, wc := range rs.watcherCaches {
+		go wc.run()
+	}
+
+	log.Info("Starting main watch processing loop")
```
caseydavenport: "Starting main event processing loop" would be more accurate - the events might come from a list or a watch.

lib/backend/watchersyncer/watchersyncer.go
```diff
@@ -0,0 +1,200 @@
...
+		// Block until there is data.
+		result := <-rs.results
+
+		// Process the data - this will append the data in subsequent calls, and action
+		// it if we hit a non-update event.
+		updates := rs.processResult(updates, result)
+
+		// Append results into the one update until we either flush the channel or we
+		// hit our fixed limit per update.
+	consolidatationloop:
```
caseydavenport: Is this logic worth it?

@fasaxc does Felix perform significantly better if we wait for a batch of updates over the syncer rather than sending them through individually? If so, do we have any idea what the right batch size to send is? 

@robbrockbank how did we pick `maxUpdatesToConsolidate`?

lib/backend/watchersyncer/watchersyncer.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+
+}
+
+// Process a result from the result channel.  We don't immediately action updates, but
+// instead start grouping them together so that we can send a larger single update to
+// Felix.
+func (rs *watcherSyncer) processResult(updates []api.Update, result interface{}) []api.Update {
+
+	// Final call of the interation, send any updates if we've collated some.
```
caseydavenport: Would it be cleaner to split this out into its own function? e.g. `sendUpdates`?

Seems like we only call this with `nil` once so not sure why we need to cram it into this function.

lib/backend/watchersyncer/watchersyncer.go
```diff
@@ -0,0 +1,200 @@
...
+		// Received an error.  Firstly, send any updates that we have grouped.
+		if len(updates) > 0 {
+			rs.callbacks.OnUpdates(updates)
+			updates = nil
+		}
+
+		// If this is a parsing error, and if the callbacks support
+		// it, then send the error update.
+		log.WithError(r).Info("Error received in syncer")
+		if ec, ok := rs.callbacks.(api.SyncerParseFailCallbacks); ok {
```
caseydavenport: If the callbacks don't support it, shouldn't we at least log out that we received an error?

lib/backend/watchersyncer/watchersyncer.go
```diff
@@ -0,0 +1,200 @@
...
+			if pe, ok := r.(cerrors.ErrorParsingDatastoreEntry); ok {
+				ec.ParseFailed(pe.RawKey, pe.RawValue)
+			}
+		}
+
+	case api.SyncStatus:
```
caseydavenport: Just so I understand, this code is built on the assumption that the individual watcherCaches can send no other types besides `api.InSync`

If so, it would be good to have an assert or something here - I can imagine one day someone will try to add a new call which throws off the value here.

lib/backend/watchersyncer/watchersyncer.go
```diff
@@ -0,0 +1,200 @@
...
+		// Received a synced event.  If we are still waiting for datastore, send a
+		// ResyncInProgress since at least one watcher has connected.
+		log.Info("Received sync event from watcher")
+		if rs.status == api.WaitForDatastore {
+			rs.sendStatusUpdate(api.ResyncInProgress)
+		}
+
+		// Increment the count of synced events.
+		rs.numSynced++
```
caseydavenport: The individual watcher caches send an `api.InSync` every time they do a resync. This means that if one of them runs in "polling mode", or even if running in regular watcher mode, we might get more than one sync status update from a given watcherCache before all of the others have finished syncing.

The logic as it is now will handle this poorly - I think we probably need a map which tracks whether or not each watcherCache is actually in sync, rather than assuming that X messages means X watcherCaches are in sync.

caseydavenport: might need to add an interface on top of the watcher cache with an `InSync` method so that we can tell when each one is in sync.

lib/backend/watchersyncer/watchercache.go
```diff
@@ -0,0 +1,296 @@
...
+}
+
+var (
+	ListRetryInterval = 1000 * time.Millisecond
+	WatchPollInterval = 5000 * time.Millisecond
+)
+
+// resourceKey is an entry in our cache.  It groups the a key with the last known
+// revision that we processed.
+type resourceKey struct {
```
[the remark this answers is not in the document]
robbrockbank: Yup sounds good to me.  I thought the name sucked too!

lib/backend/watchersyncer/watchercache.go
robbrockbank: it could - but this will work for non "resource types" too - in particular I'll want this for the allocation blocks, which I'm still on the fence about resource-ifying.

robbrockbank: the path contains more useful info IMHO, but for the mainline - we could just print out the resource type....

yeah ok.   I'll print out the valueType....   that'll work for all key types.

lib/backend/watchersyncer/watchercache.go
```diff
@@ -0,0 +1,296 @@
...
+func (wc *watcherCache) handleWatchListEvent(kvp *model.KVPair) {
+	if wc.resourceType.UpdateProcessor == nil {
+		// No update processor - handle immediately.
+		wc.handleConvertedWatchEvent(kvp)
+		return
+	}
+
+	// We have an update processor so use that to convert the event data.
+	kvps, err := wc.resourceType.UpdateProcessor.Process(kvp)
+	for _, kvp := range kvps {
```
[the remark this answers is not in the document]
robbrockbank: it's a perfectly valid thing to have both errors and data returned in the same call.  It was my intended use case for partial conversions, yes.  I don't think I actually use that feature anywhere though - so perhaps remove it later if none of the concrete processors make use of it.

lib/backend/watchersyncer/watchercache.go
```diff
@@ -0,0 +1,296 @@
...
+func (wc *watcherCache) handleAddedOrModifiedUpdate(kvp *model.KVPair) {
+	thisKey := kvp.Key
+	thisKeyString := thisKey.String()
+	thisRevision := kvp.Revision
+	wc.validateResource(thisKeyString)
+
+	// If the resource is already in our map, then this is a modified event.  Check the
+	// revision to see if we actually need to send an update.
+	if resource, ok := wc.resources[thisKeyString]; ok {
+		if resource.revision == thisRevision {
```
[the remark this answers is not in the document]
robbrockbank: yup

lib/backend/watchersyncer/watchercache.go
```diff
@@ -0,0 +1,296 @@
...
+}
+
+// handleDeletedWatchEvent sends a deleted event and removes the resource key from our cache.
+func (wc *watcherCache) handleDeletedUpdate(key model.Key) {
+	thisKeyString := key.String()
+	wc.validateResource(thisKeyString)
+
+	// If we have seen an added event for this key then send a deleted event and remove
+	// from the cache.
+	if _, ok := wc.resources[thisKeyString]; ok {
```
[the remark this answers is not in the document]
robbrockbank: It makes the processor code significantly simpler in some cases.

For example, the config resources don't also need to keep an internal cache of data to know when a field is added or removed from the config resource.  Instead all they have to do us send in a bunch of adds for the set fields and a bunch of deletes for the unset fields.

lib/backend/watchersyncer/watchersyncer.go
robbrockbank: completely random - and yup I glad you flagged this - I wanted to check with Shaun when he got back too....

lib/backend/watchersyncer/watchersyncer.go
robbrockbank: Not quite.  Each individual sub-synced will only send an InSync message once.  So when stuck in a polling loop - you will only get one sync message from that sub syncer.

I'm pretty sure I test that case too.

lib/backend/watchersyncer/watchercache.go
```diff
@@ -0,0 +1,296 @@
...
+
+// resourceKey is an entry in our cache.  It groups the a key with the last known
+// revision that we processed.
+type resourceKey struct {
+	revision string
+	key      model.Key
+}
+
+// Create a new watcherCache.
+// Note that the results channel is untyped.  The watcherSyncer expects one of the following
```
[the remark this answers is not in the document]
robbrockbank: oops it belongs alongside the struct - my bad.

lib/backend/watchersyncer/watchersyncer.go
robbrockbank: A buffered channel its sensible - it's possible that conversion between the v2 and v1 takes a chunk of time that could be done in parallel with the callback processing.  In which case having them buffered allows for more upfront concurrent processing.

lib/backend/watchersyncer/watchersyncer.go
robbrockbank: nitpick away.... :-)

lib/backend/watchersyncer/watchersyncer.go
caseydavenport: Just wondering if this is going to hurt us when we're resyncing large quantities of data. 

e.g. running a 5k node * 30 pod test = 150k pod events at start of day. We might just hit a resync loop where we fall behind because the wathcerCache for pods is stuck waiting for the queue to empty.. or maybe we're fine and we handle it quick enough anyway?

robbrockbank: The 2000 wasn't quite random....  The etcdv2 one had a buffered channel size of 20000 ... which i then misread as 2000.  :-/

lib/backend/watchersyncer/watchersyncer.go
caseydavenport: Ah yeah, you're right - I missed that.

Looks like we never set that back to false too, so there's no way we'll get multiple of those.

lib/backend/watchersyncer/watchersyncer.go
robbrockbank: I don't think having a buffered channel is making things any worse is it?   Having a unbuffered one that blocked would certainly be worse.

What's the alternative that you might be eluding to?  Or do you think the channel buffer should be bigger *back to the 20000 that the etcdv2 one was).

Probable one for @fasaxc .

lib/backend/watchersyncer/watchersyncer.go
robbrockbank: actually it wasn't random - i made it the same length as the channel capacity.  (i forgot about that)

lib/backend/watchersyncer/watchersyncer.go
robbrockbank: aha!   good spot!  I think I'd restructured the code and that was a side effect.

lib/backend/watchersyncer/watchersyncer.go
```diff
@@ -0,0 +1,200 @@
...
+			default:
+				break consolidatationloop
+			}
+		}
+
+		// Perform final processing (pass in a nil result) before we loop and hit the blocking
+		// call again.
+		updates = rs.processResult(updates, nil)
+	}
+
```
robbrockbank: nit:  extra blank line...

lib/backend/watchersyncer/watchersyncer.go
caseydavenport: Yeah, I think the question is just "is 2000 big enough or should we put it back to 20k?"

lib/backend/watchersyncer/watchercache.go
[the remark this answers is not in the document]
robbrockbank: fixed

lib/backend/watchersyncer/watchercache.go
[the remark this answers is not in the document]
robbrockbank: fixed

lib/backend/watchersyncer/watchercache.go
robbrockbank: Actually that might be easier said than done without plumbing a bunch of other stuff in.

If you don't mind, I'll leave as is and raise a post-Beta fix up to change this.

lib/backend/watchersyncer/watchercache.go
```diff
@@ -0,0 +1,296 @@
...
+		resources:    make(map[string]resourceKey, 0),
+	}
+}
+
+// run creates the watcher and loops indefinitely reading from the watcher.
+func (wc *watcherCache) run() {
+	wc.resyncAndCreateWatcher()
+	for {
+		rc := wc.watch.ResultChan()
+		wc.logger.WithField("RC", rc).Info("About to read from results channel")
```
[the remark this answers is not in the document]
robbrockbank: Fixed

lib/backend/watchersyncer/watchercache.go
```diff
@@ -0,0 +1,296 @@
...
+		wc.logger.Info("Stopping previous watcher")
+		wc.watch.Stop()
+		wc.watch = nil
+	}
+
+	for {
+		// Start the resync.  This processing loops until we create the watcher.  If the
+		// watcher continuously fails then this loop effectively becomes a polling based
+		// syncer.
+		wc.logger.Info("Starting watch resync processing")
```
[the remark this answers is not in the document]
robbrockbank: Done

lib/backend/watchersyncer/watchercache.go
```diff
@@ -0,0 +1,296 @@
...
+			wc.logger.Debug("Trigger converter resync notification")
+			wc.resourceType.UpdateProcessor.SyncStarting()
+		}
+
+		// Start the sync by Listing the current resources.
+		l, err := wc.client.List(context.Background(), wc.resourceType.ListInterface, "")
+
+		// Failed to perform the list.  Pause briefly (so we don't tight loop) and retry.
+		if err != nil {
+			wc.logger.WithError(err).Debug("Failed to perform list")
```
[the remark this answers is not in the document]
robbrockbank: Done

lib/backend/watchersyncer/watchercache.go
```diff
@@ -0,0 +1,296 @@
...
+		// Notify the converter that we are resyncing.
+		if wc.resourceType.UpdateProcessor != nil {
+			wc.logger.Debug("Trigger converter resync notification")
+			wc.resourceType.UpdateProcessor.SyncStarting()
+		}
+
+		// Start the sync by Listing the current resources.
+		l, err := wc.client.List(context.Background(), wc.resourceType.ListInterface, "")
+
+		// Failed to perform the list.  Pause briefly (so we don't tight loop) and retry.
```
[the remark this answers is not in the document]
robbrockbank: Done

lib/backend/watchersyncer/watchercache.go
```diff
@@ -0,0 +1,296 @@
...
+			// delay to avoid overloading the datastore.  If the watcher keeps failing then
+			// we are effectively operating in a polling mode, so the interval should be a
+			// sensible polling interval.
+			wc.logger.WithError(err).Debug("Failed to create watcher")
+			time.Sleep(WatchPollInterval)
+			continue
+		}
+
+		// Store the watcher and exit back to the main event loop.
+		wc.watch = w
```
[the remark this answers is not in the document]
robbrockbank: done

lib/backend/watchersyncer/watchercache.go
```diff
@@ -0,0 +1,296 @@
...
+		return
+	}
+}
+
+// dataIsNowSynced handles processing to finish synchronization.
+// If this watcher has never been synced then notify the main watcherSyncer that we've synced.
+// We may also need to send deleted messages for old resources that were not validated in the
+// resync (i.e. they must have since been deleted).
+func (wc *watcherCache) dataIsNowSynced() {
```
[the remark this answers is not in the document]
robbrockbank: Done.

lib/backend/watchersyncer/watchercache.go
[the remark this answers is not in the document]
robbrockbank: I'll change if you really want me to, otherwise I'll leave it for now.  I think if it made the code path more complicated I'd omit it, but in some ways it actually makes it simpler.

lib/backend/watchersyncer/watchercache.go
```diff
@@ -0,0 +1,296 @@
...
+	thisKeyString := thisKey.String()
+	thisRevision := kvp.Revision
+	wc.validateResource(thisKeyString)
+
+	// If the resource is already in our map, then this is a modified event.  Check the
+	// revision to see if we actually need to send an update.
+	if resource, ok := wc.resources[thisKeyString]; ok {
+		if resource.revision == thisRevision {
+			// No update to revision, so no event to send.
+			wc.logger.WithField("Key", thisKeyString).Debug("Entry not modified")
```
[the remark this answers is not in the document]
robbrockbank: Done

lib/backend/watchersyncer/watchercache.go
```diff
@@ -0,0 +1,296 @@
...
+			UpdateType: api.UpdateTypeKVDeleted,
+			KVPair: model.KVPair{
+				Key: key,
+			},
+		}}
+		delete(wc.resources, thisKeyString)
+	}
+}
+
+// validateResource validates a resource that is currently in the oldResources.
```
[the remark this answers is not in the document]
robbrockbank: done

lib/backend/watchersyncer/watchersyncer.go
robbrockbank: Done

lib/backend/watchersyncer/watchersyncer.go
robbrockbank: We do ... the log above has an Info log which includes the error.  I'll add an extra log to say the callbacks are supported ... i'm not a big fan of additional branches just to log.

lib/backend/watchersyncer/watchersyncer.go
robbrockbank: Rather than assert, I log the status we receive and no op or anything other than the InSync

lib/backend/watchersyncer/watchersyncer.go
fasaxc: It may help a bit at start-of-day, when we have to push 150k updates through but at steady-state 100-1000updates/s it shouldn't make a difference.  IIRC, sending on a channel takes about half a microsecond so if you're doing 1000 sends/s it's negligible; if you're doing 100k+ then you'll start to notice and if you're doing 1m+ then it dominates.

lib/backend/watchersyncer/watchercache.go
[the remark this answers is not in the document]
caseydavenport: It's fine. I don't really like it but I'm not bothered enough to ask you to change.

PR #562, opened 2017-10-09
Add orchrefs to libcalico-go

ozdanborne:
## Description

Add a new field to libcalico-go called `orchRefs`. This field will allow for association between a k8s node and the etcdNode object that matches it, and allow for cleanup of stale nodes.

## Release Note

```release-note
Add the orchRef field to the Node data model.
```


caseydavenport:
@ozdanborne changes look good to me - could you add a really simple test here: https://github.com/projectcalico/libcalico-go/blob/master/lib/client/node_e2e_test.go#L48

This will verify that the data can be read back from the datastore once written.

heschlie:
Closing as work was done in #614

PR #563, opened 2017-10-10
Renamed and moved IPInIPTunnelAddr field

mgleung:
## Description
Moved the `IPinIPTunnelAddr` to the `NodeBGPSpec` and renamed the attribute to specify IPv4.

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/apiv2/node.go
+++ b/lib/apiv2/node.go
@@ -37,8 +37,7 @@ type Node struct {
 // NodeSpec contains the specification for a Node resource.
 type NodeSpec struct {
 	// BGP configuration for this node.
-	BGP              *NodeBGPSpec `json:"bgp,omitempty" validate:"omitempty"`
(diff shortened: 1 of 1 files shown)


lib/apiv2/node.go
```diff
@@ -52,6 +51,8 @@ type NodeBGPSpec struct {
 	// IPv6Address is the IPv6 address and network of this node.  At least
 	// one of the IPv4 and IPv6 addresses should be specified.
 	IPv6Address string `json:"ipv6Address,omitempty" validate:"omitempty,ipv6"`
+	// IPv4IPIPTunnel is the IPv4 address of the IP in IP tunnel.
```
gunjan5: this should be `IPv4IPIPTunnelAddr`

mgleung: oops

PR #564, opened 2017-10-11
Support CRUD for v2 client when using KDD backend

doublek:
## Description
@robbrockbank, ~~I've created a new directory called k8sv2 and put all the v2 data model relevant code there. Ideally I wanted to remove k8s and have the k8sv2 folder take its place.~~ All changes are now in the `backend/k8s` folder.

~~Let me know if there is a way in which you'd like me to move code around to make it better for reviewing.~~

In addition to supporting existing items using v2 data model, compared to the the k8s backend for v1 datamodel, the k8s backend for v2 has the following modifications:

Renamed CRDs:
- `GlobalBGPConfig` -> `BGPConfiguration`
- `GlobalFelixConfig` -> `FelixConfiguration`

CRDs that stay the same:

- `BGPPeers` - The resource file is now named `bgppeers.go` instead of `globalbgppeers.go`.
- `GlobalNetworkPolicy`.
- `IPPool`

Newly created/supported CRDs:
- `ClusterInformation`

I haven't updated the syncer yet and is a copy of the v1 datamodel k8s backend implementation.

## Todos
- [x] Tests


lib/backend/k8sv2/custom/bgpconfig.go
```diff
@@ -0,0 +1,85 @@
...
+import (
+	"encoding/json"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+
+	"github.com/projectcalico/libcalico-go/lib/apiv2"
+)
+
+type BGPConfig struct {
```
robbrockbank: Can we now just use the definitions in apiv2 for all these things?

robbrockbank: same is true now for all of the CRD backed resources - i think the apiv2 definitions should be "k8s" crd compliant.

lib/backend/k8sv2/k8s.go
```diff
@@ -0,0 +1,795 @@
...
+	kubeClient := &KubeClient{
+		clientSet:       cs,
+		crdClientV1:     crdClientV1,
+		disableNodePoll: kc.K8sDisableNodePoll,
+	}
+
+	// Create the Calico sub-clients.
+	kubeClient.ipPoolClient = resources.NewIPPoolClient(cs, crdClientV1)
+	kubeClient.nodeClient = resources.NewNodeClient(cs, crdClientV1)
+	kubeClient.gnpClient = resources.NewGlobalNetworkPolicyClient(cs, crdClientV1)
```
robbrockbank: I think we need one for NetworkPolicy too, and then combine the Calico CRD-backed NetworkPolicy resources with the k8s Network Policy.  Arguably this could fall out into a different task though - as long as we are tracking that.

doublek: Yup. I realized that I missed out the network policy conversion bit. I'll push this with some shortly.

lib/backend/k8sv2/custom/bgpconfig.go
```diff
@@ -0,0 +1,85 @@
...
+	Spec            apiv2.BGPConfigurationSpec `json:"spec"`
+}
+
+type BGPConfigList struct {
+	metav1.TypeMeta `json:",inline"`
+	Metadata        metav1.ListMeta `json:"metadata"`
+	Items           []BGPConfig     `json:"items"`
+}
+
+// Required to satisfy Object interface
```
robbrockbank: btw, I don't think any of this stuff from here below is required if you use the v2 definitions.  if it seems like it is, can you let me know.

lib/backend/k8sv2/resources/customnoderesource.go
```diff
@@ -0,0 +1,461 @@
+// Copyright (c) 2017 Tigera, Inc. All rights reserved.
```
robbrockbank: Does anything use this anymore?

I believe this isn't required now we have FelixConfiguration, BGConfiguration, BGPPeer and ClusterInformation being standard CRDs.

doublek: Oops.. I committed this by misted.. Removed.

lib/backend/k8sv2/custom/bgpconfig.go
doublek: I tried to use the apiv2 definitions directly and they didn't seem to work. Probably something I did incorrectly. Let me give it another shot..

lib/backend/k8sv2/k8s.go
doublek: So, a couple of things aren't clear to me w.r.t policies in the backend.

Is the CRD-backed NetworkPolicy resource the GlobalNetworkPolicy or is it a Calico Policy (just to distinguish from K8s NetworkPolicy) but is allowed to be namespaced?
My understanding was that we have two v2 policy related resources: `GlobalNetworkPolicy` and `NetworkPolicy` and that the `GlobalNetworkPolicy` resource will be backed by a CRD and `NetworkPolicy` is backed by K8s _NetworkPolicy_. 

In the current (corresponding to the v1 style backend) _GET_/_LIST_ for a _Policy_ resource, the k8s backend combines `GlobalNetworkPolicy` (a CRD) and `NetworkPolicy` (K8s API). With the v2 datamodel, with 2 separate client resources/APIs, does this combining still make sense?

lib/backend/k8sv2/custom/bgpconfig.go
doublek: > Can we now just use the definitions in apiv2 for all these things?
^^ Was the comment from you that I accidentally deleted...

robbrockbank: that was something i wrote, so possibly :-)

lib/backend/k8sv2/conversion.go
```diff
@@ -0,0 +1,431 @@
...
+	//				Name:     containerPort.Name,
+	//				Protocol: modelProto,
+	//				Port:     uint16(containerPort.ContainerPort),
+	//			})
+	//		}
+	//	}
+	//}
+
+	// Create the key / value pair to return.
+	// TODO(doublek): Which container becomes ContainerID in WEPSpec?
```
caseydavenport: Yeah, strictly speaking we can't get the container ID when running in KDD mode.

We probably need to leave that blank, or we need the CNI plugin to apply an annotation which stores it on the Pod.

doublek: @caseydavenport You are correct. IIUC, it should be required for name construction only when using the etcd backend. Will remove the TODO.

lib/backend/k8sv2/custom/bgpconfig.go
doublek: @robbrockbank Using apiv2 definitions ended up working as expected. There was a small subtle conversion thing that I'd missed and it is all sorted out now.

test/crds.yaml
```diff
@@ -67,3 +70,55 @@ items:
       kind: GlobalNetworkPolicy
       plural: globalnetworkpolicies
       singular: globalnetworkpolicy
+- apiVersion: apiextensions.k8s.io/v1beta1
```
robbrockbank: this is a dupe, unless i'm misreading the entry above.

doublek: @robbrockbank It was a duplicate. I've removed it.

lib/backend/k8sv2/k8s.go
robbrockbank: So for v3.0 the network policy that is sync'd to Felix will come from 3 places in KDD:
-  K8s Network Policy
-  Calico Network Policy (CRD)
-  Calico Global Network Policy (CRD)

The names/selectors are adjusted on the fly in the Felix syncer, such that Felix itself isn't actually aware of these different sources of policy.

Regarding CRUD:
-  GlobalNetworkPolicy maps straight through to the CRD
-  NetworkPolicy maps through to both the Kubernetes NP and the Calico NP CRD  - so it's the union of the two - with a name prefix ensuring name uniqueness across the two sources.

doublek: Got it.

robbrockbank: Fixing up the syncer is a separate task - it might prove to be easier to use the new syncer infrastructure that i've implemented -- but would require us to implement watcher support in kdd.

PR #565, opened 2017-10-11
Add OrchRefs to Node object

heschlie:
## Description
Expanding on @ozdanborne PR #562 to add OrchRefs to libcalico-go as a field on the Node.

Add a new field to libcalico-go called `OrchRefs`. This field will allow for association between a k8s node and the etcdNode object that matches it, and allow for cleanup of stale nodes.

This PR includes the needed additions to the compat layer as the Node is a composite object. I'm not extremely familiar with the compat layer so I expect it will need some tweaking.

## Todos
- [x] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
Add the orchRef field to the Node data model.
```

### DIFF
--- a/lib/backend/compat/compat.go
+++ b/lib/backend/compat/compat.go
@@ -492,14 +492,8 @@ func (c *ModelAdaptor) getNodeSubcomponents(ctx context.Context, nk model.NodeKe
 		return err
 	}
 
-	if components, err := c.client.List(model.OrchRefListOptions{Hostname: nk.Hostname}); err == nil {
-		for _, o := range components {
-			orchRef := model.OrchRef{
-				Orchestrator: o.Key.(model.OrchRefKey).Orchestrator,
-				NodeName:     o.Value.(string),
-			}
-			nv.OrchRefs = append(nv.OrchRefs, orchRef)
-		}
+	if component, err := c.client.Get(model.OrchRefKey{Hostname: nk.Hostname}); err == nil {
+		nv.OrchRefs = component.Value.([]model.OrchRef)
 	}
 
 	return nil
@@ -663,15 +657,15 @@ func toNodeComponents(d *model.KVPair) (primary *model.KVPair, optional []*model
 		})
 	}
 	if len(n.OrchRefs) > 0 {
-		for _, orchref := range n.OrchRefs {
-			optional = append(optional, &model.KVPair{
-				Key: model.OrchRefKey{
-					Hostname:     nk.Hostname,
-					Orchestrator: orchref.Orchestrator,
-				},
-				Value: orchref.NodeName,
-			})
-		}
+		//v, err := json.Marshal(n.OrchRefs)
+		//fmt.Printf("____%s____", string(v))
+		//if err != nil {
+		//	log.WithError(err).Errorf("Failed to marshal JSON for OrchRefs in Node %s", nk.Hostname)
+		//}
+		optional = append(optional, &model.KVPair{
+			Key:   model.OrchRefKey{Hostname: nk.Hostname},
+			Value: n.OrchRefs,
+		})
 	}
 
 	return primary, optional
@@ -720,8 +714,7 @@ func toNodeDeleteComponents(nk model.NodeKey) (primary *model.KVPair, optional [
 		},
 		&model.KVPair{
 			Key: model.OrchRefKey{
-				Hostname:     nk.Hostname,
-				Orchestrator: "",
+				Hostname: nk.Hostname,
 			},
 		},
 	}
--- a/lib/backend/compat/compat.go
+++ b/lib/backend/compat/compat.go
@@ -492,7 +492,7 @@ func (c *ModelAdaptor) getNodeSubcomponents(ctx context.Context, nk model.NodeKe
 		return err
 	}
 
-	if component, err := c.client.Get(model.OrchRefKey{Hostname: nk.Hostname}); err == nil {
+	if component, err := c.client.Get(ctx, model.OrchRefKey{Hostname: nk.Hostname}, ""); err == nil {
 		nv.OrchRefs = component.Value.([]model.OrchRef)
 	}
 
--- a/lib/backend/compat/compat.go
+++ b/lib/backend/compat/compat.go
@@ -492,6 +492,16 @@ func (c *ModelAdaptor) getNodeSubcomponents(ctx context.Context, nk model.NodeKe
 		return err
 	}
 
+	if components, err := c.client.List(model.OrchRefListOptions{Hostname: nk.Hostname}); err == nil {
+		for _, o := range components {
+			orchRef := model.OrchRef{
+				Orchestrator: o.Key.(model.OrchRefKey).Orchestrator,
+				NodeName:     o.Value.(string),
+			}
+			nv.OrchRefs = append(nv.OrchRefs, orchRef)
+		}
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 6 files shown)


lib/backend/model/node.go
```diff
@@ -33,6 +33,8 @@ var (
 	typeHostIp        = rawIPType
 	matchHostMetadata = regexp.MustCompile(`^/?calico/v1/host/([^/]+)/metadata$`)
 	matchHostIp       = regexp.MustCompile(`^/?calico/v1/host/([^/]+)/bird_ip$`)
+	orchRefMatch      = regexp.MustCompile(`^/?/calico/v1/host/.*/orchestrator/(.*)`)
```
caseydavenport: Is there a reason you decided to have each ref as its own path, versus having a single per-node orchestratorrefs section?

e.g. `calico/v1/host/.*/orchrefs/` and plop the data in as a single json blob?

heschlie: Not particularly, I don't see a huge advantage one way or the other, but I'd be happy to move over to storing it as a json blob if that's what we'd prefer.

caseydavenport: I think that's what I'd prefer.

caseydavenport: @robbrockbank ?

robbrockbank: oh yes - use a single entry in etcd.  The more entries we add the more things we need to query to query a single node.

lib/backend/compat/compat.go
```diff
@@ -646,6 +650,17 @@ func toNodeComponents(d *model.KVPair) (primary *model.KVPair, optional []*model
 			},
 		})
 	}
+	if len(n.OrchRefs) > 0 {
+		//v, err := json.Marshal(n.OrchRefs)
```
heschlie: Whoops that shouldn't be there...

lib/client/node_e2e_test.go
```diff
@@ -98,7 +98,7 @@ var _ = testutils.E2eDatastoreDescribe("Node tests", testutils.DatastoreEtcdV2,
 
 			// Apply node2 with spec1.
 			By("Applying node2 with spec1")
-			_, err = c.Nodes().Update(&api.Node{Metadata: meta2, Spec: spec1})
+			_, err = c.Nodes().Apply(&api.Node{Metadata: meta2, Spec: spec1})
```
caseydavenport: Why does this need to be apply now?

heschlie: Thought I may have caught an error, the By line states `By("Applying node2 with spec1")` and we were calling update, I can revert it if this was intentional, but perhaps the comment should be updated?

lib/api/node.go
```diff
@@ -65,6 +65,17 @@ type NodeSpec struct {
 	// BGP configuration for this node.  If this omitted, the Calico node
 	// will be run in policy-only mode.
 	BGP *NodeBGPSpec `json:"bgp,omitempty" validate:"omitempty"`
+
+	// OrchRefs for this node.
+	OrchRefs []OrchRef `json:"orchRefs,omitempty" validate:"omitempty"`
+}
+
+// OrchRef is used to map a node in etcd to nodes in other orchestrators.
```
caseydavenport: nit: this shouldn't say anything about etcd - this is the Calico API structures and they don't care about what datastore is in use.

So, this should say something like "used to correlate a Calico node to its corresponding representation in a given orchestrator"

lib/backend/model/node.go
```diff
@@ -174,3 +181,40 @@ func (key HostIPKey) valueType() reflect.Type {
 func (key HostIPKey) String() string {
 	return fmt.Sprintf("Node(name=%s)", key.Hostname)
 }
+
+type OrchRefKey struct {
+	Hostname string
+}
+
+func (key OrchRefKey) defaultPath() (string, error) {
```
caseydavenport: The path here should be `/calico/v1/host/%s/orchestrator_refs`

lib/backend/model/node.go
```diff
@@ -174,3 +181,40 @@ func (key HostIPKey) valueType() reflect.Type {
...
+func (key OrchRefKey) defaultDeleteParentPaths() ([]string, error) {
+	return nil, nil
+}
+
+func (key OrchRefKey) valueType() reflect.Type {
+	return typeOrchRefs
+}
+
+func (key OrchRefKey) String() string {
+	return fmt.Sprintf("OrchRef(nodename=%s)", key.Hostname)
```
caseydavenport: This should be plural since there are multiple of them - `OrchRefs`

lib/backend/model/node.go
```diff
@@ -174,3 +181,40 @@ func (key HostIPKey) valueType() reflect.Type {
...
+}
+
+type OrchRefListOptions struct {
+	Hostname string
+}
+
+func (options OrchRefListOptions) defaultPathRoot() string {
+	return fmt.Sprintf("/calico/v1/host/%s/orchestrator", options.Hostname)
```
caseydavenport: update this guy as well to `orchestrator_refs`

lib/client/node_e2e_test.go
caseydavenport: Yep I think you might be right.

lib/client/node_e2e_test.go
```diff
@@ -166,12 +166,32 @@ var _ = testutils.E2eDatastoreDescribe("Node tests", testutils.DatastoreEtcdV2,
...
 				},
+				OrchRefs: []api.OrchRef{
+					{
+						Orchestrator: "k8s",
+						NodeName:     "node1",
+					},
+					{
+						Orchestrator: "mesos",
+						NodeName:     "node1",
+					},
```
caseydavenport: These should have different NodeName values

lib/backend/compat/compat.go
```diff
@@ -646,6 +650,12 @@ func toNodeComponents(d *model.KVPair) (primary *model.KVPair, optional []*model
 			},
 		})
 	}
+	if len(n.OrchRefs) > 0 {
+		optional = append(optional, &model.KVPair{
+			Key:   model.OrchRefKey{Hostname: nk.Hostname},
+			Value: n.OrchRefs,
```
caseydavenport: Generally, when writing json to the datastore, you want the top level structure to be a map so that it can be extended in the future. I understand that the other things in the node structure are not done like this, so maybe we can get away with it claiming prior art, but ideally instead of storing

```json
[{"nodeName": "mynode", "orchestrator": "k8s"}]
```



we'd store something like

```json
{
  "orchRefs": [{"nodeName": "mynode", "orchestrator": "k8s"}]
}
```

That does make the code a bit ickier. @robbrockbank do you have a preference? I'm tempted to say just store a list given this will change in the v2 datamodel + etcdv3 and it matches what we do for other node subcomponents.

caseydavenport:
@bcreane are you the third runner in this relay? :)

lib/backend/compat/compat.go
heschlie: Gah! I had originally done that but thought it felt clunky in the struct.

heschlie: I haven't added this on my latest commit, but if we prefer it I am happy to oblige.

To verify what we're looking for, we would want to contain the slice of `OrchRefs` in a struct, which would mean getting that slice would be `Node.OrchRefs.OrchRefs`? Or maybe something similar but with less redundant naming.

caseydavenport:
@heschlie please re-open against the [v1.x series](https://github.com/projectcalico/libcalico-go/tree/v1.x-series) branch

PR #569, opened 2017-10-13
Host Endpoint converter and code cleanup for converter functions

mgleung:
## Description
Added the code to convert v2 Host Endpoint objects into their v1 equivalents

This is the rebased code from https://github.com/robbrockbank/libcalico-go/pull/3

## Todos
- [x] Tests
- [x] Release note

## Release Note

```release-note
None required
```


lib/backend/syncersv1/updateprocessors/hostendpointprocessor.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	v1key := model.HostEndpointKey{
+		Hostname:   v2res.Spec.Node,
+		EndpointID: v2res.GetName(),
+	}
+
+	var v1value *model.HostEndpoint
+	// Deletion operations will have empty values so skip if empty
+	if v2res != nil {
```
robbrockbank: This check isn't required actually.. you actually dereference v2Res above to get the Node name so it can't be nil

robbrockbank: When using the conflict resolving cache the conversion method only gets called for adds/modifies.  The deletes are handled using the data in the cache.

PR #574, opened 2017-10-14
Add the V2 to V1 converters for Workload Endpoint, Profile, NetworkPolicy, and GlobalNetworkPolicy

mgleung:
## Description
Add the conversion code and tests for Workload Endpoint, Profile, NetworkPolicy, and GlobalNetworkPolicy resources.

## Todos
- [x] Tests
- [x] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/apiv2/workloadendpoint.go
+++ b/lib/apiv2/workloadendpoint.go
@@ -67,6 +67,8 @@ type WorkloadEndpointSpec struct {
 	InterfaceName string `json:"interfaceName,omitempty" validate:"interface"`
 	// MAC is the MAC address of the endpoint interface.
 	MAC string `json:"mac,omitempty" validate:"omitempty,mac"`
+	// Ports contains the endpoint's named ports, which may be referenced in security policy rules.
+	Ports []EndpointPort `json:"ports,omitempty" validate:"dive,omitempty"`
 }
 
 // IPNat contains a single NAT mapping for a WorkloadEndpoint resource.
--- a/lib/backend/syncersv1/updateprocessors/globalnetworkpolicyprocessor.go
+++ b/lib/backend/syncersv1/updateprocessors/globalnetworkpolicyprocessor.go
@@ -1,0 +1,47 @@
+// Copyright (c) 2017 Tigera, Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package updateprocessors
+
+import (
+	"errors"
+
+	"github.com/projectcalico/libcalico-go/lib/apiv2"
+	"github.com/projectcalico/libcalico-go/lib/backend/model"
+	"github.com/projectcalico/libcalico-go/lib/backend/watchersyncer"
+)
+
+// Create a new SyncerUpdateProcessor to sync GlobalNetworkPolicy data in v1 format for
+// consumption by Felix.
+func NewGlobalNetworkPolicyUpdateProcessor() watchersyncer.SyncerUpdateProcessor {
+	return NewSimpleUpdateProcessor(apiv2.KindGlobalNetworkPolicy, convertGlobalNetworkPolicyV2ToV1Key, convertGlobalNetworkPolicyV2ToV1Value)
+}
+
+func convertGlobalNetworkPolicyV2ToV1Key(v2key model.ResourceKey) (model.Key, error) {
+	if v2key.Name == "" {
+		return model.PolicyKey{}, errors.New("Missing Name field to create a v1 NetworkPolicy Key")
+	}
+	return model.PolicyKey{
+		Name: v2key.Name,
+	}, nil
+
+}
+
+func convertGlobalNetworkPolicyV2ToV1Value(val interface{}) (interface{}, error) {
+	v2res, ok := val.(*apiv2.GlobalNetworkPolicy)
+	if !ok {
+		return nil, errors.New("Value is not a valid GlobalNetworkPolicy resource value")
+	}
+	return convertPolicyV2ToV1Spec(v2res.Spec)
+}
--- a/lib/backend/syncersv1/updateprocessors/networkpolicyprocessor.go
+++ b/lib/backend/syncersv1/updateprocessors/networkpolicyprocessor.go
@@ -1,0 +1,86 @@
+// Copyright (c) 2017 Tigera, Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package updateprocessors
+
+import (
+	"errors"
+	"strings"
+
+	"github.com/projectcalico/libcalico-go/lib/apiv2"
+	"github.com/projectcalico/libcalico-go/lib/backend/model"
+	"github.com/projectcalico/libcalico-go/lib/backend/watchersyncer"
+)
+
+// Create a new SyncerUpdateProcessor to sync NetworkPolicy data in v1 format for
+// consumption by Felix.
+func NewNetworkPolicyUpdateProcessor() watchersyncer.SyncerUpdateProcessor {
+	return NewSimpleUpdateProcessor(apiv2.KindNetworkPolicy, convertNetworkPolicyV2ToV1Key, convertNetworkPolicyV2ToV1Value)
+}
+
+func convertNetworkPolicyV2ToV1Key(v2key model.ResourceKey) (model.Key, error) {
+	if v2key.Name == "" || v2key.Namespace == "" {
+		return model.PolicyKey{}, errors.New("Missing Name or Namespace field to create a v1 NetworkPolicy Key")
+	}
+	return model.PolicyKey{
+		Name: v2key.Namespace + "/" + v2key.Name,
+	}, nil
+
+}
+
+func convertNetworkPolicyV2ToV1Value(val interface{}) (interface{}, error) {
+	v2res, ok := val.(*apiv2.NetworkPolicy)
+	if !ok {
+		return nil, errors.New("Value is not a valid NetworkPolicy resource value")
+	}
+	return convertPolicyV2ToV1Spec(v2res.Spec)
+}
+
+func convertPolicyV2ToV1Spec(spec apiv2.PolicySpec) (interface{}, error) {
+	var irules []model.Rule
+	for _, irule := range spec.IngressRules {
+		irules = append(irules, RuleAPIV2ToBackend(irule))
+	}
+
+	var erules []model.Rule
+	for _, erule := range spec.EgressRules {
+		erules = append(erules, RuleAPIV2ToBackend(erule))
+	}
+
+	v1value := &model.Policy{
+		Order:         spec.Order,
+		InboundRules:  irules,
+		OutboundRules: erules,
+		Selector:      spec.Selector,
+		DoNotTrack:    spec.DoNotTrack,
+		PreDNAT:       spec.PreDNAT,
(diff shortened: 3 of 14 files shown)


lib/backend/syncersv1/updateprocessors/globalnetworkpolicyprocessor.go
```diff
@@ -0,0 +1,81 @@
...
+// Convert v2 KVPair to the equivalent v1 KVPair.
+func convertGlobalNetworkPolicyV2ToV1(kvp *model.KVPair) (*model.KVPair, error) {
+	// Validate against incorrect key/value kinds.  This indicates a code bug rather
+	// than a user error.
+	v2key, ok := kvp.Key.(model.ResourceKey)
+	if !ok || v2key.Kind != apiv2.KindGlobalNetworkPolicy {
+		return nil, errors.New("Key is not a valid GlobalNetworkPolicy resource key")
+	}
+
+	if kvp.Value == nil {
```
robbrockbank: I hope this isn't true, because for a delete event, the value will always be nil.

For resources where the name of the v2 resource can be mapped directly to the v1 index then the v2 key will have sufficient information to construct the v1 key.  In this case the v2Key name basically maps to the name in the v1 key.  You don't need the resource to extract that.

mgleung: Ok, I was using the Value for the name, but if it is always nil for delete, then I'll have to derive the name from the key.  Is the Key name always guaranteed to be the same as the name in the Value?

robbrockbank: yup it will be unless someone is hand editing stuff in the datastore.

robbrockbank: same with the namespace.

lib/backend/syncersv1/updateprocessors/workloadendpointprocessor.go
```diff
@@ -0,0 +1,200 @@
...
+
+func convertWorkloadEndpointV2ToV1Key(v2res *apiv2.WorkloadEndpoint) (model.WorkloadEndpointKey, error) {
+	parts := names.ExtractDashSeparatedParms(v2res.GetName(), 4)
+	if len(parts) != 4 {
+		return model.WorkloadEndpointKey{}, errors.New("Not enough information provided to create v1 Workload Endpoint Key")
+	}
+	return model.WorkloadEndpointKey{
+		Hostname:       parts[0],
+		OrchestratorID: parts[1],
+		WorkloadID:     v2res.GetNamespace() + "/" + parts[2],
```
robbrockbank: so namespace in the Key too - so you don't need the KVP Value to create the v1 Key.

lib/backend/syncersv1/updateprocessors/networkpolicyprocessor.go
```diff
@@ -0,0 +1,137 @@
...
+		for _, irule := range spec.IngressRules {
+			irules = append(irules, RuleAPIV2ToBackend(irule))
+		}
+
+		var erules []model.Rule
+		for _, erule := range spec.EgressRules {
+			erules = append(erules, RuleAPIV2ToBackend(erule))
+		}
+
+		// TODO: Am I supposed to do anything with Annotations? Are they labels?
```
robbrockbank: nope annotations are not used by Felix.

lib/backend/syncersv1/updateprocessors/processor.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/projectcalico/libcalico-go/lib/backend/watchersyncer"
+)
+
+// NewGeneralUpdateProcessor implements an update processor that only needs to take in
+// a conversion function. This is only meant to be used in cases where the Process and
+// OnSyncerStarting methods do not have any special handling.
+//
+// One of the main differences between the general update processor and the conflict
+// resolving name cache processor is that the conflict resolving name cache processor
+// handles deletes by inspecting the value in the key-value pair. The general update processor
```
robbrockbank: Actually the delete - the delete uses the cache to find the old values from an internal cache because it needs the value to construct the v1 key.

Key thing about this "general update processor" is that it's used to provide common processing for v2 resources where the v1 key can be calculated directly from the v2 resource key (i.e. just with the Kind, Name and Namespace).

lib/backend/syncersv1/updateprocessors/processor.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+}
+
+type generalUpdateProcessor struct {
+	v2Kind    string
+	converter ConvertV2ToV1
+}
+
+func (gup *generalUpdateProcessor) Process(kvp *model.KVPair) ([]*model.KVPair, error) {
+	// Extract the name of the resource.
```
robbrockbank: actually you are extracting the kind, not the name.

lib/backend/syncersv1/updateprocessors/processor.go
```diff
@@ -0,0 +1,65 @@
...
+		converter: converter,
+	}
+}
+
+type generalUpdateProcessor struct {
+	v2Kind    string
+	converter ConvertV2ToV1
+}
+
+func (gup *generalUpdateProcessor) Process(kvp *model.KVPair) ([]*model.KVPair, error) {
```
robbrockbank: As it stands, I'm not sure this generalUpdateProcess buys us much.  The only common processing is the validation of the resource Kind and then the rest of the processing is done in the converter.  And from what I can see, you do the type casting an Kind validation again in each of the individual converter functions.

It might be more useful if it took two converter functions, one to handle the key, and one to handle the value.  That way the handling of a delete can be handled generically here by only calling the convertKey method - whereas for an add/update you call both.

Also, since you are already asserting that the v2 key is a ResourceKey, the function signature for the convert key method could look more like:

`type ConvertV2KeyToV1 func(v2Key model.ResourceKey) model.Key`

and the value converter I guess would look like:

`type ConvertV2ResToV1 func(v2Obj runtime.Object) (interface{}, error)`

robbrockbank: Also, perhaps rather than GeneralUpdateProc... maybe SimpleUpdateProc... would be better since it can only be used for simple cases where the keys are directly mappable from v2 to v1.

lib/backend/syncersv1/updateprocessors/profileprocessor.go
```diff
@@ -0,0 +1,217 @@
...
+
+	"github.com/projectcalico/libcalico-go/lib/apiv2"
+	"github.com/projectcalico/libcalico-go/lib/backend/model"
+	"github.com/projectcalico/libcalico-go/lib/backend/watchersyncer"
+	cnet "github.com/projectcalico/libcalico-go/lib/net"
+)
+
+// Create a new SyncerUpdateProcessor to sync Profile data in v1 format for
+// consumption by Felix.
+func NewProfileUpdateProcessor() watchersyncer.SyncerUpdateProcessor {
```
robbrockbank: Apologies for not mentioning this earlier, but profiles are actually constructed of three different keys in the v1 data model.  Rules, Tags and Labels.

I think the Tags one can be ignored, but you'll need to also include Labels - where the Labels are assigned from the  LabelsToApply fields.  Take a look in compat.go in the backend which deals with handling the three separate v1 keys.

mgleung: Ok, I split it out into 3.  I left Tags empty, but it is always there as the first value.  Not sure if it might be better to have the return values be a map instead of a list of KVPairs so that its obvious which KVPairs are for which attributes, but I have the return as a list of 3 KVPairs to represent Tags, Labels, and Rules for now.

lib/apiv2/workloadendpoint.go
```diff
@@ -66,7 +66,8 @@ type WorkloadEndpointSpec struct {
 	// InterfaceName the name of the Linux interface on the host: for example, tap80.
 	InterfaceName string `json:"interfaceName,omitempty" validate:"interface"`
 	// MAC is the MAC address of the endpoint interface.
-	MAC string `json:"mac,omitempty" validate:"omitempty,mac"`
+	MAC   string         `json:"mac,omitempty" validate:"omitempty,mac"`
+	Ports []EndpointPort `json:"ports,omitempty" validate:"dive,omitempty"`
```
robbrockbank: Looks like it's missing the comment (same with HostEndpoint too actually).  Think you can just steal  the comment from the original v1 api def.

lib/backend/syncersv1/updateprocessors/profileprocessor.go
```diff
@@ -0,0 +1,236 @@
...
+	}
+	return out
+}
+
+// ruleActionAPIV2ToBackend converts the rule action field value from the API
+// value to the equivalent backend value.
+func ruleActionAPIV2ToBackend(action apiv2.Action) string {
+	if action == apiv2.Pass {
+		// TODO: why does this say tiers?
+		return "next-tier"
```
robbrockbank: yeah, it's correct - we can chat about that on monday.

lib/backend/syncersv1/updateprocessors/profileprocessor.go
```diff
@@ -0,0 +1,236 @@
...
+
+	v1value := &model.Profile{
+		Rules:  rules,
+		Labels: v2res.Spec.LabelsToApply,
+	}
+
+	return v1value, nil
+}
+
+func RulesAPIV2ToBackend(ars []apiv2.Rule) []model.Rule {
```
robbrockbank: Might be nice to split these out into a rules.go or similar since they are shared by profiles and the different network policy processors.

robbrockbank: not too fussed though.

lib/backend/syncersv1/updateprocessors/profileprocessor.go
```diff
@@ -0,0 +1,236 @@
...
+}
+
+// Need to create custom logic for Profile since it breaks the values into 3 separate KV Pairs:
+// Tags, Labels, and Rules.
+type profileUpdateProcessor struct {
+	v2Kind string
+}
+
+func (pup *profileUpdateProcessor) Process(kvp *model.KVPair) ([]*model.KVPair, error) {
+	// Check the v2 resource is the correct type.
```
robbrockbank: I think this processing could be simplified a little by doing more of it inline rather than splitting into separate key/value functions.  That split doesn't work so well for the profile since it is composed of three things - and passing round slices of different keys and values that then need to be joined together feels a little awkward.

Also, since we don't need the tags any more, i'd judo drop both the key and value completely.

lib/backend/syncersv1/updateprocessors/processor.go
```diff
@@ -0,0 +1,94 @@
...
+// OnSyncerStarting methods do not have any special handling.
+//
+// One of the main differences between the simple update processor and the conflict
+// resolving name cache processor is that the conflict resolving name cache processor
+// uses a cache to handle conflicts when v1 objects cannot be built up using only the
+// resource key. The simple update processor is for converting objects that can create
+// the v1 key using only the information available in the resource key.
+
+type ConvertV2ToV1Key func(v2Key model.ResourceKey) (model.Key, error)
+type ConvertV2ToV1Value func(interface{}) (interface{}, error)
```
robbrockbank: I believe all of our v2 reources will implement the runtime.Object interface, so the method signature could be further improved to:

type ConvertV2ToV1Value func(runtime.Object) (interface{}, error)

lib/backend/syncersv1/updateprocessors/processor.go
```diff
@@ -0,0 +1,94 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	var v1value interface{}
+	// Deletion events will have a value of nil. Do not convert anything for a deletion event.
+	if kvp.Value != nil {
+		v1value, err = sup.valueConverter(kvp.Value)
+		if err != nil {
+			// Currently ignore the error so that incorrect values get skipped instead of erroring out
```
robbrockbank: We're not skipping - we're treating as a delete

We should add a warning log that we're doing this, the log should include the resource (I think you can just trace out the resource key as a string ...  `log.WithField("Resource": kvp.Key).Warn("Unable to process resource data - treating as deleted")` or sim.

lib/backend/syncersv1/updateprocessors/processor.go
mgleung: I thought that the KVPair Value was set to be an interface which means that the value needs to be typecast to runtime.Object right? Should we be typecasting this to the runtime.Object interface even though we are going to typecast this to the more specific resources right after?

robbrockbank: I guess it doesn't really matter - I think i was thinking that it further limits misuse.

I'm not fussed if that doesn't change.

lib/backend/syncersv1/updateprocessors/profileprocessor.go
mgleung: Ok, I was thinking that it might've been a little easier to read if I broke off the key components into their own similarly named helper functions, but I can see how slices or maps being returned can be detrimental for that readability and a little more of a pain to maintain.

robbrockbank:
LGTM.

Can you squash please.

PR #578, opened 2017-10-16
Validate the WEP name and assign from Spec fields if unassigned

robbrockbank:
Add some validation to the WEP name

### DIFF
--- a/lib/clientv2/workloadendpoint.go
+++ b/lib/clientv2/workloadendpoint.go
@@ -16,10 +16,13 @@ package clientv2
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/projectcalico/libcalico-go/lib/apiv2"
 	"github.com/projectcalico/libcalico-go/lib/options"
 	"github.com/projectcalico/libcalico-go/lib/watch"
+	"github.com/projectcalico/libcalico-go/lib/names"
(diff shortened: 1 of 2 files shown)


lib/clientv2/workloadendpoint.go
```diff
@@ -90,3 +99,34 @@ func (r workloadEndpoints) List(ctx context.Context, opts options.ListOptions) (
...
+		Endpoint: res.Spec.Endpoint,
+		Workload: res.Spec.Workload,
+		Pod: res.Spec.Pod,
+		ContainerID: res.Spec.ContainerID,
+	}
+	expectedName, err := wepids.CalculateWorkloadEndpointName(false)
+	if err != nil {
+		return err
+	}
+	if len(res.Name) == 0 {
```
gunjan5: curious why can't we do `res.Name == ""`? That seems more readable

gunjan5: i think it should have the same result but `len` is usually for a slice or array

robbrockbank: i think its what the k8s code does more often for empty string.
 should amount to the same thing

gunjan5: Ok, not too fussed about it. Just looks a bit odd

PR #579, opened 2017-10-17
Fix up CalicoAPIConfig definition to use new resource format

robbrockbank:
Update calico api config def.

### DIFF
--- a/lib/apiconfig/apiconfig.go
+++ b/lib/apiconfig/apiconfig.go
@@ -14,26 +14,27 @@
 
 package apiconfig
 
-import "github.com/projectcalico/libcalico-go/lib/api/unversioned"
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+	"github.com/projectcalico/libcalico-go/lib/apiv2"
+)
 type DatastoreType string
 
 const (
 	EtcdV2     DatastoreType = "etcdv2"
 	EtcdV3     DatastoreType = "etcdv3"
 	Kubernetes DatastoreType = "kubernetes"
+	KindCalicoAPIConfig     = "CalicoAPIConfig"
 )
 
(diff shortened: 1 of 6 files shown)


lib/apiconfig/load.go
```diff
@@ -53,8 +53,8 @@ func LoadClientConfigFromBytes(b []byte) (*CalicoAPIConfig, error) {
 	if c.APIVersion != apiv2.GroupVersionCurrent {
 		return nil, errors.New("invalid config file: unknown APIVersion '" + c.APIVersion + "'")
 	}
-	if c.Kind != "calicoApiConfig" {
-		return nil, errors.New("invalid config file: expected kind 'calicoApiConfig', got '" + c.Kind + "'")
+	if c.Kind != KindCalicoAPIConfig {
```
tmjd: This changes the case of the first `c` to `C`, does that matter?

lib/apiconfig/apiconfig.go
```diff
@@ -71,9 +72,9 @@ type KubeConfig struct {
 // TypeMetadata initialised to the current version.
 func NewCalicoAPIConfig() *CalicoAPIConfig {
 	return &CalicoAPIConfig{
-		TypeMetadata: unversioned.TypeMetadata{
-			Kind:       "calicoApiConfig",
-			APIVersion: unversioned.VersionCurrent,
+		TypeMeta: metav1.TypeMeta{
+			Kind:       KindCalicoAPIConfig ,
```
tmjd: Changed case of first letter.

robbrockbank: Yup - all our resources are now cap first letter

lib/apiconfig/load.go
robbrockbank: nope - as expected as part of the v2 client.

PR #582, opened 2017-10-17
[WIP] Add watcher support for KDD CRD backed resources

robbrockbank:
## Description
A few sentences describing the overall goals of the pull request's commits.
Please include
- the type of fix - (e.g. bug fix, new feature, documentation)
- some details on _why_ this PR should be merged
- the details of the testing you've done on it (both manual and automated)
- which components are affected by this PR
- links to issues that this PR addresses

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


lib/backend/k8s/resources/profile.go
```diff
@@ -0,0 +1,172 @@
...
+	"github.com/projectcalico/libcalico-go/lib/backend/k8s/conversion"
+	"github.com/projectcalico/libcalico-go/lib/backend/model"
+	cerrors "github.com/projectcalico/libcalico-go/lib/errors"
+)
+
+const (
+	profileBgpIpv4AddrAnnotation = "projectcalico.org/IPv4Address"
+	profileBgpIpv6AddrAnnotation = "projectcalico.org/IPv6Address"
+	profileBgpAsnAnnotation      = "projectcalico.org/ASNumber"
+)
```
doublek: These aren't needed for the profile resource.

lib/backend/k8s/conversion/conversion.go
```diff
@@ -135,7 +126,7 @@ func (c Converter) isScheduled(pod *kapiv1.Pod) bool {
 	return pod.Spec.NodeName != ""
 }
 
-func (c Converter) isHostNetworked(pod *kapiv1.Pod) bool {
+func (c Converter) IsHostNetworked(pod *kapiv1.Pod) bool {
```
caseydavenport: Why did you make this one public but not the others? e.g. `isScheduled`, etc?

lib/backend/k8s/k8s.go
```diff
@@ -130,18 +135,116 @@ func NewKubeClient(kc *apiconfig.KubeConfig) (api.Client, error) {
...
+func (c *KubeClient) getResourceClientFromList(list model.ListInterface) (resources.K8sResourceClient, error) {
+	var client resources.K8sResourceClient
+	lt := reflect.TypeOf(list)
+	if lt == resourceListType {
+		client = c.clientsByResourceKind[list.(model.ResourceListOptions).Kind]
+	} else {
+		client = c.clientsByListType[lt]
+	}
+	if client == nil {
+		return nil, fmt.Errorf("Unhandled resource type %v", lt)
```
caseydavenport: Given this error is propagated to the user, I think this should by an UnsupportedOperation or whatever, with a more user-facing message.

Fine to punt post-beta so long as there is an issue tracking it within the v3.0 milestone.

lib/backend/k8s/k8s.go
```diff
@@ -274,332 +377,97 @@ func (c *KubeClient) Syncer(callbacks api.SyncerCallbacks) api.Syncer {
...
 }
 
-// getWorkloadEndpoint gets the WorkloadEndpoint from the k8s API based on existing Pods.
-func (c *KubeClient) getWorkloadEndpoint(ctx context.Context, k model.ResourceKey, revision string) (*model.KVPair, error) {
-	// The workloadID is of the form namespace.podname.  Parse it so we
-	// can find the correct namespace to get the pod.
-	namespace, podName := c.converter.parseWorkloadID(k.Name)
-
-	pod, err := c.clientSet.Pods(namespace).Get(podName, metav1.GetOptions{})
+// List entries in the datastore.  This may return an empty list if there are
```
caseydavenport: Comment is wrong

lib/backend/k8s/k8s.go
```diff
@@ -648,7 +516,8 @@ func (c *KubeClient) listPolicies(ctx context.Context, l model.ResourceListOptio
 	}
 
 	// List all Global Network Policies.
-	gnps, err := c.gnpClient.List(ctx, l, revision)
+	gnpClient, _ := c.getResourceClientFromList(l)
```
caseydavenport: We should be handling this error

lib/backend/k8s/k8s.go
```diff
@@ -689,7 +558,8 @@ func (c *KubeClient) getPolicy(ctx context.Context, k model.ResourceKey, revisio
 		return c.converter.NetworkPolicyToPolicy(&networkPolicy)
 	} else {
 		// This is backed by a Global Network Policy CRD.
-		return c.gnpClient.Get(ctx, k, revision)
+		gnpClient, _ := c.getResourceClientFromKey(k)
```
caseydavenport: We should handle this error

lib/backend/k8s/resources/affinityblock.go
```diff
@@ -0,0 +1,136 @@
...
+		KVPairs:  []*model.KVPair{},
+		Revision: revision,
+	}
+
+	// If a host is specified, then do an exact lookup (ip version should not be expected in the query)
+	if bl.Host != "" && bl.IPVersion == 0 {
+		// Get the node settings, we use the nodes PodCIDR as the only node affinity block.
+		node, err := c.clientSet.Nodes().Get(bl.Host, metav1.GetOptions{})
+		if err != nil {
+			return nil, K8sErrorToCalico(err, list)
```
caseydavenport: This is weird - we've got some unreachable code here.

caseydavenport: I think the code that comes after this line just needs to be removed

lib/backend/k8s/resources/client.go
```diff
@@ -63,6 +56,11 @@ type K8sResourceClient interface {
 	// Non-zero fields in the struct are used as filters.
 	List(ctx context.Context, list model.ListInterface, revision string) (*model.KVPairList, error)
 
+	// List returns a slice of KVPairs matching the input list options.
```
caseydavenport: Comment is wrong

lib/backend/k8s/resources/customresource.go
```diff
@@ -125,55 +119,30 @@ func (c *customK8sResourceClient) Update(ctx context.Context, kvp *model.KVPair)
...
+	name := resIn.GetObjectMeta().GetName()
+	logContext = logContext.WithField("Name", name)
+	logContext.Debug("Update resource by name")
+	updateError = c.restClient.Put().
+		Resource(c.resource).
+		Body(resIn).
+		Name(name).
+		Do().Into(resOut)
+	if updateError == nil {
+		// Success.
```
caseydavenport: A few lines down: https://github.com/projectcalico/libcalico-go/pull/582/files#diff-054d1c999e6b6ec72d35a9f7122ab93dR151

This is all garbage code now, right? We don't need to do a `Get` after a failed Update?

lib/backend/k8s/resources/customresource.go
```diff
@@ -363,10 +299,31 @@ func (c *customK8sResourceClient) List(ctx context.Context, list model.ListInter
...
+	}
+
+	k8sWatchClient := cache.NewListWatchFromClient(
+		c.restClient,
+		c.resource,
+		resl.Namespace,
+		fields.Everything())
+	k8sWatch, err := k8sWatchClient.WatchFunc(metav1.ListOptions{ResourceVersion: revision})
+	if err != nil {
+		return nil, err
```
caseydavenport: Do we need to convert this error to a Calico error?

lib/backend/k8s/resources/node.go
```diff
@@ -175,3 +157,113 @@ func (c *nodeClient) List(ctx context.Context, list model.ListInterface, revisio
...
 }
+
+func (c *nodeClient) Watch(ctx context.Context, list model.ListInterface, revision string) (api.WatchInterface, error) {
+	if len(list.(model.ResourceListOptions).Name) != 0 {
+		return nil, fmt.Errorf("cannot watch specific resource instance: %s", list.(model.ResourceListOptions).Name)
+	}
+
+	k8sWatch, err := c.clientSet.Nodes().Watch(metav1.ListOptions{ResourceVersion: revision})
+	if err != nil {
+		return nil, err
```
caseydavenport: Need to convert to calico error?

lib/backend/k8s/resources/node_test.go
```diff
@@ -69,32 +69,14 @@ var _ = Describe("Test Node conversion", func() {
 		bgpIpv4Address := n.Value.(*apiv2.Node).Spec.BGP.IPv4Address
 		ipInIpAddr := n.Value.(*apiv2.Node).Spec.BGP.IPv4IPIPTunnelAddr
 		asn := n.Value.(*apiv2.Node).Spec.BGP.ASNumber
-		labels := n.Value.(*apiv2.Node).Labels
```
caseydavenport: What's the cause of this one?

lib/backend/k8s/conversion/conversion.go
robbrockbank: I just updated what I needed to get things working.

lib/backend/k8s/k8s.go
robbrockbank: Oooh yes it will be - i'll fix it up!

lib/backend/k8s/k8s.go
robbrockbank: this code is changing - so I just wanted to get it compiling.

lib/backend/k8s/k8s.go
robbrockbank: same with this one.

lib/backend/k8s/conversion/conversion.go
robbrockbank: but i can make em public.

lib/backend/k8s/resources/resources.go
```diff
@@ -0,0 +1,96 @@
...
+	log "github.com/sirupsen/logrus"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+
+	"github.com/projectcalico/libcalico-go/lib/backend/model"
+)
+
+const (
+	labelsAnnotation      = "projectcalico.org/Labels"
+	annotationsAnnotation = "projectcalico.org/Annotations"
```
caseydavenport: I think these should be lowercase - `projectcalico.org/labels`

lib/backend/k8s/k8s.go
robbrockbank: Actually I think I'll fix up to not return an error, but just return nil and the calling code can tailor the error appropriately.

lib/backend/k8s/resources/resources.go
```diff
@@ -0,0 +1,96 @@
...
+	runtime.Object
+	metav1.ListMetaAccessor
+}
+
+type ConvertK8sResourceToKVPair func(Resource) (*model.KVPair, error)
+
+// Store Calico Metadata in the k8s resource annotations for non-CRD backed resources.
+// Currently this just stores Annotations and Labels and drops all other metadata
+// attributes.
+func SetK8sAnnotationsFromCalicoMetadata(k8sRes Resource, calicoRes Resource) {
```
caseydavenport: Can't tell what's wrong but the syntax highlighting looks weird here in GitHub....

lib/backend/k8s/resources/resources.go
```diff
@@ -0,0 +1,96 @@
...
+	if a == nil {
+		a = make(map[string]string)
+	}
+	if labels := calicoRes.GetObjectMeta().GetLabels(); len(labels) > 0 {
+		if lann, err := json.Marshal(labels); err == nil {
+			a[labelsAnnotation] = string(lann)
+		}
+	}
+	if annotations := calicoRes.GetObjectMeta().GetAnnotations(); len(annotations) > 0 {
+		if aann, err := json.Marshal(annotations); err == nil {
```
caseydavenport: We shouldn't be throwing away the marshalling error here - we need to log it out at least.

Above as well

lib/backend/k8s/resources/resources.go
```diff
@@ -0,0 +1,96 @@
...
+// Extract the Calico resource Metadata from the k8s resource annotations for non-CRD
+// backed resources.  This extracts the Annotations and Labels stored as a annotation,
+// and fills in the CreationTimestamp and UID from the k8s resource.
+func SetCalicoMetadataFromK8sAnnotations(calicoRes Resource, k8sRes Resource) {
+	com := calicoRes.GetObjectMeta()
+	kom := k8sRes.GetObjectMeta()
+	com.SetResourceVersion(kom.GetResourceVersion())
+	com.SetCreationTimestamp(kom.GetCreationTimestamp())
+	a := kom.GetAnnotations()
+	if len(a) == 0 {
```
caseydavenport: This feels like a pointless optimization - nodes are almost always going to have annotations, and the code that follows isn't confusing / intensive. Could we just remove this?

lib/backend/k8s/k8s.go
robbrockbank: actually I've now removed the error - so this can return a nil interface.  I actually think in this case it's assertable that it returns something (since we know what we're putting in) so I'm not convinced we need to check that since we'll just end up panicking (which is fine IMHO).

lib/backend/k8s/resources/affinityblock.go
robbrockbank: d'oh!

lib/backend/k8s/resources/watcher.go
```diff
@@ -0,0 +1,155 @@
...
+)
+
+const (
+	resultsBufSize = 100
+)
+
+func newK8sWatcherConverter(ctx context.Context, converter ConvertK8sResourceToKVPair, k8sWatch kwatch.Interface) api.WatchInterface {
+	ctx, cancel := context.WithCancel(ctx)
+	return &k8sWatcherConverter{
+		converter:  converter,
```
caseydavenport: nitty..so feel free to ignore.. but the order of these doesn't match how they're defined in the struct which makes it a bit harder to follow.

lib/backend/k8s/resources/affinityblock.go
robbrockbank: nah the first return is wrong.  we should just be translating the error and then returning the error or an empty list depending on the converted error.

lib/backend/k8s/resources/client.go
robbrockbank: oops

lib/backend/k8s/resources/customresource.go
robbrockbank: grrr - yes - looks like i removed some but not all of the bits.  Thanks!

lib/backend/k8s/resources/customresource.go
robbrockbank: I think we should, yes!

lib/backend/k8s/resources/workloadendpoint.go
```diff
@@ -0,0 +1,192 @@
...
+		}
+		ret = append(ret, kvp)
+	}
+	return &model.KVPairList{
+		KVPairs:  ret,
+		Revision: revision,
+	}, nil
+}
+
+// getWorkloadEndpoint gets the WorkloadEndpoint from the k8s API based on existing Pods.
```
caseydavenport: @doublek you're doing all of this in your PR?

lib/backend/k8s/resources/node.go
robbrockbank: yup

lib/backend/k8s/syncer.go
```diff
@@ -125,7 +127,8 @@ func (k *realKubeAPI) GlobalNetworkPolicyWatch(opts metav1.ListOptions) (watch.I
 }
 
 func (k *realKubeAPI) GlobalNetworkPolicyList() ([]*model.KVPair, string, error) {
-	kvpl, err := k.kc.gnpClient.List(context.Background(), model.PolicyListOptions{}, "")
+	gnpClient, _ := k.kc.getResourceClientFromResourceKind(apiv2.KindGlobalNetworkPolicy)
```
caseydavenport: Think we need to handle this error too

caseydavenport: and the rest in this file that we're skipping

caseydavenport: OK if @doublek is doing that in another PR

lib/backend/k8s/k8s.go
robbrockbank: i'll fix.

lib/backend/k8s/resources/node.go
```diff
@@ -175,3 +157,113 @@ func (c *nodeClient) List(ctx context.Context, list model.ListInterface, revisio
...
+	return newK8sWatcherConverter(ctx, converter, k8sWatch), nil
+}
+
+// K8sNodeToCalico converts a Kubernetes format node, with Calico annotations, to a Calico Node.
+func K8sNodeToCalico(k8sNode *kapiv1.Node) (*model.KVPair, error) {
+	// Create a new CalicoNode resource and copy the settings across from the k8s Node.
+	calicoNode := apiv2.NewNode()
+	calicoNode.ObjectMeta.Name = k8sNode.Name
+	SetCalicoMetadataFromK8sAnnotations(calicoNode, k8sNode)
+	calicoNode.Spec = apiv2.NodeSpec{
```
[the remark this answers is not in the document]
robbrockbank: yeah i guess i could tighten that up.

lib/backend/k8s/resources/node_test.go
robbrockbank: because it wasn't correct as per the current plan of storing the labels and annotations as encoded annotations.

This assumed the labels were passed straight through to the Calico node resource.  A bug.

lib/backend/k8s/resources/resources.go
robbrockbank: the node BGP annotations are not.

lib/backend/k8s/resources/resources.go
robbrockbank: looks ok to me.

robbrockbank: oh i see yeah it's not highlighted... odd

lib/backend/k8s/resources/resources.go
robbrockbank: the len == 0 is *really* covering the == nil case but I thought I'd get two birds with one stone

lib/backend/k8s/resources/customresource.go
robbrockbank: hmmm actually - I'm not sure now

the etcdv3 code does return the snapshot on a failed update - which I basically copied from the k8s api server ... so I have a feeling we may need it for integration with the aapis.  I'll check.

lib/backend/k8s/resources/workloadendpoint.go
doublek: Yes.

caseydavenport:
I think this should be closed in favor of #587

PR #586, opened 2017-10-18
Relocate api and apiv2 into apis/v1 and apisv2 respectively

bcreane:
## Description
Relocate api and apiv2 into apis/v1 and apis/v2 respectively

## Todos
- [X] Tests

## Release Note

```release-note
None required
```


tmjd:
Are the pkg names (`v1` and `v2`) how K8s names its pkgs in the apis directory structure?

tmjd:
Maybe not as part of this PR but do we want to change client and clientv2? @robbrockbank @caseydavenport Would we want to mimic this same structure for client?

bcreane:
@tmjd - yeah. You'll see something like: 
 https://github.com/kubernetes/kubernetes/blob/master/cmd/kube-apiserver/app/testing/server_test.go#L28

They'll alias the import with the apigroup+version. Our apigroup is sort of "apis" which is a little weird, but its our only api.

PR #587, opened 2017-10-18
Start adding KDD watcher support

robbrockbank:
This adds code review markups and watcher tests to the original PR #582 
It also rebases off latest master and this PR is now against master.


lib/backend/k8s/k8s.go
```diff
@@ -263,20 +250,24 @@ func (c *KubeClient) EnsureInitialized() error {
 }
 
 func (c *KubeClient) Clean() error {
```
caseydavenport: Could you add a comment (now or later) to say this method is just for the tests?

caseydavenport:
@robbrockbank LGTM

PR #588, opened 2017-10-18
Use new client-go / k8s.io/api

bcreane:
Use new client-go / k8s.io/api

## Todos
- [X] Tests

## Release Note

```release-note
None required
```


lib/apis/v2/register.go
```diff
@@ -0,0 +1,42 @@
...
+
+// Resource takes an unqualified resource and returns a Group qualified GroupResource
+func Resource(resource string) schema.GroupResource {
+	return SchemeGroupVersion.WithResource(resource).GroupResource()
+}
+
+// Adds the list of known types to api.Scheme.
+func addKnownTypes(scheme *runtime.Scheme) error {
+	scheme.AddKnownTypes(SchemeGroupVersion,
+		&BGPConfiguration{},
```
caseydavenport: @robbrockbank @bcreane do we need these here at all? Do we need more here?

bcreane: looks like we don't need either BGPConfiguration{}, or BGPConfigurationList{}.
I can clear more out of register.go tomorrow, if that works for you?

bcreane: note that we do need most of register.go for linking to succeed (I'll push an update tomorrow that elides BGPConfiguration structs).

lib/apis/v2/bgpconfig.go
```diff
@@ -25,6 +25,8 @@ const (
 	KindBGPConfigurationList = "BGPConfigurationList"
 )
 
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
```
caseydavenport: I think we should have newlines between the generator tags and our other comments, throughout this PR.

bcreane: yep, agreed. putting together a PR right now.

bcreane: note that I squashed the commits, its newlines + stripped down register.go

lib/apis/v2/register.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+// Resource takes an unqualified resource and returns a Group qualified GroupResource
+func Resource(resource string) schema.GroupResource {
+	return SchemeGroupVersion.WithResource(resource).GroupResource()
+}
+
+// Adds the list of known types to api.Scheme.
+func addKnownTypes(scheme *runtime.Scheme) error {
+	scheme.AddKnownTypes(SchemeGroupVersion,
```
bcreane: ugh, this should probably just return `nil`

bcreane:
Thanks for merging Shaun.

PR #593, opened 2017-10-18
Add generate-files make target for generating deepcopy boilerplate.

bcreane:
## Description
Add a `generate-files` make target for updating the generated deepcopy file.

## Todos
- [x] Tests

## Release Note

```release-note
None required
```


tmjd:
Looks good. Is the plan that this will only be ran manually and then the output will be checked in?

bcreane:
@tmjd - that's the plan, but we could actually add `generate-files` to the `all` target easily. The target only takes a few seconds to build and the zz_generated*.go only changes when there's a substantive change to the api.

PR #594, opened 2017-10-18
Fix workload endpoints in KDD mode

doublek:
## Description

Adds and fixes support for WorkloadEndpoints. 

- Adds Get/List support.
- Fixes WEP name parsing
- Adds watcher boilerplate

The thing that I am not 100% sure is this: earlier we used on `Apply()`, we used to patch the k8s `Pod` with an IP address. I've put this as part of the `Update` call.

## Todos
- [ ] Tests
- [x] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/backend/k8s/conversion/conversion.go
+++ b/lib/backend/k8s/conversion/conversion.go
@@ -27,6 +27,7 @@ import (
 
 	apiv2 "github.com/projectcalico/libcalico-go/lib/apis/v2"
 	"github.com/projectcalico/libcalico-go/lib/backend/model"
+	"github.com/projectcalico/libcalico-go/lib/names"
 	cnet "github.com/projectcalico/libcalico-go/lib/net"
 	"github.com/projectcalico/libcalico-go/lib/numorstring"
 	extensions "k8s.io/api/extensions/v1beta1"
@@ -51,10 +52,11 @@ func VethNameForWorkload(workload string) string {
 	return fmt.Sprintf("cali%s", hex.EncodeToString(h.Sum(nil))[:11])
 }
 
-// ParseWorkloadID extracts the Namespace and Pod name from the given workload ID.
-func (c Converter) ParseWorkloadID(workloadID string) (string, string) {
-	splits := strings.SplitN(workloadID, ".", 2)
-	return splits[0], splits[1]
+// ParseWorkloadName extracts the Node name, Orchestrator, Pod name and endpoint from the
+// given WorkloadEndpoint name.
+// The expected format for k8s is <node>-k8s-<pod>-<endpoint>
+func (c Converter) ParseWorkloadEndpointName(workloadName string) (names.WorkloadEndpointIdentifiers, error) {
+	return names.ParseWorkloadEndpointName(workloadName)
 }
 
 // ParsePolicyNameNetworkPolicy extracts the Kubernetes Namespace and NetworkPolicy that backs the given Policy.
@@ -136,10 +138,21 @@ func (c Converter) HasIPAddress(pod *kapiv1.Pod) bool {
 
 // PodToWorkloadEndpoint converts a Pod to a WorkloadEndpoint.  It assumes the calling code
 // has verified that the provided Pod is valid to convert to a WorkloadEndpoint.
+// PodToWorkloadEndpoint requires a Pods Name and Node Name to be populated. It will
+// fail to convert from a Pod to WorkloadEndpoint otherwise.
 func (c Converter) PodToWorkloadEndpoint(pod *kapiv1.Pod) (*model.KVPair, error) {
 	// Pull out the profile and workload ID based on pod name and Namespace.
 	profile := fmt.Sprintf("k8s_ns.%s", pod.ObjectMeta.Namespace)
-	workload := fmt.Sprintf("%s.%s", pod.ObjectMeta.Namespace, pod.ObjectMeta.Name)
+	wepids := names.WorkloadEndpointIdentifiers{
+		Node:         pod.Spec.NodeName,
+		Orchestrator: "k8s",
+		Endpoint:     "eth0",
+		Pod:          pod.ObjectMeta.Name,
+	}
+	wepName, err := wepids.CalculateWorkloadEndpointName(false)
+	if err != nil {
+		return nil, err
+	}
 
 	// We do, in some circumstances, want to parse Pods without an IP address.  For example,
 	// a DELETE update will not include an IP.
@@ -153,9 +166,9 @@ func (c Converter) PodToWorkloadEndpoint(pod *kapiv1.Pod) (*model.KVPair, error)
 		ipNets = append(ipNets, ipNet.String())
 	}
 
-	// Generate the interface name and MAC based on workload.  This must match
+	// Generate the interface name based on workload.  This must match
 	// the host-side veth configured by the CNI plugin.
-	interfaceName := VethNameForWorkload(workload)
+	interfaceName := VethNameForWorkload(wepName)
 
 	// Build the labels map.
 	labels := map[string]string{}
@@ -196,19 +209,20 @@ func (c Converter) PodToWorkloadEndpoint(pod *kapiv1.Pod) (*model.KVPair, error)
 	// Create the key / value pair to return.
 	kvp := model.KVPair{
 		Key: model.ResourceKey{
-			Name: pod.ObjectMeta.Name,
-			Kind: apiv2.KindWorkloadEndpoint,
+			Name:      wepName,
+			Namespace: pod.ObjectMeta.Namespace,
+			Kind:      apiv2.KindWorkloadEndpoint,
 		},
 		Value: &apiv2.WorkloadEndpoint{
 			ObjectMeta: metav1.ObjectMeta{
-				Name:      pod.ObjectMeta.Name,
+				Name:      wepName,
 				Namespace: pod.ObjectMeta.Namespace,
 				Labels:    labels,
 			},
 			Spec: apiv2.WorkloadEndpointSpec{
 				Orchestrator:  "k8s",
 				Node:          pod.Spec.NodeName,
-				Workload:      workload,
+				Pod:           pod.Name,
 				Endpoint:      "eth0",
 				InterfaceName: interfaceName,
 				Profiles:      []string{profile},
--- a/lib/backend/k8s/conversion/conversion_test.go
+++ b/lib/backend/k8s/conversion/conversion_test.go
@@ -33,11 +33,14 @@ var _ = Describe("Test parsing strings", func() {
 	// Use a single instance of the Converter for these tests.
 	c := Converter{}
 
-	It("should parse workloadIDs", func() {
-		workloadName := "Namespace.podName"
-		ns, podName := c.ParseWorkloadID(workloadName)
-		Expect(ns).To(Equal("Namespace"))
-		Expect(podName).To(Equal("podName"))
+	It("should parse WorkloadEndpoint name", func() {
+		wepName := "node-k8s-pod--name-eth0"
+		weid, err := c.ParseWorkloadEndpointName(wepName)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(weid.Node).To(Equal("node"))
+		Expect(weid.Orchestrator).To(Equal("k8s"))
+		Expect(weid.Endpoint).To(Equal("eth0"))
+		Expect(weid.Pod).To(Equal("pod-name"))
 	})
 
 	It("should parse valid policy names", func() {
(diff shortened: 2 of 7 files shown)


lib/backend/k8s/conversion/conversion_test.go
```diff
@@ -143,16 +145,18 @@ var _ = Describe("Test Pod conversion", func() {
...
 		// Assert key fields.
-		Expect(wep.Key.(model.ResourceKey).Name).To(Equal("podA"))
+		Expect(wep.Key.(model.ResourceKey).Name).To(Equal("nodeA-k8s-podA-eth0"))
+		Expect(wep.Key.(model.ResourceKey).Namespace).To(Equal("default"))
 		Expect(wep.Key.(model.ResourceKey).Kind).To(Equal(apiv2.KindWorkloadEndpoint))
 
-		Expect(wep.Value.(*apiv2.WorkloadEndpoint).Spec.Workload).To(Equal("default.podA"))
+		// TODO(doublek): Figure out the correct workload name.
+		//Expect(wep.Value.(*apiv2.WorkloadEndpoint).Spec.Workload).To(Equal("default.podA"))
+		Expect(wep.Value.(*apiv2.WorkloadEndpoint).Spec.Workload).To(Equal("nodeA-k8s-podA-eth0"))
```
doublek: @robbrockbank Can you confirm which one is the expected value for `WorkloadEndpointSpec.Workload`.

lib/backend/k8s/resources/workloadendpoint.go
```diff
@@ -47,11 +52,29 @@ func (c *WorkloadEndpointClient) Create(ctx context.Context, kvp *model.KVPair)
 }
 
 func (c *WorkloadEndpointClient) Update(ctx context.Context, kvp *model.KVPair) (*model.KVPair, error) {
-	log.Warn("Operation Update is not supported on WorkloadEndpoint type")
-	return nil, cerrors.ErrorOperationNotSupported{
-		Identifier: kvp.Key,
-		Operation:  "Create",
+	log.Debug("Received Update request on WorkloadEndpoint type")
```
doublek: I wasn't sure where to this patching of pod IP should end up. `Update` made the most sense to me and I've put it here. Let me know if this needs to be somewhere else.

lib/backend/k8s/conversion/conversion_test.go
robbrockbank: Neither - the workload isn't required for k8s - we have the Pod field instead.

Workload is used for non-cni/k8s deployments, where we don't really know what the workload is.

doublek: @robbrockbank So it is okay to leave it empty? Rather, it is expected to be empty in the case of k8s?

lib/backend/k8s/conversion/conversion_test.go
```diff
@@ -215,19 +219,20 @@ var _ = Describe("Test Pod conversion", func() {
 		Expect(err).NotTo(HaveOccurred())
 
 		// Assert key fields.
-		Expect(wep.Key.(model.ResourceKey).Name).To(Equal("podA"))
+		Expect(wep.Key.(model.ResourceKey).Name).To(Equal("nodeA-k8s-podA-eth0"))
+		Expect(wep.Key.(model.ResourceKey).Namespace).To(Equal("default"))
 		Expect(wep.Key.(model.ResourceKey).Kind).To(Equal(apiv2.KindWorkloadEndpoint))
 		// Assert value fields.
-		Expect(wep.Value.(*apiv2.WorkloadEndpoint).Spec.Workload).To(Equal("default.podA"))
+		//Expect(wep.Value.(*apiv2.WorkloadEndpoint).Spec.Workload).To(Equal("default.podA"))
```
caseydavenport: Make sure to uncomment or remove this line before merging

lib/backend/k8s/conversion/conversion_test.go
```diff
@@ -215,19 +219,20 @@ var _ = Describe("Test Pod conversion", func() {
...
 		Expect(wep.Value.(*apiv2.WorkloadEndpoint).Spec.Node).To(Equal("nodeA"))
 		Expect(wep.Value.(*apiv2.WorkloadEndpoint).Spec.Endpoint).To(Equal("eth0"))
 		Expect(wep.Value.(*apiv2.WorkloadEndpoint).Spec.Orchestrator).To(Equal("k8s"))
 		Expect(len(wep.Value.(*apiv2.WorkloadEndpoint).Spec.IPNetworks)).To(Equal(1))
 		Expect(wep.Value.(*apiv2.WorkloadEndpoint).ObjectMeta.Labels).To(Equal(map[string]string{"calico/k8s_ns": "default"}))
-		//Expect(wep.Value.(*apiv2.WorkloadEndpoint).State).To(Equal("active"))
 	})
 
-	It("should Parse a Pod with no NodeName", func() {
+	It("should not parse a Pod with no NodeName", func() {
```
caseydavenport: Hm, this is worrying.

I think it's probably fine, but I don't know off the top of my head which logic relies on this test passing as it is today. This could very well break something down the line.

lib/backend/k8s/resources/workloadendpoint.go
caseydavenport: I think Update is the right place.

lib/backend/k8s/conversion/conversion_test.go
doublek: I jumped the gun with "fixing" this test. At a minimum I should add a note about this restriction in the appropriate conversion method.
@caseydavenport Will our different clients be interested in knowing about a `Pod` that isn't assigned to a `Node` yet?

 @robbrockbank The `WorkloadEndpoint` name goes something like `<node>-k8s-<pod>-<endpoint>`. Looks like there may be cases when a Pod exists without a `Node`. Do you have any thoughts on this?

robbrockbank: Previously the WEP had the node as an explicit index, so in a way nothing has really changed.

lib/backend/k8s/conversion/conversion_test.go
robbrockbank: yes empty for k8s - the Pod field should be filled in though.

doublek: Got it. Will take care of it.

lib/backend/k8s/conversion/conversion.go
```diff
@@ -51,10 +52,12 @@ func VethNameForWorkload(workload string) string {
...
-// ParseWorkloadID extracts the Namespace and Pod name from the given workload ID.
-func (c Converter) ParseWorkloadID(workloadID string) (string, string) {
-	splits := strings.SplitN(workloadID, ".", 2)
-	return splits[0], splits[1]
+// ParseWorkloadName extracts the Node name, Orchestrator, Pod name and endpoint from the
+// given WorkloadEndpoint Resource name.
+// The expected format for k8s is <node>-k8s-<pod>-<endpoint>
+func (c Converter) ParseWorkloadName(workloadName string) (string, string, string, string) {
+	parts := names.ExtractDashSeparatedParms(workloadName, 4)
+	return parts[0], parts[1], parts[2], parts[3]
```
robbrockbank: CNI actually does a "prefix" query to find the current endpoint in a workload.

In this case it's possible to get only 3 parts to the name.  In any case this feels like it should error if it can't deconstruct the name otherwise won't it panic?

robbrockbank: i.e. For a List(), if the prefix flag is set in the ListOptions, then we expect 3 parameters.  For all other actions - we expect all 4.

lib/backend/k8s/conversion/conversion.go
```diff
@@ -196,19 +210,20 @@ func (c Converter) PodToWorkloadEndpoint(pod *kapiv1.Pod) (*model.KVPair, error)
 	// Create the key / value pair to return.
 	kvp := model.KVPair{
 		Key: model.ResourceKey{
-			Name: pod.ObjectMeta.Name,
-			Kind: apiv2.KindWorkloadEndpoint,
+			Name:      workload,
```
robbrockbank: can we use name, or wepName instead of workload?

lib/backend/k8s/conversion/conversion.go
robbrockbank: I'm wondering if we should beef up the WorkloadEndpointIdentifiers thing in /lib/names so that the values can be assigned from the name - and that way you use that and pass around a WorkloadEndpointIdentifiers struct rather than returning multiple values.

doublek: Yup. I was thinking along similar lines. It becomes unwieldy to handle the error otherwise.

lib/backend/k8s/resources/workloadendpoint.go
```diff
@@ -171,22 +161,25 @@ func (c *KubeClient) listWorkloadEndpoints(ctx context.Context, l model.Resource
...
-	// can find the correct namespace to get the pod.
-	namespace, podName := c.converter.ParseWorkloadID(k.Name)
+func (c *WorkloadEndpointClient) EnsureInitialized() error {
+	return nil
+}
 
-	pod, err := c.clientSet.Pods(namespace).Get(podName, metav1.GetOptions{})
-	if err != nil {
-		return nil, resources.K8sErrorToCalico(err, k)
+func (c *WorkloadEndpointClient) Watch(ctx context.Context, list model.ListInterface, revision string) (api.WatchInterface, error) {
```
robbrockbank: Thanks for doing the Watch!

lib/backend/k8s/resources/workloadendpoint.go
```diff
@@ -171,22 +161,25 @@ func (c *KubeClient) listWorkloadEndpoints(ctx context.Context, l model.Resource
...
+	if len(list.(model.ResourceListOptions).Name) != 0 {
+		return nil, fmt.Errorf("cannot watch specific resource instance: %s", list.(model.ResourceListOptions).Name)
 	}
 
-	// Decide if this pod should be displayed.
-	if !c.converter.IsReadyCalicoPod(pod) {
-		return nil, cerrors.ErrorResourceDoesNotExist{Identifier: k}
+	k8sWatch, err := c.clientSet.CoreV1().Pods("").Watch(metav1.ListOptions{ResourceVersion: revision})
```
robbrockbank: You need to pass in the namespace from the ListOptions here.  (granted it will be blank for the Syncer use case)

lib/backend/k8s/resources/workloadendpoint.go
```diff
@@ -146,9 +136,9 @@ func (c *KubeClient) listWorkloadEndpoints(ctx context.Context, l model.Resource
 
 	// Otherwise, enumerate all pods in all namespaces.
 	// We don't yet support hostname, orchestratorID, for the k8s backend.
-	pods, err := c.clientSet.Pods("").List(metav1.ListOptions{})
+	pods, err := c.clientSet.CoreV1().Pods("").List(metav1.ListOptions{})
```
robbrockbank: Need to pass in Namespace from the ListOptions

lib/backend/k8s/resources/workloadendpoint.go
```diff
@@ -47,11 +52,29 @@ func (c *WorkloadEndpointClient) Create(ctx context.Context, kvp *model.KVPair)
...
+	log.Debug("Received Update request on WorkloadEndpoint type")
+
+	// We only patch the existing Pod to include an IP address, if one has been
+	// set on the workload endpoint.
+	// TODO: This is only required as a workaround for an upstream k8s issue.  Once fixed,
+	// this should be a no-op. See https://github.com/kubernetes/kubernetes/issues/39113
+	ips := kvp.Value.(*apiv2.WorkloadEndpoint).Spec.IPNetworks
+	if len(ips) > 0 {
+		log.Debugf("Applying workload with IPs: %+v", ips)
+		_, _, ns, name := c.converter.ParseWorkloadName(kvp.Key.(model.ResourceKey).Name)
```
doublek: This is wrong. ns needs to extracted from key.

lib/backend/k8s/resources/workloadendpoint.go
```diff
@@ -47,11 +52,36 @@ func (c *WorkloadEndpointClient) Create(ctx context.Context, kvp *model.KVPair)
...
+		if weid.Pod == "" {
+			return nil, cerrors.ErrorInsufficientIdentifiers{Name: kvp.Key.(model.ResourceKey).Name}
+		}
+		ns := kvp.Key.(model.ResourceKey).Namespace
+		pod, err := c.clientSet.CoreV1().Pods(ns).Get(weid.Pod, metav1.GetOptions{})
+		if err != nil {
+			return nil, K8sErrorToCalico(err, kvp.Key)
+		}
+		pod.Status.PodIP = ips[0]
+		pod, err = c.clientSet.CoreV1().Pods(ns).UpdateStatus(pod)
```
caseydavenport: We can let this slide for now, but this really needs to be updated to use `Patch` - right now we're going to be overwriting other people's status updates if there are any.  There's already an issue to track, but just noting here to raise awareness!

lib/backend/k8s/resources/workloadendpoint.go
```diff
@@ -63,65 +93,35 @@ func (c *WorkloadEndpointClient) Delete(ctx context.Context, key model.Key, revi
...
+	log.Debug("Received Get request on WorkloadEndpoint type")
+	k := key.(model.ResourceKey)
 
-func (c *WorkloadEndpointClient) List(ctx context.Context, list model.ListInterface, revision string) (*model.KVPairList, error) {
-	log.Warn("Operation List is not supported on WorkloadEndpoint type")
-	return nil, cerrors.ErrorOperationNotSupported{
-		Identifier: list,
-		Operation:  "List",
+	// Parse resource name so we can get get the podName
+	weid, err := c.converter.ParseWorkloadName(key.(model.ResourceKey).Name)
```
caseydavenport: We usually call it `wepID`

lib/backend/k8s/conversion/conversion.go
```diff
@@ -196,19 +209,20 @@ func (c Converter) PodToWorkloadEndpoint(pod *kapiv1.Pod) (*model.KVPair, error)
...
-				Name:      pod.ObjectMeta.Name,
+				Name:      wepName,
 				Namespace: pod.ObjectMeta.Namespace,
 				Labels:    labels,
 			},
 			Spec: apiv2.WorkloadEndpointSpec{
 				Orchestrator:  "k8s",
 				Node:          pod.Spec.NodeName,
-				Workload:      workload,
+				Pod:           pod.Name,
```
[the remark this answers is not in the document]
robbrockbank: but the resource is namespaced so why does this name need to be namespaced?  I don't think it does

I think the converter that maps this to the v1 model needs to prepend the namespace to the name perhaps with a / -- since Felix is not restrictive on the use of /

lib/backend/k8s/resources/workloadendpoint.go
doublek: Regarding not parsing a pod with no Nodename - To ensure this behavior, I'll  check if such a FV test exist and if not add one (to the soon to be restored test file).

lib/backend/k8s/conversion/conversion.go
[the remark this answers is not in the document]
caseydavenport: make sense.

PR #596, opened 2017-10-19
Add the logic to set UID and Creation Time on the resources, add storage prefixes on policies, and store Calico metadata in the K8s annotations.

mgleung:
## Description
Added the handling for UID and CreationTimestamp to be added to the resources. Also added logic for prefixing the names internally on policies as well as logic for storing Calico metadata in the K8s annotations field.

## Todos
- [x] Tests
- [x] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/backend/etcdv3/etcdv3.go
+++ b/lib/backend/etcdv3/etcdv3.go
@@ -85,6 +85,7 @@ func NewEtcdV3Client(config *apiconfig.EtcdConfig) (api.Client, error) {
 func (c *etcdV3Client) Create(ctx context.Context, d *model.KVPair) (*model.KVPair, error) {
 	logCxt := log.WithFields(log.Fields{"model-etcdKey": d.Key, "value": d.Value, "ttl": d.TTL, "rev": d.Revision})
 	logCxt.Debug("Processing Create request")
+
 	key, value, err := getKeyValueStrings(d)
 	if err != nil {
 		return nil, err
--- a/lib/backend/k8s/resources/customresource.go
+++ b/lib/backend/k8s/resources/customresource.go
@@ -79,8 +79,15 @@ func (c *customK8sResourceClient) Create(ctx context.Context, kvp *model.KVPair)
 		return nil, K8sErrorToCalico(err, kvp.Key)
 	}
 
+	// Update the return data with the metadata populated by the (Kubernetes) datastore.
+	kvp, err = c.convertResourceToKVPair(resOut)
+	if err != nil {
+		logContext.WithError(err).Info("Error converting created K8s resource to Calico resource")
+		return nil, K8sErrorToCalico(err, kvp.Key)
+	}
 	// Update the revision information from the response.
 	kvp.Revision = resOut.GetObjectMeta().GetResourceVersion()
+
 	return kvp, nil
 }
 
@@ -122,10 +129,15 @@ func (c *customK8sResourceClient) Update(ctx context.Context, kvp *model.KVPair)
 		return nil, K8sErrorToCalico(updateError, kvp.Key)
 	}
 
-	resOut.GetObjectKind().SetGroupVersionKind(c.k8sResourceTypeMeta.GetObjectKind().GroupVersionKind())
-
+	// Update the return data with the metadata populated by the (Kubernetes) datastore.
+	kvp, err = c.convertResourceToKVPair(resOut)
+	if err != nil {
+		logContext.WithError(err).Info("Error converting created K8s resource to Calico resource")
+		return nil, K8sErrorToCalico(err, kvp.Key)
+	}
 	// Success. Update the revision information from the response.
 	kvp.Revision = resOut.GetObjectMeta().GetResourceVersion()
+
 	return kvp, nil
 }
 
@@ -198,8 +210,6 @@ func (c *customK8sResourceClient) Get(ctx context.Context, key model.Key, revisi
 		logContext.WithError(err).Info("Error getting resource")
 		return nil, K8sErrorToCalico(err, key)
 	}
-	resOut.GetObjectKind().SetGroupVersionKind(c.k8sResourceTypeMeta.GetObjectKind().GroupVersionKind())
-
 	return c.convertResourceToKVPair(resOut)
 }
 
@@ -276,8 +286,6 @@ func (c *customK8sResourceClient) List(ctx context.Context, list model.ListInter
 	items := reflect.ValueOf(elem.FieldByName("Items").Interface())
 	for idx := 0; idx < items.Len(); idx++ {
 		res := items.Index(idx).Addr().Interface().(Resource)
-		res.GetObjectKind().SetGroupVersionKind(c.k8sResourceTypeMeta.GetObjectKind().GroupVersionKind())
-
 		if kvp, err := c.convertResourceToKVPair(res); err == nil {
 			kvps = append(kvps, kvp)
 		} else {
@@ -309,7 +317,6 @@ func (c *customK8sResourceClient) Watch(ctx context.Context, list model.ListInte
 		return nil, K8sErrorToCalico(err, list)
 	}
 	toKVPair := func(r Resource) (*model.KVPair, error) {
-		r.GetObjectKind().SetGroupVersionKind(c.k8sResourceTypeMeta.GetObjectKind().GroupVersionKind())
 		return c.convertResourceToKVPair(r)
 	}
 	return newK8sWatcherConverter(ctx, resl.Kind+" (custom)", toKVPair, k8sWatch), nil
@@ -341,22 +348,31 @@ func (c *customK8sResourceClient) nameToKey(name string) (model.Key, error) {
 }
 
 func (c *customK8sResourceClient) convertResourceToKVPair(r Resource) (*model.KVPair, error) {
+	r.GetObjectKind().SetGroupVersionKind(c.k8sResourceTypeMeta.GetObjectKind().GroupVersionKind())
 	kvp := &model.KVPair{
 		Key: model.ResourceKey{
 			Name:      r.GetObjectMeta().GetName(),
 			Namespace: r.GetObjectMeta().GetNamespace(),
 			Kind:      c.resourceKind,
 		},
-		Value:    r,
 		Revision: r.GetObjectMeta().GetResourceVersion(),
 	}
 
+	if err := ConvertK8sResourceToCalicoResource(r); err != nil {
+		return kvp, err
+	}
+
+	kvp.Value = r
 	return kvp, nil
 }
 
 func (c *customK8sResourceClient) convertKVPairToResource(kvp *model.KVPair) (Resource, error) {
 	resource := kvp.Value.(Resource)
 	resource.GetObjectMeta().SetResourceVersion(kvp.Revision)
+	resOut, err := ConvertCalicoResourceToK8sResource(resource)
+	if err != nil {
+		return resOut, err
+	}
 
-	return resource, nil
+	return resOut, nil
 }
--- a/lib/backend/k8s/resources/customresource_test.go
+++ b/lib/backend/k8s/resources/customresource_test.go
@@ -25,7 +25,7 @@ import (
 	. "github.com/onsi/gomega"
 )
 
-var _ = Describe("BGP Peer conversion methods", func() {
+var _ = Describe("Custom resource conversion methods (tested using BGPPeer)", func() {
 	// Create an empty client since we are only testing conversion functions.
 	client := NewBGPPeerClient(nil, nil).(*customK8sResourceClient)
 
@@ -55,6 +55,7 @@ var _ = Describe("BGP Peer conversion methods", func() {
 	// Compatible set of KVPair and Kubernetes Resource.
 	value1 := apiv2.NewBGPPeer()
 	value1.ObjectMeta.Name = name1
+	value1.ObjectMeta.ResourceVersion = "rv"
 	value1.Spec = apiv2.BGPPeerSpec{
 		PeerIP:   peerIP1.String(),
 		ASNumber: 1212,
@@ -107,14 +108,27 @@ var _ = Describe("BGP Peer conversion methods", func() {
 		Expect(r.GetObjectMeta().GetResourceVersion()).To(Equal(res1.ObjectMeta.ResourceVersion))
 		Expect(r).To(BeAssignableToTypeOf(&apiv2.BGPPeer{}))
 		Expect(r.(*apiv2.BGPPeer).Spec).To(Equal(res1.Spec))
+
+		// Make sure to clean up the annotations on the resource
+		err = ConvertK8sResourceToCalicoResource(r)
+		Expect(err).NotTo(HaveOccurred())
 	})
 
 	It("should convert between a Kuberenetes resource and the equivalent KVPair", func() {
+		// Make sure it returns the same resource if a resource is given without Calico metadata annotations.
 		kvp, err := client.convertResourceToKVPair(res1)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(kvp.Value).To(Equal(res1))
+
+		// Convert the values into the annotations.
+		resConverted, err := ConvertCalicoResourceToK8sResource(res1)
+		Expect(err).NotTo(HaveOccurred())
+
+		kvp, err = client.convertResourceToKVPair(resConverted)
 		Expect(err).NotTo(HaveOccurred())
 		Expect(kvp.Key).To(Equal(kvp1.Key))
 		Expect(kvp.Revision).To(Equal(kvp1.Revision))
-		Expect(kvp.Value.(*apiv2.BGPPeer)).To(BeAssignableToTypeOf(&apiv2.BGPPeer{}))
+		Expect(kvp.Value).To(BeAssignableToTypeOf(&apiv2.BGPPeer{}))
 		Expect(kvp.Value).To(Equal(kvp1.Value))
 	})
 })
--- a/lib/backend/k8s/resources/resources.go
+++ b/lib/backend/k8s/resources/resources.go
@@ -27,6 +27,7 @@ import (
 const (
 	labelsAnnotation      = "projectcalico.org/labels"
 	annotationsAnnotation = "projectcalico.org/annotations"
+	metadataAnnotation    = "projectcalico.org/metadata"
 )
 
 // Interface that all Kubernetes and Calico resources implement.
@@ -78,6 +79,7 @@ func SetCalicoMetadataFromK8sAnnotations(calicoRes Resource, k8sRes Resource) {
 	kom := k8sRes.GetObjectMeta()
 	com.SetResourceVersion(kom.GetResourceVersion())
 	com.SetCreationTimestamp(kom.GetCreationTimestamp())
+	com.SetUID(kom.GetUID())
 	a := kom.GetAnnotations()
 	if a == nil {
 		return
@@ -99,4 +101,88 @@ func SetCalicoMetadataFromK8sAnnotations(calicoRes Resource, k8sRes Resource) {
 			com.SetAnnotations(annotations)
 		}
 	}
+}
+
+// Store Calico Metadata in the in the k8s resource annotations for CRD backed resources.
+// This should store all Metadata except for those stored in Annotations and Labels and
+// store them in annotations.
+func ConvertCalicoResourceToK8sResource(resIn Resource) (Resource, error) {
+	rom := resIn.GetObjectMeta()
+
+	// Make sure to remove data that is passed to Kubernetes so it is not duplicated in
+	// the annotations.
+	romCopy := &metav1.ObjectMeta{}
+	rom.(*metav1.ObjectMeta).DeepCopyInto(romCopy)
+	romCopy.Name = ""
+	romCopy.Namespace = ""
+	romCopy.ResourceVersion = ""
+	romCopy.Labels = nil
+	romCopy.Annotations = nil
+
+	// Marshal the data and store the json representation in the annotations.
+	metadataBytes, err := json.Marshal(romCopy)
+	if err != nil {
+		return nil, err
+	}
+	annotations := rom.GetAnnotations()
+	if len(annotations) == 0 {
+		annotations = make(map[string]string)
+	}
+	annotations[metadataAnnotation] = string(metadataBytes)
+
+	// Make sure to clear out all of the Calico Metadata out of the ObjectMeta except for
+	// Name, Namespace, ResourceVersion, Labels, and Annotations. Annotations is already
+	// copied so it can be set separately.
+	meta := &metav1.ObjectMeta{}
+	meta.Name = rom.GetName()
+	meta.Namespace = rom.GetNamespace()
+	meta.ResourceVersion = rom.GetResourceVersion()
+	meta.Labels = rom.GetLabels()
+
+	resOut := resIn.DeepCopyObject().(Resource)
+	romOut := resOut.GetObjectMeta()
+	meta.DeepCopyInto(romOut.(*metav1.ObjectMeta))
+	romOut.SetAnnotations(annotations)
+
+	return resOut, nil
+}
+
+// Retrieve all of the Calico Metadata from the k8s resource annotations for CRD backed
+// resources. This should remove the relevant Calico Metadata annotation when it has finished.
+func ConvertK8sResourceToCalicoResource(res Resource) error {
+	rom := res.GetObjectMeta()
+	annotations := rom.GetAnnotations()
+	if len(annotations) == 0 {
+		// Make no changes if there are no annotations to read Calico Metadata out of.
+		return nil
+	}
+	if _, ok := annotations[metadataAnnotation]; !ok {
+		// No changes if there are no annotations stored on the Resource.
+		return nil
+	}
+
+	meta := &metav1.ObjectMeta{}
+	err := json.Unmarshal([]byte(annotations[metadataAnnotation]), meta)
+	if err != nil {
+		return err
+	}
+
+	// Clear out the annotations
+	delete(annotations, metadataAnnotation)
+	if len(annotations) == 0 {
+		annotations = nil
+	}
+
+	// Manually write in the data not stored in the annotations: Name, Namespace, ResourceVersion,
+	// Labels, and Annotations so that they do not get overwritten.
+	meta.Name = rom.GetName()
+	meta.Namespace = rom.GetNamespace()
+	meta.ResourceVersion = rom.GetResourceVersion()
+	meta.Labels = rom.GetLabels()
+	meta.Annotations = annotations
+
+	// Overwrite the K8s metadata with the Calico metadata.
+	meta.DeepCopyInto(rom.(*metav1.ObjectMeta))
+
+	return nil
 }
--- a/lib/clientv2/globalnetworkpolicy.go
+++ b/lib/clientv2/globalnetworkpolicy.go
@@ -16,6 +16,7 @@ package clientv2
 
 import (
 	"context"
+	"strings"
 
 	apiv2 "github.com/projectcalico/libcalico-go/lib/apis/v2"
 	"github.com/projectcalico/libcalico-go/lib/options"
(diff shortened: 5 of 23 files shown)


lib/backend/etcdv3/etcdv3.go
```diff
@@ -85,6 +87,17 @@ func NewEtcdV3Client(config *apiconfig.EtcdConfig) (api.Client, error) {
 func (c *etcdV3Client) Create(ctx context.Context, d *model.KVPair) (*model.KVPair, error) {
 	logCxt := log.WithFields(log.Fields{"model-etcdKey": d.Key, "value": d.Value, "ttl": d.TTL, "rev": d.Revision})
 	logCxt.Debug("Processing Create request")
+
+	// Add in the UID and creation timestamp for the resource if needed since etcd does not do it itself.
```
robbrockbank: Not everything that gets here is a resource - so you can't cast to an ObjectMetaAccessor.

Any reason why you're not doing this processing in resources.go?

lib/backend/k8s/resources/bgppeer.go
```diff
@@ -71,8 +71,7 @@ func (c BGPPeerConverter) ToKVPair(r Resource) (*model.KVPair, error) {
 
 	// Clear any CRD TypeMeta fields and then create a KVPair.
 	res := apiv2.NewBGPPeer()
-	res.ObjectMeta.Name = t.ObjectMeta.Name
-	res.ObjectMeta.Namespace = t.ObjectMeta.Namespace
+	res.ObjectMeta = t.ObjectMeta
```
robbrockbank: don't think this is correct we can't simply use the CRDs Metadata

bcreane:
@mgleung - let's hold off merging this until after rc1 is cut.

lib/backend/k8s/resources/resources.go
```diff
@@ -100,3 +102,74 @@ func SetCalicoMetadataFromK8sAnnotations(calicoRes Resource, k8sRes Resource) {
...
+	res.GetObjectMeta().SetFinalizers(meta.Finalizers)
+	res.GetObjectMeta().SetOwnerReferences(meta.OwnerReferences)
+	res.GetObjectMeta().SetClusterName(meta.ClusterName)
+
+	return nil
+}
+
+// Cleans out the Calico metadata from the annotations field. Needed to reduced side effects
+// that come from needing to modify the input data.
+func CleanAnnotations(res Resource) {
```
robbrockbank: This doesn't need to be public.

It could also benefit from a better name - I wasn't sure what it was really doing at first.

robbrockbank: Maybe something like :   removeCalicoMetadataAnnotation

lib/backend/k8s/resources/resources.go
```diff
@@ -100,3 +102,74 @@ func SetCalicoMetadataFromK8sAnnotations(calicoRes Resource, k8sRes Resource) {
 		}
 	}
 }
+
+// Store Calico Metadata in the in the k8s resource annotations for CRD backed resources.
+// This should store all Metadata except for those stored in Annotations and Labels and
+// store them in annotations.
+func StoreCalicoMetadataInK8sAnnotations(res Resource) error {
+	metadataBytes, err := json.Marshal(res.GetObjectMeta().(*metav1.ObjectMeta))
```
robbrockbank: You don't need to cast this to a metav1.ObjectMeta do you?

lib/backend/k8s/resources/resources.go
```diff
@@ -100,3 +102,74 @@ func SetCalicoMetadataFromK8sAnnotations(calicoRes Resource, k8sRes Resource) {
...
+	annotations[metadataAnnotation] = string(metadataBytes)
+	res.GetObjectMeta().SetAnnotations(annotations)
+
+	// Make sure to clear out all of the Calico Metadata out of the ObjectMeta except for
+	// Name, Namespace, and ResourceVersion.
+	res.GetObjectMeta().SetSelfLink("")
+	res.GetObjectMeta().SetUID("")
+	res.GetObjectMeta().SetGeneration(int64(0))
+	var emptyTime metav1.Time
+	res.GetObjectMeta().SetCreationTimestamp(emptyTime)
```
robbrockbank: I don't think you need these var lines.  You can just do:  `res.GetObjectMeta().SetCreationTimestamp(metav1.Time{})`

same for the others below

lib/backend/k8s/resources/customresource.go
```diff
@@ -60,6 +60,8 @@ func (c *customK8sResourceClient) Create(ctx context.Context, kvp *model.KVPair)
 
 	// Convert the KVPair to the K8s resource.
 	resIn, err := c.convertKVPairToResource(kvp)
+	// Clean out the annotations for the resource given since that's only used for our internal storage.
+	defer CleanAnnotations(resIn)
```
robbrockbank: I don't think this ClearAnnotations belongs here.  Everything that needs doing should be in the convertKVPairToResource.

lib/backend/k8s/resources/customresource.go
```diff
@@ -99,6 +110,8 @@ func (c *customK8sResourceClient) Update(ctx context.Context, kvp *model.KVPair)
 	var updateError error
 	// Convert the KVPair to a K8s resource.
 	resIn, err := c.convertKVPairToResource(kvp)
+	// Clean out the annotations for the resource given since that's only used for our internal storage.
+	defer CleanAnnotations(resIn)
```
robbrockbank: Same here.

lib/backend/k8s/resources/customresource.go
robbrockbank: And in this case it feels particularly strange to use defer to do this since you are assuming that resIn is what we return below - which makes things hard to maintain.  This would be better handled just before we return the result.

In any case - the convertResourceToKVPair should take care of the converting back to the Calico ObjectMeta.

lib/backend/k8s/resources/resources.go
```diff
@@ -100,3 +102,74 @@ func SetCalicoMetadataFromK8sAnnotations(calicoRes Resource, k8sRes Resource) {
...
+	annotations := res.GetObjectMeta().GetAnnotations()
+	if len(annotations) == 0 {
+		annotations = make(map[string]string)
+	}
+	annotations[metadataAnnotation] = string(metadataBytes)
+	res.GetObjectMeta().SetAnnotations(annotations)
+
+	// Make sure to clear out all of the Calico Metadata out of the ObjectMeta except for
+	// Name, Namespace, and ResourceVersion.
+	res.GetObjectMeta().SetSelfLink("")
```
robbrockbank: this would have been less code if you'd first got the ObjectMeta in a local variable (and potentially less operations depending on how clever the compiler is)

lib/backend/k8s/resources/resources.go
robbrockbank: IN fact I think I would just scrap this method and build it into the method above - that way those two methods are more symmetric.

One updates Metadata for K8s
One updates Metadata for Calico

robbrockbank: With that in mind, perhaps naming those methods something like:
```
ConvertMetadataCalicoToK8s()
ConvertMetadataK8sToCalico()
```

lib/backend/k8s/resources/resources.go
```diff
@@ -100,3 +102,74 @@ func SetCalicoMetadataFromK8sAnnotations(calicoRes Resource, k8sRes Resource) {
...
+		return errors.New("Attempting to create a resource from a resource without Calico metadata set in the annotations.")
+	}
+	meta := &metav1.ObjectMeta{}
+	err := json.Unmarshal([]byte(annotations[metadataAnnotation]), meta)
+	if err != nil {
+		return err
+	}
+
+	// Overwrite the K8s metadata with the Calico metadata.
+	res.GetObjectMeta().SetSelfLink(meta.SelfLink)
```
robbrockbank: we never store a self link

lib/backend/k8s/resources/resources.go
```diff
@@ -100,3 +102,74 @@ func SetCalicoMetadataFromK8sAnnotations(calicoRes Resource, k8sRes Resource) {
...
+	var emptyTime metav1.Time
+	res.GetObjectMeta().SetCreationTimestamp(emptyTime)
+	res.GetObjectMeta().SetDeletionTimestamp(nil)
+	res.GetObjectMeta().SetDeletionGracePeriodSeconds(nil)
+	res.GetObjectMeta().SetInitializers(nil)
+	var emptyStringSlice []string
+	res.GetObjectMeta().SetFinalizers(emptyStringSlice)
+	var emptyReferences []metav1.OwnerReference
+	res.GetObjectMeta().SetOwnerReferences(emptyReferences)
+	res.GetObjectMeta().SetClusterName("")
```
robbrockbank: All of this field copying fields like a maintenance hassle - especially if k8s decide to add more fields to the Metadata.

I think a better a more maintainable way to do this would be to use the DeepCopyInto() method defined on the ObjectMeta struct.  You could create your own local "empty" ObjectMeta with just the fields that should be included:  ie. the name, namespace, labels and annotations - and then use DeepCopyInto to copy over the Metadata in the resource.  This is more resilient to new fields being added.

robbrockbank: Same goes for both of these conversion functions.

robbrockbank: I think you should also avoid storing duplicate copies of everything - so I think the ObjectMeta that is stored in the annotation should have the Name, Namespace, TypeMeta, Annotations and Labels removed before marshaling into json.

Obviously you'll need to add those fields back in the other function.

lib/clientv2/globalnetworkpolicy.go
```diff
@@ -113,3 +147,21 @@ func defaultPolicyTypesField(spec *apiv2.PolicySpec) {
...
+func removePrefixFromGlobalPolicy(policy *apiv2.GlobalNetworkPolicy) {
+	name := policy.GetObjectMeta().GetName()
+	if !strings.HasPrefix(name, "knp.") {
+		parts := strings.SplitN(name, ".", 2)
+		resetGlobalPolicyName(policy, parts[1])
+	}
+}
+
+func resetGlobalPolicyName(res *apiv2.GlobalNetworkPolicy, originalName string) {
+	res.GetObjectMeta().SetName(originalName)
```
robbrockbank: Not sure what the point of this function is.

lib/backend/k8s/resources/resources.go
robbrockbank: Actually, do we need these to be two separate methods?

Can't we just include this processing in the ConvertResourceToKVPair and ConvertKVPairToResource functions.

robbrockbank: Hmm, given that this will only be used by the CRD backed resources, I wonder if this should just be part of the CRD conversion methods.  Don't mind it staying here though if you prefer but you should tighten up the method names though to indicate what they're doing.

lib/backend/k8s/resources/customresource.go
```diff
@@ -79,8 +81,17 @@ func (c *customK8sResourceClient) Create(ctx context.Context, kvp *model.KVPair)
 		return nil, K8sErrorToCalico(err, kvp.Key)
 	}
 
+	resOut.GetObjectKind().SetGroupVersionKind(c.k8sResourceTypeMeta.GetObjectKind().GroupVersionKind())
+
+	// Update the return data with the metadata populated by the (Kubernetes) datastore.
+	kvp, err = c.convertResourceToKVPair(resOut)
```
robbrockbank: Do we need this convertResourceToKVPair function - can't we just call directly through to the resource.go functions to convert the resource?  Seems like there's some surplus code here or further down the line.

robbrockbank: or if you are gonna keep it then you should include the other bits above and below in the function.

lib/clientv2/globalnetworkpolicy.go
```diff
@@ -41,8 +42,16 @@ type globalnetworkpolicies struct {
 // representation of the GlobalNetworkPolicy, and an error, if there is any.
 func (r globalnetworkpolicies) Create(ctx context.Context, res *apiv2.GlobalNetworkPolicy, opts options.SetOptions) (*apiv2.GlobalNetworkPolicy, error) {
 	defaultPolicyTypesField(&res.Spec)
+
+	// Properly prefix the name
```
robbrockbank: I think we should have converter methods in both directions.

We are disallowing tier names with a kip prefix - so you can use the same check in both directions to add or remove the prefix.

robbrockbank: Oh just realized - we only need this for NetworkPolicies not GlobalNetworkPolicies

lib/backend/k8s/resources/resources.go
mgleung: This logic used to be built into the above methods but then I realized that it also needed to be present in the `customK8sResourceClient` if we wanted to ensure that the `resource` input is not modified (since ideally functions should not leave side effects on the inputs). If we're fine with the input data being modified, then I can move the logic back into the above methods and fix the tests. I was really only made aware of this from the new k8s tests.

lib/backend/k8s/resources/resources.go
mgleung: I tried using `metav1.Time{}` but it gave me an empty Time object instead of the zero value for a Time object.  Ideally I'm hoping to remove the `CreationTimestamp` and other fields as if they didn't exist instead of filling in the values with empty ones. I did a poor job naming the variables there though so I can fix that.

lib/backend/k8s/resources/customresource.go
mgleung: I have it deferred in these functions because I noticed that the input data to the functions was being modified. Since the `Object` interface we're dealing with doesn't have a `DeepCopy` method defined on it, it makes it difficult to modify the contents of the metadata without modifying the input data. I thought it would be cleaner to not let input data to functions modify the input data but if we're always expecting these functions and their outputs to be used in a controlled manner, then I don't think it matters.

lib/backend/k8s/resources/resources.go
mgleung: I thought about using the `DeepCopyInto` method to create the metadata, but the interface definitions of the `Resource` do not have any methods to replace the `ObjectMeta` struct do they? The issue is I think we need to use the resources for these since the concrete value of each resource is different depending on the type.

lib/clientv2/globalnetworkpolicy.go
mgleung: Since we are internally prefixing our name data, this prevents the external output from reflecting the internal prefixing that we are adding.

lib/backend/k8s/resources/customresource.go
mgleung: We could put this function inline, but since it is called multiple times throughout the other methods, I thought it would be cleaner to keep it all together. There's no real reason that this specifically exists other than to prevent lines of code from being rewritten.

lib/backend/k8s/resources/resources.go
robbrockbank: No you can't replace the Metadata unfortunately - but you can deep copy into it, and the GetObjectMetadata() returns you the pointer to the ObjectMeta.

lib/backend/k8s/resources/customresource.go
robbrockbank: I have a couple of concerns.  Firstly doing this in a defer feels odd to me because it makes the flow of the code hard to follow.  It's clearer if you Modify the ObjectMeta, send the request, Update the ObjectMeta from the response data.  So really the use of defer is more that it's difficult to follow the flow here.  Is your concern that in the error case we've modified the resource that was given to us?  I thought DeepCopy was on the interface now?

lib/clientv2/globalnetworkpolicy.go
robbrockbank: Actually I see it's because you needed a function to wrap this up for a defer.

lib/backend/k8s/resources/customresource.go
robbrockbank: Yup - that makes sense.  I was more worried that we seem to have a bit more redirection than I think we need - that was all.  Because the convertResource then calls out to another convert method to do the heavy lifting

lib/backend/k8s/resources/resources.go
mgleung: Thought I had to cast it to a concrete type instead of just having it marshal from an interface. Just tested it, marshaling from the interface works.  I'll fix that.

lib/backend/k8s/resources/resources.go
mgleung: Just retested it. Must've gotten it confused with the slices. `metav1.Time{}` is the empty/zero value.  I'll make that change as well.

lib/backend/k8s/resources/resources.go
mgleung: Good point, I'll fix that.

lib/backend/k8s/resources/customresource.go
mgleung: Yup, my concern was that in the error case (or any case without that function running, I had it in a `defer` just to make sure it was kicked off) we are modifying the input data. I see, I got the `runtime.Object` interface mixed up with the `metav1.Object` interface. I'll change this back.

lib/backend/k8s/resources/customresource.go
```diff
@@ -347,16 +363,25 @@ func (c *customK8sResourceClient) convertResourceToKVPair(r Resource) (*model.KV
...
+	}
+
+	kvp.Value = r
 	return kvp, nil
 }
 
 func (c *customK8sResourceClient) convertKVPairToResource(kvp *model.KVPair) (Resource, error) {
 	resource := kvp.Value.(Resource)
+	// TODO: Figure out if Resource version is suposed to be overwritten by K8s or not.
 	resource.GetObjectMeta().SetResourceVersion(kvp.Revision)
```
robbrockbank: We do need to pass down the resource version into the resource definition that we pass to k8s.  This is used to perform "atomic" operations.  The K8s response will have an updated resource version.

lib/backend/k8s/resources/resources.go
```diff
@@ -100,3 +103,81 @@ func SetCalicoMetadataFromK8sAnnotations(calicoRes Resource, k8sRes Resource) {
 		}
 	}
 }
+
+// Store Calico Metadata in the in the k8s resource annotations for CRD backed resources.
+// This should store all Metadata except for those stored in Annotations and Labels and
+// store them in annotations.
+func StoreCalicoMetadataInK8sAnnotations(resIn Resource) (Resource, error) {
```
robbrockbank: Not sure about these function names.  Maybe these should just be ConvertCalicoResourceToK8sResource and ConvertK8sResourceToCalicoResource - obviously we still need to document what that conversion entails.

lib/backend/k8s/resources/customresource.go
robbrockbank: The ResourceVersion should be extracted from the response and included in the KVPair that is returned.

mgleung: Sounds good. Wasn't sure if I needed to store the version in the annotations as well, but from your answer, it seems we don't need to.

lib/backend/k8s/resources/resources.go
```diff
@@ -100,3 +103,81 @@ func SetCalicoMetadataFromK8sAnnotations(calicoRes Resource, k8sRes Resource) {
...
+
+	return resOut, nil
+}
+
+// Retrieve all of the Calico Metadata from the k8s resource annotations for CRD backed
+// resources. This should remove the relevant Calico Metadata annotation when it has finished.
+func GetCalicoMetadataFromK8sAnnotations(res Resource) error {
+	rom := res.GetObjectMeta()
+	annotations := rom.GetAnnotations()
+	if len(annotations) == 0 {
```
robbrockbank: I think we need to be more forgiving about the annotation not being present.  For example, our k8s manifest may want to include some of the CRDs for pre-population.  In that case we should return the labels/annotations/uuid/name/namespace from the resource but nothing else.

robbrockbank: When we write data through libcalico-go we do want to persist the values that are given to us - so as not to potentially confuse the AAPIS -  but reading stuff is less important because that will include data that has already been configured and the AAPIS will learn about that through its watch.

lib/backend/k8s/resources/resources.go
```diff
@@ -100,3 +103,81 @@ func SetCalicoMetadataFromK8sAnnotations(calicoRes Resource, k8sRes Resource) {
...
+
+// Store Calico Metadata in the in the k8s resource annotations for CRD backed resources.
+// This should store all Metadata except for those stored in Annotations and Labels and
+// store them in annotations.
+func StoreCalicoMetadataInK8sAnnotations(resIn Resource) (Resource, error) {
+	rom := resIn.GetObjectMeta()
+
+	// Make sure to remove data that is passed to Kubernetes so it is not duplicated in
+	// the annotations.
+	romCopy := &metav1.ObjectMeta{}
```
robbrockbank: Whilst this feels like a little more cumbersome - I'm much happier about that way this is being done now - it feels much more maintainable.  Good stuff.

lib/clientv2/globalnetworkpolicy.go
```diff
@@ -16,6 +16,7 @@ package clientv2
```
robbrockbank: GlobalNetworkPolicies don't need any of this special name handling since we don't have any k8s based policies show up here.

robbrockbank: Apologies - I thought I'd written that in one of my comments and it looks like i did not.

robbrockbank: Oh actually, yes I did.

mgleung: Oh oops missed that one. My bad.

lib/clientv2/networkpolicy.go
```diff
@@ -41,8 +42,16 @@ type networkPolicies struct {
 // representation of the NetworkPolicy, and an error, if there is any.
 func (r networkPolicies) Create(ctx context.Context, res *apiv2.NetworkPolicy, opts options.SetOptions) (*apiv2.NetworkPolicy, error) {
 	defaultPolicyTypesField(&res.Spec)
+
+	// Properly prefix the name
+	name := res.GetObjectMeta().GetName()
+	defer resetPolicyName(res, name)
```
robbrockbank: You don't need this defer anymore.

I think things would feel a lot more symmetric here if you had a convertPolicyNameForStorage and convertPolicyNameFromStorage  or something similar.   You've got one method in one direction and on in another.

That way I think this code would look something like:

`res.GetObjectMeta().SetName(convertPolicyNameForStorage(res.GetObjectMeta().GetName()))`

and just before you return the result you do the same with the opposite conversion function.  That feels like it hides less away.

Obviously for the list you'd need to go through each Item and convert those.

lib/clientv2/networkpolicy.go
```diff
@@ -92,3 +124,42 @@ func (r networkPolicies) List(ctx context.Context, opts options.ListOptions) (*a
...
+			res.Spec.Types = []apiv2.PolicyType{apiv2.PolicyTypeEgress}
+		} else {
+			// Policy has both ingress and egress rules, so apply this policy to both
+			// ingress and egress.
+			res.Spec.Types = []apiv2.PolicyType{apiv2.PolicyTypeIngress, apiv2.PolicyTypeEgress}
+		}
+	}
+}
+
+func removePrefixFromPolicies(res *apiv2.NetworkPolicyList) {
```
robbrockbank: I would just inline this function in the List.

lib/clientv2/networkpolicy.go
robbrockbank: I'm not suggesting mine is the right way, but it feels cleaner to me than what we have at the moment since we have a number of different methods used at different times performing subtly different things.

lib/clientv2/networkpolicy.go
```diff
@@ -92,3 +124,42 @@ func (r networkPolicies) List(ctx context.Context, opts options.ListOptions) (*a
 func (r networkPolicies) Watch(ctx context.Context, opts options.ListOptions) (watch.Interface, error) {
 	return r.client.resources.Watch(ctx, opts, apiv2.KindNetworkPolicy)
 }
+
+func (r networkPolicies) defaultTypesField(res *apiv2.NetworkPolicy) {
```
robbrockbank: Where did this change come from?

robbrockbank: I think this is redundant now.

lib/clientv2/networkpolicy.go
mgleung: I agree that it would be easier to read with symmetric name conversion functions.  Thinking about it, the only reason I can see keeping the way its written now is if for some reason the return value from the client for some reason does not follow the `<prefix>.<name>` format. How valid of a case would it be that data is injected into etcd (I'm not sure if you can do this to KDD) that would not follow our internal prefixing structure and is named something like `<something>.<random>`?

mgleung: Actually, that's a silly concern because if we convert correctly, then this wouldn't find the strange name that they might manually inject.

lib/clientv2/globalnetworkpolicy.go
mgleung: From our conversation about this, I will be leaving the prefixing logic in the GlobalNetworkPolicy as well since we need it for non k8s based policies as well.

robbrockbank: Yup - my bad.

lib/backend/k8s/resources/customresource_test.go
```diff
@@ -107,11 +108,25 @@ var _ = Describe("BGP Peer conversion methods", func() {
...
+		// Make sure to clean up the annotations on the resource
+		err = ConvertK8sResourceToCalicoResource(r)
+		Expect(err).NotTo(HaveOccurred())
 	})
 
 	It("should convert between a Kuberenetes resource and the equivalent KVPair", func() {
+		// Make sure it returns the same resource if a resource is given without Calico metadata annotations.
 		kvp, err := client.convertResourceToKVPair(res1)
 		Expect(err).NotTo(HaveOccurred())
+		Expect(kvp.Value.(*apiv2.BGPPeer)).To(BeAssignableToTypeOf(&apiv2.BGPPeer{}))
```
robbrockbank: I think you mean this to be `Expect(kvp.Value).To(BeAssignableToTypeOf(&apiv2.BGPPeer{}))`

i.e. without casting the Value - otherwise if it isn't a BGPPeer that cast will panic.

robbrockbank: Given you are doing a value comparison below - I don't think this particular Expect adds any value.

lib/backend/k8s/resources/resources.go
```diff
@@ -100,3 +102,82 @@ func SetCalicoMetadataFromK8sAnnotations(calicoRes Resource, k8sRes Resource) {
...
+// resources. This should remove the relevant Calico Metadata annotation when it has finished.
+func ConvertK8sResourceToCalicoResource(res Resource) error {
+	rom := res.GetObjectMeta()
+	annotations := rom.GetAnnotations()
+	if len(annotations) == 0 {
+		// Make no changes if there are no annotations to read Calico Metadata out of.
+		return nil
+	}
+	meta := &metav1.ObjectMeta{}
+	err := json.Unmarshal([]byte(annotations[metadataAnnotation]), meta)
```
robbrockbank: you should also check if metadataAnnotation is a valid key in the annotations.

lib/clientv2/networkpolicy.go
```diff
@@ -71,8 +123,19 @@ func (r networkPolicies) Delete(ctx context.Context, namespace, name string, opt
 // Get takes name of the NetworkPolicy, and returns the corresponding NetworkPolicy object,
 // and an error if there is any.
 func (r networkPolicies) Get(ctx context.Context, namespace, name string, opts options.GetOptions) (*apiv2.NetworkPolicy, error) {
-	out, err := r.client.resources.Get(ctx, opts, apiv2.KindNetworkPolicy, namespace, name)
+	// Properly prefix the name
+	prefixedName := name
+	if !strings.HasPrefix(name, "knp.") {
```
robbrockbank: Why isn't this knp check part of the convertPolicyNameForStorage?

lib/clientv2/networkpolicy.go
```diff
@@ -92,3 +165,7 @@ func (r networkPolicies) List(ctx context.Context, opts options.ListOptions) (*a
 func (r networkPolicies) Watch(ctx context.Context, opts options.ListOptions) (watch.Interface, error) {
 	return r.client.resources.Watch(ctx, opts, apiv2.KindNetworkPolicy)
 }
+
+func resetPolicyName(res *apiv2.NetworkPolicy, originalName string) {
```
robbrockbank: Don't think this is needed anymore

lib/backend/k8s/resources/customresource.go
```diff
@@ -79,8 +79,17 @@ func (c *customK8sResourceClient) Create(ctx context.Context, kvp *model.KVPair)
 		return nil, K8sErrorToCalico(err, kvp.Key)
 	}
 
+	resOut.GetObjectKind().SetGroupVersionKind(c.k8sResourceTypeMeta.GetObjectKind().GroupVersionKind())
```
robbrockbank: Shouldn't this be part of the conversion?

PR #599, opened 2017-10-19
Default v2 PolicySpec.Types field similarly as for v1

neiljerram:
Default the Types field for v2 PolicySpec similarly as we do for v1 at https://github.com/projectcalico/libcalico-go/blob/master/lib/converter/policy.go#L60

### DIFF
--- a/lib/backend/syncersv1/updateprocessors/networkpolicyprocessor.go
+++ b/lib/backend/syncersv1/updateprocessors/networkpolicyprocessor.go
@@ -67,24 +67,7 @@ func convertPolicyV2ToV1Spec(spec apiv2.PolicySpec) (interface{}, error) {
 		PreDNAT:       spec.PreDNAT,
 		Types:         policyTypesAPIV2ToBackend(spec.Types),
 	}
-	if len(spec.Types) == 0 {
-		// Default the Types field according to what inbound and outbound rules are present
-		// in the policy.
-		if len(spec.EgressRules) == 0 {
-			// Policy has no egress rules, so apply this policy to ingress only.  (Note:
-			// intentionally including the case where the policy also has no ingress
-			// rules.)
-			v1value.Types = []string{string(apiv2.PolicyTypeIngress)}
-		} else if len(spec.IngressRules) == 0 {
-			// Policy has egress rules but no ingress rules, so apply this policy to
-			// egress only.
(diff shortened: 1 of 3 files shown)


lib/backend/syncersv1/updateprocessors/networkpolicyprocessor.go
```diff
@@ -67,7 +67,24 @@ func convertPolicyV2ToV1Spec(spec apiv2.PolicySpec) (interface{}, error) {
 		PreDNAT:       spec.PreDNAT,
 		Types:         policyTypesAPIV2ToBackend(spec.Types),
 	}
-
+	if len(spec.Types) == 0 {
```
fasaxc: I'm not 100% sure of this.  Previously, we did this defaulting on write but this is on the syncer's read path.  Previously, on the read path, we defaulted to `[ingress, egress]`, I think.

neiljerram: Do you know where the write path for v2 is?

fasaxc:
LGTM but probably best to get @robbrockbank's opinion on whether we're adding defaulting it he right place.  Also, do we need some UT for this?

lib/clientv2/globalnetworkpolicy.go
```diff
@@ -40,6 +40,7 @@ type globalnetworkpolicies struct {
 // Create takes the representation of a GlobalNetworkPolicy and creates it.  Returns the stored
 // representation of the GlobalNetworkPolicy, and an error, if there is any.
 func (r globalnetworkpolicies) Create(ctx context.Context, res *apiv2.GlobalNetworkPolicy, opts options.SetOptions) (*apiv2.GlobalNetworkPolicy, error) {
+	r.defaultTypesField(res)
```
fasaxc: What about Update()?  Do we need to do Get() too (we had that in v1)?

neiljerram: For Get(): similarly if the object concerned was created through here.  If it's coming from an existing DB, I think the defaulting for that case happens elsewhere.

robbrockbank: I assume someone can "undo" the change with an update no?  feels like we probably should do this in the update too.

PR #601, opened 2017-10-19
Fix resource etcdv3 paths to follow k8s standard

robbrockbank:
Update the resource paths to follow k8s path format:

If the resource is namespace scoped:
`<prefix>/<group_name>/<resources>/<namespace>/<resource_name>`

If the resource is cluster scoped:
`<prefix>/<group_name>/<resources>/<resource_name>`

Where,
prefix can be any given string constant. In case of Kubernetes Core resources its "registry"
group_name is a string identifying the API group. In case of Kubernetes Core resources it is empty "". By convention the group names are typically in domain name form. The Kubernetes group reserves the right to use empty group, all single word names ("extensions", "apps"), and any group name ending "*.k8s.io" for its sole use.
resources is a string identifying the resource kind name in plural.

### DIFF
--- a/lib/backend/model/resource.go
+++ b/lib/backend/model/resource.go
@@ -26,23 +26,89 @@ import (
 	"github.com/projectcalico/libcalico-go/lib/namespace"
 )
 
+// Name/type information about a single resource.
+type resourceInfo struct {
+	typeOf reflect.Type
+	plural string
+	kind string
+}
+
 var (
-	matchGlobalResource     = regexp.MustCompile("^/calico/resources/v2/([^/]+)/([^/]+)$")
-	matchNamespacedResource = regexp.MustCompile("^/calico/resources/v2/([^/]+)/([^/]+)/([^/]+)$")
-	kindToType              = map[string]reflect.Type{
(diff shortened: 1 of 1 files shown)


lib/backend/model/resource.go
```diff
@@ -26,24 +26,90 @@ import (
 	"github.com/projectcalico/libcalico-go/lib/namespace"
```
caseydavenport: Why alias this as `apiv2` and not just `api`?

Seems like since we're only going to support a single api version, that's the best way to go.

caseydavenport: Maybe `api` is wrong too - `capi` or something to indicate it's the Calico API vs `kapi` for k8s api? idc.

robbrockbank: urm iunno - i thought that was done as part of Eriks/Brendans file move

bcreane: yeah, that <didn't> happen during https://github.com/projectcalico/libcalico-go/pull/586. I aliased the import rather than change all the existing `apiv2` namespace references to `v2` or `capiv2`.

caseydavenport: ok, let's deal with it later.

robbrockbank:
closed/reopened to trigger build

PR #602, opened 2017-10-19
Refactor the shared resource methods for KDD to reduce code duplication

mgleung:
## Description
Refactor the CustomResourceClient so that instead of depending on various converters to do the same operations on different types, that all of the operations can be done in the same function instead of being split out per type.

## Todos
- [x] Tests
- [x] Documentation
- [x] Release note

## Release Note

```release-note
None required
```


lib/backend/k8s/resources/customresource.go
```diff
@@ -310,3 +326,63 @@ func (c *customK8sResourceClient) Watch(ctx context.Context, list model.ListInte
...
+	kvp := &model.KVPair{
+		Key: model.ResourceKey{
+			Name:      r.GetObjectMeta().GetName(),
+			Namespace: r.GetObjectMeta().GetNamespace(),
+			Kind:      resourceKind,
+		},
+		Revision: r.GetObjectMeta().GetResourceVersion(),
+	}
+
+	var v2resource Resource
```
robbrockbank: This isn't necessary.  The resource that is passed in is already the correct type.

lib/backend/k8s/resources/customresource.go
```diff
@@ -310,3 +326,63 @@ func (c *customK8sResourceClient) Watch(ctx context.Context, list model.ListInte
...
+	v2resource.GetObjectMeta().SetNamespace(r.GetObjectMeta().GetNamespace())
+
+	kvp.Value = v2resource
+	return kvp, nil
+}
+
+func convertKVPairToResource(kvp *model.KVPair, resourceKind string) (Resource, error) {
+	var resource Resource
+	switch resourceKind {
+	case apiv2.KindBGPPeer:
```
robbrockbank: none of this switching should be necessary.  You already have the correct type in your hand

lib/backend/k8s/resources/customresource.go
```diff
@@ -302,11 +306,55 @@ func (c *customK8sResourceClient) Watch(ctx context.Context, list model.ListInte
...
+	return newK8sWatcherConverter(ctx, toKVPair, k8sWatch), nil
 }
 
 // EnsureInitialized is a no-op since the CRD should be
 // initialized in advance.
 func (c *customK8sResourceClient) EnsureInitialized() error {
 	return nil
 }
+
+func ListInterfaceToKey(l model.ListInterface) model.Key {
```
robbrockbank: these should be lower case now, and to avoid ambiguity the names should either be more descriptive or make them methods attached to the struct rather than stand alone functions.

Personally I'd go with the latter.

PR #603, opened 2017-10-19
Types field defaulting is needed for NetworkPolicy as well as GlobalN…

neiljerram:
…etworkPolicy


lib/clientv2/networkpolicy.go
```diff
@@ -40,6 +40,7 @@ type networkPolicies struct {
 // Create takes the representation of a NetworkPolicy and creates it.  Returns the stored
 // representation of the NetworkPolicy, and an error, if there is any.
 func (r networkPolicies) Create(ctx context.Context, res *apiv2.NetworkPolicy, opts options.SetOptions) (*apiv2.NetworkPolicy, error) {
+	r.defaultTypesField(res)
```
robbrockbank: as per last PR - not sure why this isn't in the update.  Surely someone can "undo" the defaulting in an update no?

PR #604, opened 2017-10-20
Use generic Felix syncer in KDD and add both KDD and etcdv3 syncer tests

robbrockbank:
Updated the e2e BGP syncer tests to include KDD


robbrockbank:
Including in another PR

PR #607, opened 2017-10-20
Re-Add k8s backend test file

doublek:
## Description
PR #564 removed a couple of test files. This PR reintroduces them and moves some of the test code to be v2 data model compatible.

## Todos
- [x] Verify GNP vs NP tests
- [ ] Agree upon Syncer test status for now

## Release Note

```release-note
None required
```


lib/backend/k8s/k8s_fv_test.go
```diff
@@ -0,0 +1,1519 @@
...
+			Expect(kvps.KVPairs[0].Value).To(Equal(kvp1b.Value))
+		})
+
+		By("Deleting an existing BGP Peer", func() {
+			_, err := c.Delete(ctx, kvp1a.Key, "")
+			Expect(err).NotTo(HaveOccurred())
+		})
+	})
+
+	// TODO(doublek): Test the BGP annotations of a Node resource
```
caseydavenport: Is something causing these to fail?

PR #608, opened 2017-10-20
Tolerate named ports with duplicate names on the same endpoint.

fasaxc:
## Description
Note: PR is against v1.x-series, which is a branch cut from just before the merge of the v2 datamodel.

@ozdanborne Found that there are kubernetes Pod specs in the wild that have duplicate named ports so our validation was too aggressive.

## Todos
- [x] Tests
- [x] Documentation
- [x] Release note

## Release Note

```release-note
None required
```


fasaxc:
Checked the docs, we didn't mention the uniqueness constraint.

PR #612, opened 2017-10-20
Update IPPool definition

robbrockbank:
move IPIPMode to top level field in IPPool
add cluster name field to cluster information resource


lib/apis/v2/zz_generated.deepcopy.go
```diff
@@ -19,11 +19,10 @@
 package v2
 
 import (
-	reflect "reflect"
-
 	numorstring "github.com/projectcalico/libcalico-go/lib/numorstring"
 	conversion "k8s.io/apimachinery/pkg/conversion"
 	runtime "k8s.io/apimachinery/pkg/runtime"
+	reflect "reflect"
```
gunjan5: this is a weird aliasing, but I'm guessing it's generated code so we can't do anything

lib/apis/v2/ippool.go
```diff
@@ -37,25 +37,16 @@ type IPPool struct {
 type IPPoolSpec struct {
 	// The pool CIDR.
 	CIDR string `json:"cidr" validate:"omitempty,cidr"`
-	// Contains configuration for ipip tunneling for this pool. If not specified,
-	// then ipip tunneling is disabled for this pool.
-	IPIP *IPIPConfiguration `json:"ipip,omitempty"`
+	// Contains configuration for IPIP tunneling for this pool. If not specified,
```
gunjan5: Since this is the source of truth for our APIs, should we add `This can be one of "Never", "Always" or "CrossSubnet".` in the comment?

robbrockbank: isn't that implied from the type?

gunjan5: not CrossSubnet

PR #619, opened 2017-10-22
Various fixes/workarounds to kdd based policy and namespace selection

robbrockbank:
Bug fixes:

-  Wasn't handling namespace selection correctly (was only in KDD code, but is now non-KDD only)
-  Policy name from k8s policy should not have the namespace in it anymore
-  Added orchestrator label so that k8s-backed policy only applies to k8s WEPs.

Also:
-  Some name prefix changes

### DIFF
--- a/lib/backend/k8s/conversion/constants.go
+++ b/lib/backend/k8s/conversion/constants.go
@@ -1,0 +1,21 @@
+// Copyright (c) 2017 Tigera, Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package conversion
+
+const (
+	NamespaceLabelPrefix       = "pcns."
+	NamespaceProfileNamePrefix = "kns."
+	K8sNetworkPolicyNamePrefix = "knp.default."
+)
--- a/lib/backend/k8s/conversion/conversion.go
+++ b/lib/backend/k8s/conversion/conversion.go
@@ -38,6 +38,13 @@ var (
 	protoTCP = kapiv1.ProtocolTCP
 )
 
+type selectorType int8
+
+const (
+	SelectorNamespace selectorType = iota
+	SelectorPod
+)
+
 //TODO: make this private and expose a public conversion interface instead
 type Converter struct {
 }
@@ -59,22 +66,6 @@ func (c Converter) ParseWorkloadEndpointName(workloadName string) (names.Workloa
 	return names.ParseWorkloadEndpointName(workloadName)
 }
 
-// ParsePolicyNameNetworkPolicy extracts the Kubernetes Namespace and NetworkPolicy that backs the given Policy.
-func (c Converter) ParsePolicyNameNetworkPolicy(name string) (string, string, error) {
-	// Policies backed by NetworkPolicies have form "knp.default.<ns_name>.<np_name>"
-	if !strings.HasPrefix(name, "knp.default.") {
-		// This is not backed by a Kubernetes NetworkPolicy.
-		return "", "", fmt.Errorf("Policy %s not backed by a NetworkPolicy", name)
-	}
-
-	splits := strings.SplitN(strings.TrimPrefix(name, "knp.default."), ".", 2)
-	if len(splits) != 2 {
-		return "", "", fmt.Errorf("Name does not include both Namespace and NetworkPolicy: %s", name)
-	}
-	// Return Namespace, NetworkPolicy name.
-	return splits[0], splits[1], nil
-}
-
 // NamespaceToProfile converts a Namespace to a Calico Profile.  The Profile stores
 // labels from the Namespace which are inherited by the WorkloadEndpoints within
 // the Profile. This Profile also has the default ingress and egress rules, which are both 'allow'.
@@ -82,12 +73,12 @@ func (c Converter) NamespaceToProfile(ns *kapiv1.Namespace) (*model.KVPair, erro
 	// Generate the labels to apply to the profile, using a special prefix
 	// to indicate that these are the labels from the parent Kubernetes Namespace.
 	labels := map[string]string{}
-	for k, v := range ns.ObjectMeta.Labels {
-		labels[fmt.Sprintf("pcns.%s", k)] = v
+	for k, v := range ns.Labels {
+		labels[NamespaceLabelPrefix+k] = v
 	}
 
 	// Create the profile object.
-	name := fmt.Sprintf("k8s_ns.%s", ns.ObjectMeta.Name)
+	name := NamespaceProfileNamePrefix + ns.Name
 	profile := apiv2.NewProfile()
 	profile.ObjectMeta = metav1.ObjectMeta{
 		Name:              name,
@@ -95,8 +86,8 @@ func (c Converter) NamespaceToProfile(ns *kapiv1.Namespace) (*model.KVPair, erro
 		UID:               ns.UID,
 	}
 	profile.Spec = apiv2.ProfileSpec{
-		IngressRules:  []apiv2.Rule{apiv2.Rule{Action: apiv2.Allow}},
-		EgressRules:   []apiv2.Rule{apiv2.Rule{Action: apiv2.Allow}},
+		IngressRules:  []apiv2.Rule{{Action: apiv2.Allow}},
+		EgressRules:   []apiv2.Rule{{Action: apiv2.Allow}},
 		LabelsToApply: labels,
 	}
 
@@ -107,7 +98,7 @@ func (c Converter) NamespaceToProfile(ns *kapiv1.Namespace) (*model.KVPair, erro
 			Kind: apiv2.KindProfile,
 		},
 		Value:    profile,
-		Revision: ns.ObjectMeta.ResourceVersion,
+		Revision: ns.ResourceVersion,
 	}
 	return &kvp, nil
 }
@@ -146,12 +137,12 @@ func (c Converter) HasIPAddress(pod *kapiv1.Pod) bool {
 // fail to convert from a Pod to WorkloadEndpoint otherwise.
 func (c Converter) PodToWorkloadEndpoint(pod *kapiv1.Pod) (*model.KVPair, error) {
 	// Pull out the profile and workload ID based on pod name and Namespace.
-	profile := fmt.Sprintf("k8s_ns.%s", pod.ObjectMeta.Namespace)
+	profile := NamespaceProfileNamePrefix + pod.Namespace
 	wepids := names.WorkloadEndpointIdentifiers{
 		Node:         pod.Spec.NodeName,
-		Orchestrator: "k8s",
+		Orchestrator: apiv2.OrchestratorKubernetes,
 		Endpoint:     "eth0",
-		Pod:          pod.ObjectMeta.Name,
+		Pod:          pod.Name,
 	}
 	wepName, err := wepids.CalculateWorkloadEndpointName(false)
 	if err != nil {
@@ -174,12 +165,14 @@ func (c Converter) PodToWorkloadEndpoint(pod *kapiv1.Pod) (*model.KVPair, error)
 	// the host-side veth configured by the CNI plugin.
 	interfaceName := VethNameForWorkload(wepName)
 
-	// Build the labels map.
-	labels := map[string]string{}
-	if pod.ObjectMeta.Labels != nil {
-		labels = pod.ObjectMeta.Labels
+	// Build the labels map.  Start with the pod labels, and append two additional labels for
+	// namespace and orchestrator matches.
+	labels := pod.Labels
+	if labels == nil {
+		labels = make(map[string]string, 2)
 	}
-	labels["calico/k8s_ns"] = pod.ObjectMeta.Namespace
+	labels[apiv2.LabelNamespace] = pod.Namespace
+	labels[apiv2.LabelOrchestrator] = apiv2.OrchestratorKubernetes
 
 	// Map any named ports through.
 	var endpointPorts []apiv2.EndpointPort
@@ -214,7 +207,7 @@ func (c Converter) PodToWorkloadEndpoint(pod *kapiv1.Pod) (*model.KVPair, error)
 	wep := apiv2.NewWorkloadEndpoint()
 	wep.ObjectMeta = metav1.ObjectMeta{
 		Name:              wepName,
-		Namespace:         pod.ObjectMeta.Namespace,
+		Namespace:         pod.Namespace,
 		CreationTimestamp: pod.CreationTimestamp,
 		UID:               pod.UID,
 		Labels:            labels,
@@ -234,11 +227,11 @@ func (c Converter) PodToWorkloadEndpoint(pod *kapiv1.Pod) (*model.KVPair, error)
 	kvp := model.KVPair{
 		Key: model.ResourceKey{
 			Name:      wepName,
-			Namespace: pod.ObjectMeta.Namespace,
+			Namespace: pod.Namespace,
 			Kind:      apiv2.KindWorkloadEndpoint,
 		},
 		Value:    wep,
-		Revision: pod.ObjectMeta.ResourceVersion,
+		Revision: pod.ResourceVersion,
 	}
 	return &kvp, nil
 }
@@ -246,7 +239,7 @@ func (c Converter) PodToWorkloadEndpoint(pod *kapiv1.Pod) (*model.KVPair, error)
 // K8sNetworkPolicyToCalico converts a k8s NetworkPolicy to a model.KVPair.
 func (c Converter) K8sNetworkPolicyToCalico(np *extensions.NetworkPolicy) (*model.KVPair, error) {
 	// Pull out important fields.
-	policyName := fmt.Sprintf("knp.default.%s.%s", np.ObjectMeta.Namespace, np.ObjectMeta.Name)
+	policyName := fmt.Sprintf(K8sNetworkPolicyNamePrefix + np.Name)
 
 	// We insert all the NetworkPolicy Policies at order 1000.0 after conversion.
 	// This order might change in future.
@@ -255,13 +248,13 @@ func (c Converter) K8sNetworkPolicyToCalico(np *extensions.NetworkPolicy) (*mode
 	// Generate the ingress rules list.
 	var ingressRules []apiv2.Rule
 	for _, r := range np.Spec.Ingress {
-		ingressRules = append(ingressRules, c.k8sRuleToCalico(r.From, r.Ports, np.ObjectMeta.Namespace, true)...)
+		ingressRules = append(ingressRules, c.k8sRuleToCalico(r.From, r.Ports, np.Namespace, true)...)
 	}
 
 	// Generate the egress rules list.
 	var egressRules []apiv2.Rule
 	for _, r := range np.Spec.Egress {
-		egressRules = append(egressRules, c.k8sRuleToCalico(r.To, r.Ports, np.ObjectMeta.Namespace, false)...)
+		egressRules = append(egressRules, c.k8sRuleToCalico(r.To, r.Ports, np.Namespace, false)...)
 	}
 
 	// Calculate Types setting.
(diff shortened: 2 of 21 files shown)


lib/backend/k8s/conversion/constants.go
```diff
@@ -0,0 +1,22 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package conversion
+
+const (
+	NamespaceLabelPrefix       = "pcns."
+	Orchestrator               = "k8s"
```
caseydavenport: For better readability I think this should be `OrchestratorKubernetes` or similar.

Probably fine for now since there's only one value, but we might eventually have `OrchestratorMesos` etc

robbrockbank: Well if we keep the constant here then it is scoped by teh KDD package

If you want me to move to apiv2 then I'll Add a set of values and make this OrchestratorKubernetes

robbrockbank: just say the word and i'll do either

robbrockbank: Actually it should be in apiv2!  I'll move it!

lib/backend/k8s/conversion/conversion.go
```diff
@@ -239,7 +226,7 @@ func (c Converter) PodToWorkloadEndpoint(pod *kapiv1.Pod) (*model.KVPair, error)
 // K8sNetworkPolicyToCalico converts a k8s NetworkPolicy to a model.KVPair.
 func (c Converter) K8sNetworkPolicyToCalico(np *extensions.NetworkPolicy) (*model.KVPair, error) {
 	// Pull out important fields.
-	policyName := fmt.Sprintf("knp.default.%s.%s", np.ObjectMeta.Namespace, np.ObjectMeta.Name)
+	policyName := fmt.Sprintf( K8sNetworkPolicyNamePrefix + np.ObjectMeta.Namespace)
```
caseydavenport: looks like an extra space here

caseydavenport: Plus this should be `Name` not `Namespace`

lib/backend/k8s/conversion/conversion.go
```diff
@@ -313,14 +300,16 @@ func (c Converter) K8sNetworkPolicyToCalico(np *extensions.NetworkPolicy) (*mode
...
-		selectors = append(selectors, fmt.Sprintf("calico/k8s_ns == '%s'", *ns))
+func (c Converter) k8sSelectorToCalico(s *metav1.LabelSelector, selectorType selectorType) string {
+	// If this is a namespace selector, the labels need to be prefixed with the profile namespace
+	// prefix.
+	var selectors []string
+	var prefix string
+	if selectorType == SelectorNamespace {
+		prefix = NamespaceLabelPrefix
+	}
+	if selectorType == SelectorPod {
```
caseydavenport: This isn't quite right - the pod selector needs to be ANDed with that pod's namespace as well, not just with the entire cluster.

We still need this selector: `("projectcalico.org/namespace == '%s', *ns)`

caseydavenport: Also, I think _all_ selectors generated by this function need the `projectcalico.org/orchestrator == 'k8s'` bit.

robbrockbank: Ah but policy now has a namespace, so this rule does not need to be part of the policy selector.

However, instead - in the conversion to the v1 model - that's where we insert the additional selector to match on the namespace.

robbrockbank: (at least that's my thinking)

caseydavenport: Ok, so looks like the `namespace == foo` bit is added in the networkpolicy processor, is that right?

That makes sense.  I still think all of them need the orchestrator label

lib/backend/syncersv1/updateprocessors/networkpolicyprocessor.go
```diff
@@ -58,11 +59,22 @@ func convertPolicyV2ToV1Spec(spec apiv2.PolicySpec) (interface{}, error) {
 		erules = append(erules, RuleAPIV2ToBackend(erule))
 	}
 
+	// If this policy is namespaced, then add a namespace selector.
+	selector := spec.Selector
+	if ns != "" {
+		nsSelector := fmt.Sprintf("%s == '$s'", apiv2.NamespaceLabel, ns)
```
caseydavenport: making a note: we're going to need corresponding changes to the CNI plugin to ensure it's writing the correct labels.

caseydavenport: This should be a `%s` not a `$s`

lib/backend/k8s/conversion/conversion.go
robbrockbank: yup - i think you are correct about the orchestrator -- adding now

lib/backend/syncersv1/updateprocessors/workloadendpointprocessor.go
```diff
@@ -114,6 +114,11 @@ func convertWorkloadEndpointV2ToV1Value(val interface{}) (interface{}, error) {
 		})
 	}
 
+	// Felix does not have the concept of namespaces, so add the namespace as a Label.  We'll
```
caseydavenport: Hm, I wasn't expecting this code here, but I'm not sure what to make of it.

My expectation was that each plugin would write the namespace in.  This is nice because then they don't have to, but it also makes for an asymmetrical UX between defining a policy using certain label and being unable to query and see that label in, say `calicoctl`.

lib/backend/k8s/conversion/conversion.go
```diff
@@ -207,6 +190,10 @@ func (c Converter) PodToWorkloadEndpoint(pod *kapiv1.Pod) (*model.KVPair, error)
 		}
 	}
 
+	// Add the orchestrator label to the workload endpoint labels.
+	labels := pod.ObjectMeta.Labels
+	labels[apiv2.OrchestratorLabel] = Orchestrator
```
caseydavenport: I think it would be good to show the Namespace label here as well - that way users can see that the wep has that label on it.

lib/backend/k8s/conversion/conversion.go
```diff
@@ -207,6 +190,10 @@ func (c Converter) PodToWorkloadEndpoint(pod *kapiv1.Pod) (*model.KVPair, error)
 		}
 	}
 
+	// Add the orchestrator label to the workload endpoint labels.
+	labels := pod.ObjectMeta.Labels
+	labels[apiv2.OrchestratorLabel] = apiv2.OrchestratorKubernetes
```
caseydavenport: I think your latest round erased my comment here, but I think it would be good to add the Namespace label to this map as well.

It means that people who query this through calicoctl will see it.  We probably want to do that in the CNI plugin as well so it works with etcd.

lib/backend/syncersv1/updateprocessors/workloadendpointprocessor.go
caseydavenport: If anything, this logic should go in the defaulting for WorkloadEndpoints so that whenever a WEP is written it gets this label on it (plus PodToWorkloadEndpoint for KDD).

That's a better approach I think.

robbrockbank: hmm not sure i agree about not writing in the namespace.

Otherwise the concept of namespace vs global resources is completely dependent on the behavior of the plugin - that doesn't make sense to me.

Eventually we should add the concept of namespace to felix, so we can have namespace selectors and then all this label manipulation for the namespace goes away - that's not the case if the orchestrator is trying to control that.

robbrockbank: I could possible buy that we return the namespace label in the policy definition - but that would be nothing to do with the orchestrator though - more just something in the client to make it clear to users what is going on.

robbrockbank: In fact I would go one further and say that we should be adding the Orchestrator label here too, rather than within the KDD code - since we know the orchestrator from the name/Spec.

And in the client Get/List code insert the same labels so that people can at least see labels we are "auto-inserting".

If it isn't added generically then the behavior of what the namespace implies is different depending on the orchestrator - but I think we probably want that behavior defined.

lib/backend/syncersv1/updateprocessors/workloadendpointprocessor.go
```diff
@@ -114,17 +114,26 @@ func convertWorkloadEndpointV2ToV1Value(val interface{}) (interface{}, error) {
...
+	if labels == nil {
+		labels = make(map[string]string, 2)
+	}
+	labels[apiv2.LabelNamespace] = v2res.Namespace
+	labels[apiv2.LabelOrchestrator] = v2res.Spec.Orchestrator
+
 	v1value := &model.WorkloadEndpoint{
 		State:            "active",
 		Name:             v2res.Spec.InterfaceName,
-		ActiveInstanceID: v2res.Spec.ContainerID,
```
caseydavenport: Why this change?

robbrockbank: It's not required for Felix.

caseydavenport: Yep. I was thinking along the lines of "what if another syncer client comes along that wants it" - e.g. I'd like to eventually update the kube-controllers code to use a syncer rather than a periodic poll, and it does care about this field.

I think I wouldn't use an updateprocessor to convert to v1 in that case though, so it's probably fine.

lib/backend/k8s/conversion/conversion.go
```diff
@@ -171,13 +162,6 @@ func (c Converter) PodToWorkloadEndpoint(pod *kapiv1.Pod) (*model.KVPair, error)
 	// the host-side veth configured by the CNI plugin.
 	interfaceName := VethNameForWorkload(wepName)
 
-	// Build the labels map.
-	labels := map[string]string{}
-	if pod.ObjectMeta.Labels != nil {
-		labels = pod.ObjectMeta.Labels
-	}
-	labels["calico/k8s_ns"] = pod.ObjectMeta.Namespace
```
caseydavenport: I think we still need to set the "magic" labels in this function - there are other places in our code - probably just kube-controllers - that relies on this function returning the same set of labels that you would see when doing a List/Get through libcalico.

lib/backend/k8s/conversion/conversion_test.go
```diff
@@ -410,12 +391,12 @@ var _ = Describe("Test NetworkPolicy conversion", func() {
 		By("parsing the policy", func() {
 			pol, err = c.K8sNetworkPolicyToCalico(&np)
 			Expect(err).NotTo(HaveOccurred())
-			Expect(pol.Key.(model.ResourceKey).Name).To(Equal("knp.default.default.testPolicy"))
+			Expect(pol.Key.(model.ResourceKey).Name).To(Equal("knp.default.testPolicy"))
```
caseydavenport: We should add in an assertion on the Namespace being "default" as well.

caseydavenport: And related, make sure we have a test that checks the updateprocessor correctly namespaces the selectors when given a NetworkPolicy (if we don't already)

lib/backend/k8s/conversion/conversion.go
robbrockbank: sounds good.

in that case - i'll just make sure we set the labels in the clientv2 as well and persist them and then remove the syncer magic.

lib/backend/k8s/conversion/conversion_test.go
robbrockbank: Good call!  Uncovered a conversion bug.

lib/backend/k8s/conversion/conversion_test.go
```diff
@@ -1049,6 +1054,10 @@ var _ = Describe("Test NetworkPolicy conversion (k8s <= 1.7, no policyTypes)", f
 			pol, err = c.K8sNetworkPolicyToCalico(&np)
 			Expect(err).NotTo(HaveOccurred())
 			Expect(pol.Key.(model.ResourceKey).Name).To(Equal("knp.default.testPolicy"))
+			Expect(pol.Key.(model.ResourceKey).Namespace).To(Equal("default"))
+			Expect(pol.Value.(*apiv2.NetworkPolicy).Name).To(Equal("knp.default.testPolicy"))
+			Expect(pol.Value.(*apiv2.NetworkPolicy).Namespace).To(Equal("default"))
+			Expect(int(*pol.Value.(*apiv2.NetworkPolicy).Spec.Order)).To(Equal(1000))
```
caseydavenport: Looks like a duplicate Expect here

robbrockbank: one is checking the key, the other the metadata

lib/backend/k8s/conversion/conversion.go
```diff
@@ -218,10 +211,10 @@ func (c Converter) PodToWorkloadEndpoint(pod *kapiv1.Pod) (*model.KVPair, error)
 			ObjectMeta: metav1.ObjectMeta{
 				Name:      wepName,
 				Namespace: pod.ObjectMeta.Namespace,
-				Labels:    labels,
+				Labels:    pod.ObjectMeta.Labels,
```
caseydavenport: nit, but just `pod.Labels` seems to be what we do elsewhere.

lib/backend/k8s/conversion/conversion_test.go
```diff
@@ -227,7 +213,7 @@ var _ = Describe("Test Pod conversion", func() {
 		Expect(wep.Value.(*apiv2.WorkloadEndpoint).Spec.Endpoint).To(Equal("eth0"))
 		Expect(wep.Value.(*apiv2.WorkloadEndpoint).Spec.Orchestrator).To(Equal("k8s"))
 		Expect(len(wep.Value.(*apiv2.WorkloadEndpoint).Spec.IPNetworks)).To(Equal(1))
-		Expect(wep.Value.(*apiv2.WorkloadEndpoint).ObjectMeta.Labels).To(Equal(map[string]string{"calico/k8s_ns": "default"}))
+		Expect(wep.Value.(*apiv2.WorkloadEndpoint).ObjectMeta.Labels).To(BeNil())
```
caseydavenport: Don't think this test is right - should have the orchestrator / namespace labels yeah?

lib/backend/k8s/conversion/conversion_test.go
caseydavenport: @robbrockbank Did you add this test? I don't see it.

caseydavenport: Oh it's way below. ok.

lib/backend/k8s/conversion/conversion.go
robbrockbank: yup - was fixing that up where I noticed - guess this one slipped by

lib/backend/k8s/conversion/conversion_test.go
```diff
@@ -1067,12 +1053,16 @@ var _ = Describe("Test NetworkPolicy conversion (k8s <= 1.7, no policyTypes)", f
 		By("parsing the policy", func() {
 			pol, err = c.K8sNetworkPolicyToCalico(&np)
 			Expect(err).NotTo(HaveOccurred())
-			Expect(pol.Key.(model.ResourceKey).Name).To(Equal("knp.default.default.testPolicy"))
+			Expect(pol.Key.(model.ResourceKey).Name).To(Equal("knp.default.testPolicy"))
+			Expect(pol.Key.(model.ResourceKey).Namespace).To(Equal("default"))
+			Expect(pol.Value.(*apiv2.NetworkPolicy).Name).To(Equal("knp.default.testPolicy"))
+			Expect(pol.Value.(*apiv2.NetworkPolicy).Namespace).To(Equal("default"))
+			Expect(int(*pol.Value.(*apiv2.NetworkPolicy).Spec.Order)).To(Equal(1000))
```
caseydavenport: Duplicate Expect()

lib/backend/k8s/conversion/conversion_test.go
robbrockbank: well that is sad.  Let me investigate.

lib/backend/syncersv1/updateprocessors/networkpolicyprocessor.go
```diff
@@ -58,11 +59,22 @@ func convertPolicyV2ToV1Spec(spec apiv2.PolicySpec) (interface{}, error) {
...
 	}
 
+	// If this policy is namespaced, then add a namespace selector.
+	selector := spec.Selector
+	if ns != "" {
+		nsSelector := fmt.Sprintf("%s == '%s'", apiv2.LabelNamespace, ns)
+		if selector == "" {
+			selector = nsSelector
+		} else {
+			selector += " && " + nsSelector
```
caseydavenport: I don't _think_ we generate any `||` clauses for these selectors.

If we did, we'd need `selector = fmt.Sprintf("%s && (%s)", nsSelector, selector)` or similar.

caseydavenport: To make sure the grouping is correct.

lib/backend/syncersv1/updateprocessors/workloadendpointprocessor.go
```diff
@@ -114,10 +114,12 @@ func convertWorkloadEndpointV2ToV1Value(val interface{}) (interface{}, error) {
 		})
 	}
 
+	// Add the namespace and orchestrator labels to the set of labels we give Felix.  We
+	// don't need to persist these, and we need to make sure they are not user-configurable
```
caseydavenport: Think this comment is out of date now

robbrockbank: yup.  ugh tired.

lib/backend/syncersv1/updateprocessors/networkpolicyprocessor.go
robbrockbank: yes i agree - and I think we DO need this for the user defined policy case

lib/backend/syncersv1/updateprocessors/rules.go
```diff
@@ -46,6 +49,45 @@ func RuleAPIV2ToBackend(ar apiv2.Rule) model.Rule {
...
+	if ns != "" && (ar.Source.Selector != "" || ar.Source.NotSelector != "") {
+		logCxt := log.WithFields(log.Fields{
+			"Namespace": ns,
+			"Selector": ar.Source.Selector,
+			"NotSelector": ar.Source.NotSelector,
+		})
+		logCxt.Debug("Maybe update source Selector to include namespace")
+		if !strings.Contains(ar.Source.Selector, "kns.") {
+			logCxt.Debug("Updating source selector")
+			if ar.Source.Selector == "" {
```
caseydavenport: At this point we've already identified that `ar.Source.Selector != ""` so we can remove this check.

lib/backend/syncersv1/updateprocessors/rules.go
```diff
@@ -46,6 +49,45 @@ func RuleAPIV2ToBackend(ar apiv2.Rule) model.Rule {
...
 		notICMPType = ar.NotICMP.Type
 	}
+	
+	// If we have any selector specified, then we may need to add the namespace selector.
+	// We do this if this policy is namespaced AND if the Selector does not have any other
+	// k8s namespace (profile label) selector in it.
+	// TODO this is TEMPORARY CODE:  We currently do a simple regex to search for kns. in the
+	// selector to see if we are performing k8s namespace queries.
+	nsSelector := fmt.Sprintf("%s == '%s'", apiv2.LabelNamespace, ns)
+	if ns != "" && (ar.Source.Selector != "" || ar.Source.NotSelector != "") {
```
caseydavenport: We're not doing anything with `NotSelector` below. I think we need to namespace that as well (if present) using the same logic as for `Selector` in order to be technically correct.

lib/backend/syncersv1/updateprocessors/rules.go
robbrockbank: no we haven't

robbrockbank: first check is if either are != ""

And the ".kns" check is a negated check

caseydavenport: Yeah you're right. Carry on.

lib/backend/syncersv1/updateprocessors/rules.go
caseydavenport: Discussed offline and this is fine.

PR #622, opened 2017-10-23
Adding OrchRefs to master

heschlie:
## Description
Adding `OrchRefs` to the Calico Node object.

## Todos
- [x] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
Added OrchRefs to the Node model to help track Nodes.
```


lib/apis/v2/node.go
```diff
@@ -41,6 +41,17 @@ type Node struct {
 type NodeSpec struct {
 	// BGP configuration for this node.
 	BGP *NodeBGPSpec `json:"bgp,omitempty" validate:"omitempty"`
+
```
caseydavenport: @heschlie could you update or add a test here: https://github.com/projectcalico/libcalico-go/blob/master/lib/clientv2/node_e2e_test.go

heschlie: Thought I had moved the old test over, will get it added.

heschlie: Ah I did move over the OrchRef additions to the NodeSpecs in the top of that test. Are the CRUD tests for Node not sufficient coverage of the OrchRefs? I can add some targeted tests if we don't think so.

heschlie: ping @caseydavenport

caseydavenport: Ah, somehow I didn't see that before.

Yes, that's fine.

caseydavenport:
We may want to add this into the conversion logic: https://github.com/projectcalico/libcalico-go/blob/master/lib/backend/k8s/resources/node.go#L181

I think it's fine to go as is for now though.

PR #623, opened 2017-10-23
Added tests to check Rule conversion code separately from policies

mgleung:
## Description
Cleaned up the tests so that instead of writing the `Expect` statements for the same set of data for two different policy tests, the conversion code tests for rules are separate.  This makes allows us to tell if the update processor is broken or the rule conversion code is broken separately.

## Todos
- [x] Tests
- [x] Documentation
- [x] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/backend/syncersv1/updateprocessors/rules_test.go
+++ b/lib/backend/syncersv1/updateprocessors/rules_test.go
@@ -1,0 +1,190 @@
+package updateprocessors_test
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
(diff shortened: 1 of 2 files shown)


lib/backend/syncersv1/updateprocessors/rules_test.go
```diff
@@ -0,0 +1,191 @@
+package updateprocessors_test
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	apiv2 "github.com/projectcalico/libcalico-go/lib/apis/v2"
+	//"github.com/projectcalico/libcalico-go/lib/backend/model"
```
robbrockbank: should remove this

PR #628, opened 2017-10-24
Enable IPIP globally if IPIP-enabled pool is configured.

robbrockbank:
## Description
A few sentences describing the overall goals of the pull request's commits.
Please include
- the type of fix - (e.g. bug fix, new feature, documentation)
- some details on _why_ this PR should be merged
- the details of the testing you've done on it (both manual and automated)
- which components are affected by this PR
- links to issues that this PR addresses

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/clientv2/ippool.go
+++ b/lib/clientv2/ippool.go
@@ -49,11 +49,19 @@ func (r ipPools) Create(ctx context.Context, res *apiv2.IPPool, opts options.Set
 		return nil, err
 	}
 
+	// Enable IPIP globally if required.  Do this before the Create so if it fails the user
+	// can retry the same command.
+	err := r.maybeEnableIPIP(ctx, res)
+	if err != nil {
+		return nil, err
+	}
+
 	out, err := r.client.resources.Create(ctx, opts, apiv2.KindIPPool, res)
 	if out != nil {
 		return out.(*apiv2.IPPool), err
 	}
 	return nil, err
+
 }
 
 // Update takes the representation of a IPPool and updates it. Returns the stored
@@ -61,6 +69,13 @@ func (r ipPools) Create(ctx context.Context, res *apiv2.IPPool, opts options.Set
 func (r ipPools) Update(ctx context.Context, res *apiv2.IPPool, opts options.SetOptions) (*apiv2.IPPool, error) {
 	// Validate the IPPool updating the resource.
 	if err := r.validateAndSetDefaults(res); err != nil {
+		return nil, err
+	}
+
+	// Enable IPIP globally if required.  Do this before the Update so if it fails the user
+	// can retry the same command.
+	err := r.maybeEnableIPIP(ctx, res)
+	if err != nil {
 		return nil, err
 	}
 
@@ -198,7 +213,7 @@ func (_ ipPools) validateAndSetDefaults(pool *apiv2.IPPool) error {
 		})
 	}
 
-	// If return the errors if we have one or more validation errors.
+	// Return the errors if we have one or more validation errors.
 	if len(errFields) > 0 {
 		return cerrors.ErrorValidation{
 			ErroredFields: errFields,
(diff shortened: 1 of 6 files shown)


lib/clientv2/ippool.go
```diff
@@ -50,10 +50,13 @@ func (r ipPools) Create(ctx context.Context, res *apiv2.IPPool, opts options.Set
...
 
 	out, err := r.client.resources.Create(ctx, opts, apiv2.KindIPPool, res)
-	if out != nil {
-		return out.(*apiv2.IPPool), err
+	if err != nil {
+		return nil, err
 	}
-	return nil, err
+
+	// Enable IPIP globally if required.
```
caseydavenport: Should we do this before creating the IP Pool? That way we can fail before we successfully create the pool, giving the user another chance.

Otherwise, they might successfully create the pool but it still won't work.

robbrockbank: Ah yes a good point! Ta.

lib/clientv2/ippool.go
```diff
@@ -207,3 +213,64 @@ func (_ ipPools) validateAndSetDefaults(pool *apiv2.IPPool) error {
 
 	return nil
 }
+
+// maybeEnableIPIP turns enables globally IPIP if a default setting is not already configured
```
caseydavenport: typo: turns enables

lib/clientv2/ippool.go
```diff
@@ -207,3 +213,64 @@ func (_ ipPools) validateAndSetDefaults(pool *apiv2.IPPool) error {
...
+// and the pool has IPIP enabled.
+func (c ipPools) maybeEnableIPIP(ctx context.Context, pool *apiv2.IPPool) error {
+	logCxt := log.WithField("IPIPMode", pool.Spec.IPIPMode)
+	if pool.Spec.IPIPMode == apiv2.IPIPModeNever {
+		logCxt.Debug("IPIP is not enabled for this pool - no need to check global setting")
```
caseydavenport: Does this really need the above context logger? We say right in the log that it's not enabled...

robbrockbank: Not strictly - but it does help verify the logic is correct

lib/clientv2/ippool.go
```diff
@@ -207,3 +213,64 @@ func (_ ipPools) validateAndSetDefaults(pool *apiv2.IPPool) error {
...
+		return nil
+	}
+
+	var err error
+	ipEnabled := true
+	for i := 0; i < maxApplyRetries; i++ {
+		logCxt.WithField("Retry", i).Debug("Checking global IPIP setting")
+		res, err := c.client.FelixConfigurations().Get(ctx, "default", options.GetOptions{})
```
caseydavenport: I suspect one day we'll have validation to prevent people from creating FelixConfiguration with name `foo` - it either needs to be `default` or `node.X`?

robbrockbank: i believe we already have that

robbrockbank: oh. no we don't

lib/clientv2/ippool.go
```diff
@@ -207,3 +213,64 @@ func (_ ipPools) validateAndSetDefaults(pool *apiv2.IPPool) error {
...
+			res.Name = "default"
+		} else if res.Spec.IpInIpEnabled != nil {
+			// A value for the default config is set so leave unchanged.  It may be set to false,
+			// so log the actual value - but we shouldn't update it if someone has explicitly
+			// disabled it globally.
+			log.WithField("IpInIpEnabled", res.Spec.IpInIpEnabled).Debug("Global IpInIpEnabled setting is already configured")
+			return nil
+		}
+
+		// Enable IpInIp  and do the Create or Update.
```
caseydavenport: typo: two spaces

lib/clientv2/ippool.go
robbrockbank: in that case, yes one day.

caseydavenport:
@robbrockbank tests are also failing.

robbrockbank:
yes I know - I'm still working on the PR.

PR #631, opened 2017-10-25
Move pod status update to Create

caseydavenport:
## Description

The CNI plugin calls Create() in kdd mode, not Update(). So, we should just do the IP update logic in Create() instead.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/backend/k8s/resources/workloadendpoint.go
+++ b/lib/backend/k8s/resources/workloadendpoint.go
@@ -44,11 +44,36 @@ type WorkloadEndpointClient struct {
 }
 
 func (c *WorkloadEndpointClient) Create(ctx context.Context, kvp *model.KVPair) (*model.KVPair, error) {
-	log.Warn("Operation Create is not supported on WorkloadEndpoint type")
-	return nil, cerrors.ErrorOperationNotSupported{
-		Identifier: kvp.Key,
-		Operation:  "Create",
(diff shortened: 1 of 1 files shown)


lib/backend/k8s/resources/workloadendpoint.go
```diff
@@ -84,6 +76,14 @@ func (c *WorkloadEndpointClient) Update(ctx context.Context, kvp *model.KVPair)
 	return kvp, nil
 }
 
+func (c *WorkloadEndpointClient) Update(ctx context.Context, kvp *model.KVPair) (*model.KVPair, error) {
```
doublek: Without digging too much, the CNI plugin also seems to perform Update the WEP resource, and I am not sure which scenario this would get called.

caseydavenport: Hmm, I was assuming it would only ever call Create() because it will never have a resourceVersion, and this code gets executed: https://github.com/projectcalico/cni-plugin/blob/master/utils/utils.go#L68

HOWEVER, the more I think about it, the more I think there may be scenarios where Update() gets called.

Let's perform this logic in both places.

caseydavenport:
Will open another to commonize the code

PR #633, opened 2017-10-26
Release pool affinities, check for overlapping pools and check CIDR cannot be altered

robbrockbank:
WIP.  Need to fix up UTs.

This PR adds:
-  Validation of the CIDR:
   -  CIDR cannot change on an Update
   -  CIDR cannot overlap with existing pools
-  Releasing pool affinities when the pool is deleted.

### DIFF
--- a/lib/clientv2/ippool.go
+++ b/lib/clientv2/ippool.go
@@ -16,14 +16,17 @@ package clientv2
 
 import (
 	"context"
+	"fmt"
 	"net"
+	"time"
+
+	log "github.com/sirupsen/logrus"
 
 	apiv2 "github.com/projectcalico/libcalico-go/lib/apis/v2"
 	cerrors "github.com/projectcalico/libcalico-go/lib/errors"
 	cnet "github.com/projectcalico/libcalico-go/lib/net"
 	"github.com/projectcalico/libcalico-go/lib/options"
 	"github.com/projectcalico/libcalico-go/lib/watch"
-	log "github.com/sirupsen/logrus"
 )
 
 // IPPoolInterface has methods to work with IPPool resources.
@@ -45,7 +48,7 @@ type ipPools struct {
 // representation of the IPPool, and an error, if there is any.
 func (r ipPools) Create(ctx context.Context, res *apiv2.IPPool, opts options.SetOptions) (*apiv2.IPPool, error) {
 	// Validate the IPPool before creating the resource.
-	if err := r.validateAndSetDefaults(res); err != nil {
+	if err := r.validateAndSetDefaults(ctx, res, nil); err != nil {
 		return nil, err
 	}
 
@@ -67,14 +70,20 @@ func (r ipPools) Create(ctx context.Context, res *apiv2.IPPool, opts options.Set
 // Update takes the representation of a IPPool and updates it. Returns the stored
 // representation of the IPPool, and an error, if there is any.
 func (r ipPools) Update(ctx context.Context, res *apiv2.IPPool, opts options.SetOptions) (*apiv2.IPPool, error) {
+	// Get the existing settings, so that we can validate the CIDR has not changed.
+	old, err := r.Get(ctx, res.Name, options.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+
 	// Validate the IPPool updating the resource.
-	if err := r.validateAndSetDefaults(res); err != nil {
+	if err := r.validateAndSetDefaults(ctx, res, old); err != nil {
 		return nil, err
 	}
 
 	// Enable IPIP globally if required.  Do this before the Update so if it fails the user
 	// can retry the same command.
-	err := r.maybeEnableIPIP(ctx, res)
+	err = r.maybeEnableIPIP(ctx, res)
 	if err != nil {
 		return nil, err
 	}
@@ -88,6 +97,65 @@ func (r ipPools) Update(ctx context.Context, res *apiv2.IPPool, opts options.Set
 
 // Delete takes name of the IPPool and deletes it. Returns an error if one occurs.
 func (r ipPools) Delete(ctx context.Context, name string, opts options.DeleteOptions) (*apiv2.IPPool, error) {
+	// Deleting a pool requires a little care because of existing endpoints
+	// using IP addresses allocated in the pool.  We do the deletion in
+	// the following steps:
+	// -  disable the pool so no more IPs are assigned from it
+	// -  remove all affinities associated with the pool
+	// -  delete the pool
+
+	// Get the pool so that we can find the CIDR associated with it.
+	pool, err := r.Get(ctx, name, options.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	logCxt := log.WithFields(log.Fields{
+		"CIDR": pool.Spec.CIDR,
+		"Name": name,
+	})
+
+	// If the pool is active, set the disabled flag to ensure we stop allocating from this pool.
+	if !pool.Spec.Disabled {
+		logCxt.Info("Disabling pool to release affinities")
+		pool.Spec.Disabled = true
+
+		// If the Delete has been called with a ResourceVersion then use that to perform the
+		// update - that way we'll catch update conflicts (we could actually check here, but
+		// the most likely scenario is there isn't one - so just pass it through and let the
+		// Update handle any conflicts).
+		if opts.ResourceVersion != "" {
+			pool.ResourceVersion = opts.ResourceVersion
+		}
+		if _, err := r.Update(ctx, pool, options.SetOptions{}); err != nil {
+			return nil, err
+		}
+
+		// Reset the resource version before the actual delete since the version of that resource
+		// will now have been updated.
+		opts.ResourceVersion = ""
+	}
+
+	// Release affinities associated with this pool.  We do this even if the pool was disabled
+	// (since it may have been enabled at one time, and if there are no affine blocks created
+	// then this will be a no-op).  We've already validated the CIDR so we know it will parse.
+	if _, cidrNet, err := cnet.ParseCIDR(pool.Spec.CIDR); err == nil {
+		logCxt.Info("Releasing pool affinities")
+
+		// Pause for a short period before releasing the affinities - this gives any in-progress
+		// allocations an opportunity to finish.
+		time.Sleep(500 * time.Millisecond)
+		err = r.client.IPAM().ReleasePoolAffinities(ctx, *cidrNet)
+
+		// Depending on the datastore, IPAM may not be supported.  If we get a not supported
+		// error, then continue.  Any other error, fail.
+		if _, ok := err.(cerrors.ErrorOperationNotSupported); !ok && err != nil {
+			return nil, err
+		}
+	}
+
+	// And finally, delete the pool.
+	logCxt.Info("Deleting pool")
 	out, err := r.client.resources.Delete(ctx, opts, apiv2.KindIPPool, noNamespace, name)
 	if out != nil {
 		return out.(*apiv2.IPPool), err
@@ -120,12 +188,14 @@ func (r ipPools) Watch(ctx context.Context, opts options.ListOptions) (watch.Int
 	return r.client.resources.Watch(ctx, opts, apiv2.KindIPPool)
 }
 
-// validateAndSetDefaults validates IPPool fields.
-func (_ ipPools) validateAndSetDefaults(pool *apiv2.IPPool) error {
+// validateAndSetDefaults validates IPPool fields and sets default values that are
+// not assigned.
+// The old pool will be unassigned for a Create.
+func (r ipPools) validateAndSetDefaults(ctx context.Context, new, old *apiv2.IPPool) error {
 	errFields := []cerrors.ErroredField{}
 
 	// Spec.CIDR field must not be empty.
-	if pool.Spec.CIDR == "" {
+	if new.Spec.CIDR == "" {
 		return cerrors.ErrorValidation{
 			ErroredFields: []cerrors.ErroredField{{
 				Name:   "IPPool.Spec.CIDR",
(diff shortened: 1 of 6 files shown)


lib/clientv2/ippool.go
```diff
@@ -18,12 +18,14 @@ import (
...
 
+	log "github.com/sirupsen/logrus"
+
 	apiv2 "github.com/projectcalico/libcalico-go/lib/apis/v2"
 	cerrors "github.com/projectcalico/libcalico-go/lib/errors"
 	cnet "github.com/projectcalico/libcalico-go/lib/net"
 	"github.com/projectcalico/libcalico-go/lib/options"
 	"github.com/projectcalico/libcalico-go/lib/watch"
-	log "github.com/sirupsen/logrus"
+	"fmt"
```
caseydavenport: probably put this up with "net" et al.

lib/clientv2/ippool.go
```diff
@@ -88,6 +96,46 @@ func (r ipPools) Update(ctx context.Context, res *apiv2.IPPool, opts options.Set
...
+	// Deleting a pool requires a little care because of existing endpoints
+	// using IP addresses allocated in the pool.  We do the deletion in
+	// the following steps:
+	// -  disable the pool so no more IPs are assigned from it
+	// -  remove all affinities associated with the pool
+	// -  delete the pool
+	logCxt := log.WithField("Name", name)
+
+	// Get the pool so that we can find the CIDR associated with it.  If the pool is
+	// enabled we'll need to release affinities associated with the pool.
```
caseydavenport: Shouldn't we just do this even if the pool is enabled? 

Can a pool go from enabled to disabled while there are IPs allocated from it?

lib/clientv2/ippool.go
```diff
@@ -88,6 +96,46 @@ func (r ipPools) Update(ctx context.Context, res *apiv2.IPPool, opts options.Set
...
+		if _, err := r.Update(ctx, pool, options.SetOptions{}); err != nil {
+			return nil, err
+		}
+
+		// Convert the CIDR to a network.  This shouldn't fail, but if it does then there
+		// should be no need to release affinities since we can't assigning from this invalid
+		// pool.
+		if _, cidrNet, err := cnet.ParseCIDR(pool.Spec.CIDR); err != nil {
+			logCxt.Debugf("Releasing pool affinities")
+			err = r.client.IPAM().ReleasePoolAffinities(context.Background(), *cidrNet)
```
caseydavenport: Shouldn't we pass in `ctx` instead of a new context?

lib/clientv2/ippool.go
```diff
@@ -121,7 +169,7 @@ func (r ipPools) Watch(ctx context.Context, opts options.ListOptions) (watch.Int
 }
 
 // validateAndSetDefaults validates IPPool fields.
-func (_ ipPools) validateAndSetDefaults(pool *apiv2.IPPool) error {
+func (r ipPools) validateAndSetDefaults(ctx context.Context, pool *apiv2.IPPool, previousCIDR string) error {
```
caseydavenport: Might be nicer to define the function like so:

```go
func (r ipPools) validateAndSetDefaults(ctx, new, old *apiv2.IPPool) error 
```

That way we can do arbitrary validation checks on an Update() call.

caseydavenport: Looking at the rest of this code, my gut feeling is that it probably wants to be split up into a few different functions.

e.g.

```go
func setDefaults(pool *IPPool) {}
func validateCreate(new *IPPool) {}
func validateUpdate(new, old *IPPool) {}
```

The `validateCreate` and `validateUpdate` could still call through to commonized `validateIPPool` or something, but I think it will make the logic that only applies to one or the other a lot more clear.

lib/clientv2/ippool.go
```diff
@@ -145,6 +193,41 @@ func (_ ipPools) validateAndSetDefaults(pool *apiv2.IPPool) error {
...
+	if previousCIDR != "" && previousCIDR != pool.Spec.CIDR {
+		errFields = append(errFields, cerrors.ErroredField{
+			Name:   "IPPool.Spec.CIDR",
+			Reason: "IPPool CIDR can not be modified",
+		})
+	}
+
+	// If there was no previous CIDR, check that it does not overlap with any
+	// other pool CIDRs.
+	if previousCIDR == "" {
```
caseydavenport: This is essentially saying "only perform this validation on Create", right?

lib/clientv2/ippool.go
```diff
@@ -145,6 +193,41 @@ func (_ ipPools) validateAndSetDefaults(pool *apiv2.IPPool) error {
 		}
 	}
 
+	// Normalize the CIDR before persisting.
+	pool.Spec.CIDR = cidr.String()
+
+	// If there was a previous CIDR, validate the CIDR has not changed.  Since we
+	// are using noramlized CIDRs we can just do a simple string comparison.
+	if previousCIDR != "" && previousCIDR != pool.Spec.CIDR {
```
caseydavenport: This is essentially saying "Only perform this validation on Update", right?

lib/clientv2/ippool.go
robbrockbank: yup i'd already done that one :-)

lib/clientv2/ippool.go
robbrockbank: hmmm - yeah actually now that you mention it that's a good idea!  it also closes down a window where failure just after disabling the pool means we wouldn't release affinities next attempt!

lib/clientv2/ippool.go
robbrockbank: ha!  yeah absolutely -- copied from elsewhere and didn't spot that.

lib/clientv2/ippool.go
robbrockbank: the setDefaults would be better elsewhere I agree.

the validate, less clear - it's nice to be able to return the list of validation errors in one go which is what we attempt to do.  It gets complicated if you do validateCreate and validateUpdate since there'd be a some duplication, or we'd need to have another common validation method.

lib/clientv2/ippool.go
robbrockbank: Yup.

lib/clientv2/ippool.go
robbrockbank: Yup yup

lib/clientv2/ippool.go
robbrockbank: I'll ponder though.

caseydavenport: It's an extra function call, but it allows us to more clearly write different validation logic for Create vs Update, whereas right now we need to have a bunch of branches within a single validation function, which I find less good.

robbrockbank: Ok, so having actually dug a little deeper....  as it stands *most* of the validation code could be split up as Create or Update validation with a small amount of overlap.  My concern with doing that is if we relax the restrictions about changing CIDR or duplicate CIDRs etc then we'd need to undo that change.

robbrockbank: I'd rather change this as per your first suggestion which I like.

caseydavenport: We wouldn't _need_ to undo the change.

I still think it's a much better structure, even if the validation performed by each is identical. But I'm happy to let it go for now.

robbrockbank: ok well actually let me change as per your initial suggestion.  I actually think the whole validation side of things will need to change anyways because we want calicoctl to be able to as much validation as it can before sending in the requests - at the moment having this validation so intertwined with the Create/update processing prevents that.  So in that sense having it split out as you suggest sounds like a good idea - but I think I'd rather do that as part of a "fix validation" task which we do have coming up.

lib/clientv2/ippool_e2e_test.go
```diff
@@ -399,20 +426,7 @@ var _ = testutils.E2eDatastoreDescribe("IPPool tests", testutils.DatastoreAll, f
...
-			be.Clean()
-			testWatcher4.ExpectEvents(apiv2.KindIPPool, []watch.Event{
-				{
-					Type:     watch.Deleted,
-					Previous: outRes1,
-				},
-				{
-					Type:     watch.Deleted,
-					Previous: outRes3,
+					Object: outRes4,
```
robbrockbank: I've removed the last test - it'll be a bit of a pain to update, and we already test the delete everything case in all of the other resource types - so we don't gain anything by testing for this resource too.

lib/clientv2/ippool.go
```diff
@@ -88,6 +97,60 @@ func (r ipPools) Update(ctx context.Context, res *apiv2.IPPool, opts options.Set
...
+		logCxt.Debugf("Releasing pool affinities")
+		err = r.client.IPAM().ReleasePoolAffinities(context.Background(), *cidrNet)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	// And finally, delete the pool.
+	logCxt.Debugf("Deleting pool")
+	opts.ResourceVersion = ""
```
robbrockbank: drat - this isn't needed here - since I reset it above only if the pool needs disabling.

lib/clientv2/ippool.go
```diff
@@ -145,22 +210,58 @@ func (_ ipPools) validateAndSetDefaults(pool *apiv2.IPPool) error {
 		}
 	}
 
+	// Normalize the CIDR before persisting.
+	new.Spec.CIDR = cidr.String()
+
+	// If there was a previous pool then this must be an Update, validate that the
+	// CIDR has not changed.  Since we are using noramlized CIDRs we can just do a
```
neiljerram: noramlized -> normalized

lib/clientv2/ippool.go
```diff
@@ -145,22 +210,58 @@ func (_ ipPools) validateAndSetDefaults(pool *apiv2.IPPool) error {
...
+		for _, otherPool := range allPools.Items {
+			_, otherCIDR, err := cnet.ParseCIDR(otherPool.Spec.CIDR)
+			if err != nil {
+				log.WithField("Name", otherPool.Name).WithError(err).Error("IPPool is configured with an invalid CIDR")
+				continue
+			}
+			if otherCIDR.IsNetOverlap(cidr.IPNet) {
+				errFields = append(errFields, cerrors.ErroredField{
+					Name:   "IPPool.Spec.CIDR",
+					Reason: fmt.Sprintf("IPPool CIDR overlaps with IPPool(%s) CIDR %s", otherPool.Name, otherPool.Spec.CIDR),
```
neiljerram: Note: this causes an existing calicoctl ST to fail:
```
FAIL: Test namespace is handled as expected for each resource type [with data={'kind': 'IPPool', 'spec': {'ipi...data': {'name': 'ippool-name1'}}, namespaced=False].
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/nose_parameterized/parameterized.py", line 392, in standalone_func
    return func(*(a + p.args), **p.kwargs)
  File "/code/tests/st/calicoctl/test_crud.py", line 364, in test_namespace
    rc.assert_no_error()
  File "/code/tests/st/utils/utils.py", line 139, in assert_no_error
    "command=" + self.command + "\noutput=" + self.output
AssertionError: Expected no error running command; 
command=export ETCD_ENDPOINTS=http://192.168.1.16:2379; export ETCD_CA_CERT_FILE=; export ETCD_CERT_FILE=; export ETCD_KEY_FILE=; export DATASTORE_TYPE=etcdv3;  /code/dist/calicoctl create -f /tmp/input-data
output=Failed to create 'IPPool' resource: error with field IPPool.Spec.CIDR = '<nil>' (IPPool CIDR overlaps with IPPool(name1) CIDR 10.0.1.0/24)
```

(https://semaphoreci.com/calico/calicoctl/branches/pull-request-1742/builds/1)

lib/clientv2/ippool.go
```diff
@@ -88,6 +96,61 @@ func (r ipPools) Update(ctx context.Context, res *apiv2.IPPool, opts options.Set
...
+	if err != nil {
+		return nil, err
+	}
+
+	logCxt := log.WithFields(log.Fields{
+		"CIDR": pool.Spec.CIDR,
+		"Name": name,
+	})
+
+	// If the pool is active, set the disabled flag to ensure we stop allocating from this pool.
```
neiljerram: Why disable the pool if we are just about to delete it anyway?  Is this needed to allow ReleasePoolAffinities to work?

lib/clientv2/ippool.go
```diff
@@ -145,22 +210,58 @@ func (_ ipPools) validateAndSetDefaults(pool *apiv2.IPPool) error {
...
+	// Normalize the CIDR before persisting.
+	new.Spec.CIDR = cidr.String()
+
+	// If there was a previous pool then this must be an Update, validate that the
+	// CIDR has not changed.  Since we are using noramlized CIDRs we can just do a
+	// simple string comparison.
+	if old != nil && old.Spec.CIDR != new.Spec.CIDR {
+		errFields = append(errFields, cerrors.ErroredField{
+			Name:   "IPPool.Spec.CIDR",
+			Reason: "IPPool CIDR can not be modified",
```
neiljerram: can not -> cannot

lib/clientv2/ippool.go
neiljerram: If this is an intended restriction now (which I think it is!), it should be documented at https://docs.projectcalico.org/master/reference/calicoctl/resources/ippool.  (As well as fixing up the ST.)

lib/clientv2/ippool.go
robbrockbank: hmmm, tries to remember the dim and distant past.. (this bit was copied over from the old client)

I believe this is to reduce the window of new affine blocks being claimed from the pool during the affine block cleanup which involves multiple read/writes to etcd.

However, there is still a window where it can happen.  Short of introducing more locking and more reads/writes during the IPAM affine block claiming code I think it'll be super hard to close this window completely - but this certainly reduces chances.  Also, claiming affine blocks should be a lot rarer than claiming IPs - and I don't believe we had any known issues with this approach since it was added to the last client.  But yes, perhaps we don't gain a whole lot from Disabling first.

lib/clientv2/ippool.go
robbrockbank: d'ohhhhhhh   - silly me!

thanks!  I'll jump on this.

1. is because none of the validation, including the stuff I've just added includes the field value.

I think there are a few fixes required here.  Firstly to add the values to the validation errors, secondly for the validation error to handle more gracefully a value not being present - I don't think we want to display nil values in the error messages.

lib/clientv2/ippool.go
```diff
@@ -88,6 +97,65 @@ func (r ipPools) Update(ctx context.Context, res *apiv2.IPPool, opts options.Set
...
+
+	// Release affinities associated with this pool.  We do this even if the pool was disabled
+	// (since it may have been enabled at one time, and if there are no affine blocks created
+	// then this will be a no-op).  We've already validated the CIDR so we know it will parse.
+	if _, cidrNet, err := cnet.ParseCIDR(pool.Spec.CIDR); err == nil {
+		logCxt.Info("Releasing pool affinities")
+
+		// Pause for a short period before releasing the affinities - this gives any in-progress
+		// allocations an opportunity to finish.
+		time.Sleep(500 * time.Millisecond)
```
caseydavenport: Shouldn't this sleep be after the `r.Update` above which disables the pool?

We only need to do this if we're just now disabling the pool, I think.

robbrockbank: deleting before the release is a safer bet?  could be someones workflow is Disable, Delete.

PR #636, opened 2017-10-27
Include TypeMeta in resource returned/converted by NetworkPolicy CRD Client

doublek:
## Description
The NetworkPolicy client wasn't passing some parameters that the CustomResourceClient requires to include TypeMeta. Pass them in and enable E2E tests that catch this bug.

## Todos
- [x] Tests
- [x] Release note

## Release Note

```release-note
None required
```


robbrockbank:
Are we missing something in our tests that we didn't spot this sooner?

PR #642, opened 2017-10-31
Don't allow creating resource with no name

### LINKED ISSUE #641: Creating a pool without a name should return an error
Creating a pool without a name should return an error

## Expected Behavior

I'd expect an error to be returned.

## Current Behavior

Pool creation silently fails

## Possible Solution

Pool creation should return an error when no name is supplied.

```
pool := api.IPPool{
				Spec: api.IPPoolSpec{
					CIDR: "192.168.0.0/16",
				},
			}
			_, err := calicoClient.IPPools().Create(context.Background(), &pool, options.SetOptions{})
			Expect(err).To(HaveOccurred())
```
## Context

Was trying to create a pool then create an affinity block within the pool. But was receiving an error that `The requested CIDR (192.168.0.0/24) is not within any configured pools.`. Realized that my pool wasn't being created because I had forgotten to give the pool a name. But it was silently failing pool creation.

## Your Environment

* libcalico-go version:

neiljerram:
Fixes #641

### DIFF
--- a/lib/clientv2/ippool_e2e_test.go
+++ b/lib/clientv2/ippool_e2e_test.go
@@ -58,6 +58,19 @@ var _ = testutils.E2eDatastoreDescribe("IPPool tests", testutils.DatastoreAll, f
 		IPIPMode: apiv2.IPIPModeNever,
 	}
 
+	It("should error when creating an IPPool with no name", func() {
+		c, err := clientv2.New(config)
+		Expect(err).NotTo(HaveOccurred())
+
+		pool := apiv2.IPPool{
+			Spec: apiv2.IPPoolSpec{
+				CIDR: "192.168.0.0/16",
+			},
+		}
+		_, err = c.IPPools().Create(ctx, &pool, options.SetOptions{})
+		Expect(err).To(HaveOccurred())
+	})
+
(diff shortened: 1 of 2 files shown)


caseydavenport:
Looks good to me, but @robbrockbank should look at this to see how it overlaps with any other validation work in the pipeline.

robbrockbank:
LGTM too, but this does overlap with the validation logic - and should be covered by the validation handling.  We can merge now if we think it makes sense to, or hold off and wait for the validator updates that should be coming within the next day or two.

caseydavenport:
@robbrockbank so long as the validation logic covers this, then I think we should fix #641 in that PR rather than this one.

I don't think there's any urgency to fixing this so long as it gets done.

neiljerram:
@robbrockbank @caseydavenport OTOH, this fix is ready now, and might usefully catch something (misconceived setup, I guess) before the rest of validation lands; and it can always be reverted by the validation PR.  So I think there's a case for merging this anyway - WDYT?

caseydavenport:
SGTM - let's merge and make sure to revert / move into the validation code once it exists.

CC @gunjan5

PR #644, opened 2017-10-31
Add NamespaceSelector to NetworkPolicy and GNP

caseydavenport:
## Description

Things I haven't done in this PR and don't intend to (though we might want to some day):

- Add `NotNamespaceSelector`
- Support a top-level `NamespaceSelector` on GNP

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
NetworkPolicy and GlobalNetworkPolicy include a new NamespaceSelector in rules
```

### DIFF
--- a/lib/apis/v2/policy.go
+++ b/lib/apis/v2/policy.go
@@ -79,6 +79,7 @@ type EntityRule struct {
 	// Nets is an optional field that restricts the rule to only apply to traffic that
 	// originates from (or terminates at) IP addresses in any of the given subnets.
 	Nets []string `json:"nets,omitempty" validate:"omitempty,dive,cidr"`
+
 	// Selector is an optional field that contains a selector expression (see Policy for
 	// sample syntax).  Only traffic that originates from (terminates at) endpoints matching
 	// the selector will be matched.
@@ -96,6 +97,19 @@ type EntityRule struct {
 	// The effect is that the latter will accept packets from non-Calico sources whereas the
 	// former is limited to packets from Calico-controlled endpoints.
 	Selector string `json:"selector,omitempty" validate:"omitempty,selector"`
+
+	// NamespaceSelector is an optional field that contains a selector expression. Only traffic
+	// that originates from (or terminates at) endpoints within the selected namespaces will be
+	// matched. When both NamespaceSelector and Selector are defined on the same rule, then only
+	// workload endpoints that are matched by both selectors will be selected by the rule.
+	//
+	// For NetworkPolicy, an empty NamespaceSelector implies that the Selector is limited to selecting
+	// only workload endpoints in the same namespace as the NetworkPolicy.
+	//
+	// For GlobalNetworkPolicy, an empty NamespaceSelector implies the Selector applies to workload
+	// endpoints across all namespaces.
+	NamespaceSelector string `json:"namespaceSelector,omitempty" validate:"omitempty,selector"`
+
 	// Ports is an optional field that restricts the rule to only apply to traffic that has a
 	// source (destination) port that matches one of these ranges/values. This value is a
 	// list of integers or strings that represent ranges of ports.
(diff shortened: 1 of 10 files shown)


lib/backend/syncersv1/updateprocessors/rules.go
```diff
@@ -116,6 +131,12 @@ func RuleAPIV2ToBackend(ar apiv2.Rule, ns string) model.Rule {
 	}
 }
 
+// parseNamespaceSelector takes a v2 namespace selector and returns the appropriate v1 representation
+// by prefix the keys with the `pcns.` prefix. For example, `k == 'v'` becomes `pcns.k == 'v'`.
+func parseNamespaceSelector(s string) string {
+	return "has(TODO)"
```
caseydavenport: @fasaxc I'd like to chat with you to figure out the right way to do this. We need to go through the selector and prefix any labels with `pcns.`, then return the modified selector. 

Looks like the lib/selector package and friends should have the right tools to do this.. but I can't quite find the right incantation, so might need some tweaking to make that work.

lib/selector/parser/ast.go
```diff
@@ -40,13 +43,52 @@ func (l MapAsLabels) Get(labelName string) (value string, present bool) {
...
+
+// PrefixVisitor implements the Visitor interface to allow prefixing of
+// label names within a selector.
+type PrefixVisitor struct {
+	Prefix string
+}
+
+func (v PrefixVisitor) Visit(n interface{}) {
+	log.Debugf("PrefixVisitor visiting node %#v", n)
+	switch n.(type) {
```
fasaxc: Use `switch np := n.(type)`, then you don't need the type assertions in each case.

lib/selector/parser/ast.go
```diff
@@ -214,6 +281,10 @@ func (node NotNode) Evaluate(labels Labels) bool {
 	return !node.Operand.Evaluate(labels)
 }
 
+func (node *NotNode) AcceptVisitor(v Visitor) {
+	node.Operand.AcceptVisitor(v)
```
fasaxc: It doesn't matter for your use case, but I think, in general, you should do `v.Visit(node)` before visiting the operand.

(Same for the other non-leaf nodes)

lib/selector/parser/ast.go
```diff
@@ -63,6 +105,10 @@ func (sel selectorRoot) EvaluateLabels(labels Labels) bool {
 	return sel.root.Evaluate(labels)
 }
 
+func (sel *selectorRoot) AcceptVisitor(v Visitor) {
+	sel.root.AcceptVisitor(v)
+}
+
 func (sel selectorRoot) String() string {
```
fasaxc: If we're starting to use these structs as pointers, we should change all the methods to have pointer receivers.

lib/apis/v2/policy.go
```diff
@@ -162,18 +163,27 @@ type EntityRule struct {
 	// The effect is that the latter will accept packets from non-Calico sources whereas the
 	// former is limited to packets from Calico-controlled endpoints.
 	Selector string `json:"selector,omitempty" validate:"omitempty,selector"`
+
+	// NamespaceSelector is an optional field that contains a selector expression. Only traffic
+	// that originates from (or terminates at) endpoints within the selected namespaces will be
+	// matched.
```
fasaxc: How is this defaulted for GNP and NP?

caseydavenport: For both it is defaulted to `""`, which indicates not to perform any namespace selection.

The implied behavior is different for the two resources:
- NP - implies select only within my own namespace
- GNP - implies select across all namespaces.

lib/selector/parser/ast.go
caseydavenport: Neat, I didn't know about that.

So, it does a type check _and_ a cast in one operation?

lib/selector/parser/ast.go
caseydavenport: ok, wasn't sure if that was desired or not - the other methods don't modify the node, so thought keeping the same was OK.

lib/selector/parser/ast.go
caseydavenport: d'oh yes, I meant to do this but forgot.

lib/apis/v2/policy.go
fasaxc: Sorry, was meaning, would it make sense to add a note to the comment to explain that `""` means different things in different types of policy?

caseydavenport: ah, yeah. can do.

lib/backend/syncersv1/updateprocessors/rules.go
```diff
@@ -116,6 +133,20 @@ func RuleAPIV2ToBackend(ar apiv2.Rule, ns string) model.Rule {
 	}
 }
 
+// parseNamespaceSelector takes a v2 namespace selector and returns the appropriate v1 representation
+// by prefix the keys with the `pcns.` prefix. For example, `k == 'v'` becomes `pcns.k == 'v'`.
```
robbrockbank: prefix (ing)

robbrockbank: isn't it better to prefix with "pcns/" (i.e. `/` rather than `.`) - that means we don't have to limit labels to not include the `pcns.` prefix.

caseydavenport: I discussed with @lxpollitt and decided to leave it as-is.

We'd still need to prevent people from labeling with `pcns/` (labels allow for the `/` character).

PR #646, opened 2017-11-01
Annotate natOutgoing used to be nat-outgoing in v1 api

bcreane:
## Description
Annotate natOutgoing used to be nat-outgoing in v1 api.

## Todos
- [ ] Tests -- I'm going to assume that upgrades, when they're supported, will test this.

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/apis/v2/ippool.go
+++ b/lib/apis/v2/ippool.go
@@ -42,7 +42,7 @@ type IPPoolSpec struct {
 	IPIPMode IPIPMode `json:"ipipMode,omitempty" validate:"omitempty,ipipmode"`
 	// When nat-outgoing is true, packets sent from Calico networked containers in
 	// this pool to destinations outside of this pool will be masqueraded.
-	NATOutgoing bool `json:"natOutgoing,omitempty"`
+	NATOutgoing bool `json:"natOutgoing,omitempty" confignamev1:"nat-outgoing"`
 	// When disabled is true, Calico IPAM will not assign addresses from this pool.
 	Disabled bool `json:"disabled,omitempty"`
 }


lib/apis/v2/ippool.go
```diff
@@ -42,7 +42,7 @@ type IPPoolSpec struct {
 	IPIPMode IPIPMode `json:"ipipMode,omitempty" validate:"omitempty,ipipmode"`
 	// When nat-outgoing is true, packets sent from Calico networked containers in
 	// this pool to destinations outside of this pool will be masqueraded.
-	NATOutgoing bool `json:"natOutgoing,omitempty"`
+	NATOutgoing bool `json:"natOutgoing,omitempty" confignamev1:"nat-outgoing"`
```
gunjan5: not sure if this is needed for IPPool fields. @robbrockbank?

robbrockbank: The confignamev1 annotation is only used for the config resources where we used to have a separate entry for each config key and now we have a single resource *and* have changed the names of some of the keys.

For IP pools we just convert the IPPool struct to the v1 struct as before.  The old YAML name should not matter.

robbrockbank: @bcreane :  is this trying to fix an issue with IPPools?

bcreane: @robbrockbank - nah. I noticed the field name changed with v2 and that other changed field names were annotated similarly. This fixes nothing, so we can discard the PR if you don't need to track `nat-outing` -> `natOutgoing`.

gunjan5: Yeah I think we can close this PR

bcreane:
closing ...

PR #647, opened 2017-11-02
Remove dependency on vendor directory from Makefile.

fasaxc:
## Description

This bit me because the vendor dir didn't update after I changed branches.

- Depending on a directory doesn't work because directory mtimes are
  updated outside of make.

- Rework the vendor target to use a touch file and add an update-vendor target
  to rebuild glide.lock.  (This is still done automatically if glide.lock is
  removed but I needed a second target to keep the old behaviour once the
  vendor target had been fixed.)

- Use a flag to ensure the vendor dir is only rebuilt once.  I couldn't think
  of a way to avoid this given the way make dependencies work.  There's no
  allowance for more than one target being able to complete a task.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,26 +16,49 @@ LIBCALICO-GO_PKG  = github.com/projectcalico/libcalico-go
 TOP_SRC_DIR       = lib
 MY_UID           := $(shell id -u)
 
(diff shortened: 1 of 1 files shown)


caseydavenport:
@fasaxc is this ready to go?

PR #648, opened 2017-11-02
Fix proto port json def

robbrockbank:
Set the json tags for the ProtoPort fields to ensure capitalization is correct.

### DIFF
--- a/lib/apis/v2/felixconfig.go
+++ b/lib/apis/v2/felixconfig.go
@@ -187,8 +187,8 @@ type FelixConfigurationSpec struct {
 }
 
 type ProtoPort struct {
-	Protocol string
-	Port     uint16
+	Protocol string `json:"protocol"`
+	Port     uint16 `json:"port"`
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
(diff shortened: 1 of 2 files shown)


lib/apis/v2/felixconfig.go
```diff
@@ -187,8 +187,8 @@ type FelixConfigurationSpec struct {
 }
 
 type ProtoPort struct {
-	Protocol string
-	Port     uint16
+	Protocol string `json:"protocol"`
+	Port     uint16 `json:"port"`
```
matthewdupre: How does this work if people don't want to specify the port, just the protocol?

caseydavenport: I think we want `omitempty`

robbrockbank: oh ok - didn't realize that was a thing.  okey dokey - i'll update.

robbrockbank: i assume that's just for ports.

do we need to be able to specify ranges of ports or multiple ports in the same thing or just put this back as a string that we'll have to validate or???

robbrockbank: @matthewdupre @caseydavenport ^^^

Issue #637, opened 2017-10-27
v2 datamodel KDD syncer isn't filtering host-networked pods


fasaxc:
Host networked pods are being passed through to Felix, with a made-up `caliABCD` interface name, for example:

```
2017-10-27 02:46:23.388 [DEBUG][76] conversion.go 110: Pod is host networked. pod="kube-proxy-b5l2l"
...
2017-10-27 02:46:23.397 [DEBUG][76] validation_filter.go 50: Validating KV pair. key=WorkloadEndpoint(node=fd-crc-kubeadm-master, orchestrator=k8s, workload=kube-system/kube-proxy-b5l2l, name=eth0) value=&model.WorkloadEndpoint{State:"active", Name:"cali9b0375b0140", ActiveInstanceID:"", Mac:(*net.MAC)(nil), ProfileIDs:[]string{"kns.kube-system"}, IPv4Nets:[]net.IPNet{net.IPNet{IPNet:net.IPNet{IP:net.IP{0xa, 0x80, 0x0, 0x2}, Mask:net.IPMask{0xff, 0xff, 0xff, 0xff}}}}, IPv6Nets:[]net.IPNet(nil), IPv4NAT:[]model.IPNAT(nil), IPv6NAT:[]model.IPNAT(nil), Labels:map[string]string{"controller-revision-hash":"3888521348", "k8s-app":"kube-proxy", "pod-template-generation":"1", "projectcalico.org/namespace":"kube-system", "projectcalico.org/orchestrator":"k8s"}, IPv4Gateway:(*net.IP)(nil), IPv6Gateway:(*net.IP)(nil), Ports:[]model.EndpointPort{}}
```

## Expected Behavior

Host networked pods shouldn't be passed through to Felix (or if there's a reason for passing them through, we need to adjust Felix).

## Current Behavior

Felix programs dispatch and per-endpoint chains for the phantom interfaces.  I don't think that has any functional impact, but it does add extra fluff to the dataplane.

## Possible Solution

Filter them out in the update processor?

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Your Environment

```
/ # /sbin/versions
confd v1.0.0-alpha1-rc3
BIRD version 1.6.3
libnetwork-plugin: v1.1.0-28-g20f511e
calico-bgp-daemon: v0.2.2

Felix:
2017-10-27 03:37:51.938 [INFO][656] logutils.go 83: Early screen log level set to debug
[...]

PR #649, opened 2017-11-03
Redfine when a Pod is a valid Calico Workload Endpoint

robbrockbank:
Update List/Watch to return KDD WEP that have a Node assigned and are not Host networked.  This removes the requirement of an IP address being assigned.

Fixes https://github.com/projectcalico/libcalico-go/issues/637

### DIFF
--- a/lib/backend/k8s/resources/watcher.go
+++ b/lib/backend/k8s/resources/watcher.go
@@ -93,7 +93,7 @@ func (crw *k8sWatcherConverter) processK8sEvents() {
 		case event := <-crw.k8sWatch.ResultChan():
 			e := crw.convertEvent(event)
 			select {
-			case crw.resultChan <- e:
+			case crw.resultChan <- *e:
 				crw.logCxt.Debug("Kubernetes event converted and sent to backend watcher")
 
 				// If this is an error event, check to see if it's a terminating one (the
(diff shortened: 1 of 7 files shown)


lib/backend/k8s/resources/watcher.go
```diff
@@ -118,48 +118,52 @@ func (crw *k8sWatcherConverter) processK8sEvents() {
...
 		if err != nil {
 			crw.logCxt.WithError(err).Warning("Error converting Kubernetes resource to Calico resource")
-			return api.WatchEvent{
+			return &api.WatchEvent{
 				Type:  api.WatchError,
 				Error: err,
 			}
 		}
+		if kvp == nil {
+			crw.logCxt.Info("Event converted to a no-op")
```
caseydavenport: Probably a debug log?

lib/backend/syncersv1/updateprocessors/processor.go
```diff
@@ -74,16 +80,17 @@ func (sup *simpleUpdateProcessor) Process(kvp *model.KVPair) ([]*model.KVPair, e
...
 			log.WithField("Resource", kvp.Key).Warn("Unable to process resource data - treating as deleted")
-			return []*model.KVPair{
-				&model.KVPair{
-					Key: v1key,
-				},
-			}, nil
+			return []*model.KVPair{{Key: v1key}}, nil
+		}
+		if v1value == nil {
+			// If the value is filtered out by the processor, treat as a delete.
```
caseydavenport: Doesn't this mean that any time we get any update for any host networked pod, we're going to send a Delete over the syncer?

robbrockbank: no - the UpdateProcessor contract with the watchersyncer explicitly allows multiple adds/deletes events and will send the appropriate event based on what is in the cache.

So if it gets a delete and it is not in the cache then the event is swallowed.
If we get a modified and the event is not in the cache it'll get converted to an Add.
If we get a delete and it is in the cache then we'll send a Delete.

PR #650, opened 2017-11-03
Delete all associated node data on delete

ozdanborne:
## Description

Remove all associated node data on node delete. This is effecitvely a port of [the same v1 behavior](https://github.com/projectcalico/libcalico-go/blob/master/lib/client/node.go#L80-L132) to v2.

## Todos
- [x] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/clientv2/node.go
+++ b/lib/clientv2/node.go
@@ -18,8 +18,12 @@ import (
 	"context"
 
 	apiv2 "github.com/projectcalico/libcalico-go/lib/apis/v2"
+	"github.com/projectcalico/libcalico-go/lib/names"
 	"github.com/projectcalico/libcalico-go/lib/options"
 	"github.com/projectcalico/libcalico-go/lib/watch"
+	"github.com/projectcalico/libcalico-go/lib/net"
+	"github.com/projectcalico/libcalico-go/lib/errors"
+	"fmt"
 )
 
 // NodeInterface has methods to work with Node resources.
@@ -59,6 +63,97 @@ func (r nodes) Update(ctx context.Context, res *apiv2.Node, opts options.SetOpti
 
 // Delete takes name of the Node and deletes it. Returns an error if one occurs.
 func (r nodes) Delete(ctx context.Context, name string, opts options.DeleteOptions) (*apiv2.Node, error) {
+	pname, err := names.WorkloadEndpointIdentifiers{Node: name}.CalculateWorkloadEndpointName(true)
+	if err != nil {
+		return nil, err
+	}
+
+	// Get all weps belonging to the node
+	weps, err := r.client.WorkloadEndpoints().List(ctx, options.ListOptions{
+		Prefix: true,
+		Name: pname,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	// Collate all IPs across all endpoints, and then release those IPs.
+	ips := []net.IP{}
+	for _, wep := range weps.Items {
+		// The prefix match is unfortunately not a perfect match on the Node (since it is theoretically possible for
+		// another node to match the prefix (e.g. a node name of the format <thisnode>-foobar would also match a prefix
+		// search of the node <thisnode>). Therefore, we will also need to check that the Spec.Node field matches the Node.
+		if wep.Spec.Node != name {
+			continue
+		}
+		for _, ip := range wep.Spec.IPNetworks {
+			ips = append(ips, *net.ParseIP(ip))
+		}
+	}
+	_, err = r.client.IPAM().ReleaseIPs(context.Background(), ips)
+	switch err.(type) {
+	case nil, errors.ErrorResourceDoesNotExist, errors.ErrorOperationNotSupported:
+	default:
+		return nil, err
+	}
+
+	// Delete the weps.
+	for _, wep := range weps.Items {
+		if wep.Spec.Node != name {
+			continue
+		}
+
+		_, err = r.client.WorkloadEndpoints().Delete(ctx, wep.Namespace, wep.Name, options.DeleteOptions{})
+		switch err.(type) {
+		case nil, errors.ErrorResourceDoesNotExist, errors.ErrorOperationNotSupported:
+		default:
+			return nil, err
+		}
+	}
+
+	// Remove the node from the IPAM data if it exists.
+	err = r.client.IPAM().RemoveIPAMHost(ctx, name)
+	switch err.(type) {
+	case nil, errors.ErrorResourceDoesNotExist, errors.ErrorOperationNotSupported:
+	default:
+		return nil, err
+	}
+
+	// Remove BGPPeers.
+	bgpPeers, err := r.client.BGPPeers().List(ctx, options.ListOptions{})
+	if err != nil {
+		return nil, err
+	}
+	for _, peer := range bgpPeers.Items {
+		if peer.Spec.Node != name {
+			continue
+		}
+		_, err = r.client.BGPPeers().Delete(ctx, peer.Name, options.DeleteOptions{})
+		switch err.(type) {
+		case nil, errors.ErrorResourceDoesNotExist, errors.ErrorOperationNotSupported:
+		default:
+			return nil, err
+		}
+	}
+
+	// Delete felix configuration
+	nodeConfName := fmt.Sprintf("node.%s", name)
+	_, err = r.client.FelixConfigurations().Delete(ctx, nodeConfName, options.DeleteOptions{})
+	switch err.(type) {
+	case nil, errors.ErrorResourceDoesNotExist, errors.ErrorOperationNotSupported:
+	default:
+		return nil, err
+	}
+
+	// Delete bgp configuration
+	_, err = r.client.BGPConfigurations().Delete(ctx, nodeConfName, options.DeleteOptions{})
+	switch err.(type) {
+	case nil, errors.ErrorResourceDoesNotExist, errors.ErrorOperationNotSupported:
+	default:
+		return nil, err
+	}
+
+	// Delete the node.
 	out, err := r.client.resources.Delete(ctx, opts, apiv2.KindNode, noNamespace, name)
 	if out != nil {
 		return out.(*apiv2.Node), err
--- a/lib/names/workloadendpoint.go
+++ b/lib/names/workloadendpoint.go
@@ -135,18 +135,11 @@ func (ids WorkloadEndpointIdentifiers) getSegments() ([]segment, error) {
 	endp := segment{value: ids.Endpoint, field: "endpoint", structField: "Endpoint"}
 	workl := segment{value: ids.Workload, field: "workload", structField: "Workload"}
 
-	// Node and Orchestrator are *always* required.
+	// Node is *always* required.
 	if len(node.value) == 0 {
 		return nil, cerrors.ErrorValidation{
 			ErroredFields: []cerrors.ErroredField{
 				{Name: node.field, Reason: "field should be assigned"},
-			},
-		}
-	}
-	if len(orch.value) == 0 {
-		return nil, cerrors.ErrorValidation{
-			ErroredFields: []cerrors.ErroredField{
-				{Name: orch.field, Reason: "field should be assigned"},
 			},
 		}
 	}
--- a/lib/names/workloadendpoint_test.go
+++ b/lib/names/workloadendpoint_test.go
@@ -130,7 +130,7 @@ var _ = DescribeTable("WorkloadEndpoint name construction, name prefix",
 		Node:     "node-1",
 		Pod:      "pod-name",
 		Endpoint: "eth0",
-	}, "", "orchestrator"),
+	}, "node--1-", ""),
 )
 
 var _ = DescribeTable("WorkloadEndpoint name matching",
@@ -154,7 +154,7 @@ var _ = DescribeTable("WorkloadEndpoint name matching",
 		Node:     "node-1",
 		Pod:      "pod-name",
 		Endpoint: "eth0",
-	}, "", false, "orchestrator"),
+	}, "", false, ""),
 	Entry("Valid k8s endpoint and matching name", names.WorkloadEndpointIdentifiers{
 		Node:         "node-1",
 		Orchestrator: "k8s",
--- a/lib/clientv2/node_e2e_test.go
+++ b/lib/clientv2/node_e2e_test.go
@@ -16,6 +16,7 @@ package clientv2_test
 
 import (
 	"time"
+	"net"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/ginkgo/extensions/table"
@@ -31,6 +32,10 @@ import (
 	"github.com/projectcalico/libcalico-go/lib/options"
 	"github.com/projectcalico/libcalico-go/lib/testutils"
 	"github.com/projectcalico/libcalico-go/lib/watch"
+	cnet "github.com/projectcalico/libcalico-go/lib/net"
+	"github.com/projectcalico/libcalico-go/lib/ipam"
+	"github.com/projectcalico/libcalico-go/lib/backend/model"
+	"fmt"
 )
 
 var _ = testutils.E2eDatastoreDescribe("Node tests", testutils.DatastoreEtcdV3, func(config apiconfig.CalicoAPIConfig) {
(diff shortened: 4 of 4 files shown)


lib/clientv2/node.go
```diff
@@ -59,6 +63,65 @@ func (r nodes) Update(ctx context.Context, res *apiv2.Node, opts options.SetOpti
...
+
+	// Collate all IPs across all endpoints, and then release those IPs.
+	ips := []net.IP{}
+	for _, wep := range weps.Items {
+		for _, ip := range wep.Spec.IPNetworks {
+			ips = append(ips, *net.ParseIP(ip))
+		}
+	}
+	_, err = r.client.IPAM().ReleaseIPs(context.Background(), ips)
+	fmt.Printf("Releasing: %v", ips)
```
caseydavenport: Should update to use the logger (if actually needed - does the IPAM code log this out?)

caseydavenport: same with other `fmt.Printf` calls in this file

lib/names/workloadendpoint.go
```diff
@@ -143,13 +143,6 @@ func (ids WorkloadEndpointIdentifiers) getSegments() ([]segment, error) {
 			},
 		}
 	}
-	if len(orch.value) == 0 {
```
caseydavenport: Why were we requiring this before? @robbrockbank?

Also, the comment on line 138 needs updating to match this change.

lib/clientv2/node.go
ozdanborne: Oops, leftover debug statements.

lib/names/workloadendpoint.go
ozdanborne: I don't think I implemented this correctly. I think we can pass when _just_ a node is specified, but we should error when all identifiers are provided besides orchestrator. right?

lib/clientv2/node.go
```diff
@@ -59,6 +62,63 @@ func (r nodes) Update(ctx context.Context, res *apiv2.Node, opts options.SetOpti
 
 // Delete takes name of the Node and deletes it. Returns an error if one occurs.
 func (r nodes) Delete(ctx context.Context, name string, opts options.DeleteOptions) (*apiv2.Node, error) {
+	pname, err := names.WorkloadEndpointIdentifiers{Node: name}.CalculateWorkloadEndpointName(true)
```
robbrockbank: I think the WEPs, IP and IPAMHost only need to be done for !KDD

lib/clientv2/node.go
```diff
@@ -59,6 +62,63 @@ func (r nodes) Update(ctx context.Context, res *apiv2.Node, opts options.SetOpti
...
+	}
+	_, err = r.client.IPAM().ReleaseIPs(context.Background(), ips)
+	if err != nil {
+		return nil, err
+	}
+
+	// Delete the weps.
+	for _, wep := range weps.Items {
+		_, err = r.client.WorkloadEndpoints().Delete(ctx, wep.Namespace, wep.Name, options.DeleteOptions{})
+		if err != nil {
```
robbrockbank: You could ignore DoesNotExist error type here.

lib/clientv2/node.go
```diff
@@ -59,6 +62,63 @@ func (r nodes) Update(ctx context.Context, res *apiv2.Node, opts options.SetOpti
...
+		Prefix: true,
+		Name: pname,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	// Collate all IPs across all endpoints, and then release those IPs.
+	ips := []net.IP{}
+	for _, wep := range weps.Items {
```
robbrockbank: The prefix match is unfortunately not a perfect match on the Node (since it is theoretically possible for another node to match the prefix (e.g.  a node name of the format `<thisnode>-foobar` would also match a prefix search of the node `<thisnode>`).  Therefore, in this branch and the branch where you delete the WEPs you will also need to check that the Spec.Node field matches the Node.

lib/clientv2/node.go
```diff
@@ -59,6 +62,63 @@ func (r nodes) Update(ctx context.Context, res *apiv2.Node, opts options.SetOpti
...
+	}
+	for _, peer := range bgpPeers.Items {
+		if peer.Spec.Node == name {
+			_, err = r.client.BGPPeers().Delete(ctx, peer.Name, options.DeleteOptions{})
+			if _, ok := err.(errors.ErrorResourceDoesNotExist); ok {
+				return nil, err
+			}
+		}
+	}
+
```
robbrockbank: I think we should also delete FelixConfiguration and BGPConfiguration that has been created for the node - they both have the name "node.<thisnode>"

lib/clientv2/node.go
ozdanborne: Isn't it up to the underlying backend driver to no-op on that?

lib/names/workloadendpoint_test.go
```diff
@@ -150,11 +150,6 @@ var _ = DescribeTable("WorkloadEndpoint name matching",
 		Pod:          "pod-name",
 		Endpoint:     "eth0",
 	}, "", false, "node"),
-	Entry("Orchestrator is missing", names.WorkloadEndpointIdentifiers{
```
ozdanborne: @robbrockbank how would you expect this code to work if everything but an orchestrator was provided? An error?

lib/clientv2/node.go
robbrockbank: kinda - so if you want to go that route then you need to make sure you handle OperationNotSupported errors and just swallow them.

However, for KDD we don't do anything clever about prefix lists for WEPs, so you are basically gonna list all of the workload endpoints (Pods) and then do nothing with it.  Seems like a bit of an overkill.

lib/clientv2/node.go
```diff
@@ -59,6 +63,98 @@ func (r nodes) Update(ctx context.Context, res *apiv2.Node, opts options.SetOpti
...
+		return nil, err
+	}
+
+	// Collate all IPs across all endpoints, and then release those IPs.
+	ips := []net.IP{}
+	for _, wep := range weps.Items {
+		// The prefix match is unfortunately not a perfect match on the Node (since it is theoretically possible for
+		// another node to match the prefix (e.g. a node name of the format <thisnode>-foobar would also match a prefix
+		// search of the node <thisnode>). Therefore, we will also need to check that the Spec.Node field matches the Node.
+		if wep.Spec.Node == name {
```
robbrockbank: to reduce nesting, it would be better to do the opposite test and just continue.

```
if wep.Spec.Node != name {
    continue
}
```

lib/clientv2/node.go
```diff
@@ -59,6 +63,98 @@ func (r nodes) Update(ctx context.Context, res *apiv2.Node, opts options.SetOpti
...
+			if _, ok := err.(errors.ErrorResourceDoesNotExist); ok {
+				return nil, err
+			}
+		}
+	}
+
+	// Remove the node from the IPAM data if it exists.
+	err = r.client.IPAM().RemoveIPAMHost(ctx, name)
+	if err != nil {
+		if _, ok := err.(errors.ErrorResourceDoesNotExist); ok {
```
robbrockbank: I think this logic is the wrong way round, it should be `!ok`

also, you can combine these two ifs:

```
if _, ok := err.(errors.ErrorResourceDoesNotExist); !ok && err != nil {
}
```

lib/clientv2/node.go
```diff
@@ -59,6 +63,98 @@ func (r nodes) Update(ctx context.Context, res *apiv2.Node, opts options.SetOpti
...
+
+	// Delete felix configuration
+	fConfs, err := r.client.FelixConfigurations().List(ctx, options.ListOptions{})
+	if err != nil {
+		return nil, err
+	}
+	for _, fc := range fConfs.Items {
+		if fc.ObjectMeta.Name == fmt.Sprintf("node.%s", name) {
+			_, err = r.client.FelixConfigurations().Delete(ctx, fc.ObjectMeta.Name, options.DeleteOptions{})
+			if _, ok := err.(errors.ErrorResourceDoesNotExist); ok {
```
robbrockbank: logic is wrong.  Should be `!ok && err != nil`

lib/clientv2/node.go
```diff
@@ -59,6 +63,98 @@ func (r nodes) Update(ctx context.Context, res *apiv2.Node, opts options.SetOpti
...
+
+	// Delete bgp configuration
+	bgpConfs, err := r.client.BGPConfigurations().List(ctx, options.ListOptions{})
+	if err != nil {
+		return nil, err
+	}
+	for _, bgpc := range bgpConfs.Items {
+		if bgpc.ObjectMeta.Name == fmt.Sprintf("node.%s", name)  {
+			_, err = r.client.BGPConfigurations().Delete(ctx, bgpc.ObjectMeta.Name, options.DeleteOptions{})
+			if _, ok := err.(errors.ErrorResourceDoesNotExist); ok {
```
robbrockbank: ditto.

lib/clientv2/node.go
```diff
@@ -59,6 +63,98 @@ func (r nodes) Update(ctx context.Context, res *apiv2.Node, opts options.SetOpti
...
+	_, err = r.client.IPAM().ReleaseIPs(context.Background(), ips)
+	if err != nil {
+		return nil, err
+	}
+
+	// Delete the weps.
+	for _, wep := range weps.Items {
+		if wep.Spec.Node == name {
+			_, err = r.client.WorkloadEndpoints().Delete(ctx, wep.Namespace, wep.Name, options.DeleteOptions{})
+			if _, ok := err.(errors.ErrorResourceDoesNotExist); ok {
```
robbrockbank: `!ok && err != nil`

lib/clientv2/node.go
robbrockbank: similar in a couple of places below too.

lib/clientv2/node.go
robbrockbank: prolly need to add a test which creates and then deletes a node so that there isn't any of the other per-node config kicking around.

lib/clientv2/node.go
```diff
@@ -59,6 +63,98 @@ func (r nodes) Update(ctx context.Context, res *apiv2.Node, opts options.SetOpti
...
+	}
+
+	// Delete felix configuration
+	fConfs, err := r.client.FelixConfigurations().List(ctx, options.ListOptions{})
+	if err != nil {
+		return nil, err
+	}
+	for _, fc := range fConfs.Items {
+		if fc.ObjectMeta.Name == fmt.Sprintf("node.%s", name) {
+			_, err = r.client.FelixConfigurations().Delete(ctx, fc.ObjectMeta.Name, options.DeleteOptions{})
```
robbrockbank: You're deleting the wrong name here.

lib/clientv2/node.go
```diff
@@ -59,6 +63,98 @@ func (r nodes) Update(ctx context.Context, res *apiv2.Node, opts options.SetOpti
...
+	}
+
+	// Delete bgp configuration
+	bgpConfs, err := r.client.BGPConfigurations().List(ctx, options.ListOptions{})
+	if err != nil {
+		return nil, err
+	}
+	for _, bgpc := range bgpConfs.Items {
+		if bgpc.ObjectMeta.Name == fmt.Sprintf("node.%s", name)  {
+			_, err = r.client.BGPConfigurations().Delete(ctx, bgpc.ObjectMeta.Name, options.DeleteOptions{})
```
robbrockbank: You're deleting the wrong name here

robbrockbank: Maybe just construct the string "node.<nodename>" before you start deleting the Felix and BGP config and use that to delete both.

lib/clientv2/node.go
robbrockbank: actually scrap that last comment - we have  a test that simply adds and deletes a node without any of the secondary config -- not sure why that wasn't causing failures though.

lib/clientv2/node.go
```diff
@@ -59,6 +63,85 @@ func (r nodes) Update(ctx context.Context, res *apiv2.Node, opts options.SetOpti
...
+			continue
+		}
+		_, err = r.client.BGPPeers().Delete(ctx, peer.Name, options.DeleteOptions{})
+		if _, ok := err.(errors.ErrorResourceDoesNotExist); !ok && err != nil {
+			return nil, err
+		}
+	}
+
+	// Delete felix configuration
+	nodeConfName := fmt.Sprintf("node.%s", name)
```
ozdanborne: @robbrockbank does this need to be dash-escaped?

robbrockbank: nope

lib/clientv2/node.go
```diff
@@ -59,7 +63,117 @@ func (r nodes) Update(ctx context.Context, res *apiv2.Node, opts options.SetOpti
...
+		if wep.Spec.Node != name {
+			continue
+		}
+		for _, ip := range wep.Spec.IPNetworks {
+			ips = append(ips, *net.ParseIP(ip))
+		}
+	}
+	_, err = r.client.IPAM().ReleaseIPs(context.Background(), ips)
+	if err != nil {
+		if _, ok := err.(errors.ErrorResourceDoesNotExist); !ok {
```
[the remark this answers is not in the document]
robbrockbank: Not sure if the `case nil` is good practice or not, but certainly seems to work and means you don't have to do the `err != nil` comparison separately.

lib/clientv2/node.go
```diff
@@ -59,7 +63,117 @@ func (r nodes) Update(ctx context.Context, res *apiv2.Node, opts options.SetOpti
...
+		if wep.Spec.Node != name {
+			continue
+		}
+
+		_, err = r.client.WorkloadEndpoints().Delete(ctx, wep.Namespace, wep.Name, options.DeleteOptions{})
+		if err != nil {
+			if _, ok := err.(errors.ErrorResourceDoesNotExist); ok {
+				continue
+			}
+			if _, ok := err.(errors.ErrorOperationNotSupported); ok {
```
robbrockbank: Probably want to break here rather than continue.

lib/clientv2/node.go
```diff
@@ -59,7 +63,117 @@ func (r nodes) Update(ctx context.Context, res *apiv2.Node, opts options.SetOpti
...
+				continue
+			}
+			return nil, err
+		}
+	}
+
+	// Remove the node from the IPAM data if it exists.
+	err = r.client.IPAM().RemoveIPAMHost(ctx, name)
```
robbrockbank: Not sure what happens with IPAM for k8s - does this simply no-op for k8s, or does it return an OperationNotSupported?

lib/clientv2/node.go
caseydavenport: yep, agreed.

lib/clientv2/node.go
caseydavenport: @ozdanborne could you double check? I'd expect it to return OpNotSupported, but could be wrong.

ozdanborne: Looks like it just does a KV get, which can return a `ErrorOperationNotSupported` for k8s backend

lib/clientv2/node.go
```diff
@@ -92,10 +92,10 @@ func (r nodes) Delete(ctx context.Context, name string, opts options.DeleteOptio
 	}
 	_, err = r.client.IPAM().ReleaseIPs(context.Background(), ips)
 	if err != nil {
-		if _, ok := err.(errors.ErrorResourceDoesNotExist); !ok {
-			if _, ok := err.(errors.ErrorOperationNotSupported); !ok {
-				return nil, err
-			}
+		switch err.(type) {
+		case nil, errors.ErrorResourceDoesNotExist, errors.ErrorOperationNotSupported:
```
caseydavenport: Since you're checking for `nil` here, this doesn't need to be inside an `if err != nil` clause

lib/clientv2/node.go
```diff
@@ -59,7 +63,103 @@ func (r nodes) Update(ctx context.Context, res *apiv2.Node, opts options.SetOpti
...
+	if err != nil {
+		return nil, err
+	}
+	for _, peer := range bgpPeers.Items {
+		if peer.Spec.Node != name {
+			continue
+		}
+		_, err = r.client.BGPPeers().Delete(ctx, peer.Name, options.DeleteOptions{})
+		switch err.(type) {
+		case nil, errors.ErrorResourceDoesNotExist, errors.ErrorOperationNotSupported:
```
robbrockbank: Peers are supported on all datastore - so I don't think we should have the OperationNotSupported here as it'll potentially hide future bugs if we introduce another datastore.

robbrockbank: Same applies to BGPConfiguration and FelixConfiguration

lib/clientv2/node.go
```diff
@@ -59,7 +63,103 @@ func (r nodes) Update(ctx context.Context, res *apiv2.Node, opts options.SetOpti
...
+	switch err.(type) {
+	case nil, errors.ErrorResourceDoesNotExist, errors.ErrorOperationNotSupported:
+	default:
+		return nil, err
+	}
+
+	// Delete the node.
 	out, err := r.client.resources.Delete(ctx, opts, apiv2.KindNode, noNamespace, name)
+	switch err.(type) {
+	case nil, errors.ErrorResourceDoesNotExist, errors.ErrorOperationNotSupported:
```
robbrockbank: When deleting the actual node we should return the ResourceDoesNotExist error if there was no per-node configuration to delete at all.

The OperationNotSupported is more interesting as we may have deleted some other config.  @caseydavenport - what do you reckon for this error - perhaps return a not supported if no other config was deleted?

lib/clientv2/node.go
ozdanborne: I'm confused - It will only hide `OperationNotSupported` errors.  Isn't that what we want?

lib/clientv2/node.go
ozdanborne: Looks like we didn't swallow any errors on the v1 api: https://github.com/projectcalico/libcalico-go/blob/master/lib/client/node.go#L80-L132

lib/clientv2/node.go
robbrockbank: I'm saying that OperationNotSupported should never happen for BGPPeers, BGPConfig and FelixConfig on either KDD or etcdv3

We could leave it in place since it won't make any effective difference for now, but i don't really think we should be swallowing errors that we don't expect.

robbrockbank: ah leave it - perhaps it's safer in the long run

ozdanborne: I'll take the `git blame` when someday it comes back to bit us

ozdanborne:
I got a verbal LGTM from @caseydavenport

PR #651, opened 2017-11-03
re-watch rather than re-sync

robbrockbank:
## Description
If the reason for the watch failure was closure by the remote end, retry watching before doing a full resync.

-  In the KDD watcher code, if the channel was closed by the remote, flag that in the error
-  In the watchersyncer, keep track of the "last seen" revision
-  In the watchersyncer, if the channel was closed by the remote, start watching from the last seen revision.

I have also added a 30min timeout for the CRDs - the default if not specified is 1 minute.  In the code review comments I point to a couple of places in the k8s code the show this timeout behavior.

I have not added additional handling of different error types - I think that would require significant testing overhead that is not going to be possible within the timeframe.

### DIFF
--- a/lib/backend/k8s/resources/watcher.go
+++ b/lib/backend/k8s/resources/watcher.go
@@ -19,7 +19,7 @@ import (
 	"fmt"
 	"sync/atomic"
 
-	log "github.com/sirupsen/logrus"
+	"github.com/sirupsen/logrus"
 	kwatch "k8s.io/apimachinery/pkg/watch"
 
 	"github.com/projectcalico/libcalico-go/lib/backend/api"
@@ -39,7 +39,7 @@ func newK8sWatcherConverter(
 ) api.WatchInterface {
 	ctx, cancel := context.WithCancel(ctx)
 	wc := &k8sWatcherConverter{
-		logCxt:     log.WithField("resource", name),
+		logCxt:     logrus.WithField("resource", name),
 		converter:  converter,
 		k8sWatch:   k8sWatch,
 		context:    ctx,
@@ -51,7 +51,7 @@ func newK8sWatcherConverter(
 }
 
 type k8sWatcherConverter struct {
-	logCxt     *log.Entry
+	logCxt     *logrus.Entry
 	converter  ConvertK8sResourceToKVPair
 	k8sWatch   kwatch.Interface
 	context    context.Context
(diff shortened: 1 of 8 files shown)


lib/backend/k8s/resources/watcher.go
```diff
@@ -90,18 +90,34 @@ func (crw *k8sWatcherConverter) processK8sEvents() {
 
 	for {
 		select {
-		case event := <-crw.k8sWatch.ResultChan():
-			e := crw.convertEvent(event)
+		case event, ok := <-crw.k8sWatch.ResultChan():
+			var e *api.WatchEvent
+			if !ok {
+				// The channel is closed so send a terminating watche event indicating the watch was
```
neiljerram: watche -> watcher (I guess)

lib/backend/watchersyncer/watchercache.go
```diff
@@ -122,53 +130,73 @@ func (wc *watcherCache) resyncAndCreateWatcher() {
...
+			if performFullResync {
+				wc.logger.Debug("Wait for watch poll interval")
+				time.Sleep(WatchPollInterval)
+			} else {
+				wc.logger.Debug("Wait for list retry interval")
+				time.Sleep(ListRetryInterval)
+			}
+
+			// Failure to watch, trigger a full resync.
+			performFullResync = true
```
neiljerram: Hmm, I'm not fully convinced by the logic and comments here.

1. If "some resource types cannot be watched", that is presumably something that we can predict for each resource type?  In that case - and given that it feels like <failure to watch a resource that we don't expect to be watchable> should be treated quite differently from <failure to watch a resource that *should* be watchable> - then it feels like canBeWatched should be a property of this object that is set by whatever toplevel it is that creates an object for each resource type.

2. Then, if this was a watcher for which canBeWatched is false, why would we even try to watch it at all?

3. In the specific code just above, why sometimes sleep for ListRetryInterval, instead of WatchPollInterval, when just about to do a full resync?

4. For resource types that can be watched, what are the conditions that can cause a Watch() call to fail?  Is doing a full resync the right recovery, for all of those conditions?

robbrockbank: 1. Yup we certainly could add that additional flag and I had considered it but omitted it originally due to time contraints, and in the case of not being able to watch we'd get an OperationNotSupported error back from KDD (since all etcdv3 resources are watchable) without any round trips to the API server.  So yes something I can add if you think it's necessary at this stage.

2. Yup we could do, just more stuff to plumb through and didn't seem immediately necessary since the cost of getting an operation not supported should be pretty small since it won't require any round trips to etcd. 

3.  The Watch poll is longer than the List retry interval - I just wanted to minimize the delay when doing an initial re-watch that wasn't actually possible because (for example) of compaction (the revision is no longer available) - in this case we shouldn't need to wait that long - in actual fact in this scenario I don't think I should be pausing at all - so i think I should pause for the Watch interval if we were doing a full resend, and if not don't pause at all.  WDYT?

robbrockbank: 4.  Yup I agree and I did start looking into the different error cases that we should handle by retrying the watch rather than retrying the full resync.  However - that felt like a scarier change to rush in to a release without decent amounts of testing.  The code change in this PR, I believe, reinstates the same effective behavior that you'd implemented in the old KDD code - that was certainly the intention, but if that's not true then that is an issue of itself.

robbrockbank: In terms of etcd error types, I think the only error type that requires a full resync is a 401 but I think I'd rather make a change like that post-Beta when we can do a bunch of failure tests and check recovery is sensible.

robbrockbank: OK, I did some code tracing through kubernetes and the API server.  I _think_ I've found the source of the 1 minute timeout:

https://github.com/kubernetes/apiextensions-apiserver/blob/master/pkg/apiserver/customresource_handler.go#L187

Which then gets used the API server rest handler:

https://github.com/kubernetes/apiserver/blob/release-1.8/pkg/endpoints/handlers/rest.go#L359

From looking at this code, I think it should be sufficient to ensure we add a sensible TimeoutSeconds on the ListOptions that we use for CRDs, so I will add that too as part of this PR.

PR #652, opened 2017-11-06
Fix validation for ICMP type and code in backend rules.

fasaxc:
## Description

- 0 (ping reply) should have been allowed.
- 255 (unused) should have been foridden due to lack of kernel support.

Frontend API models were already correct.

Fixes the validation part of https://github.com/projectcalico/felix/issues/1604

## Todos
- [x] Tests
- [x] Documentation
- [x] Release note

## Release Note

```release-note
Fix incorrect validation of ICMP type to allow 0, which is a valid value.
```


caseydavenport:
@fasaxc does this need to go into v2.x as well?

fasaxc:
@caseydavenport Yes, it probably should: #654

PR #655, opened 2017-11-06
Add ready flag to ClusterInformation.

fasaxc:
## Description

@robbrockbank Does this look vaguely sane?

## Todos
- [x] Check whether changing the type of the ready flag is an issue for Typha/Felix upgrade
- [x] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/backend/syncersv1/updateprocessors/clusterinfoprocessor.go
+++ b/lib/backend/syncersv1/updateprocessors/clusterinfoprocessor.go
@@ -27,8 +27,24 @@ func NewClusterInfoUpdateProcessor() watchersyncer.SyncerUpdateProcessor {
 	return NewConfigUpdateProcessor(
 		reflect.TypeOf(apiv2.ClusterInformationSpec{}),
 		DisallowAnnotations,
-		func(node, name string) model.Key { return model.HostConfigKey{Hostname: node, Name: name} },
-		func(name string) model.Key { return model.GlobalConfigKey{Name: name} },
-		nil,
+		func(node, name string) model.Key {
+			if name == "DatastoreReady" {
+				return nil
+			}
+			return model.HostConfigKey{Hostname: node, Name: name}
+		},
+		func(name string) model.Key {
+			if name == "DatastoreReady" {
+				return model.ReadyFlagKey{}
+			}
+			return model.GlobalConfigKey{Name: name}
+		},
+		map[string]ConfigFieldValueToV1ModelValue{
+			"DatastoreReady": datastoreReadyToBool,
+		},
 	)
+}
+
+func datastoreReadyToBool(value interface{}) interface{} {
+	return value.(bool)
 }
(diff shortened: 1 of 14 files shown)


lib/backend/syncersv1/updateprocessors/clusterinfoprocessor.go
```diff
@@ -27,8 +27,20 @@ func NewClusterInfoUpdateProcessor() watchersyncer.SyncerUpdateProcessor {
 	return NewConfigUpdateProcessor(
 		reflect.TypeOf(apiv2.ClusterInformationSpec{}),
 		DisallowAnnotations,
-		func(node, name string) model.Key { return model.HostConfigKey{Hostname: node, Name: name} },
-		func(name string) model.Key { return model.GlobalConfigKey{Name: name} },
+		func(node, name string) model.Key {
+			if name == "DatastoreReady" {
+				return model.HostReadyFlagKey{
```
robbrockbank: Do you think it's worth having the Host ready flag? or should we just handle this returning nil?

lib/clientv2/client.go
```diff
@@ -153,12 +165,134 @@ func (p poolAccessor) GetEnabledPools(ipVersion int) ([]net.IPNet, error) {
 // Most Calico deployment scenarios will automatically implicitly invoke this
 // method and so a general consumer of this API can assume that the datastore
 // is already initialized.
-func (c client) EnsureInitialized() error {
+func (c client) EnsureInitialized(ctx context.Context, calicoVersion, clusterType string) error {
```
robbrockbank: Looks like we're missing a UT for this.

lib/clientv2/client.go
```diff
@@ -153,12 +165,134 @@ func (p poolAccessor) GetEnabledPools(ipVersion int) ([]net.IPNet, error) {
...
+	if err := c.ensureClusterInformation(ctx, calicoVersion, clusterType); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+const globalClusterInfoName = "default"
+
+// ensureClusterInformation ensures that the ClusterInformation fields i.e. ClusterType,
```
robbrockbank: Given that we have this now, should we be calling this from the Node create method to reinstate the HostEndpoint use case.

robbrockbank:
@fasaxc Looks sane to me - couple of comments inline.

lib/backend/syncersv1/updateprocessors/clusterinfoprocessor.go
fasaxc: Think I prefer nil, let's do that.

lib/backend/syncersv1/updateprocessors/configurationprocessor.go
```diff
@@ -61,7 +61,7 @@ func NewConfigUpdateProcessor(
 	allowAnnotations bool,
 	nodeConfigKeyFn NodeConfigKeyFn,
 	globalConfigKeyFn GlobalConfigKeyFn,
-	stringifiers map[string]ValueToStringFn,
+	stringifiers map[string]ConfigFieldValueToV1ModelValue,
```
robbrockbank: stringifiers -> valueConverters

PR #656, opened 2017-11-07
Fix updateprocessor conversion of rules

caseydavenport:
## Description

Update processor wasn't handling rules which were not namespaced + had no namespace selector. 

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


lib/backend/syncersv1/updateprocessors/rules.go
```diff
@@ -90,6 +90,9 @@ func RuleAPIV2ToBackend(ar apiv2.Rule, ns string) model.Rule {
 		} else {
 			srcSelector = sourceNSSelector
 		}
+	} else {
```
mgleung: would it be easier to set these values on line 78 and only overwrite it if there is a NS selector?

caseydavenport: yes, I considered that approach. I thought this might be easier to read, but happy to change if you'd like.

PR #660, opened 2017-11-08
Combine CRD+K8s NP resource versions in the Calico NP resource version

robbrockbank:
## Description

This PR updates the KDD NetworkPolicy client to use a ResourceVersion for the NP resources that is a combination of the CRD backed NP and the k8s native NP.  

The primary use case is for the WatcherSyncer code, but I've plumbed this through to all of the CRUD operations.

### DIFF
--- a/lib/backend/k8s/resources/networkpolicy.go
+++ b/lib/backend/k8s/resources/networkpolicy.go
@@ -62,16 +62,15 @@ func NewNetworkPolicyClient(c *kubernetes.Clientset, r *rest.RESTClient) K8sReso
 	return &networkPolicyClient{
 		clientSet: c,
 		crdClient: crdClient,
-		converter: conversion.Converter{},
 	}
 }
 
 // Implements the api.Client interface for NetworkPolicys.
 type networkPolicyClient struct {
+	conversion.Converter
 	resourceName string
 	clientSet    *kubernetes.Clientset
 	crdClient    *customK8sResourceClient
-	converter    conversion.Converter
 }
 
 func (c *networkPolicyClient) Create(ctx context.Context, kvp *model.KVPair) (*model.KVPair, error) {
@@ -84,11 +83,19 @@ func (c *networkPolicyClient) Create(ctx context.Context, kvp *model.KVPair) (*m
 			Operation:  "Create",
 		}
 	}
-	return c.crdClient.Create(ctx, kvp)
+
+	kvp, err := c.crdClient.Create(ctx, kvp)
+	if kvp != nil {
+		// Convert the revision to the combined CRD/k8s revision - the k8s rev will be empty, but this
+		// format will allow the revision to be passed into List and Watch calls.
+		kvp.Revision = c.JoinNetworkPolicyRevisions(kvp.Revision, "")
+	}
+	return kvp, err
 }
 
 func (c *networkPolicyClient) Update(ctx context.Context, kvp *model.KVPair) (*model.KVPair, error) {
 	log.Debug("Received Update request on NetworkPolicy type")
+
 	key := kvp.Key.(model.ResourceKey)
 	if strings.HasPrefix(key.Name, conversion.K8sNetworkPolicyNamePrefix) {
 		// We don't support Update of a Kubernetes NetworkPolicy.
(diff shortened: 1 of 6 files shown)


lib/backend/k8s/resources/networkpolicy.go
```diff
@@ -97,7 +105,22 @@ func (c *networkPolicyClient) Update(ctx context.Context, kvp *model.KVPair) (*m
...
+	// The revision, if supplied, will be a combination of CRD and k8s-backed revisions.  Extract
+	// the CRD rev and use that for the update.
+	crdRev, _, err := splitRevision(kvp.Revision)
+	if err != nil {
+		return nil, err
+	}
+	kvp.Revision = crdRev
+	kvp, err = c.crdClient.Update(ctx, kvp)
+
+	// Convert the revision back to the combined CRD/k8s revision - the k8s rev will be empty, but this
```
caseydavenport: nit: the comment should be inside the `kvp != nil` block, I think.

lib/backend/k8s/resources/networkpolicy.go
```diff
@@ -130,26 +166,48 @@ func (c *networkPolicyClient) Get(ctx context.Context, key model.Key, revision s
...
+			kvp.Revision = joinRevisions("", kvp.Revision)
+		}
+		return kvp, err
 	} else {
-		return c.crdClient.Get(ctx, k, revision)
+		// The revision, if supplied, will be a combination of CRD and k8s-backed revisions.  Extract
+		// the CRD rev and use that for the update.
+		kvp, err := c.crdClient.Get(ctx, k, crdRev)
+
+		// Convert the revision back to the combined CRD/k8s revision - the k8s rev will be empty.
```
caseydavenport: also here: comment should be inside the if statement.

lib/backend/k8s/resources/networkpolicy.go
```diff
@@ -130,26 +166,48 @@ func (c *networkPolicyClient) Get(ctx context.Context, key model.Key, revision s
...
+		kvp, err := c.converter.K8sNetworkPolicyToCalico(&networkPolicy)
+
+		// Convert the revision back to the combined CRD/k8s revision - the CRD rev will be empty.
+		if kvp != nil {
+			kvp.Revision = joinRevisions("", kvp.Revision)
+		}
+		return kvp, err
 	} else {
-		return c.crdClient.Get(ctx, k, revision)
+		// The revision, if supplied, will be a combination of CRD and k8s-backed revisions.  Extract
```
caseydavenport: comment seems wrong, we're not extracting it here.

lib/clientv2/networkpolicy_e2e_test.go
```diff
@@ -77,20 +77,28 @@ var _ = testutils.E2eDatastoreDescribe("NetworkPolicy tests", testutils.Datastor
 			be.Clean()
 
 			By("Updating the NetworkPolicy before it is created")
+			var rv string
+			if config.Spec.DatastoreType != apiconfig.Kubernetes {
+				rv = "1234"
+			} else {
+				// Resource version for KDD is a combination of both the CRD and K8s NP backed
+				// resources separated by a slash.
+				rv = "1234/1234"
```
caseydavenport: Should probably use `joinRevision`?

lib/clientv2/networkpolicy_e2e_test.go
```diff
@@ -77,20 +77,28 @@ var _ = testutils.E2eDatastoreDescribe("NetworkPolicy tests", testutils.Datastor
 			be.Clean()
 
 			By("Updating the NetworkPolicy before it is created")
+			var rv string
+			if config.Spec.DatastoreType != apiconfig.Kubernetes {
```
caseydavenport: A bit sad that we need the special casing here, though I suppose it's just for this specific test.

etcd complains if you pass a RV with a `/`?

robbrockbank: meh - if it was a different datastore we might need to pass in another format of the RV so it doesn't feel too icky to me.

lib/clientv2/networkpolicy_e2e_test.go
robbrockbank: we can do - i'll have to make all that public though

lib/backend/k8s/resources/networkpolicy.go
robbrockbank: Ah! I originally had the extraction code in both branches and moved it out - looks like i didn't move the comment - my bad.  Removing!

PR #661, opened 2017-11-08
[WIP] Add validation for V2 resources

gunjan5:
## Todos
- [x] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Add validation for V2 resources
```

### DIFF
--- a/lib/testutils/rules.go
+++ b/lib/testutils/rules.go
@@ -20,82 +20,81 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-var ipv4 = 4
-var ipv6 = 6
-var strProtocol1 = numorstring.ProtocolFromString("icmp")
-var strProtocol2 = numorstring.ProtocolFromString("udp")
-var numProtocol1 = numorstring.ProtocolFromInt(240)
-
-var portRange, singlePort, namedPort numorstring.Port
+var InRule1, InRule2, EgressRule1, EgressRule2 apiv2.Rule
 
 func init() {
-	var err error
-	portRange, err = numorstring.PortFromRange(10, 20)
+	portRange, err := numorstring.PortFromRange(10, 20)
 	if err != nil {
 		logrus.WithError(err).Panic("Failed to create port range")
 	}
-	singlePort = numorstring.SinglePort(1024)
-	namedPort = numorstring.NamedPort("named-port")
-}
+	singlePort := numorstring.SinglePort(1024)
+	namedPort := numorstring.NamedPort("named-port")
 
-var icmpType1 = 100
-var icmpCode1 = 200
+	ipv4 := 4
+	ipv6 := 6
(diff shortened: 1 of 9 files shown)


lib/validator/validator.go
```diff
@@ -36,22 +36,28 @@ import (
...
+	namespacedNameRegex   = regexp.MustCompile(`^[a-zA-Z0-9_./-]{1,128}$`)
+	interfaceRegex        = regexp.MustCompile("^[a-zA-Z0-9_-]{1,15}$")
+	actionRegex           = regexp.MustCompile("^(allow|deny|log|pass)$")
+	backendActionRegex    = regexp.MustCompile("^(allow|deny|log|next-tier|)$")
+	protocolRegex         = regexp.MustCompile("^(tcp|udp|icmp|icmpv6|sctp|udplite)$")
+	ipipModeRegex         = regexp.MustCompile("^(Always|CrossSubnet|Never)$")
+	bgpLogLevel           = regexp.MustCompile("^(Info|Warning|Debug|Error|Fatal|Trace)$")
+	felixLogLevel         = regexp.MustCompile("^(Info|Warning|Debug|Error|Critical)$")
+	datastoreType         = regexp.MustCompile("^(etcdv3|kubernetes)$")
+	dropacceptreturnRegex = regexp.MustCompile("^(Drop|Accept|Return)$")
```
bcreane: On https://docs.projectcalico.org/master/reference/calicoctl/resources/felixconfig, drop/accept/return is denoted case-insensitive. The regex isn't case insensitive, what's the expected behavior?

gunjan5: Yup, we need to change the doc so the yaml config follows the standard casing that's in this regex, once this PR is approved we'll need to update the doc page. I'll create a task for that, thanks @bcreane

bcreane: There are also some doc changes following from Rob's tweak PR: https://github.com/projectcalico/libcalico-go/pull/664. I can take the combined stories unless someone else is hankering for that.

gunjan5: Good idea @bcreane, you can take it. Thanks!

lib/testutils/rules.go
```diff
@@ -20,82 +20,81 @@ import (
...
-	Protocol:  &strProtocol1,
-	ICMP:      &icmp1,
-	Source: apiv2.EntityRule{
-		Nets:     []string{cidr1},
-		Selector: "label1 == 'value1'",
-		Ports: []numorstring.Port{
-			portRange,
-			singlePort,
-			namedPort,
+	InRule1 = apiv2.Rule{
```
gunjan5: the reason i had to pull these inside the init() was because when it's a var outside an init() golang initializes the vars with the zero values for some reason, so it may look unrelated but the tests wont work without these things being inside the init() function

bcreane:
closing in favor of https://github.com/projectcalico/libcalico-go/pull/667

PR #664, opened 2017-11-08
Tweak structure field definitions for consistency

robbrockbank:
## Description
Update the apiv2 field names for consistency across structs and with the json tag name.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note
- [ ] Rev CNI Plugin


lib/apis/v2/felixconfig.go
```diff
@@ -130,9 +130,9 @@ type FelixConfigurationSpec struct {
 	// [Default: INFO]
 	LogSeveritySys string `json:"logSeveritySys,omitempty"`
 
-	IpInIpEnabled *bool `json:"ipInIpEnabled,omitempty"`
+	IPIPEnabled *bool `json:"ipInIpEnabled,omitempty" confignamev1:"IpInIpEnabled"`
 	// IpInIpMTU is the MTU to set on the tunnel device. See Configuring MTU [Default: 1440]
-	IpInIpMtu *int `json:"ipInIpMtu,omitempty"`
+	IPIPMtu *int `json:"ipInIpMtu,omitempty" confignamev1:"IpInIpMtu"`
```
robbrockbank: `IPIPMTU`?  urgh

lib/apis/v2/felixconfig.go
```diff
@@ -187,8 +187,8 @@ type FelixConfigurationSpec struct {
 }
 
 type ProtoPort struct {
-	Protocol string
-	Port     uint16
+	Protocol string `json:"protocol"`
```
bcreane: there they are, my favorite `json` tags.

lib/apis/v1/policy.go
```diff
@@ -123,15 +123,15 @@ type PolicySpec struct {
 
 	// Types indicates whether this policy applies to ingress, or to egress, or to both.  When
 	// not explicitly specified (and so the value on creation is empty or nil), Calico defaults
-	// Types according to what IngressRules and EgressRules are present in the policy.  The
+	// Types according to what Ingress and Egress are present in the policy.  The
```
bcreane: this change lines up nicely with k8s - type: `NetworkPolicy[In|E]gressRule` -> `Rule` and now `[In|E]gress` == `[In|E]ngress`.

lib/apis/v2/zz_generated.deepcopy.go
```diff
@@ -382,7 +381,7 @@ func (in *ClusterInformation) DeepCopyInto(out *ClusterInformation) {
 	*out = *in
 	out.TypeMeta = in.TypeMeta
 	in.ObjectMeta.DeepCopyInto(&out.ObjectMeta)
-	out.Spec = in.Spec
+	in.Spec.DeepCopyInto(&out.Spec)
```
bcreane: Good catch to re-run `make deepcopy-gen`.

lib/apis/v2/felixconfig.go
bcreane: `IPIPMTU` is internally consistent - all letters in the two acronyms are capitalized. But fugly as all get-out.

lib/apis/v2/felixconfig.go
caseydavenport: Don't we want `omitempty` on these?

robbrockbank: I think we do on the port, yes - but possibly not on the Protocol?  It depends if we want a "missing" protocol to mean TCP or whether we should just be explicit about it.

@bcreane what was the decision there?

robbrockbank: hmmm, actually shouldn't it be a *int since we want to be able to not assign it, or a value of 0 means unassigned (since I don't think 0 is a value port number is it?)

robbrockbank: actually reading the Felix code the port must be specified and I think we also want the protocol specified too.  So neither should be omitempty.

robbrockbank:
One more commit for this PR to:
-  update a few comments
-  handle the value "none" when an empty slice of failsafe ports is specified
-  a few updates from the rebase

bcreane:
changes look good. approved again.

PR #665, opened 2017-11-09
Use WEP name and namespace to uniquely generate the vethname

gunjan5:
## Description

Part of https://github.com/projectcalico/cni-plugin/issues/410 

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/backend/k8s/conversion/conversion.go
+++ b/lib/backend/k8s/conversion/conversion.go
@@ -50,12 +50,12 @@ type Converter struct {
 }
 
 // VethNameForWorkload returns a deterministic veth name
-// for the given Kubernetes workload.
-func VethNameForWorkload(workload string) string {
+// for the given Kubernetes workload (WEP) name and namespace.
+func VethNameForWorkload(namespace, name string) string {
 	// A SHA1 is always 20 bytes long, and so is sufficient for generating the
 	// veth name and mac addr.
 	h := sha1.New()
-	h.Write([]byte(workload))
+	h.Write([]byte(fmt.Sprintf("%s/%s", namespace, name)))
 	return fmt.Sprintf("cali%s", hex.EncodeToString(h.Sum(nil))[:11])
 }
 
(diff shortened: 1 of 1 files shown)


lib/backend/k8s/conversion/conversion.go
```diff
@@ -173,7 +173,7 @@ func (c Converter) PodToWorkloadEndpoint(pod *kapiv1.Pod) (*model.KVPair, error)
 
 	// Generate the interface name based on workload.  This must match
 	// the host-side veth configured by the CNI plugin.
-	interfaceName := VethNameForWorkload(wepName)
+	interfaceName := VethNameForWorkload(wepName + pod.Namespace)
```
robbrockbank: For paranoia sake - better to join these together with, say,  a `/` separator

gunjan5: hmm ok, not sure why it's necessary but I feel the same way now that you pointed it out

robbrockbank: i don't think it'll cause issues with the current code because of the fact that the endpoint is always eth0

but suppose in the future we want `eth1` on namespace `0x` and `eth10` on namespace `x`.  Adds up to the same thing.

lib/backend/k8s/conversion/conversion.go
```diff
@@ -173,7 +173,7 @@ func (c Converter) PodToWorkloadEndpoint(pod *kapiv1.Pod) (*model.KVPair, error)
 
 	// Generate the interface name based on workload.  This must match
 	// the host-side veth configured by the CNI plugin.
-	interfaceName := VethNameForWorkload(wepName)
+	interfaceName := VethNameForWorkload(fmt.Sprintf("%s/%s", wepName, pod.Namespace))
```
robbrockbank: If this is called from elsewhere and the result needs to be the same, then VethNameForWorkload should be updated to take namespace as a parameter and it should do the concatenation.

gunjan5: This is only the case for k8s code path in CNI. For the non-k8s code path, we use `cali+first 11 digits of the ContainerID` for the vethname

gunjan5: o ok I misread that comment, I'll change it

lib/backend/k8s/conversion/conversion.go
```diff
@@ -50,12 +50,12 @@ type Converter struct {
 }
 
 // VethNameForWorkload returns a deterministic veth name
-// for the given Kubernetes workload.
-func VethNameForWorkload(workload string) string {
+// for the given Kubernetes workload (WEP) name and namespace.
+func VethNameForWorkload(namespace, workload string) string {
```
caseydavenport: nit: the argument `workload` should be called `name` now

PR #666, opened 2017-11-09
Handle no-op events in KDD watcher 

robbrockbank:
## Description
Recently added code to filter out host-networked pods by converting the pod event to a no-op.  Looks like this wasn't handled all the way up the stack.

### DIFF
--- a/lib/backend/k8s/resources/watcher.go
+++ b/lib/backend/k8s/resources/watcher.go
@@ -106,6 +106,10 @@ func (crw *k8sWatcherConverter) processK8sEvents() {
 			} else {
 				// We have a valid event, so convert it.
 				e = crw.convertEvent(event)
+				if e == nil {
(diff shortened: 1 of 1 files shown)


lib/backend/k8s/resources/watcher.go
```diff
@@ -106,6 +106,10 @@ func (crw *k8sWatcherConverter) processK8sEvents() {
 			} else {
 				// We have a valid event, so convert it.
 				e = crw.convertEvent(event)
+				if e == nil {
+					crw.logCxt.Debug("Event converted to a no-op")
+					continue
+				}
```
robbrockbank: This was causing a panic on line 116, presumably because of the `*e`

bcreane: makes sense.

PR #667, opened 2017-11-09
Add validation for v3 resources

bcreane:
## Description
Enable resource validation on v3 resources.
* Added/removed some resources (e.g. +GlobalNetworkPolicy, +Felix, -Node)
* Renamed protocols to be all caps, in alignment with K8s (e.g. `tcp` -> `TCP`)
* Many fields renamed to CamelCase (`ingress` -> `Ingress`). See the regexes in `validator.go` for details.

## Todos
- [x] Tests
- [ ] Documentation -- documentation needs to be updated to reflect the new syntax
- [x] Release note

## Release Note

```release-note
Add validation for V3 resources.
```


lib/apis/v2/profile.go
```diff
@@ -48,7 +48,7 @@ type ProfileSpec struct {
 	// An option set of labels to apply to each endpoint (in addition to their own labels)
 	// referencing this profile.  If labels configured on the endpoint have keys matching those
 	// labels inherited from the profile, the endpoint label values take precedence.
-	LabelsToApply map[string]string `json:"labelsToApply,omitempty" protobuf:"bytes,11,rep,name=labels"`
+	LabelsToApply map[string]string `json:"labelsToApply,omitempty" protobuf:"bytes,11,rep,name=labels" validate:"omitempty,labelsToApply"`
```
robbrockbank: labelsToApply -> labels I think.  I don't think the validator needs to be something different to the standard labels validator does it?

robbrockbank: where did that protobufs incantation come from.  Not really sure it should be here.

lib/apis/v2/bgpconfig.go
```diff
@@ -37,11 +37,10 @@ type BGPConfiguration struct {
 	Spec BGPConfigurationSpec `json:"spec,omitempty"`
 }
 
-// TODO: Add validation on LogSeverityScreen for valid values
 // BGPConfigurationSpec contains the values of the BGP configuration.
 type BGPConfigurationSpec struct {
 	// LogSeverityScreen is the log severity above which logs are sent to the stdout. [Default: INFO]
-	LogSeverityScreen string `json:"logSeverityScreen,omitempty" validate:"omitempty" confignamev1:"loglevel"`
+	LogSeverityScreen string `json:"logSeverityScreen,omitempty" validate:"omitempty,bgpLogLevel" confignamev1:"loglevel"`
```
robbrockbank: I think we can get away with a single logLevel validator no?

lib/validator/validator.go
```diff
@@ -36,22 +36,28 @@ import (
...
-	protocolRegex       = regexp.MustCompile("^(tcp|udp|icmp|icmpv6|sctp|udplite)$")
-	ipipModeRegex       = regexp.MustCompile("^(always|cross-subnet|)$")
-	reasonString        = "Reason: "
-	poolSmallIPv4       = "IP pool size is too small (min /26) for use with Calico IPAM"
-	poolSmallIPv6       = "IP pool size is too small (min /122) for use with Calico IPAM"
-	poolUnstictCIDR     = "IP pool CIDR is not strictly masked"
-	overlapsV4LinkLocal = "IP pool range overlaps with IPv4 Link Local range 169.254.0.0/16"
-	overlapsV6LinkLocal = "IP pool range overlaps with IPv6 Link Local range fe80::/10"
-	protocolPortsMsg    = "rules that specify ports must set protocol to TCP or UDP"
+	labelRegex            = regexp.MustCompile(`^` + tokenizer.LabelKeyMatcher + `$`)
```
robbrockbank: The label regexes should exactly match those in kubernetes now, rather than the one in the tokenizer code.

lib/validator/validator.go
```diff
@@ -36,22 +36,28 @@ import (
...
+	labelValueRegex       = regexp.MustCompile("^[a-zA-Z0-9]?([a-zA-Z0-9_.-]{0,61}[a-zA-Z0-9])?$")
+	nameRegex             = regexp.MustCompile("^[a-zA-Z0-9_.-]{1,128}$")
```
robbrockbank: name regex should match the k8s one.

and in addition, Calico Network policies names cannot have a "." in them unless they start with "knp.default."

lib/validator/validator.go
```diff
@@ -36,22 +36,28 @@ import (
...
+	labelsToApplyValRegex = regexp.MustCompile("^[a-zA-Z0-9_.-]{0,128}$")
+	namespacedNameRegex   = regexp.MustCompile(`^[a-zA-Z0-9_./-]{1,128}$`)
+	interfaceRegex        = regexp.MustCompile("^[a-zA-Z0-9_-]{1,15}$")
+	actionRegex           = regexp.MustCompile("^(allow|deny|log|pass)$")
```
robbrockbank: Allow / Deny / Log / Pass

lib/validator/validator.go
```diff
@@ -36,22 +36,28 @@ import (
...
+	backendActionRegex    = regexp.MustCompile("^(allow|deny|log|next-tier|)$")
```
robbrockbank: I don't think the backendActionRegex is a thing anymore.

lib/validator/validator.go
```diff
@@ -36,22 +36,28 @@ import (
...
+	protocolRegex         = regexp.MustCompile("^(tcp|udp|icmp|icmpv6|sctp|udplite)$")
```
robbrockbank: Hmmm, should these be uppercase?  What does k8s do for this?

lib/validator/validator.go
```diff
@@ -92,16 +100,26 @@ func init() {
...
+	registerFieldValidator("labelsToApply", validateLabelsToApply)
 	registerFieldValidator("scopeglobalornode", validateScopeGlobalOrNode)
 	registerFieldValidator("ipversion", validateIPVersion)
 	registerFieldValidator("ipipmode", validateIPIPMode)
 	registerFieldValidator("policytype", validatePolicyType)
+	registerFieldValidator("bgpLogLevel", validateBGPLogLevel)
+	registerFieldValidator("felixLogLevel", validateFelixLogLevel)
+	registerFieldValidator("dropacceptreturn", validateFelixEtoHAction)
+	registerFieldValidator("acceptreturn", validateAcceptReturn)
+	registerFieldValidator("ipv4", validateIPv4Address)
```
robbrockbank: I thought the validator had built-ins for ip, ipv4 and ipv6?

robbrockbank: same should be true for cidrs I think (although if we allow IPs interchangeably with CIDRs then I daresay we'd need to handle that).

lib/validator/validator.go
```diff
@@ -256,6 +312,35 @@ func validateProtocol(v *validator.Validate, structLevel *validator.StructLevel)
 	}
 }
 
+func validateIPv4Address(v *validator.Validate, topStruct reflect.Value, currentStructOrField reflect.Value, field reflect.Value, fieldType reflect.Type, fieldKind reflect.Kind, param string) bool {
```
robbrockbank: Ah I see... maybe these should be validate IPOrCIDRv4 or something similar.

lib/apis/v2/profile.go
robbrockbank: I suspect I added it by copying the labels field from the v1 metadata....

bcreane: yeah, I bet protobuf is wrong, will delete. when you say labelsToApply, you mean just the validator?

bcreane: btw, the regex is different:
`labelsToApplyValRegex = regexp.MustCompile("^[a-zA-Z0-9_.-]{0,128}$")`
`labelValueRegex       = regexp.MustCompile("^[a-zA-Z0-9]?([a-zA-Z0-9_.-]{0,61}[a-zA-Z0-9])?$")`

seems like gunjan went to some trouble to distinguish between the two types of labels.

lib/validator/validator.go
```diff
@@ -230,10 +289,7 @@ func validateScopeGlobalOrNode(v *validator.Validate, topStruct reflect.Value, c
 func validatePolicyType(v *validator.Validate, topStruct reflect.Value, currentStructOrField reflect.Value, field reflect.Value, fieldType reflect.Type, fieldKind reflect.Kind, param string) bool {
 	s := field.String()
 	log.Debugf("Validate policy type: %s", s)
-	if s == string(api.PolicyTypeIngress) {
-		return true
-	}
-	if s == string(api.PolicyTypeEgress) {
+	if s == string(api.PolicyTypeIngress) || s == string(api.PolicyTypeEgress) {
```
mgleung: Should this be the uppercased values of `Ingress` and `Egress` (as it is done here and according to k8s docs https://kubernetes.io/docs/concepts/services-networking/network-policies/) or lowercased (as per the docs https://docs.projectcalico.org/v3.0/reference/calicoctl/resources/networkpolicy)?

@bcreane @robbrockbank @caseydavenport  WDYT?

lib/apis/v2/bgpconfig.go
bcreane: looks like bgp accepts a wider range of logging options:
`bgpLogLevel           = regexp.MustCompile("^(Info|Warning|Debug|Error|Fatal|Trace)$")`

versus:
`felixLogLevel         = regexp.MustCompile("^(Info|Warning|Debug|Error|Critical)$")`

unless Critical/Trace are incorrect, I'd say we need both regex's

lib/validator/validator.go
bcreane: I found what looks like the k8s regex for labels in apimachinery: https://goo.gl/Sv4h6t

We shouldn't try to replicate what they're doing at this point since the approach is significantly different. How about removing reference to the `tokenizer` and having a standalone regex?

lib/validator/validator.go
bcreane: do you mean the `metadata` name field? we're not currently validating metadata.

lib/validator/validator.go
bcreane: good catch, will fix

lib/validator/validator.go
bcreane: agreed, no one is using it. will delete.

bcreane: oops, my mistake. In fact we do need distinct action and backendAction regexes. Turns out Felix is using `next-tier` and those regexes are not interchangeable, and both are required.

lib/validator/validator.go
bcreane: Looks like all-caps: https://kubernetes.io/docs/concepts/services-networking/network-policies/

Shall we do the same?

lib/validator/validator.go
bcreane: We use IPv4 and IPv4+CIDR interchangeably which `validator.IPv4` sadly does not support. Let's stick with our home-grown validators since they're efficient, minimal and reflect what the backend is using as well.

robbrockbank: Yup that's fine - but perhaps we could rename it - cidrOrIP or something similar

lib/validator/validator.go
bcreane: yeah good idea, will rename our validator functions.

lib/validator/validator.go
bcreane: yeah, we'll go with `Ingress` and `Egress` just like k8s.

lib/validator/validator.go
bcreane: how is: `validateIPv4orCIDRAddress` for example?
Should really be `validateIPv4OrCIDRAddress` but that kind of looks like 40.

lib/validator/validator.go
bcreane: done

lib/apis/v2/bgpconfig.go
robbrockbank: I think both Critical and Trace are incorrect.

They both use the libcalico-go logutils to set log level and that basically filters down to logrus.

Neither Critical nor Trace are options in the logrus logging levels.

I think Felix will basically handle CRITICAL as "Panic" but only on the grounds that it would not recognize Critical and then default to Panic (along with a warning log).

I suspect Felix needs updating to change CRITICAL to PANIC.

And the BGP one is just wrong

lib/apis/v2/policy.go
```diff
@@ -42,12 +42,12 @@ type Rule struct {
 	// a specific IP protocol. Required if any of the EntityRules contain Ports
 	// (because ports only apply to certain protocols).
 	//
-	// Must be one of these string values: "tcp", "udp", "icmp", "icmpv6", "sctp", "udplite"
+	// Must be one of these string values: "TCP", "UDP", "ICMP", "ICMPV6", "SCtp", "UDPLITE"
```
robbrockbank: SCTP
AND PROBABLY UDPLite

robbrockbank: not sure about the ICMP one tho

lib/backend/k8s/resources/networkpolicy.go
```diff
@@ -398,16 +398,40 @@ func (npw *networkPolicyWatcher) processNPEvents() {
 		atomic.AddUint32(&npw.terminated, 1)
 	}()
 
-	var e api.WatchEvent
-	var isCRDEvent bool
-	var value interface{}
 	for {
+		var ok bool
```
robbrockbank: what's this doing here?  rebase include this commit accidentally?

bcreane: yeah, the rebase is a little wonky. sorry about that.

lib/validator/validator.go
robbrockbank: yup  - that's what i was trying to suggest :-)  Key thing is that we only allow the same names/labels etc.

lib/validator/validator.go
robbrockbank: we should be validating the metadata.  I think there is a subtask for that no?  Or did that go to me?  I forget now.

robbrockbank: The only things we *have* to validate in the Metadata are the name, labels and annotations.

All of the other stuff is only configurable through the  AAPIS, so we don't really need to validate it here.

lib/validator/validator.go
```diff
@@ -36,22 +36,28 @@ import (
...
+	labelValueRegex       = regexp.MustCompile("^[a-zA-Z0-9]?([a-zA-Z0-9_.-]{0,61}[a-zA-Z0-9])?$")
+	nameRegex             = regexp.MustCompile("^[a-zA-Z0-9_.-]{1,128}$")
+	labelsToApplyValRegex = regexp.MustCompile("^[a-zA-Z0-9_.-]{0,128}$")
+	namespacedNameRegex   = regexp.MustCompile(`^[a-zA-Z0-9_./-]{1,128}$`)
+	interfaceRegex        = regexp.MustCompile("^[a-zA-Z0-9_-]{1,15}$")
+	actionRegex           = regexp.MustCompile("^(Allow|Deny|Log|Pass)$")
+	backendActionRegex    = regexp.MustCompile("^(Allow|Deny|Log|next-tier|)$")
+	protocolRegex         = regexp.MustCompile("^(TCP|UDP|ICMP|ICMPV6|SCTP|UDPLITE)$")
+	ipipModeRegex         = regexp.MustCompile("^(Always|CrossSubnet|Never)$")
+	bgpLogLevel           = regexp.MustCompile("^(Info|Warning|Debug|Error|Fatal|Trace)$")
```
robbrockbank: I think these should both be `Info|Warning|Debug|Error|Fatal`

lib/validator/validator.go
bcreane: `labelValueRegex` evaluates identically to apimachinery's `IsQualifiedName()`. I've removed the dependency on `tokenizer`.

lib/validator/validator.go
bcreane: we discussed addressing metadata in a subsequent PR since it's not clear how to use our current approach with the k8s `Metadata` types. The better longer term approach is to adopt `apimachinery` validation where appropriate.

lib/validator/validator.go
```diff
@@ -36,22 +36,28 @@ import (
...
+	labelValueRegex       = regexp.MustCompile("^[a-zA-Z0-9]?([a-zA-Z0-9_.-]{0,61}[a-zA-Z0-9])?$")
+	nameRegex             = regexp.MustCompile("^[a-zA-Z0-9_.-]{1,128}$")
+	labelsToApplyValRegex = regexp.MustCompile("^[a-zA-Z0-9_.-]{0,128}$")
+	namespacedNameRegex   = regexp.MustCompile(`^[a-zA-Z0-9_./-]{1,128}$`)
+	interfaceRegex        = regexp.MustCompile("^[a-zA-Z0-9_-]{1,15}$")
+	actionRegex           = regexp.MustCompile("^(Allow|Deny|Log|Pass)$")
+	backendActionRegex    = regexp.MustCompile("^(Allow|Deny|Log|next-tier|)$")
+	protocolRegex         = regexp.MustCompile("^(TCP|UDP|ICMP|ICMPv6|SCTP|UDPLite)$")
+	ipipModeRegex         = regexp.MustCompile("^(Always|CrossSubnet|Never)$")
+	bgpLogLevel           = regexp.MustCompile("^(Info|Warning|Debug|Error|Fatal|Trace)$")
```
bcreane: `Trace` is a valid `bird` logging level: http://bird.network.cz/?get_doc&f=bird-3.html. Consequently no changes required.

lib/converter/rule.go
```diff
@@ -177,7 +177,7 @@ func ruleBackendToAPI(br model.Rule) api.Rule {
 // ruleActionAPIToBackend converts the rule action field value from the API
 // value to the equivalent backend value.
 func ruleActionAPIToBackend(action string) string {
-	if action == "pass" {
+	if action == "Pass" {
```
fasaxc: If we're preserving the v1 semantics, we should push things to lower-case in these functions.  If we're not, then you'll need to analyse Felix to make sure it doesn't rely on case.

lib/numorstring/numorstring_test.go
```diff
@@ -84,8 +84,8 @@ func init() {
 		// Protocol tests.  Invalid integer values will be stored as strings.
 		Entry("should accept 0 protocol as int", "0", protocolType, numorstring.ProtocolFromInt(0)),
 		Entry("should accept 255 protocol as int", "255", protocolType, numorstring.ProtocolFromInt(255)),
-		Entry("should accept tcp protocol as string", "\"tcp\"", protocolType, numorstring.ProtocolFromString("tcp")),
-		Entry("should accept tcp protocol as string", "\"tcp\"", protocolType, numorstring.ProtocolFromString("tcp")),
+		Entry("should accept tcp protocol as string", "\"TCP\"", protocolType, numorstring.ProtocolFromString("TCP")),
```
fasaxc: Is lower case still allowed?  Should have UTs either way.

My preference would be to be permissive but we should probably sync with whatever k8s does.

lib/testutils/rules.go
```diff
@@ -20,82 +20,81 @@ import (
...
-	namedPort = numorstring.NamedPort("named-port")
-}
+	singlePort := numorstring.SinglePort(1024)
+	namedPort := numorstring.NamedPort("named-port")
 
-var icmpType1 = 100
-var icmpCode1 = 200
+	ipv4 := 4
+	ipv6 := 6
+	strProtocol1 := numorstring.ProtocolFromString("TCP")
```
fasaxc: We do this a lot in the Felix UTs/FVs too so either need to be permissive or fix those up too.

lib/validator/validator.go
```diff
@@ -36,22 +35,28 @@ import (
...
+	felixLogLevel         = regexp.MustCompile("^(Info|Warning|Debug|Error|Critical)$")
```
fasaxc: IIRC logrus levels are debug, info, warning, error, fatal, panic. Looks like felix also validates to `Info|Warning|Debug|Error|Critical` which is wrong.  That said, maybe we shouldn't just follow the particular log library that we're using; maybe we should map `critical` to `fatal` (which includes `panic`) instead?

bcreane: discussed with Rob and we'll fixup the regex to replace `Critical` with `Fatal`.

bcreane: remapping is done and added tests to validate

lib/validator/v3/validator.go
```diff
@@ -0,0 +1,729 @@
...
+	labelValueRegex       = regexp.MustCompile("^[a-zA-Z0-9]?([a-zA-Z0-9_.-]{0,61}[a-zA-Z0-9])?$")
+	nameRegex             = regexp.MustCompile("^[a-zA-Z0-9_.-]{1,128}$")
+	labelsToApplyValRegex = regexp.MustCompile("^[a-zA-Z0-9_.-]{0,128}$")
+	namespacedNameRegex   = regexp.MustCompile(`^[a-zA-Z0-9_./-]{1,128}$`)
+	interfaceRegex        = regexp.MustCompile("^[a-zA-Z0-9_-]{1,15}$")
+	actionRegex           = regexp.MustCompile("^(Allow|Deny|Log|Pass)$")
+	backendActionRegex    = regexp.MustCompile("^(Allow|Deny|Log|next-tier|)$")
+	protocolRegex         = regexp.MustCompile("^(TCP|UDP|ICMP|ICMPv6|SCTP|UDPLite)$")
+	ipipModeRegex         = regexp.MustCompile("^(Always|CrossSubnet|Never)$")
+	bgpLogLevel           = regexp.MustCompile("^(Info|Warning|Debug|Error|Fatal|Trace)$")
```
robbrockbank: Still not sure where Trace comes from for bgpLogLevel.

nit:  Can we either order alphabetically or in order of severity :-)  (it was probably already like this - in which case I'm prolly the one to blame)

lib/validator/v3/validator.go
```diff
@@ -0,0 +1,729 @@
...
+
+var (
+	labelRegex            = regexp.MustCompile("^[a-zA-Z0-9_./-]{1,512}$")
+	labelValueRegex       = regexp.MustCompile("^[a-zA-Z0-9]?([a-zA-Z0-9_.-]{0,61}[a-zA-Z0-9])?$")
+	nameRegex             = regexp.MustCompile("^[a-zA-Z0-9_.-]{1,128}$")
+	labelsToApplyValRegex = regexp.MustCompile("^[a-zA-Z0-9_.-]{0,128}$")
+	namespacedNameRegex   = regexp.MustCompile(`^[a-zA-Z0-9_./-]{1,128}$`)
+	interfaceRegex        = regexp.MustCompile("^[a-zA-Z0-9_-]{1,15}$")
+	actionRegex           = regexp.MustCompile("^(Allow|Deny|Log|Pass)$")
+	backendActionRegex    = regexp.MustCompile("^(Allow|Deny|Log|next-tier|)$")
```
robbrockbank: backendActionRegex not required in the v3 validator.

lib/converter/rule.go
```diff
@@ -187,9 +187,9 @@ func ruleActionAPIToBackend(action string) string {
 // value to the equivalent API value.
 func ruleActionBackendToAPI(action string) string {
 	if action == "" {
-		return "allow"
+		return "Allow"
```
robbrockbank: Not sure any changes are required in this file.  The converter code was used for the v1 API -> v1 backend conversion, but I don't think anything uses that anymore.  In any case, the action in the v1 API is lowercase, so this should need updating for these case changes.

lib/apis/v3/felixconfig.go
```diff
@@ -113,22 +110,22 @@ type FelixConfigurationSpec struct {
...
+	IptablesMangleAllowAction   string `json:"iptablesMangleAllowAction,omitempty" validate:"omitempty,acceptreturn"`
 	// LogPrefix is the log prefix that Felix uses when rendering LOG rules. [Default: calico-packet]
 	LogPrefix string `json:"logPrefix,omitempty"`
 
 	// LogFilePath is the full path to the Felix log. Set to none to disable file logging. [Default: /var/log/calico/felix.log]
 	LogFilePath string `json:"logFilePath,omitempty"`
 
 	// LogSeverityFile is the log severity above which logs are sent to the log file. [Default: INFO]
-	LogSeverityFile string `json:"logSeverityFile,omitempty"`
+	LogSeverityFile string `json:"logSeverityFile,omitempty" validate:"omitempty,felixLogLevel"`
```
robbrockbank: nit: Our validation tags are a mix of all lower and camelcase.  Would be nice to choose one or t'other.

lib/validator/v3/validator.go
bcreane: Will reorder from most severe to least: `(Fatal|Error|Warning|Info|Debug)`
Also sounds like `Trace` isn't supported/necessary, so I'll combine Felix and Bgp logging regexes. I'll leave the validation tags and functions separate for now, but we can combine them if necessary later.

lib/validator/v3/validator.go
bcreane: backendActionRegex and associated validation function `validateBackendAction` are used in a bunch of places. What's your take on the change that should happen if we get rid of this? Note that just substituting `actionRegex` causes a bunch of UT failures - note the empty string after `next-tier`.

lib/converter/rule.go
bcreane: Yes, good catch: reverting this change and also a couple of changes I made to a v1 api test.

lib/apis/v3/felixconfig.go
bcreane: will fix

lib/apis/v3/felixconfig.go
```diff
@@ -113,22 +110,22 @@ type FelixConfigurationSpec struct {
...
 	// LogSeverityScreen is the log severity above which logs are sent to the stdout. [Default: INFO]
-	LogSeverityScreen string `json:"logSeverityScreen,omitempty"`
+	LogSeverityScreen string `json:"logSeverityScreen,omitempty" validate:"omitempty,felixLogLevel"`
 	// LogSeveritySys is the log severity above which logs are sent to the syslog. Set to NONE for no logging to syslog.
 	// [Default: INFO]
-	LogSeveritySys string `json:"logSeveritySys,omitempty"`
+	LogSeveritySys string `json:"logSeveritySys,omitempty" validate:"omitempty,felixLogLevel"`
 
 	IPIPEnabled *bool `json:"ipipEnabled,omitempty" confignamev1:"IpInIpEnabled"`
 	// IPIPMTU is the MTU to set on the tunnel device. See Configuring MTU [Default: 1440]
```
robbrockbank: `ProtoPort` needs some validation tags
`     |`
`     |`
`     V`

lib/backend/k8s/conversion/conversion.go
```diff
@@ -192,9 +192,9 @@ func (c Converter) PodToWorkloadEndpoint(pod *kapiv1.Pod) (*model.KVPair, error)
 				var modelProto numorstring.Protocol
 				switch containerPort.Protocol {
 				case kapiv1.ProtocolUDP:
-					modelProto = numorstring.ProtocolFromString("udp")
+					modelProto = numorstring.ProtocolFromString("UDP")
```
robbrockbank: Hmm, thought I'd added a comment but perhaps I didn't

I don't believe this package needs updating - as it was used for converting between the v1 API and the v1 backend data

lib/backend/k8s/conversion/conversion_test.go
```diff
@@ -149,8 +149,8 @@ var _ = Describe("Test Pod conversion", func() {
 		}
 		Expect(wep.Value.(*apiv3.WorkloadEndpoint).ObjectMeta.Labels).To(Equal(expectedLabels))
 
-		nsProtoTCP := numorstring.ProtocolFromString("tcp")
-		nsProtoUDP := numorstring.ProtocolFromString("udp")
+		nsProtoTCP := numorstring.ProtocolFromString("TCP")
```
robbrockbank: Again, I don't believe any of the conversion package needs to change

lib/backend/syncersv1/felixsyncer/felixsyncer_e2e_test.go
```diff
@@ -70,36 +70,36 @@ var _ = testutils.E2eDatastoreDescribe("Felix syncer tests", testutils.Datastore
 				syncTester.ExpectData(model.KVPair{
 					Key: model.ProfileRulesKey{ProfileKey: model.ProfileKey{Name: "kns.default"}},
 					Value: &model.ProfileRules{
-						InboundRules:  []model.Rule{{Action: "allow"}},
-						OutboundRules: []model.Rule{{Action: "allow"}},
+						InboundRules:  []model.Rule{{Action: "Allow"}},
```
robbrockbank: Everything in model (apart from the resources key and list) are basically the v1 model.  Therefore the values in these structs should be the v1 values (i.e. lower case)

lib/backend/syncersv1/updateprocessors/networkpolicyprocessor.go
```diff
@@ -81,5 +80,5 @@ func policyTypesAPIV2ToBackend(ptypes []apiv3.PolicyType) []string {
 }
 
 func policyTypeAPIV2ToBackend(ptype apiv3.PolicyType) string {
-	return strings.ToLower(string(ptype))
+	return string(ptype)
```
robbrockbank: The backend values need to remain unchanged.

lib/backend/syncersv1/updateprocessors/rules.go
```diff
@@ -153,6 +154,14 @@ func parseNamespaceSelector(s string) string {
 	return updated
 }
 
+// normalizeProto converts a v1 protocol string to a v3 protocol string
```
robbrockbank: Doesn't this convert a camelcase v3 protocol to a v1 lowercase protocol?

Maybe normalize is not the correct word - after all, what is normal in this case?

lib/backend/syncersv1/updateprocessors/rules_test.go
```diff
@@ -36,30 +36,30 @@ var _ = Describe("Test the Rules Conversion Functions", func() {
 				Code: &incode,
 			},
 			Source: apiv3.EntityRule{
-				Nets:        []string{"10.100.10.1"},
+				Nets:        []string{"10.100.10.1/32"},
```
robbrockbank: Are we no longer allowing IP addresses in here?

lib/client/client.go
```diff
@@ -31,7 +31,7 @@ import (
 	"github.com/projectcalico/libcalico-go/lib/backend/model"
 	"github.com/projectcalico/libcalico-go/lib/ipam"
 	"github.com/projectcalico/libcalico-go/lib/net"
-	"github.com/projectcalico/libcalico-go/lib/validator"
+	"github.com/projectcalico/libcalico-go/lib/validator/v3"
```
robbrockbank: this is the v1 client, so I think for accuracy this should be v1 not v3.

robbrockbank: also better to have aliased this as validator since vx doesn't give much away.

lib/clientv3/node.go
```diff
@@ -95,7 +97,13 @@ func (r nodes) Delete(ctx context.Context, name string, opts options.DeleteOptio
 			continue
 		}
 		for _, ip := range wep.Spec.IPNetworks {
-			ips = append(ips, *net.ParseIP(ip))
+			ipAddr, _, err := cnet.ParseCIDROrIP(ip)
+			if err == nil {
+				ips = append(ips, *ipAddr)
+			} else {
+				// Validation for wep insists upon CIDR, so we should always succeed
+				log.WithError(err).Warnf("Failed to parse IP Address: %s", ip)
```
robbrockbank: IP Address -> CIDR or Network

lib/validator/v1/validator.go
```diff
@@ -411,6 +411,7 @@ func validateICMPFields(v *validator.Validate, structLevel *validator.StructLeve
 func validateRule(v *validator.Validate, structLevel *validator.StructLevel) {
 	rule := structLevel.CurrentStruct.Interface().(api.Rule)
 
+    log.Infof("BCREANE: Inside validateRule: rule=%+v", rule)
```
robbrockbank: I assume these logs will disappear?

lib/apis/v3/felixconfig.go
bcreane: added validation + tests for ProtoPort to ensure Protocol is either `TCP` or `UDP`.

lib/validator/v3/validator.go
```diff
@@ -0,0 +1,729 @@
...
+		seenV6 = seenV6 || v6
+	}
+
+	scanNets(rule.Source.Nets, "Source.Nets")
+	scanNets(rule.Source.NotNets, "Source.NotNets")
+	scanNets(rule.Destination.Nets, "Destination.Nets")
+	scanNets(rule.Destination.NotNets, "Destination.NotNets")
+}
+
+func validateBackendRule(v *validator.Validate, structLevel *validator.StructLevel) {
```
robbrockbank: Shouldn't need any backend functions

robbrockbank:
Missing validation tests
BGPPeerSpec (test IP field, AS number)
Now that IPs and CIDRs are string fields, we should be validating that invalid IP and CIDR strings are rejected.


Not sure ClusterInfo needs validate tags, unless we intend to validate them in some way

There is some static validation code still in the IPPool which should be covered by the Validator now.  The dynamic checks (such as overlapping pools still needs to be handled in the IPPool processing)

For string fields that can be one of a set of values, should we be defining a new Type and a set of allowed values?  We do this for the IPIPMode field, but we don’t for other fields (e.g. Action, DropLogReturn thingummy)


BGPConfig has a validateDefaultOnlyFields which should probably be moved into the validator, validating the entire BGPConfig struct (otherwise calicoctl won’t be able to make the same static checks)

lib/validator/v1/validator.go
bcreane: gone. glad you noticed them!! not the sort of infamy I'm looking for.

lib/backend/syncersv1/felixsyncer/felixsyncer_e2e_test.go
bcreane: should be fixed in recent commit.

lib/backend/syncersv1/updateprocessors/rules.go
bcreane: A profound question, worthy of a lengthy dissertation. 
Renamed to `convertV3ProtocolToV1`

lib/clientv3/node.go
bcreane: fixed

lib/backend/syncersv1/updateprocessors/rules_test.go
bcreane: correct.

lib/validator/v3/validator.go
bcreane: all backend validation removed from v3 validator. Note that since v1 validator shares some of the backend rule code with v3, I needed to disable a couple of v1 tests as well.

lib/apis/v1/ippool.go
```diff
@@ -75,7 +75,7 @@ type IPIPConfiguration struct {
 	// addresses within this pool.  A mode of "cross-subnet" will only use IPIP
 	// tunneling when the destination node is on a different subnet to the
 	// originating node.  The default value (if not specified) is "always".
-	Mode ipip.Mode `json:"mode,omitempty" validate:"ipipmode"`
+	Mode ipip.Mode `json:"mode,omitempty" validate:"ipIpMode"`
```
robbrockbank: nit: ipIp -> ipip

lib/apis/v3/felixconfig.go
robbrockbank: ok - i guess I was expecting that you'd just add a regex validator to the Protocol field.  No worries though - as long as it's validated.

lib/apis/v3/hostendpoint.go
```diff
@@ -58,7 +58,7 @@ type HostEndpointSpec struct {
 	// A list of identifiers of security Profile objects that apply to this endpoint. Each
 	// profile is applied in the order that they appear in this list.  Profile rules are applied
 	// after the selector-based security policy.
-	Profiles []string `json:"profiles,omitempty" validate:"omitempty,dive,namespacedname"`
+	Profiles []string `json:"profiles,omitempty" validate:"omitempty,dive,namespacedName"`
```
robbrockbank: No - profile names are no longer namespaced - so standard Name regex now.

lib/apis/v3/workloadendpoint.go
```diff
@@ -37,9 +37,9 @@ type WorkloadEndpoint struct {
 // WorkloadEndpointMetadata contains the specification for a WorkloadEndpoint resource.
 type WorkloadEndpointSpec struct {
 	// The name of the orchestrator.
-	Orchestrator string `json:"orchestrator,omitempty" validate:"omitempty,namespacedname"`
+	Orchestrator string `json:"orchestrator,omitempty" validate:"omitempty,namespacedName"`
```
robbrockbank: Just name now I think.  i don't think we need this namespaced (although theoretically it shouldn't matter)

lib/apis/v3/workloadendpoint.go
```diff
@@ -37,9 +37,9 @@ type WorkloadEndpoint struct {
...
 	// The name of the workload.
-	Workload string `json:"workload,omitempty" validate:"omitempty,namespacedname"`
+	Workload string `json:"workload,omitempty" validate:"omitempty,namespacedName"`
```
robbrockbank: Again, I think this can just be a name regex.

lib/apis/v3/workloadendpoint.go
```diff
@@ -65,7 +65,7 @@ type WorkloadEndpointSpec struct {
 	// A list of security Profile resources that apply to this endpoint. Each profile is
 	// applied in the order that they appear in this list.  Profile rules are applied
 	// after the selector-based security policy.
-	Profiles []string `json:"profiles,omitempty" validate:"omitempty,dive,namespacedname"`
+	Profiles []string `json:"profiles,omitempty" validate:"omitempty,dive,namespacedName"`
```
robbrockbank: ditto.

lib/backend/model/rule.go
```diff
@@ -107,7 +107,7 @@ func (r Rule) String() string {
 	if r.Action != "" {
 		parts = append(parts, r.Action)
 	} else {
-		parts = append(parts, "allow")
+		parts = append(parts, "Allow")
```
robbrockbank: probably "allow" since this is the backend model.

lib/backend/syncersv1/felixsyncer/felixsyncer_e2e_test.go
robbrockbank: still seeing uppercase here.

lib/converter/rule.go
robbrockbank: still not expecting any changes in this file - since it's only used by the v1 client (I hope!)

caseydavenport:
This was merged in #680

PR #668, opened 2017-11-09
Fix etcdv3 watcher so that we get notified when the watch channel closes

caseydavenport:
## Description

So we get some indication that we're not connected to the datastore anymore. 

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -30,9 +30,9 @@ import:
   subpackages:
   - patricia
 - package: k8s.io/apimachinery
-  version: release-1.8 
+  version: release-1.8
 - package: k8s.io/api
-  version: release-1.8 
+  version: release-1.8
 - package: k8s.io/client-go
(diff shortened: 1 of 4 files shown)


caseydavenport:
Yeah, this doesn't seem to do it. I've tried both removing conntrack entries, dropping traffic in the raw table, and also just killing etcd flat out.

In each case, Felix continues to spin without any indication that the watcher has lost its connection.

caseydavenport:
The kubernetes API server also seems to have this problem. I checked the code and verified the behavior as well.

Looks like both Calico and the apiserver recover from a severed connectiono when it comes back, but we don't spot that the connection went away.

I think we'll have to wait until etcd v3.3, with support for keepalives, to get this working.

caseydavenport:
Closing in favor of https://github.com/projectcalico/libcalico-go/pull/777

PR #673, opened 2017-11-10
[WIP] Rename v2 API to v3

robbrockbank:
## Description

Renaming all of our v2 APIs to be v3 instead.

I've done this as two separate commits.
-   The first is a blanket replace of v2 -> v3
-   The second is undoing changes that were not required

(this way you can see what i've explicitly excluded).

## Todos
- [x] Tests
- [ ] Documentation
- [ ] Release note
- [ ] Update every other repo under the sun

## Release Note

```release-note
None required
```


lib/backend/k8s/conversion/conversion_test.go
```diff
@@ -1095,7 +1095,7 @@ var _ = Describe("Test NetworkPolicy conversion (k8s <= 1.7, no policyTypes)", f
 								PodSelector: &metav1.LabelSelector{
 									MatchLabels: map[string]string{
 										"k":  "v",
-										"k2": "v2",
+										"k2": "v3",
```
robbrockbank: didn't really need to change this - but more hassle that it's worth to undo just this.

lib/backend/compat/compat.go
```diff
@@ -732,7 +732,7 @@ func toDatastoreGlobalBGPConfigKey(key model.GlobalBGPConfigKey) model.GlobalBGP
 }
 
 // toDatastoreGlobalBGPConfigList modifies the Global BGP Config List interface to the one required by
-// the datastore (for back-compatibility with what is expected in teh etcdv2 datastore driver).
+// the datastore (for back-compatibility with what is expected in teh etcdv3 datastore driver).
```
bcreane: spotted an old typo: `teh`

lib/backend/k8s/conversion/conversion_test.go
```diff
@@ -272,7 +272,7 @@ var _ = Describe("Test NetworkPolicy conversion", func() {
...
```
bcreane: this looks questionable

lib/backend/k8s/conversion/conversion_test.go
```diff
@@ -435,7 +435,7 @@ var _ = Describe("Test NetworkPolicy conversion", func() {
 							extensions.NetworkPolicyPeer{
 								PodSelector: &metav1.LabelSelector{
 									MatchLabels: map[string]string{
-										"k2": "v2",
+										"k2": "v3",
```
bcreane: not sure this is right

lib/backend/k8s/conversion/conversion_test.go
```diff
@@ -515,7 +515,7 @@ var _ = Describe("Test NetworkPolicy conversion", func() {
...
```
bcreane: same question here ..

lib/backend/k8s/conversion/conversion_test.go
```diff
@@ -708,7 +708,7 @@ var _ = Describe("Test NetworkPolicy conversion", func() {
 						metav1.LabelSelectorRequirement{
 							Key:      "k",
 							Operator: metav1.LabelSelectorOpIn,
-							Values:   []string{"v1", "v2"},
+							Values:   []string{"v1", "v3"},
```
bcreane: and here ...

lib/backend/k8s/conversion/conversion_test.go
bcreane: ah okay, you can disregard previous comments on spurious `LabelSelector` changes

lib/backend/k8s/k8s_test.go
```diff
@@ -39,27 +39,27 @@ import (
 
 var (
 	zeroOrder                  = float64(0.0)
-	calicoAllowPolicyModelSpec = capiv2.GlobalNetworkPolicySpec{
+	calicoAllowPolicyModelSpec = capiv3.GlobalNetworkPolicySpec{
```
bcreane: ideally `capiv2` becomes `apiv3`

lib/hwm/hwm.go
```diff
@@ -17,7 +17,7 @@ package hwm
 
 import (
 	log "github.com/sirupsen/logrus"
-	"gopkg.in/tchap/go-patricia.v2/patricia"
+	"gopkg.in/tchap/go-patricia.v3/patricia"
```
bcreane: this looks wrong

lib/backend/compat/compat.go
robbrockbank: yeah saw that and thought meh.  I've actually just reverted that file.  We can probably bin the combat module completely - but wanted to keep it around until I'd done the migration code.

lib/backend/k8s/conversion/conversion_test.go
robbrockbank: yeah I think you saw my comment.  This change wasn't necessary but it doesn't actually make any difference - we're just testing a different label value.

robbrockbank: i'll change.  it's beginning to bug me now.

PR #676, opened 2017-11-14
Revert calculation of veth name in kdd

robbrockbank:
## Description

The calculation of the veth name has changed - this reverts that change.

IIUC without that change, upgrading the calico/node on a node that has Calico networked Pods - those pods will not be networked correctly after upgrade because Felix view of the interface name will differ from the interface name that was actually created.

Also added #677 to test against the release-1.7 branch.  Don't need to commit that one unless we think we should - I just added as a sanity check for this PR.

## Todos
- [ ] cni
- [ ] felix

```release-note
None required
```

### DIFF
--- a/lib/backend/k8s/conversion/conversion_test.go
+++ b/lib/backend/k8s/conversion/conversion_test.go
@@ -163,6 +163,11 @@ var _ = Describe("Test Pod conversion", func() {
 			// Unknown protocol port is ignored.
 		))
 
+		// Assert the interface name is fixed.  The calculation of this name should be consistent
+		// between releases otherwise there will be issues upgrading a node with networked Pods.
+		// If this fails, fix the code not this expect!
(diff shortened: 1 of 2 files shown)


lib/backend/k8s/conversion/conversion_test.go
```diff
@@ -163,14 +163,19 @@ var _ = Describe("Test Pod conversion", func() {
 			// Unknown protocol port is ignored.
 		))
 
+		// Assert the interface name is fixed.  The calculation of this name should be consistent
```
caseydavenport: Might want to add this same test to the v1.x-series branch too.

robbrockbank: Ah i thought I had, but I added to the release-1.7 not the 1.x.  Oops.

bcreane:
@robbrockbank - do we need to release note this change?

robbrockbank:
no - it's just a change to the tests.

PR #683, opened 2017-11-16
Fix that node name regex didn't match upper case.

fasaxc:
## Todos
- [x] Tests
- [x] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


fasaxc:
Apparently, this is WAD

PR #685, opened 2017-11-16
Fix error message when the minimum port of a range is invalid

neiljerram:
For example, prior to this change, for YAML

kind: NetworkPolicy
metadata:
  name: policy2
spec:
  Egress:
  - action: Deny
    destination: {}
    protocol: TCP
    source:
      ports:
      - 0:65535
  Ingress:
  - action: Allow
    destination: {}
    protocol: UDP
    source: {}
  order: 100000
  selector: ''

calicoctl says:

Failed to execute command: error with field Port = '65535' (port range invalid, port number must be between 0 and 65536)


caseydavenport:
@bcreane @neiljerram is there an easy test we can write for this?

neiljerram:
Thanks for adding that test, Brendan!

PR #687, opened 2017-11-17
Fixed issues with not all API functions being consistent

mgleung:
## Description
Fixes some consistency errors with the API and cases where it could be used:
1. Return values from `Create`, `Apply`, `Update` requests will mirror the exact information that was stored, not just the information that was passed in.
2. `List` and `Watch` functions will handle name prefixing properly for policies so that the return values from `Create` or `Update` requests can be used in those functions.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/backend/etcdv3/etcdv3.go
+++ b/lib/backend/etcdv3/etcdv3.go
@@ -124,6 +124,9 @@ func (c *etcdV3Client) Create(ctx context.Context, d *model.KVPair) (*model.KVPa
 		return existing, cerrors.ErrorResourceAlreadyExists{Identifier: d.Key}
 	}
 
+	v, err := model.ParseValue(d.Key, []byte(value))
+	cerrors.PanicIfErrored(err, "Unexpected error parsing stored datastore entry: %v", value)
+	d.Value = v
 	d.Revision = strconv.FormatInt(txnResp.Header.Revision, 10)
 
 	return d, nil
@@ -182,6 +185,9 @@ func (c *etcdV3Client) Update(ctx context.Context, d *model.KVPair) (*model.KVPa
 		return existing, cerrors.ErrorResourceUpdateConflict{Identifier: d.Key}
 	}
 
+	v, err := model.ParseValue(d.Key, []byte(value))
+	cerrors.PanicIfErrored(err, "Unexpected error parsing stored datastore entry: %v", value)
+	d.Value = v
 	d.Revision = strconv.FormatInt(txnResp.Header.Revision, 10)
 
 	return d, nil
@@ -207,6 +213,9 @@ func (c *etcdV3Client) Apply(d *model.KVPair) (*model.KVPair, error) {
 		return nil, cerrors.ErrorDatastoreError{Err: err}
 	}
 
+	v, err := model.ParseValue(d.Key, []byte(value))
+	cerrors.PanicIfErrored(err, "Unexpected error parsing stored datastore entry: %v", value)
+	d.Value = v
 	d.Revision = strconv.FormatInt(resp.Header.Revision, 10)
 
 	return d, nil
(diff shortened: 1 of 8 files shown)


lib/backend/etcdv3/conversion.go
```diff
@@ -91,9 +94,22 @@ func etcdToKVPair(key model.Key, ekv *mvccpb.KeyValue) (*model.KVPair, error) {
 		}
 	}
 
+	if _, ok := v.(*apiv3.GlobalNetworkPolicy); ok {
+		v.(metav1.ObjectMetaAccessor).GetObjectMeta().SetName(convertPolicyNameFromStorage(v.(metav1.ObjectMetaAccessor).GetObjectMeta().GetName()))
```
shatrugna: you could break this into two stmts for easier reading.

lib/backend/etcdv3/conversion.go
```diff
@@ -91,9 +94,22 @@ func etcdToKVPair(key model.Key, ekv *mvccpb.KeyValue) (*model.KVPair, error) {
...
+	}
+
 	return &model.KVPair{
 		Key:      key,
 		Value:    v,
 		Revision: strconv.FormatInt(ekv.ModRevision, 10),
 	}, nil
 }
+
+func convertPolicyNameFromStorage(name string) string {
```
shatrugna: isn't this the same function thats already defined? lets just move this into a common package to avoid duplication.

lib/backend/etcdv3/etcdv3.go
```diff
@@ -124,6 +124,15 @@ func (c *etcdV3Client) Create(ctx context.Context, d *model.KVPair) (*model.KVPa
 		return existing, cerrors.ErrorResourceAlreadyExists{Identifier: d.Key}
 	}
 
+	v, err := model.ParseValue(d.Key, []byte(value))
+	if err != nil {
```
shatrugna: this should technically never err. hence, do we panic instead?

lib/backend/etcdv3/conversion.go
mgleung: I added these as a part of the WIP since I was testing the functionality before I started moving functions into different files and refactoring. I totally agree that these functions should be shared, but while I'm trying out a somewhat hackey solution, I'll leave them there for now for testing.

lib/backend/etcdv3/etcdv3.go
mgleung: I think this is leftover from something else I tried and forgot to remove. Out of curiosity though, what is the benefit of throwing a panic vs returning an error here? From what I can tell, we're in trouble regardless since an error here means we successfully stored unparseable data, which would make me think the correct action to take would be to run a `Delete` and then throw the error to let the user know.

mgleung: I think you're right, a panic might be more appropriate here since I think the error message doesn't really convey "we have stored your validated info but for some reason cannot read it back".

lib/backend/etcdv3/conversion.go
```diff
@@ -91,9 +94,23 @@ func etcdToKVPair(key model.Key, ekv *mvccpb.KeyValue) (*model.KVPair, error) {
 		}
 	}
 
+	if _, ok := v.(*apiv3.GlobalNetworkPolicy); ok {
```
shatrugna: don't forget NetworkPolicy

mgleung: Since this is a WIP, I wanted to check one first before I do the other. I'll definitely remember to add it in if this approach looks good, but it seems a little hackey to me and I'm looking for a better way.

matthewdupre:
FYI @robbrockbank we'd like to get this in before any significant post-beta changes go in so there's something to work off that's close to beta.

lib/backend/etcdv3/etcdv3.go
```diff
@@ -124,6 +124,11 @@ func (c *etcdV3Client) Create(ctx context.Context, d *model.KVPair) (*model.KVPa
...
+		cerrors.PanicIfErrored(err, "Unexpected error parsing stored datastore entry: %v", value)
```
robbrockbank: You don't need the if `err != nil` check to use PanicIfErrored

mgleung: good to know

PR #691, opened 2017-11-21
Added logic to remove prefixes from watcher output

mgleung:
## Description
Make it so the internal name prefixes on the policy objects are stripped out and consistent with what the user interacts with.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/clientv3/bgpconfig.go
+++ b/lib/clientv3/bgpconfig.go
@@ -110,7 +110,7 @@ func (r bgpConfigurations) List(ctx context.Context, opts options.ListOptions) (
 // Watch returns a watch.Interface that watches the BGPConfiguration that
 // match the supplied options.
 func (r bgpConfigurations) Watch(ctx context.Context, opts options.ListOptions) (watch.Interface, error) {
-	return r.client.resources.Watch(ctx, opts, apiv3.KindBGPConfiguration)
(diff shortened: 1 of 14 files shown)


robbrockbank:
This is a much bigger change than I was expecting.  If you had defined the Watch converter just within the scope of the clientv3 package (which I think is all that was required) then you wouldn't have needed to make public the ResourceObject and ResourceList interfaces.  Any particular reason you did that?

Whilst I'm not against adding those two additional interfaces to the apiv3 package it might be good to double check that (a) the deepcopygen isn't going to barf and (b) that other people are ok with this being part of the public API.

lib/watch/interface.go
```diff
@@ -30,6 +31,12 @@ type Interface interface {
 	ResultChan() <-chan Event
 }
 
+// WatcherConverter represents a formatter for calico resources returned by Watch.
+type WatcherConverter interface {
```
robbrockbank: I don't think this belongs here.  We aren't adding conversion to general watcher interface - just the interface clientv3 watcher.

mgleung: Fair enough. I'll move them around.

robbrockbank:
Looks fine - but I'm not sure why we are doing this.  I would like to understand that.

PR #692, opened 2017-11-22
Add v1 to v3 conversion code and tests for IPPool

gunjan5:
## Description

V1 to V3 conversion code and tests for v1->v3 upgrade for IPPool resource.
This code only converts v1 backend model KVpair to 3 backend KVPair for etcd

## Todos
- [x] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/upgrade/etcd/conversionv1v3/ippool.go
+++ b/lib/upgrade/etcd/conversionv1v3/ippool.go
@@ -1,0 +1,75 @@
+// Copyright (c) 2017 Tigera, Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package conversionv1v3
+
+import (
+	"fmt"
+	"strings"
+
+	log "github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	apiv3 "github.com/projectcalico/libcalico-go/lib/apis/v3"
+	"github.com/projectcalico/libcalico-go/lib/backend/model"
+	"github.com/projectcalico/libcalico-go/lib/ipip"
+	cnet "github.com/projectcalico/libcalico-go/lib/net"
+)
+
+func ConvertIPPool(kvp *model.KVPair) (*model.KVPair, error) {
+	pool, ok := kvp.Value.(model.IPPool)
+	if !ok {
+		return nil, fmt.Errorf("value is not a valid BGPPeer resource key")
+	}
+
+	return &model.KVPair{
+		Key: model.ResourceKey{
+			Name: cidrToName(pool.CIDR),
+			Kind: strings.ToLower(apiv3.KindIPPool),
+		},
+		Value: apiv3.IPPool{
+			ObjectMeta: v1.ObjectMeta{Name: cidrToName(pool.CIDR)},
+			Spec: apiv3.IPPoolSpec{
+				CIDR:        pool.CIDR.String(),
+				IPIPMode:    convertIPIPMode(pool.IPIPMode, pool.IPIPInterface),
+				NATOutgoing: pool.Masquerade,
+				Disabled:    pool.Disabled,
+			},
+		},
+	}, nil
+}
+
+func convertIPIPMode(mode ipip.Mode, ipipInterface string) apiv3.IPIPMode {
+	ipipMode := strings.ToLower(string(mode))
+
+	if ipipInterface == "" {
+		return apiv3.IPIPModeNever
+	} else if ipipMode == "cross-subnet" {
+		return apiv3.IPIPModeCrossSubnet
+	}
+	return apiv3.IPIPModeAlways
+}
+
+func cidrToName(cidr cnet.IPNet) string {
+	name := strings.Replace(cidr.String(), ".", "-", 3)
+	name = strings.Replace(name, ":", "-", 7)
+	name = strings.Replace(name, "/", "-", 1)
+
+	log.WithFields(log.Fields{
+		"Name":  name,
+		"IPNet": cidr.String(),
+	}).Debug("Converted IPNet to resource name")
+
+	return name
+}
--- a/lib/upgrade/etcd/conversionv1v3/ippool_test.go
+++ b/lib/upgrade/etcd/conversionv1v3/ippool_test.go
@@ -1,0 +1,213 @@
+// Copyright (c) 2017 Tigera, Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package conversionv1v3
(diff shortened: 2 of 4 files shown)


lib/upgrade/etcd/conversionv1v3/ippool.go
```diff
@@ -0,0 +1,80 @@
...
+	default:
+		log.WithFields(log.Fields{
+			"IPIPMode": mode,
+		}).Fatalf("Unrecognized IPIP mode")
+	}
+
+	return apiv3.IPIPModeNever
+}
+
+func cidrToName(cidr cnet.IPNet) string {
```
robbrockbank: Hmm, I wonder if our KDD 2.x code is bugged.  We currently don't expand the CIDR, but I think we actually should have to avoid leading double dashes.

For example `::aabb/96 -> --aabb/96`

robbrockbank: So I guess we should be expanding the IPs for this.  I don't think we need to do anything for 2.x though and KDD doesn't need migrating.

lib/upgrade/etcd/conversionv1v3/ippool.go
```diff
@@ -0,0 +1,80 @@
...
+				NATOutgoing: pool.Masquerade,
+				Disabled:    pool.Disabled,
+			},
+		},
+	}, nil
+}
+
+func convertIPIPMode(mode ipip.Mode) apiv3.IPIPMode {
+	switch strings.ToLower(string(mode)) {
+	case "", "off":
```
robbrockbank: Is "off" a valid value in 2.x ?  I don't think so.

lib/upgrade/etcd/conversionv1v3/ippool.go
```diff
@@ -0,0 +1,80 @@
...
+		return apiv3.IPIPModeNever
+	case "cross-subnet":
+		return apiv3.IPIPModeCrossSubnet
+	case "always":
```
robbrockbank: Hmm, just looking at the v2 code and I think there is another bug in the Get processing.  The IPIPMode was added at a later stage so that means it's possible to have a backend model that has the Tunnel interface set (i.e. IPIP is enabled), but that there is no IPIPMode field -- in which case we should default to "Always".

Therefore I think "tunl" set and no IPIPMode also corresponds to "Always".

The lack of a "tunl" interface names is what corresponds to "Never"

lib/upgrade/etcd/conversionv1v3/ippool.go
```diff
@@ -0,0 +1,80 @@
...
+import (
+	"fmt"
+	"strings"
+
+	apiv3 "github.com/projectcalico/libcalico-go/lib/apis/v3"
+	"github.com/projectcalico/libcalico-go/lib/backend/model"
+	"github.com/projectcalico/libcalico-go/lib/ipip"
+	cnet "github.com/projectcalico/libcalico-go/lib/net"
+	log "github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/apis/meta/v1"
```
robbrockbank: Most of libcalico-go groups into 3.

built-ins
third party
projectcalico

lib/upgrade/etcd/conversionv1v3/ippool.go
gunjan5: do we really need to expand the IPs? Since these are just names, can't we just add a prefix like `ippool-` and then `s/:/-/g`?

lib/upgrade/etcd/conversionv1v3/ippool_test.go
```diff
@@ -0,0 +1,210 @@
...
+				Name: "10-0-0-1-24",
+				Kind: "ippool",
+			},
+			Value: apiv3.IPPool{
+				ObjectMeta: v1.ObjectMeta{
+					Name: "10-0-0-1-24",
+				},
+				Spec: apiv3.IPPoolSpec{
+					CIDR:        "10.0.0.1/24",
+					IPIPMode:    apiv3.IPIPModeNever,
```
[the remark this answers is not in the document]
gunjan5: right, I can't find the bird template that does that but I remember it should be Always.
 hmm so for the default IPPools, when it's not specified, we set it to `Never`, that shouldn't cause any issue with the conversion, but it seems like an inconsistent behavior

lib/upgrade/etcd/conversionv1v3/ippool.go
gunjan5: ah yes, not in v1 (had it in v3 for a bit). doesn't hurt, but I'll remove it

lib/upgrade/etcd/conversionv1v3/ippool.go
robbrockbank: we could prefix with `ipv4.` or `ipv6.`

robbrockbank: seems a bit odd to prefix with ippool since the resource gives that away

lib/upgrade/etcd/conversionv1v3/ippool_test.go
```diff
@@ -154,7 +154,7 @@ var poolTable = []struct {
 				},
 				Spec: apiv3.IPPoolSpec{
 					CIDR:        "6.6.6.6/26",
-					IPIPMode:    apiv3.IPIPModeNever,
+					IPIPMode:    apiv3.IPIPModeAlways,
```
robbrockbank: Nope still wrong.

The IPIP interface field is what determines whether IPIP is enabled or not.

```
IPIP interface  IPIPMode        v3 IPIPMode
not set              -              Never
set                    not set          Always
set                    cross-subnet    CrossSubnet
set                    not cross-subnet      Always
```

lib/upgrade/etcd/conversionv1v3/ippool.go
```diff
@@ -41,29 +42,26 @@ func ConvertIPPool(kvp *model.KVPair) (*model.KVPair, error) {
...
 	}, nil
 }
 
-func convertIPIPMode(mode ipip.Mode) apiv3.IPIPMode {
-	switch strings.ToLower(string(mode)) {
-	case "", "off":
+func convertIPIPMode(mode ipip.Mode, ipipInterface string) apiv3.IPIPMode {
+	ipipMode := strings.ToLower(string(mode))
+
+	if ipipInterface != "" && ipipMode == "" || ipipMode == "always" {
```
robbrockbank:
```
if ipipInterface == "" {
  return apiv3.IPIPModeNever
}
if ipipMode == "cross-subnet" {
  return apiv3.IPIPModeCrossSubnet
}
return apiv3.IPIPModeAlways
```

lib/upgrade/etcd/conversionv1v3/ippool_test.go
robbrockbank: i've added what i think is the correct logic above.

lib/upgrade/etcd/conversionv1v3/ippool.go
```diff
@@ -0,0 +1,75 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package conversionv1v3
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/projectcalico/libcalico-go/lib/ipip"
```
robbrockbank: projectcalico import is in the wrong block.

lib/upgrade/etcd/conversionv1v3/ippool_test.go
```diff
@@ -0,0 +1,213 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package conversionv1v3
+
+import (
+	"testing"
+
+	. "github.com/onsi/gomega"
+	"github.com/projectcalico/libcalico-go/lib/ipip"
```
robbrockbank: ditto.

lib/upgrade/etcd/conversionv1v3/ippool.go
```diff
@@ -0,0 +1,75 @@
...
+
+func cidrToName(cidr cnet.IPNet) string {
+	name := strings.Replace(cidr.String(), ".", "-", 3)
+	name = strings.Replace(name, ":", "-", 7)
+	name = strings.Replace(name, "/", "-", 1)
+
+	log.WithFields(log.Fields{
+		"Name":  name,
+		"IPNet": cidr.String(),
+	}).Debug("Converting IPNet to resource name")
```
robbrockbank: nit:  Converting -> Converted

PR #693, opened 2017-11-22
Support for ServiceAccount in policy

saumoh:
## Description
New feature to support service accounts match in GNP and NP policy.
Service accounts are a basic unit off which security policies are based off. It will be useful for calico policy to support rules that enable users to use them as selectors.
- Implemented using profiles. Profiles now watch both namespaces and service accounts.
- Support is behind a new env variable ALPHA_FEATURE ("serviceaccounts")
- Limited to KDD for now.

Testing:
- Written a few unit tests and have also tested GNP/NP with service account matches.

## Todos
- [ ] Tests
- [ ] Documentation
- [x ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/apiconfig/apiconfig.go
+++ b/lib/apiconfig/apiconfig.go
@@ -15,6 +15,8 @@
 package apiconfig
 
 import (
+	"strings"
+
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	apiv3 "github.com/projectcalico/libcalico-go/lib/apis/v3"
@@ -26,6 +28,7 @@ const (
 	EtcdV3              DatastoreType = "etcdv3"
 	Kubernetes          DatastoreType = "kubernetes"
 	KindCalicoAPIConfig               = "CalicoAPIConfig"
+	AlphaFeatureSA                    = "serviceaccounts"
 )
 
 // CalicoAPIConfig contains the connection information for a Calico CalicoAPIConfig resource
@@ -44,6 +47,8 @@ type CalicoAPIConfigSpec struct {
 	EtcdConfig
 	// Inline the k8s config fields.
 	KubeConfig
+	// Alpha Feature set: comma separated list of alpha features that are enabled.
+	AlphaFeatures string `json:"alphafeatures" envconfig:"ALPHA_FEATURES" default:""`
 }
 
 type EtcdConfig struct {
@@ -75,4 +80,18 @@ func NewCalicoAPIConfig() *CalicoAPIConfig {
 			APIVersion: apiv3.GroupVersionCurrent,
 		},
 	}
+}
+
+// IsAlphaFeatureSet checks if the comma separated features have the
+// name set in it.
+func IsAlphaFeatureSet(features, name string) bool {
+
+	fs := strings.Split(features, ",")
+	for _, f := range fs {
+		if f == name {
+			return true
+		}
+	}
+
+	return false
 }
--- a/lib/apis/v3/policy.go
+++ b/lib/apis/v3/policy.go
@@ -129,6 +129,21 @@ type EntityRule struct {
 	// Since only some protocols have ports, if any ports are specified it requires the
 	// Protocol match in the Rule to be set to "TCP" or "UDP".
 	NotPorts []numorstring.Port `json:"notPorts,omitempty" validate:"omitempty,dive"`
+
+	// ServiceAccounts is an optional field that restricts the rule to only apply to traffic that originates from (or
+	// terminates at) a pod running as a matching service account.
+	ServiceAccounts *ServiceAccountMatch `json:"serviceAccounts,omitempty" validate:"omitempty"`
+}
+
+type ServiceAccountMatch struct {
+	// Names is an optional field that restricts the rule to only apply to traffic that originates from (or terminates
+	// at) a pod running as a service account whose name is in the list.
+	Names []string `json:"names,omitempty" validate:"omitempty"`
+
+	// Selector is an optional field that restricts the rule to only apply to traffic that originates from
+	// (or terminates at) a pod running as a service account that matches the given label selector.
+	// If both Names and Selector are specified then they are AND'ed.
+	Selector string `json:"selector,omitempty" validate:"omitempty"`
 }
 
 type Action string
--- a/lib/backend/k8s/conversion/constants.go
+++ b/lib/backend/k8s/conversion/constants.go
@@ -15,7 +15,9 @@
 package conversion
 
 const (
-	NamespaceLabelPrefix       = "pcns."
-	NamespaceProfileNamePrefix = "kns."
-	K8sNetworkPolicyNamePrefix = "knp.default."
+	NamespaceLabelPrefix            = "pcns."
+	NamespaceProfileNamePrefix      = "kns."
+	K8sNetworkPolicyNamePrefix      = "knp.default."
+	ServiceAccountLabelPrefix       = "pcsa."
+	ServiceAccountProfileNamePrefix = "ksa."
 )
--- a/lib/backend/k8s/conversion/conversion.go
+++ b/lib/backend/k8s/conversion/conversion.go
@@ -25,6 +25,7 @@ import (
 	kapiv1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+	"github.com/projectcalico/libcalico-go/lib/apiconfig"
 	apiv3 "github.com/projectcalico/libcalico-go/lib/apis/v3"
 	"github.com/projectcalico/libcalico-go/lib/backend/model"
 	"github.com/projectcalico/libcalico-go/lib/names"
@@ -47,6 +48,8 @@ const (
 
 //TODO: make this private and expose a public conversion interface instead
 type Converter struct {
+	// AlphaSA set to true if the alpha feature for serviceaccounts is enabled.
+	AlphaSA bool
 }
 
 // VethNameForWorkload returns a deterministic veth name
@@ -98,7 +101,7 @@ func (c Converter) NamespaceToProfile(ns *kapiv1.Namespace) (*model.KVPair, erro
 			Kind: apiv3.KindProfile,
 		},
 		Value:    profile,
-		Revision: ns.ResourceVersion,
+		Revision: c.JoinProfileRevisions(ns.ResourceVersion, ""),
 	}
 	return &kvp, nil
 }
@@ -146,8 +149,18 @@ func (c Converter) HasIPAddress(pod *kapiv1.Pod) bool {
 // PodToWorkloadEndpoint requires a Pods Name and Node Name to be populated. It will
 // fail to convert from a Pod to WorkloadEndpoint otherwise.
 func (c Converter) PodToWorkloadEndpoint(pod *kapiv1.Pod) (*model.KVPair, error) {
-	// Pull out the profile and workload ID based on pod name and Namespace.
-	profile := NamespaceProfileNamePrefix + pod.Namespace
+
+	// Get all the profiles that apply
+	var profiles []string
+
+	// Pull out the Namespace based profile off the pod name and Namespace.
+	profiles = append(profiles, NamespaceProfileNamePrefix+pod.Namespace)
+
+	// Pull out the Serviceaccount based profile off the pod SA and namespace
+	if c.AlphaSA && pod.Spec.ServiceAccountName != "" {
+		profiles = append(profiles, serviceAccountNameToProfileName(pod.Spec.ServiceAccountName, pod.Namespace))
+	}
+
 	wepids := names.WorkloadEndpointIdentifiers{
 		Node:         pod.Spec.NodeName,
 		Orchestrator: apiv3.OrchestratorKubernetes,
@@ -183,6 +196,10 @@ func (c Converter) PodToWorkloadEndpoint(pod *kapiv1.Pod) (*model.KVPair, error)
 	}
 	labels[apiv3.LabelNamespace] = pod.Namespace
 	labels[apiv3.LabelOrchestrator] = apiv3.OrchestratorKubernetes
+
+	if c.AlphaSA && pod.Spec.ServiceAccountName != "" {
+		labels[apiv3.LabelServiceAccount] = pod.Spec.ServiceAccountName
+	}
 
 	// Map any named ports through.
 	var endpointPorts []apiv3.EndpointPort
@@ -228,7 +245,7 @@ func (c Converter) PodToWorkloadEndpoint(pod *kapiv1.Pod) (*model.KVPair, error)
 		Pod:           pod.Name,
 		Endpoint:      "eth0",
 		InterfaceName: interfaceName,
-		Profiles:      []string{profile},
+		Profiles:      profiles,
 		IPNetworks:    ipNets,
 		Ports:         endpointPorts,
 	}
@@ -557,5 +574,106 @@ func (c Converter) SplitNetworkPolicyRevision(rev string) (crdNPRev string, k8sN
 
 	crdNPRev = revs[0]
 	k8sNPRev = revs[1]
+	return
+}
+
+// serviceAccountNameToProfileName creates a profile name that is a join
+// of 'ksa.' + namespace + "." + serviceaccount name.
+func serviceAccountNameToProfileName(sa, namespace string) string {
+	// Need to incorporate the namespace into the name of the sa based profile
+	// to make them globally unique
+	if namespace == "" {
+		namespace = "default"
+	}
+	return ServiceAccountProfileNamePrefix + namespace + "." + sa
+}
+
+// ServiceAccountToProfile converts a ServiceAccount to a Calico Profile.  The Profile stores
+// labels from the ServiceAccount which are inherited by the WorkloadEndpoints within
+// the Profile.
+func (c Converter) ServiceAccountToProfile(sa *kapiv1.ServiceAccount) (*model.KVPair, error) {
+	// Generate the labels to apply to the profile, using a special prefix
+	// to indicate that these are the labels from the parent Kubernetes ServiceAccount.
+	labels := map[string]string{}
+	for k, v := range sa.ObjectMeta.Labels {
+		labels[ServiceAccountLabelPrefix+k] = v
+	}
+
+	name := serviceAccountNameToProfileName(sa.Name, sa.Namespace)
+	profile := apiv3.NewProfile()
+	profile.ObjectMeta = metav1.ObjectMeta{
+		Name:              name,
+		CreationTimestamp: sa.CreationTimestamp,
+		UID:               sa.UID,
+	}
+	profile.Spec = apiv3.ProfileSpec{
+		LabelsToApply: labels,
+	}
+
+	// Embed the profile in a KVPair.
+	kvp := model.KVPair{
+		Key: model.ResourceKey{
+			Name: name,
+			Kind: apiv3.KindProfile,
+		},
+		Value:    profile,
+		Revision: c.JoinProfileRevisions("", sa.ResourceVersion),
+	}
+	return &kvp, nil
+}
+
+// ProfileNameToServiceAccount extracts the ServiceAccount name from the given Profile name.
+func (c Converter) ProfileNameToServiceAccount(profileName string) (ns, sa string, err error) {
+
+	// Profile objects backed by ServiceAccounts have form "ksa.<namespace>.<sa_name>"
+	if !strings.HasPrefix(profileName, ServiceAccountProfileNamePrefix) {
+		// This is not backed by a Kubernetes ServiceAccount.
+		err = fmt.Errorf("Profile %s not backed by a ServiceAccount", profileName)
+		return
+	}
+
+	names := strings.SplitN(profileName, ".", 3)
+	if len(names) != 3 {
+		err = fmt.Errorf("Profile %s is not formatted correctly", profileName)
+		return
+	}
+
+	ns = names[1]
+	sa = names[2]
+	return
+}
+
+// JoinProfileRevisions constructs the revision from the individual namespace and serviceaccount
+// revisions.
+// This is conditional on the feature flag for serviceaccount set or not.
+func (c Converter) JoinProfileRevisions(nsRev, saRev string) string {
+	if c.AlphaSA == false {
+		return nsRev
+	}
+
+	return nsRev + "/" + saRev
+}
+
+// SplitProfileRevision extracts the namespace and serviceaccount revisions from the combined
+// revision returned on the KDD service account based profile.
+// This is conditional on the feature flag for serviceaccount set or not.
+func (c Converter) SplitProfileRevision(rev string) (nsRev string, saRev string, err error) {
+	if rev == "" {
+		return
+	}
+
+	if c.AlphaSA == false {
+		nsRev = rev
+		return
+	}
+
+	revs := strings.Split(rev, "/")
+	if len(revs) != 2 {
+		err = fmt.Errorf("ResourceVersion is not valid: %s", rev)
+		return
+	}
+
+	nsRev = revs[0]
+	saRev = revs[1]
 	return
 }
--- a/lib/backend/k8s/conversion/conversion.go
+++ b/lib/backend/k8s/conversion/conversion.go
@@ -25,7 +25,6 @@ import (
 	kapiv1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
-	"github.com/projectcalico/libcalico-go/lib/apiconfig"
 	apiv3 "github.com/projectcalico/libcalico-go/lib/apis/v3"
 	"github.com/projectcalico/libcalico-go/lib/backend/model"
 	"github.com/projectcalico/libcalico-go/lib/names"
--- a/lib/backend/k8s/conversion/conversion_test.go
+++ b/lib/backend/k8s/conversion/conversion_test.go
@@ -57,6 +57,30 @@ var _ = Describe("Test parsing strings", func() {
 		Expect(err).To(HaveOccurred())
 		Expect(ns).To(Equal(""))
 	})
+
+	It("should parse valid sa profile names", func() {
+		name := "ksa.default.test"
+		ns, sa, err := c.ProfileNameToServiceAccount(name)
+		Expect(sa).To(Equal("test"))
+		Expect(ns).To(Equal("default"))
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	It("should parse valid sa profile names with dot(.)", func() {
+		name := "ksa.default.test.foo"
+		ns, sa, err := c.ProfileNameToServiceAccount(name)
+		Expect(sa).To(Equal("test.foo"))
+		Expect(ns).To(Equal("default"))
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	It("should not parse invalid sa profile names", func() {
+		name := "ns.projectcalico.org/default"
+		ns, sa, err := c.ProfileNameToServiceAccount(name)
+		Expect(err).To(HaveOccurred())
+		Expect(ns).To(Equal(""))
+		Expect(sa).To(Equal(""))
+	})
 })
 
 var _ = Describe("Test selector conversion", func() {
@@ -204,6 +228,7 @@ var _ = Describe("Test Pod conversion", func() {
 		Expect(wep.Value.(*apiv3.WorkloadEndpoint).Spec.Orchestrator).To(Equal("k8s"))
 		Expect(len(wep.Value.(*apiv3.WorkloadEndpoint).Spec.IPNetworks)).To(Equal(1))
 		Expect(wep.Value.(*apiv3.WorkloadEndpoint).Spec.IPNetworks[0]).To(Equal("192.168.0.1/32"))
+		Expect(len(wep.Value.(*apiv3.WorkloadEndpoint).Spec.Profiles)).To(Equal(1))
 		expectedLabels := map[string]string{
 			"labelA":                         "valueA",
 			"labelB":                         "valueB",
@@ -285,6 +310,7 @@ var _ = Describe("Test Pod conversion", func() {
 		Expect(wep.Value.(*apiv3.WorkloadEndpoint).Spec.Endpoint).To(Equal("eth0"))
 		Expect(wep.Value.(*apiv3.WorkloadEndpoint).Spec.Orchestrator).To(Equal("k8s"))
 		Expect(len(wep.Value.(*apiv3.WorkloadEndpoint).Spec.IPNetworks)).To(Equal(1))
+		Expect(len(wep.Value.(*apiv3.WorkloadEndpoint).Spec.Profiles)).To(Equal(1))
 		Expect(wep.Value.(*apiv3.WorkloadEndpoint).ObjectMeta.Labels).To(Equal(map[string]string{
 			"projectcalico.org/namespace":    "default",
 			"projectcalico.org/orchestrator": "k8s",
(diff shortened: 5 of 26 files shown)


lib/backend/k8s/conversion/constants.go
```diff
@@ -15,7 +15,9 @@
 package conversion
 
 const (
-	NamespaceLabelPrefix       = "pcns."
-	NamespaceProfileNamePrefix = "kns."
-	K8sNetworkPolicyNamePrefix = "knp.default."
+	NamespaceLabelPrefix            = "pcns."
+	NamespaceProfileNamePrefix      = "kns."
+	K8sNetworkPolicyNamePrefix      = "knp.default."
+	ServiceAccountLabelPrefix       = "pcsa/"
```
robbrockbank: just curious as to why these are `/` separated rather than `.`

lib/backend/model/rule.go
```diff
@@ -39,16 +39,20 @@ type Rule struct {
...
+	SrcSelector   string             `json:"src_selector,omitempty" validate:"omitempty,selector"`
+	SrcPorts      []numorstring.Port `json:"src_ports,omitempty" validate:"omitempty,dive"`
+	SrcSANames    string             `json:"src_sa_name,omitempty" validate:"omitempty"`
+	SrcSASelector string             `json:"src_sa_selector,omitempty" validate:"omitempty"`
+	DstTag        string             `json:"dst_tag,omitempty" validate:"omitempty,tag"`
+	DstSelector   string             `json:"dst_selector,omitempty" validate:"omitempty,selector"`
+	DstNet        *net.IPNet         `json:"dst_net,omitempty" validate:"omitempty"`
+	DstNets       []*net.IPNet       `json:"dst_nets,omitempty" validate:"omitempty"`
+	DstPorts      []numorstring.Port `json:"dst_ports,omitempty" validate:"omitempty,dive"`
+	DstSANames    string             `json:"dst_sa_name,omitempty" validate:"omitempty"`
```
robbrockbank: Are there Felix changes for this too?  Just wondered why we needed to change the Felix model for this.

lib/backend/k8s/conversion/constants.go
saumoh: labels allow '.' in them :-( namespaces do not. and names also allow '.' in them.

lib/backend/model/rule.go
saumoh: There are no felix changes. tbh i was not sure if this is needed or not. i noticed we don't do it for namespaceSelector. 
Is it that all the sa based selector are going to get munged into the selector and therefore we don't need these?

lib/converter/rule.go
```diff
@@ -73,6 +73,17 @@ func ruleAPIToBackend(ar api.Rule) model.Rule {
 		})
 	}
 
+	var srcSANames, srcSASelector, dstSANames, dstSASelector string
+	if ar.Source.ServiceAccounts != nil {
```
saumoh: @robbrockbank could u please confirm if I remove SrcSANames and SrcSASelectors then we will not lose any state when going from API to backend. I am guessing the logic is v3->v1 (and now everything is in the selector) and then to backend.

lib/apis/v1/rule.go
```diff
@@ -138,6 +138,20 @@ type EntityRule struct {
...
+
+	// ServiceAccounts is an optional field that restricts the rule to only apply to traffic that originates from (or
+	// terminates at) a pod running as a matching service account.
+	ServiceAccounts *ServiceAccountMatch `json:"serviceAccounts,omitempty" validate:"omitempty"`
+}
+
+type ServiceAccountMatch struct {
+	// Names is an optional field that restricts the rule to only apply to traffic that originates from (or terminates
+	// at) a pod running as a service account whose name is in the list.
+	Names string `json:"names,omitempty" validate:"omitempty"`
```
robbrockbank: What is the format of this?  Should it be a slice of strings?

lib/backend/syncersv1/updateprocessors/rules.go
```diff
@@ -154,6 +203,47 @@ func parseNamespaceSelector(s string) string {
...
+	updated := parsedSelector.String()
+	log.WithFields(log.Fields{"original": sam.Selector, "updated": updated}).Debug("Updated service account selector")
+	if len(sam.Names) == 0 {
+		return updated
+	}
+
+	// Convert the list of ServiceAccounts to selector
+	var namesSelector, comma string
+	namesSelector = fmt.Sprintf("%s in { ", apiv3.LabelServiceAccount)
+	for _, name := range sam.Names {
```
robbrockbank: Isn't Names a string?

lib/apis/v1/rule.go
robbrockbank: Do we need this in the v1 API?  Isn't this all just for Calico v3?

lib/apis/v3/policy.go
```diff
@@ -129,6 +129,21 @@ type EntityRule struct {
...
+
+	// ServiceAccounts is an optional field that restricts the rule to only apply to traffic that originates from (or
+	// terminates at) a pod running as a matching service account.
+	ServiceAccounts *ServiceAccountMatch `json:"serviceAccounts,omitempty" validate:"omitempty"`
+}
+
+type ServiceAccountMatch struct {
+	// Names is an optional field that restricts the rule to only apply to traffic that originates from (or terminates
+	// at) a pod running as a service account whose name is in the list.
+	Names []string `json:"names,omitempty" validate:"omitempty"`
```
robbrockbank: Aha!   here it is, it is a string slice.

lib/backend/syncersv1/updateprocessors/rules.go
robbrockbank: Ah I see, I was looking at the v1 definition - which I think is surplus code.

Since Names is just a strings slice, you can use strings.Join to join together the list of values with a ", " delimiter.

lib/backend/k8s/conversion/constants.go
robbrockbank: @caseydavenport  what's your take on using / here rather than . ?

@saumoh don't we return these names to the user - and AAIK you won't be able to have slashes in the names.  From the code I think you are joining the prefix + namespace + SAP name.  Since the namespace cannot have dots then I think using a dot separator will not result in any conflicts.

lib/backend/k8s/conversion/conversion.go
```diff
@@ -555,7 +570,78 @@ func (c Converter) SplitNetworkPolicyRevision(rev string) (crdNPRev string, k8sN
...
-	k8sNPRev = revs[1]
+	RevA = revs[0]
+	RevB = revs[1]
+	return
+}
+
+// serviceAccountNameToProfileName creates a profile name that is a join
+// of 'ksa/' + namespace + "/" + serviceaccount name.
+// namespace can only have [a-z0-9] and '-'. But serviceaccount name may have
+// '.'.
```
robbrockbank: I don't think it matters that the SA can have a dot since it is the last section in the name.

lib/backend/syncersv1/updateprocessors/rules.go
saumoh: phew... so to confirm i should nuke the code in v1/rule.go?! Right.
i'll use stings.Join.

lib/backend/syncersv1/updateprocessors/rules.go
```diff
@@ -95,6 +110,23 @@ func RuleAPIV2ToBackend(ar apiv3.Rule, ns string) model.Rule {
 		}
 	}
 
+	// Append sourceSASelector
+	if sourceSASelector != "" && (ar.Source.Selector != "" || ar.Source.NotSelector != "" || ar.Source.ServiceAccounts != nil) {
```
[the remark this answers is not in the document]
caseydavenport: So we probably need to calculate `selector` and `svcAcctSelector`, and then do something _roughly_ like this.

```
selectors := []string{}
if svcAcctSelector:
  selecotrs = append(selectors, svcAccountSelector)
if selector:
  selectors = append(selectors, selector)
if (selector || notSelector || svcAccountSelector) && nsSelector:
  selectors = []string{nsSelector, selectors...}

selector = strings.Join("&&", selectors)
```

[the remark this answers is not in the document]
caseydavenport: And to make this just a little bit harder for you... we need to gate this all on the feature flag somehow :p

lib/backend/syncersv1/updateprocessors/rules.go
```diff
@@ -154,6 +203,47 @@ func parseNamespaceSelector(s string) string {
 	return updated
 }
 
+// parseServiceAccounts takes a v2 service account selector and returns the appropriate v1 representation
+// by prefixing the keys with the `pcsa/` prefix. For example, `k == 'v'` becomes `pcsa/k == 'v'`.
```
caseydavenport: comment is slightly wrong, in that it also handles the entire match struct - including names.

lib/backend/syncersv1/updateprocessors/rules.go
```diff
@@ -154,6 +203,47 @@ func parseNamespaceSelector(s string) string {
...
+func parseServiceAccounts(sam *apiv3.ServiceAccountMatch) string {
+
+	parsedSelector, err := parser.Parse(sam.Selector)
```
caseydavenport: We only need to do this when `sam.Selector != ""`, right?

lib/backend/syncersv1/updateprocessors/rules.go
```diff
@@ -154,6 +203,47 @@ func parseNamespaceSelector(s string) string {
...
+	namesSelector = fmt.Sprintf("%s in { ", apiv3.LabelServiceAccount)
+	for _, name := range sam.Names {
+		namesSelector = fmt.Sprintf("%s%s'%s'", namesSelector, comma, name)
+		comma = ", "
+	}
+	namesSelector = fmt.Sprintf("%s }", namesSelector)
+
+	// A list of Service account names are AND'd with the selectors.
+	// TBD: U sure about that?
+	selectors := updated + " && " + namesSelector
```
caseydavenport: This isn't quite right, because `updated` could be an empty string if no selector was given (just names).

lib/backend/syncersv1/updateprocessors/rules.go
```diff
@@ -154,6 +203,47 @@ func parseNamespaceSelector(s string) string {
...
+
+	parsedSelector.AcceptVisitor(parser.PrefixVisitor{Prefix: conversion.ServiceAccountLabelPrefix})
+	updated := parsedSelector.String()
+	log.WithFields(log.Fields{"original": sam.Selector, "updated": updated}).Debug("Updated service account selector")
+	if len(sam.Names) == 0 {
+		return updated
+	}
+
+	// Convert the list of ServiceAccounts to selector
+	var namesSelector, comma string
```
caseydavenport: Might be cleaner to do something like this: https://play.golang.org/p/K9L10y2dLp

Not sure if it's just me, but seems more digestible to me and is fewer lines of code.

lib/backend/k8s/conversion/constants.go
saumoh: @robbrockbank  i see what u mean. I can use '.' for both the ServiceAccountLabelPrefix and the ServiceAccountProfileNamePrefix.
I'll change the code in ProfileNameToServiceAccount to use strings.SplitN()

lib/backend/syncersv1/updateprocessors/rules.go
```diff
@@ -154,6 +203,47 @@ func parseNamespaceSelector(s string) string {
...
+		comma = ", "
+	}
+	namesSelector = fmt.Sprintf("%s }", namesSelector)
+
+	// A list of Service account names are AND'd with the selectors.
+	// TBD: U sure about that?
+	selectors := updated + " && " + namesSelector
+	log.Debugf("SA Selector pre-parse is: %s", selectors)
+
+	parsedSelector, err = parser.Parse(selectors)
```
caseydavenport: Ah, is this in order to eliminate the potential empty `updated` part? Does that work OK?

lib/backend/syncersv1/updateprocessors/rules_test.go
```diff
@@ -296,7 +298,38 @@ var _ = Describe("Test the Rules Conversion Functions", func() {
 		})
 
 		By("generating the correct destination selector", func() {
-			Expect(rulev1.SrcSelector).To(Equal(e))
+			Expect(rulev1.DstSelector).To(Equal(e))
+		})
+	})
+
+	It("should parse a serviceaccount selector", func() {
```
caseydavenport: I think we also need a few other tests here:

- should parse a rule with both a serviceaccount and a namespace selector.
- should parse a rule with both a serviceaccount and wep selector.
- should parse a rule with all three.

lib/backend/syncersv1/updateprocessors/workloadendpointprocessor_test.go
```diff
@@ -275,4 +276,51 @@ var _ = Describe("Test the WorkloadEndpoint update processor", func() {
 			Key: v1WorkloadEndpointKey1,
 		}))
 	})
+
+	It("should filter out a WEP with namespace or serviceAccount labels", func() {
```
caseydavenport: Nice, thanks for adding this test.

lib/clientv3/globalnetworkpolicy.go
```diff
@@ -140,6 +152,27 @@ func (r globalNetworkPolicies) Watch(ctx context.Context, opts options.ListOptio
 	return r.client.resources.Watch(ctx, opts, apiv3.KindGlobalNetworkPolicy)
 }
 
+func (r globalNetworkPolicies) checkAlphaFeatures(res *apiv3.GlobalNetworkPolicy) error {
+	if apiconfig.IsAlphaFeatureSet(r.client.config.Spec.AlphaFeatures, apiconfig.AlphaFeatureSA) == false {
```
caseydavenport: nit: instead of `== false`, do `!apiconfig.IsAlphaFeatureSet`

lib/clientv3/globalnetworkpolicy.go
```diff
@@ -48,6 +51,11 @@ func (r globalNetworkPolicies) Create(ctx context.Context, res *apiv3.GlobalNetw
 		return nil, err
 	}
 
+	// Check if it is permitted to create the network policy with the specific alpha feature support.
```
caseydavenport: I think I'd prefer to move this code into the generic validation (see line 50 above), but I'll defer to @robbrockbank on that one.

caseydavenport: Probably here: https://github.com/projectcalico/libcalico-go/blob/master/lib/validator/v3/validator.go#L617

saumoh: @casey @robbrockbank I've combined the checks into one place instead of both gnp and np having the same code.
i don't quite understand how to make it part of the validator though. The check needs to reference the alpha_feature configuration flag and it is not obvious to me how to get that context into the validator.
If it's not straight forward then maybe this is something we can handle post merge? wdyt

lib/backend/k8s/conversion/conversion.go
```diff
@@ -555,7 +570,78 @@ func (c Converter) SplitNetworkPolicyRevision(rev string) (crdNPRev string, k8sN
...
 	}
 
-	crdNPRev = revs[0]
-	k8sNPRev = revs[1]
+	RevA = revs[0]
+	RevB = revs[1]
+	return
+}
+
+// serviceAccountNameToProfileName creates a profile name that is a join
```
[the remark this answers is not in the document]
robbrockbank: Yup that's correct.  I said much the same thing.

lib/clientv3/globalnetworkpolicy.go
robbrockbank: I think it's fine revisiting this later as we don't currently have the infrastructure to do this in the generic validator.

lib/converter/rule.go
```diff
@@ -151,25 +152,25 @@ func ruleBackendToAPI(br model.Rule) api.Rule {
...
-			NotSelector: br.NotDstSelector,
-			NotPorts:    br.NotDstPorts,
+			Tag:             br.DstTag,
+			Nets:            br.AllDstNets(),
+			Selector:        br.DstSelector,
+			Ports:           br.DstPorts,
+			NotTag:          br.NotDstTag,
+			NotNets:         br.AllNotDstNets(),
+			NotSelector:     br.NotDstSelector,
+			NotPorts:        br.NotDstPorts,
```
robbrockbank: looks like this file hasn't actually changed now.  I assume running `goimports` will effectively back this out.

lib/apis/v3/policy.go
```diff
@@ -129,6 +138,21 @@ type EntityRule struct {
 	// Since only some protocols have ports, if any ports are specified it requires the
 	// Protocol match in the Rule to be set to "TCP" or "UDP".
 	NotPorts []numorstring.Port `json:"notPorts,omitempty" validate:"omitempty,dive"`
+
+	// ServiceAccounts is an optional field that restricts the rule to only apply to traffic that originates from (or
+	// terminates at) a pod running as a matching service account.
+	ServiceAccounts *ServiceAccountMatch `json:"serviceAccounts,omitempty" validate:"omitempty"`
+}
+
+type ServiceAccountMatch struct {
```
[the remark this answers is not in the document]
saumoh: @caseydavenport If it's a empty SAMatch then it will behave as if there is no SAmatch mentioned at all. I've added a ut for it.

lib/apis/v3/workloadendpoint.go
```diff
@@ -72,6 +72,8 @@ type WorkloadEndpointSpec struct {
 	MAC string `json:"mac,omitempty" validate:"omitempty,mac"`
 	// Ports contains the endpoint's named ports, which may be referenced in security policy rules.
 	Ports []EndpointPort `json:"ports,omitempty" validate:"dive,omitempty"`
+	// Service account that is associated with the endpoint.
+	ServiceAccount string `json:"serviceaccount,omitempty" validate:"omitempty,name"`
```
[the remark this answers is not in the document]
saumoh: Right now it's just a visual/ui thing to show that the wep has a SA on it.

lib/apis/v3/policy.go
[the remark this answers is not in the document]
caseydavenport: We may want to revisit that, but fine for alpha (and maybe GA too)
I think it's a bit odd - I'd expect it to select all service accounts, I _think_.

lib/backend/k8s/conversion/conversion.go
```diff
@@ -555,7 +573,72 @@ func (c Converter) SplitNetworkPolicyRevision(rev string) (crdNPRev string, k8sN
...
+// of 'ksa.' + namespace + "." + serviceaccount name.
+func serviceAccountNameToProfileName(sa, namespace string) string {
+	// Need to incorporate the namespace into the name of the sa based profile
+	// to make them globally unique
+	if namespace == "" {
```
caseydavenport: Is there a particular case where this is expected?

I think that the namespace will _always_ be populated (even in the "default" case).

This code doesn't hurt, but wondering if there was a specific reason it was added, or just being safe.

lib/backend/k8s/conversion/conversion.go
```diff
@@ -555,7 +573,72 @@ func (c Converter) SplitNetworkPolicyRevision(rev string) (crdNPRev string, k8sN
...
+		namespace = "default"
+	}
+	return ServiceAccountProfileNamePrefix + namespace + "." + sa
+}
+
+// ServiceAccountToProfile converts a ServiceAccount to a Calico Profile.  The Profile stores
+// labels from the ServiceAccount which are inherited by the WorkloadEndpoints within
+// the Profile. This Profile also has the default ingress and egress rules, which are both 'allow'.
```
caseydavenport: Comment is out of date now - doesn't contain ingress / egress rules.

lib/backend/k8s/conversion/conversion.go
```diff
@@ -555,7 +573,72 @@ func (c Converter) SplitNetworkPolicyRevision(rev string) (crdNPRev string, k8sN
...
+func (c Converter) ServiceAccountToProfile(sa *kapiv1.ServiceAccount) (*model.KVPair, error) {
+	// Generate the labels to apply to the profile, using a special prefix
+	// to indicate that these are the labels from the parent Kubernetes Namespace.
```
caseydavenport: comment out of date - should say "ServiceAccount" instead of "Namespace"

lib/backend/k8s/conversion/conversion_test.go
```diff
@@ -1789,46 +1817,204 @@ var _ = Describe("Test Namespace conversion", func() {
...
+		Expect(len(Ingress)).To(Equal(0))
+		Expect(len(Egress)).To(Equal(0))
+
+		// Check labels.
+		labels := p.Value.(*apiv3.Profile).Spec.LabelsToApply
+		Expect(labels["pcsa.foo"]).To(Equal("bar"))
+		Expect(labels["pcsa.roger"]).To(Equal("rabbit"))
+	})
+
+	It("should parse a ServiceAccount in Namespace to a Profile", func() {
```
caseydavenport: Is it even possible to have a ServiceAccount that's _not_ in a Namespace?

lib/backend/k8s/conversion/conversion_test.go
```diff
@@ -1789,46 +1817,204 @@ var _ = Describe("Test Namespace conversion", func() {
...
 		Expect(err).To(HaveOccurred())
 	})
 })
+
+var _ = Describe("Test ServiceAccount conversion", func() {
+
+	// Use a single instance of the Converter for these tests.
+	c := Converter{AlphaSA: true}
+
+	It("should parse a ServiceAccount to a Profile", func() {
```
caseydavenport: I think this is actually testing the "default" namespace case.

If so, this test case is a bit contrived, because in a real system the objects returned will always have the ObjectMeta.Namespace field populated (even for "default")

caseydavenport: i.e. it's not any different from the "in a namespace" test below.

lib/backend/k8s/resources/profile.go
```diff
@@ -66,22 +71,79 @@ func (c *profileClient) Delete(ctx context.Context, key model.Key, revision stri
...
+
+func (c *profileClient) getServiceAccount(ctx context.Context, name, revision string) (*model.KVPair, error) {
+	namespace, serviceAccountName, err := c.ProfileNameToServiceAccount(name)
+	if err != nil {
+		return nil, err
+	}
+
+	serviceAccount, err := c.clientSet.CoreV1().ServiceAccounts(namespace).Get(serviceAccountName, metav1.GetOptions{ResourceVersion: revision})
+	if err != nil {
+		return nil, err
```
caseydavenport: Anywhere we're querying the k8s API we'll need to wrap in `k8sErrorToCalico` 

like so: https://github.com/colabsaumoh/libcalico-go/blob/4395fc8d5600d87b740ff82e44af1106b5d9814b/lib/backend/k8s/resources/profile.go#L182

lib/backend/k8s/resources/profile.go
```diff
@@ -109,24 +171,54 @@ func (c *profileClient) List(ctx context.Context, list model.ListInterface, revi
...
 			log.Errorf("Unable to convert k8s Namespace to Calico Profile: Namespace=%s: %v", ns.Name, err)
 			continue
 		}
 		kvps = append(kvps, kvp)
 	}
+
+	if c.AlphaSA == false {
+		return &model.KVPairList{
+			KVPairs:  kvps,
+			Revision: namespaces.ResourceVersion,
```
robbrockbank: Is this correct?  I think everywhere else you assume the revision will be a "dual revision" even if the feature is not enabled.

lib/backend/k8s/conversion/conversion.go
```diff
@@ -555,7 +573,72 @@ func (c Converter) SplitNetworkPolicyRevision(rev string) (crdNPRev string, k8sN
...
+	}
+
+	// Embed the profile in a KVPair.
+	kvp := model.KVPair{
+		Key: model.ResourceKey{
+			Name: name,
+			Kind: apiv3.KindProfile,
+		},
+		Value:    profile,
+		Revision: sa.ResourceVersion,
```
robbrockbank: Shouldn't this also be a "dual version"?

lib/backend/k8s/resources/profile.go
```diff
@@ -148,7 +251,212 @@ func (c *profileClient) Watch(ctx context.Context, list model.ListInterface, rev
...
+			}
+
+		case <-pw.context.Done():
+			log.Info("Process watcher done event during watch event in kdd client")
+			return
+		}
+	}
+}
+
+func (c *profileClient) joinRevs(ns, sa string) string {
```
robbrockbank: Not convinced about these functions.

Firstly, they are specific to the AlphaSA feature, but the name of functions does not indicate that, and these functions are accessible to the whole package not just this file - so a little confusing.

Secondly, I'm not even sure these are being used are they?

I would probably have kept the old SplitNetworkPolicyRevision and JoinNetworkPolicyRevision functions and implemented separately names ones for Profiles.  That way it's a bit easier to follow, and there are fewer changes in code that doesn't need to be impacted by this feature.

lib/backend/k8s/resources/profile.go
```diff
@@ -109,24 +171,54 @@ func (c *profileClient) List(ctx context.Context, list model.ListInterface, revi
 		}, nil
 	}
 
+	nsRev, saRev, err := c.splitRev(revision)
```
caseydavenport: I suppose this is fine, though ideally I'd like the revision joining/splitting to be behind the feature flag as well.

@saumoh how hard/disruptive would that be?

lib/backend/k8s/resources/profile.go
```diff
@@ -109,24 +171,54 @@ func (c *profileClient) List(ctx context.Context, list model.ListInterface, revi
...
+	// TBD: narrow down to only to the required namespace
+	serviceaccounts, err = c.clientSet.CoreV1().ServiceAccounts(kapiv1.NamespaceAll).List(metav1.ListOptions{ResourceVersion: saRev})
+	if err != nil {
+		return nil, K8sErrorToCalico(err, nl)
+	}
+
+	for _, sa := range serviceaccounts.Items {
+		kvp, err := c.getSaKv(&sa)
+		if err != nil {
+			log.Errorf("Unable to convert k8s service account to Calico Profile: %s: %v", sa.Name, err)
```
caseydavenport: nit - we generally like to use `log.WithError(err).Errorf(...)`

lib/backend/k8s/resources/profile.go
```diff
@@ -135,11 +227,22 @@ func (c *profileClient) EnsureInitialized() error {
...
 	}
 
-	k8sWatch, err := c.clientSet.CoreV1().Namespaces().Watch(metav1.ListOptions{ResourceVersion: revision})
+	nsRev, saRev, err := c.splitRev(revision)
+	if err != nil {
+		return nil, err
+	}
+
+	log.WithFields(log.Fields{
+		"NSPROFRev": nsRev,
```
caseydavenport: nit: `NSNPROFRev` is a bit hard to parse - let's just use the name of the variable `nsRev` and `saRev`

lib/backend/k8s/resources/profile.go
```diff
@@ -135,11 +227,22 @@ func (c *profileClient) EnsureInitialized() error {
...
+		"SAPROFRev": saRev,
+	}).Info("Watching two resources at individual revisions")
```
caseydavenport: This log isn't accurate when the alpha feature is disabled.

lib/backend/k8s/resources/profile.go
```diff
@@ -148,7 +251,212 @@ func (c *profileClient) Watch(ctx context.Context, list model.ListInterface, rev
...
+	if pw.k8sSAWatch != nil {
+		terminated = terminated && pw.k8sSAWatch.HasTerminated()
+	}
+
+	return terminated
+}
+
+// Loop to process the events stream from the underlying k8s Watcher and convert them to
+// backend KVPs.
+func (pw *profileWatcher) processProfileEvents() {
```
caseydavenport: Seems fine for alpha, but I suspect we're duplicating a lot of code between this and the NP stuff?

Might want to revisit commonizing some of this when we move out of alpha. @robbrockbank wdyt?

lib/backend/k8s/resources/profile.go
saumoh: @robbrockbank good points. 
I will revert to using to distinct SplitNetworkPolicyRevision and JoinNetworkPolicyRevision for NP's and another pair for Profiles that use the alphaSA flag.
For context the ask is to keep the profile revision to be "single version" when the feature flag is off.

lib/backend/k8s/resources/profile.go
saumoh: @caseydavenport I'll make a split/join specific to Profiles that take the feature flag into consideration.

lib/backend/model/rule.go
```diff
@@ -39,16 +39,16 @@ type Rule struct {
...
-	SrcNet      *net.IPNet         `json:"src_net,omitempty" validate:"omitempty"`
-	SrcNets     []*net.IPNet       `json:"src_nets,omitempty" validate:"omitempty"`
-	SrcSelector string             `json:"src_selector,omitempty" validate:"omitempty,selector"`
-	SrcPorts    []numorstring.Port `json:"src_ports,omitempty" validate:"omitempty,dive"`
-	DstTag      string             `json:"dst_tag,omitempty" validate:"omitempty,tag"`
-	DstSelector string             `json:"dst_selector,omitempty" validate:"omitempty,selector"`
-	DstNet      *net.IPNet         `json:"dst_net,omitempty" validate:"omitempty"`
-	DstNets     []*net.IPNet       `json:"dst_nets,omitempty" validate:"omitempty"`
-	DstPorts    []numorstring.Port `json:"dst_ports,omitempty" validate:"omitempty,dive"`
+	SrcTag        string             `json:"src_tag,omitempty" validate:"omitempty,tag"`
```
caseydavenport: Was this a `gofmt` change?

lib/clientv3/alphafeaturecheck.go
```diff
@@ -0,0 +1,45 @@
...
+	"errors"
+	"fmt"
+
+	"github.com/projectcalico/libcalico-go/lib/apiconfig"
+	apiv3 "github.com/projectcalico/libcalico-go/lib/apis/v3"
+)
+
+// checkAlphaFeatures is used to check if GNP and NP rules are compatible with
+// alphaFeatures flag 
+func checkAlphaFeatures(alphaConfig string, ingress, egress []apiv3.Rule) error {
```
caseydavenport: I think this probably belongs in the validation package.

I know in the comment below we say that might be too much work for now, which I suppose is OK, but I'd still argue that even if we can't shoehorn it into the validator properly, we should at least expose this as its own public interface in the validation package and import it from there.

We can rename this `validateNoServiceAccountRules` and then the caller can decide whether or not to call this based on the enabled alpha features.

@robbrockbank @saumoh WDYT?

lib/clientv3/globalnetworkpolicy.go
caseydavenport: Is the problem just that the validator doesn't have the apiconfig available to it?

lib/backend/syncersv1/updateprocessors/rules.go
```diff
@@ -40,6 +40,45 @@ func RulesAPIV2ToBackend(ars []apiv3.Rule, ns string) []model.Rule {
...
+	} else if ns != "" {
+		// No namespace selector was given and this is a namespaced network policy,
+		// so the rule applies only to its own namespace.
+		nsSelector = fmt.Sprintf("%s == '%s'", apiv3.LabelNamespace, ns)
+	}
+
+	var selectors []string
+
+	// Determine which service account selector.
+	if er.ServiceAccounts != nil {
```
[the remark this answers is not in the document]
saumoh: @caseydavenport because we check for the flag at lib/clientv3/[gnp|np] layer. it is safe to assume that when we come to this point that the ServiceAccounts match would be nil if the alphaSA flag was not set.

lib/backend/model/rule.go
saumoh: ya. There is no code change here. i'd added some values, then removed them...should i do 'go fmt' and recommit?

lib/backend/syncersv1/updateprocessors/rules.go
```diff
@@ -140,20 +155,56 @@ func RuleAPIV2ToBackend(ar apiv3.Rule, ns string) model.Rule {
...
+	selectors := fmt.Sprintf("%s in { '%s' }", apiv3.LabelServiceAccount, names)
+
+	// A list of Service account names are AND'd with the selectors.
+	if updated != "" {
+		selectors = updated + " && " + selectors
+	}
+	log.Debugf("SA Selector pre-parse is: %s", selectors)
+
+	// TBD: Do i need this section now or is selectors ready for consumption?
+	parsedSelector, err := parser.Parse(selectors)
```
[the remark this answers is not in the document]
saumoh: if it is not a big impact then let's keep it.

lib/backend/syncersv1/updateprocessors/rules.go
```diff
@@ -140,20 +155,56 @@ func RuleAPIV2ToBackend(ar apiv3.Rule, ns string) model.Rule {
...
+		return updated
+	}
+
+	// Convert the list of ServiceAccounts to selector
+	names := strings.Join(sam.Names, "', '")
+	selectors := fmt.Sprintf("%s in { '%s' }", apiv3.LabelServiceAccount, names)
+
+	// A list of Service account names are AND'd with the selectors.
+	if updated != "" {
+		selectors = updated + " && " + selectors
```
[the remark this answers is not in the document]
saumoh: yup. good catches all around! thx

lib/clientv3/globalnetworkpolicy.go
saumoh: @caseydavenport @robbrockbank The validator does not have apiconfig is the basic issue.
For now let me move this check to the validator as `validateNoServiceAccountRules` and then invoke it from gnp and np.

lib/backend/syncersv1/updateprocessors/rules.go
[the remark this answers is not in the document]
caseydavenport: I think this is still an issue because the client code runs as a totally separate process from this.

So, someone could enable the feature on the client, disable it in Felix, and not expect Felix to interpret it.

lib/backend/model/rule.go
caseydavenport: yep

lib/backend/syncersv1/updateprocessors/rules.go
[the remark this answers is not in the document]
saumoh: k. i see your point. let me see if i can get the flag all the way down :-)

lib/backend/syncersv1/updateprocessors/rules.go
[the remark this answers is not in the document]
caseydavenport: Afraid I don't know the impact too well, but @fasaxc might!

lib/backend/syncersv1/updateprocessors/rules.go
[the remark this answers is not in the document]
saumoh: @caseydavenport It is going to be a bit hairy getting the `AlphaFeatures` peculating all the way down here. given the tight timing and the remaining review, can we -for now- assume that the client is the gatekeeper? I can add this as an enhancement after this pr is merged. wdyt?

lib/backend/syncersv1/updateprocessors/rules.go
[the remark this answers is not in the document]
fasaxc: Parsing and stringifying selectors is pretty fast; I wouldn't worry unless this is being done 100s of times/s.  I believe we do it just to validate the selector elsewhere.  It's nice to check that the selector we've generated is valid early.

lib/backend/syncersv1/updateprocessors/rules.go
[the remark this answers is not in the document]
caseydavenport: SGTM.

As discussed, we can also just mandate an env var be set, and use `os.Getenv` to avoid having to pass the flag around to so many places.

lib/backend/syncersv1/updateprocessors/rules.go
```diff
@@ -140,20 +165,62 @@ func RuleAPIV2ToBackend(ar apiv3.Rule, ns string) model.Rule {
...
+	parsedSelector, err := parser.Parse(selectors)
+	if err != nil {
+		log.WithError(err).Errorf("Failed to parse service account Names: %s", selectors)
+		return ""
+	}
+
+	selectors = parsedSelector.String()
+
+	// A list of Service account names are AND'd with the selectors.
+	if updated != "" {
```
caseydavenport: Not necessarily wrong, but curious why this is done after the parsing above, rather than before?

lib/backend/syncersv1/updateprocessors/rules_test.go
```diff
@@ -296,7 +299,119 @@ var _ = Describe("Test the Rules Conversion Functions", func() {
...
+
+	It("should parse a serviceaccount match", func() {
+		os.Setenv("ALPHA_FEATURES", "serviceaccounts")
+		srce := fmt.Sprintf("(%s == 'namespace') && ((%skey == \"value1\") && (%s in {\"%s\", \"%s\"}))", apiv3.LabelNamespace, conversion.ServiceAccountLabelPrefix, apiv3.LabelServiceAccount, "sa1", "sa2")
+		dste := fmt.Sprintf("(pcns.nskey == \"nsvalue\") && ((%skey == \"value2\") && (%s in {\"%s\"}))", conversion.ServiceAccountLabelPrefix, apiv3.LabelServiceAccount, "sa3")
+
+		r := apiv3.Rule{
+			Action: apiv3.Allow,
+			Source: apiv3.EntityRule{
+				ServiceAccounts: &apiv3.ServiceAccountMatch{Names: []string{"sa1", "sa2"},
```
caseydavenport: nit: Names should be on the next line.

{
  Names: ...
  Selector: ...
}

lib/backend/syncersv1/updateprocessors/rules_test.go
```diff
@@ -296,7 +299,119 @@ var _ = Describe("Test the Rules Conversion Functions", func() {
...
+		By("generating the correct source selector", func() {
+			Expect(rulev1.SrcSelector).To(Equal(srce))
+		})
+
+		By("generating the correct destination selector", func() {
+			Expect(rulev1.DstSelector).To(Equal(dste))
+		})
+	})
+
+	It("should parse a serviceaccount match with default namespace and no namespace selector", func() {
```
caseydavenport: this should say "global" rather than default namespace - passing `""` to `RuleAPIV2ToBackend` means a GNP, not the default namespace.

For the default ns, it will get "default" passed in.

lib/backend/syncersv1/updateprocessors/rules.go
saumoh: the `parser.Parse()` is right now just used to normalize the SA names so that the quotes et. all are corrected. If the `parser.Parse()` would have been expensive i would have just pretty printed it here.

lib/backend/k8s/resources/profile.go
robbrockbank: most of this will vanish once we start using API groups to handle this - that's why i thought this was fine to leave as is for now.

lib/backend/syncersv1/updateprocessors/rules.go
```diff
@@ -140,20 +165,62 @@ func RuleAPIV2ToBackend(ar apiv3.Rule, ns string) model.Rule {
...
 }
 
+// parseServiceAccounts takes a v3 service account match and returns the appropriate v1 representation
+// by converting the list of service account names into a set of service account with
+// key: "projectcalico.org/serviceaccount" in { 'sa-1', 'sa-2' } AND
+// by prefixing the keys with the `pcsa.` prefix. For example, `k == 'v'` becomes `pcsa.k == 'v'`.
+// NOTE: Check if the env variable ALPHA_FEATURES is set or not before converting the ServiceAccountMatch to selector
+func parseServiceAccounts(sam *apiv3.ServiceAccountMatch) string {
+
+	if apiconfig.IsAlphaFeatureSet(os.Getenv("ALPHA_FEATURES"), apiconfig.AlphaFeatureSA) == false {
```
robbrockbank: Shouldn't this use the client apiconfig rather than the environment?  What if a user sets this in the config file but not in the environment - this check will fail.

lib/apiconfig/apiconfig.go
```diff
@@ -76,3 +81,17 @@ func NewCalicoAPIConfig() *CalicoAPIConfig {
 		},
 	}
 }
+
+// IsAlphaFeatureSet checks if the comma separated features have the
+// name set in it.
+func IsAlphaFeatureSet(features, name string) bool {
```
robbrockbank: Just trying to follow the flow of the config here.  Shouldn't this really be a method attached to the apiconfig struct - this ties in with the question that you are accessing the ALPHA_FEATURES environment somewhere in the code whereas everywhere else you access it via the apconfig.

Also, should this return false for the Service account feature if the datastore type is not KDD?

lib/backend/syncersv1/updateprocessors/rules.go
saumoh: yes. ideally! getting apiconfig spec all the way down to updateprocessors was looking a bit hairy.
In talking with @caseydavenport for the interim we came up with using the env. This is going to get invoked in caliconode and there we do set the `ALPHA_FEATURES`
The plan is to make the required changes to make that happen post-merge of this PR.

lib/apiconfig/apiconfig.go
saumoh: yes on this too. I do plan to make this a method attached to apiconfig spec. that will make it easier to take into account the datastore.

saumoh:
Since I can't seem to use git rebase correctly :-( . closing this PR and using https://github.com/projectcalico/libcalico-go/pull/707 for the merge.

PR #694, opened 2017-11-23
Add Hostname helper

caseydavenport:
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/names/hostname.go
+++ b/lib/names/hostname.go
@@ -1,0 +1,16 @@
+package names
+
+import (
+	"os"
+	"strings"
+)
+
+// Hostname returns this hosts hostname, converting to lowercase so that
+// it is valid for use in the Calico API.
+func Hostname() (string, error) {
+	if h, err := os.Hostname(); err != nil {
+		return "", err
+	} else {
+		return strings.ToLower(strings.TrimSpace(h)), nil
+	}
+}


lib/names/hostname.go
```diff
@@ -0,0 +1,15 @@
+package names
+
+import (
+	"os"
+	"strings"
+)
+
+// Hostname returns this hosts hostname, converting to lowercase so that
```
robbrockbank: I think the IPAM package should be update to use this too.  Although one niggle with that is that we need IPAM to work on v2 and v3 simultaneously (in order to implement the proposed CNI changes)

lib/names/hostname.go
```diff
@@ -0,0 +1,15 @@
...
+// it is valid for use in the Calico API.
+func Hostname() (string, error) {
+	if h, err = os.Hostname(); err != nil {
+		return "", err
+	}
+	return strings.ToLower(strings.TrimSpace(h)), nil
```
robbrockbank: either need this in an else or you need to pull the assignment out of the if statement.

lib/names/hostname.go
```diff
@@ -0,0 +1,15 @@
...
+
+import (
+	"os"
+	"strings"
+)
+
+// Hostname returns this hosts hostname, converting to lowercase so that
+// it is valid for use in the Calico API.
+func Hostname() (string, error) {
+	if h, err = os.Hostname(); err != nil {
```
robbrockbank: = -> :=

lib/names/hostname.go
robbrockbank: whilst this covers immediate use case I'd like to be paranoid and handle things like leading/terminating dashes or dots as part of the upgrade work.

lib/names/hostname.go
caseydavenport: suppose the v3 IPAM call should use this, v2 should continue as usual.

caseydavenport:
@robbrockbank CI is passing now, good to go?

PR #697, opened 2017-11-27
Pro-actively mark the ready flag resync as done.

fasaxc:
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


fasaxc:
Are you sure?  I couldn't see anywhere that `needsResync[KEY_RS]` was set to false (and we don't rewrite the whole map any more?)

bcreane:
My comment was vague - I meant we're initializing to `true` here: https://github.com/fasaxc/libcalico-go/blob/c221c64a1303ed64fd2ae4e1e28963808ad1788a/lib/backend/k8s/syncer.go#L181 and we should possibly now initialize to false.

bcreane:
@caseydavenport - can you take a look?

caseydavenport:
Hm, this seems fine to me.

PR #699, opened 2017-11-29
WorkloadEndpoint V1 to V3 conversion

heschlie:
## Description

First round for WEP conversions. I wanted to get a PR up for feedback on the conversion so I could get some feedback. I'm not 100% confident in how I'm doing the conversions.

## Todos
- [x] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/upgrade/etcd/conversionv1v3/workloadendpoint.go
+++ b/lib/upgrade/etcd/conversionv1v3/workloadendpoint.go
@@ -1,0 +1,271 @@
+// Copyright (c) 2017 Tigera, Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package conversionv1v3
+
+import (
+	"fmt"
+	"strings"
+
+	log "github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	apiv1 "github.com/projectcalico/libcalico-go/lib/apis/v1"
+	"github.com/projectcalico/libcalico-go/lib/apis/v1/unversioned"
+	apiv3 "github.com/projectcalico/libcalico-go/lib/apis/v3"
+	"github.com/projectcalico/libcalico-go/lib/backend/model"
+	"github.com/projectcalico/libcalico-go/lib/names"
+	"github.com/projectcalico/libcalico-go/lib/net"
+)
+
+type WorkloadEndpoint struct{}
+
+// APIV1ToBackendV1 converts v1 WorkloadEndpoint API to v1 WorkloadEndpoint KVPair.
+func (_ WorkloadEndpoint) APIV1ToBackendV1(rIn unversioned.Resource) (*model.KVPair, error) {
+	ah := rIn.(apiv1.WorkloadEndpoint)
+	k, err := convertMetadataToKey(ah.Metadata)
+	if err != nil {
+		return nil, err
+	}
+
+	// IP networks are stored in the datastore in separate IPv4 and IPv6
+	// fields.  We normalise the network to ensure the IP is correctly
+	// masked.
+	ipv4Nets := []net.IPNet{}
+	ipv6Nets := []net.IPNet{}
+	for _, n := range ah.Spec.IPNetworks {
+		n = *(n.Network())
+		if n.Version() == 4 {
+			ipv4Nets = append(ipv4Nets, n)
+		} else {
+			ipv6Nets = append(ipv6Nets, n)
+		}
+	}
+
+	ipv4NAT := []model.IPNAT{}
+	ipv6NAT := []model.IPNAT{}
+	for _, n := range ah.Spec.IPNATs {
+		nat := model.IPNAT{IntIP: n.InternalIP, ExtIP: n.ExternalIP}
+		if n.InternalIP.Version() == 4 {
+			ipv4NAT = append(ipv4NAT, nat)
+		} else {
+			ipv6NAT = append(ipv6NAT, nat)
+		}
+	}
+
+	var ports []model.EndpointPort
+	for _, port := range ah.Spec.Ports {
+		ports = append(ports, model.EndpointPort{
+			Name:     port.Name,
+			Protocol: port.Protocol,
+			Port:     port.Port,
+		})
+	}
+
+	d := model.KVPair{
+		Key: k,
+		Value: &model.WorkloadEndpoint{
+			Labels:           ah.Metadata.Labels,
+			ActiveInstanceID: ah.Metadata.ActiveInstanceID,
+			State:            "active",
+			Name:             ah.Spec.InterfaceName,
+			Mac:              ah.Spec.MAC,
+			ProfileIDs:       ah.Spec.Profiles,
+			IPv4Nets:         ipv4Nets,
+			IPv6Nets:         ipv6Nets,
+			IPv4NAT:          ipv4NAT,
+			IPv6NAT:          ipv6NAT,
+			IPv4Gateway:      ah.Spec.IPv4Gateway,
+			IPv6Gateway:      ah.Spec.IPv6Gateway,
+			Ports:            ports,
+		},
+		Revision: ah.Metadata.Revision,
+	}
+
+	log.Debugf("Converted: %+v\n To: %+v", ah, d)
+
+	return &d, nil
+}
+
+// BackendV1ToAPIV3 converts v1 WorkloadEndpoint KVPair to v3 API.
+func (_ WorkloadEndpoint) BackendV1ToAPIV3(kvp *model.KVPair) (Resource, error) {
+	wepKey, ok := kvp.Key.(model.WorkloadEndpointKey)
+	if !ok {
+		return nil, fmt.Errorf("value is not a valid WorkloadEndpoint resource key")
+	}
+	wepValue, ok := kvp.Value.(*model.WorkloadEndpoint)
+	if !ok {
+		return nil, fmt.Errorf("value is not a valid WorkloadEndpoint resource Value")
+	}
+
+	labels := convertLabels(wepValue.Labels)
+	namespace := "default"
+
+	var pod string
+	var container string
+	var workload string
+
+	// Populate our values based on the orchestrator.
+	switch wepKey.OrchestratorID {
+	case "k8s":
+		namespace, pod = getPodNamespaceName(wepKey.WorkloadID)
+	case "cni":
+		container = wepKey.WorkloadID
+	case "libnetwork":
+		workload = "libnetwork"
+	default:
+		workload = convertName(wepKey.WorkloadID)
+	}
+
+	ipNets := convertIPNetworks(wepValue.IPv4Nets)
+	ipNets = append(ipNets, convertIPNetworks(wepValue.IPv6Nets)...)
+
+	ipNats := convertIPNATs(wepValue.IPv4NAT)
+	ipNats = append(ipNats, convertIPNATs(wepValue.IPv6NAT)...)
+
+	wep := apiv3.NewWorkloadEndpoint()
+
+	wep.ObjectMeta = v1.ObjectMeta{
+		Namespace: namespace,
+		Labels:    labels,
+	}
+	wep.Spec = apiv3.WorkloadEndpointSpec{
+		Orchestrator:  convertName(wepKey.OrchestratorID),
+		Workload:      workload,
+		Node:          convertName(wepKey.Hostname),
+		Pod:           pod,
+		ContainerID:   container,
+		Endpoint:      convertName(wepKey.EndpointID),
+		IPNetworks:    ipNets,
+		IPNATs:        ipNats,
+		Profiles:      convertProfiles(wepValue.ProfileIDs),
+		InterfaceName: wepValue.Name,
+		Ports:         convertPorts(wepValue.Ports),
+	}
+
+	if wepValue.IPv4Gateway != nil {
+		wep.Spec.IPv4Gateway = wepValue.IPv4Gateway.String()
+	}
+	if wepValue.IPv6Gateway != nil {
+		wep.Spec.IPv6Gateway = wepValue.IPv6Gateway.String()
+	}
+	if wepValue.Mac != nil {
+		wep.Spec.MAC = wepValue.Mac.String()
+	}
+
+	// Figure out the new name based on WEP fields.
+	wepids := names.WorkloadEndpointIdentifiers{
+		Node:         wep.Spec.Node,
+		Orchestrator: wep.Spec.Orchestrator,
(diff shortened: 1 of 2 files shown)


lib/upgrade/etcd/conversionv1v3/workloadendpoint_test.go
```diff
@@ -0,0 +1,278 @@
...
+			Expect(v1KVPResult.Value.(*model.WorkloadEndpoint).IPv4Gateway).To(Equal(entry.v1KVP.Value.(model.WorkloadEndpoint).IPv4Gateway))
+			Expect(v1KVPResult.Value.(*model.WorkloadEndpoint).IPv6Gateway).To(Equal(entry.v1KVP.Value.(model.WorkloadEndpoint).IPv6Gateway))
+			Expect(v1KVPResult.Value.(*model.WorkloadEndpoint).Ports).To(Equal(entry.v1KVP.Value.(model.WorkloadEndpoint).Ports))
+
+			// Test and assert v1 backend to v3 API logic.
+			v3APIResult, err := w.BackendV1ToAPIV3(entry.v1KVP)
+			Expect(err).NotTo(HaveOccurred(), entry.description)
+			Expect(v3APIResult.(*apiv3.WorkloadEndpoint).ObjectMeta.Name).To(Equal(entry.v3API.ObjectMeta.Name))
+			Expect(v3APIResult.(*apiv3.WorkloadEndpoint).ObjectMeta.Labels).To(Equal(entry.v3API.ObjectMeta.Labels))
+			Expect(v3APIResult.(*apiv3.WorkloadEndpoint).Spec.Orchestrator).To(Equal(entry.v3API.Spec.Orchestrator))
```
gunjan5: can't you assert the whole spec field like this: `Expect(v3APIResult.(*apiv3.WorkloadEndpoint).Spec).To(Equal(entry.v3API.Spec))`

ozdanborne: Yeah @heschlie check it out, I did that on mine: https://github.com/projectcalico/libcalico-go/pull/701/files#diff-d468c767450a173a2e6514311967e164R150

heschlie: yea it works fine for the apiv3 resource, but for some reason it does not with the apiv1 Spec, even though they are identical  ¯\\_(ツ)_/¯

lib/upgrade/etcd/conversionv1v3/workloadendpoint.go
```diff
@@ -0,0 +1,258 @@
...
+		return nil, fmt.Errorf("value is not a valid WorkloadEndpoint resource key")
+	}
+	wepValue, ok := kvp.Value.(model.WorkloadEndpoint)
+	if !ok {
+		return nil, fmt.Errorf("value is not a valid WorkloadEndpoint resource Value")
+	}
+
+	labels := convertLabels(wepValue.Labels)
+	namespace := labels["projectcalico.org/namespace"]
+	pod := getPodName(wepKey.WorkloadID, namespace)
```
robbrockbank: Whether we have a Pod, ContainerID or WorkloadID depends on the orchestrator.  Whilst we don't need this logic now I think it should be relatively simple to implement now.

For k8s, there is no WorkloadID.
For CNI, the workloadID -> ContainerID, (workloadID is left blank)
For libnetwork, the workloadID is libnetwork(Pod + containerID are blank)
For everything else, the workloadID == workloadID in v3 (Pod + containerID are blank)

robbrockbank: if it proves to be a big lump of work then don't bother - but in that case we should error if the orchestrator is anything other than "k8s"

lib/upgrade/etcd/conversionv1v3/workloadendpoint.go
```diff
@@ -0,0 +1,258 @@
...
+			IPv6Nets:         ipv6Nets,
+			IPv4NAT:          ipv4NAT,
+			IPv6NAT:          ipv6NAT,
+			IPv4Gateway:      ah.Spec.IPv4Gateway,
+			IPv6Gateway:      ah.Spec.IPv6Gateway,
+			Ports:            ports,
+		},
+		Revision: ah.Metadata.Revision,
+	}
+
```
gunjan5: can you add a log line at the end of both conversions so we know what it converted to and from?

lib/upgrade/etcd/conversionv1v3/workloadendpoint.go
```diff
@@ -0,0 +1,258 @@
...
+	return v3Profiles
+}
+
+func getPodName(workload string, namespace string) string {
+	if strings.HasPrefix(workload, namespace) {
+		parts := strings.Split(workload, ".")
+		pod := strings.Join(parts[1:], "")
+		return pod
+	} else {
```
gunjan5: don't need the `else`

lib/upgrade/etcd/conversionv1v3/workloadendpoint_test.go
```diff
@@ -0,0 +1,396 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package conversionv1v3
+
+import (
+	"testing"
+
+	. "github.com/onsi/gomega"
+	n "net"
```
gunjan5: we usually import golang's net as it is and projectcalico net package as `cnet`

lib/upgrade/etcd/conversionv1v3/workloadendpoint.go
heschlie: @gunjan5 I added some debug logs that just print the struct out, is that sufficient?

heschlie:
I pulled in the name conversion stuff from @robbrockbank and removed mine. Rob I took a swing at the orchestrator logic, might still not be 100% there.

lib/upgrade/etcd/conversionv1v3/workloadendpoint.go
gunjan5: looks good! I think we should make them all uniform (for all the converters), but that can be done later

lib/upgrade/etcd/conversionv1v3/workloadendpoint.go
```diff
@@ -0,0 +1,276 @@
...
+			IPv4NAT:          ipv4NAT,
+			IPv6NAT:          ipv6NAT,
+			IPv4Gateway:      ah.Spec.IPv4Gateway,
+			IPv6Gateway:      ah.Spec.IPv6Gateway,
+			Ports:            ports,
+		},
+		Revision: ah.Metadata.Revision,
+	}
+
+	log.Debugf("Converted: %+v\n To: %+v", ah, d)
```
caseydavenport: @gunjan5 you've asked a few people to do this, they all look different.

Can you take an action to standardize the log format? Really thing these logs should live in the calling code anyway - that way we only need to write it ONCE.

lib/upgrade/etcd/conversionv1v3/workloadendpoint.go
```diff
@@ -0,0 +1,276 @@
...
+	return &d, nil
+}
+
+// BackendV1ToAPIV3 converts v1 WorkloadEndpoint KVPair to v3 API.
+func (_ WorkloadEndpoint) BackendV1ToAPIV3(kvp *model.KVPair) (Resource, error) {
+	wepKey, ok := kvp.Key.(model.WorkloadEndpointKey)
+	if !ok {
+		return nil, fmt.Errorf("value is not a valid WorkloadEndpoint resource key")
+	}
+	wepValue, ok := kvp.Value.(model.WorkloadEndpoint)
```
caseydavenport: This needs to be `*model.WorkloadEndpoint`

caseydavenport: https://github.com/projectcalico/libcalico-go/pull/699/files#diff-2a00b82daa68b03a73f32cdbd44596acR78

lib/upgrade/etcd/conversionv1v3/workloadendpoint.go
```diff
@@ -0,0 +1,276 @@
...
+
+	var pod string
+	var container string
+	var workload string
+
+	// Populate our values based on the orchestrator.
+	switch wepKey.OrchestratorID {
+	case "k8s":
+		namespace = labels["projectcalico.org/namespace"]
+		workload = convertName(wepKey.WorkloadID)
```
caseydavenport: For a k8s Pod, the name is going to be of the form `namespace.name`.

So, we need to do a split here before converting the name in order to split out the two different parts. We should then use the `namespace` prefix as the value for namespace, and the name as the workload name.

lib/upgrade/etcd/conversionv1v3/workloadendpoint.go
```diff
@@ -0,0 +1,276 @@
...
+		pod = getPodName(workload, namespace)
+	case "cni":
+		container = wepValue.ActiveInstanceID
```
caseydavenport: This won't work - we only populate the ActiveInstanceID for Kubernetes weps, so container will always be `""`

lib/upgrade/etcd/conversionv1v3/workloadendpoint.go
```diff
@@ -0,0 +1,276 @@
...
+		workload = convertName(wepKey.WorkloadID)
+	}
+
+	ipNets := convertIPNetworks(wepValue.IPv4Nets)
+	ipNets = append(ipNets, convertIPNetworks(wepValue.IPv6Nets)...)
+
+	ipNats := convertIPNATs(wepValue.IPv4NAT)
+	ipNats = append(ipNats, convertIPNATs(wepValue.IPv6NAT)...)
+
+	wep := apiv3.WorkloadEndpoint{
```
caseydavenport: I think you'll need a `wep := apiv3.NewWorkloadEndpoint()` first, and then modify that one.

lib/upgrade/etcd/conversionv1v3/workloadendpoint.go
```diff
@@ -0,0 +1,276 @@
...
+		ObjectMeta: v1.ObjectMeta{
+			Namespace: namespace,
+			Labels:    labels,
+		},
+		Spec: apiv3.WorkloadEndpointSpec{
+			Orchestrator:  convertName(wepKey.OrchestratorID),
+			Workload:      workload,
+			Node:          convertName(wepKey.Hostname),
```
caseydavenport: CC @robbrockbank another place we're going to need `convertNodeName`

lib/upgrade/etcd/conversionv1v3/workloadendpoint.go
```diff
@@ -0,0 +1,276 @@
...
+			Pod:           pod,
+			ContainerID:   container,
+			Endpoint:      convertName(wepKey.EndpointID),
+			IPNetworks:    ipNets,
+			IPNATs:        ipNats,
+			Profiles:      convertProfiles(wepValue.ProfileIDs),
```
caseydavenport: You should share this with @ozdanborne - he need the `convertProfiles` function for host endpoints as well.

lib/upgrade/etcd/conversionv1v3/workloadendpoint.go
```diff
@@ -0,0 +1,276 @@
...
+			InterfaceName: wepValue.Name,
+			MAC:           wepValue.Mac.String(),
```
caseydavenport: Mac is a pointer - will this work on a pointer that is nil?

lib/upgrade/etcd/conversionv1v3/workloadendpoint.go
```diff
@@ -0,0 +1,276 @@
...
+
+	return ipNATs
+}
+
+// convertProfiles updates the Kubernetes namespace portion from "k8s_ns" to "kns" for each profile.
+func convertProfiles(v1Profiles []string) []string {
+	var v3Profiles []string
+	for _, p := range v1Profiles {
+		if strings.HasPrefix(p, "k8s_ns.") {
+			parts := strings.Split(p, ".")
```
caseydavenport: Rather than use `Split`, I'd recommend `TrimPrefix` -  I think it's safer in this scenario.


Or use `SplitN`

lib/upgrade/etcd/conversionv1v3/workloadendpoint.go
```diff
@@ -0,0 +1,276 @@
...
+	}
+
+	return v3Profiles
+}
+
+// getPodName pulls the pod name out for k8s based installs, the pod name is mixed in with the
+// WorkloadID as `namespace.pod_name` so we separate them out.
+func getPodName(workload string, namespace string) string {
+	if strings.HasPrefix(workload, namespace) {
+		parts := strings.Split(workload, ".")
```
caseydavenport: Just use `strings.SplitN(workload, 2)` - it will only split along the first dot.

lib/upgrade/etcd/conversionv1v3/workloadendpoint.go
```diff
@@ -0,0 +1,276 @@
...
+			v3Profiles = append(v3Profiles, p)
+		}
+	}
+
+	return v3Profiles
+}
+
+// getPodName pulls the pod name out for k8s based installs, the pod name is mixed in with the
+// WorkloadID as `namespace.pod_name` so we separate them out.
+func getPodName(workload string, namespace string) string {
```
caseydavenport: I think this should be `getPodNamespaceName` and it should return both, rather than take the namespace as an argument.

Just split after the first dot.

lib/upgrade/etcd/conversionv1v3/workloadendpoint_test.go
```diff
@@ -0,0 +1,393 @@
...
+	{
+		description: "fully populated WEP",
+		v1API: apiv1.WorkloadEndpoint{
+			Metadata: apiv1.WorkloadEndpointMetadata{
+				Name:             "eth0",
+				Workload:         "default.frontend-5gs43",
+				Orchestrator:     "k8s",
+				Node:             "TestNode",
+				ActiveInstanceID: "1337495556942031415926535",
+				Labels:           makeLabelsV1(),
```
caseydavenport: I recommend adding a test without any labels.

A "minimalistic" test, if you will - only the bare minimum, just in case such a wep exists.

lib/upgrade/etcd/conversionv1v3/workloadendpoint.go
```diff
@@ -0,0 +1,272 @@
...
+
+	var pod string
+	var container string
+	var workload string
+
+	// Populate our values based on the orchestrator.
+	switch wepKey.OrchestratorID {
+	case "k8s":
+		namespace, pod = getPodNamespaceName(wepKey.WorkloadID)
+		workload = convertName(wepKey.WorkloadID)
```
caseydavenport: I think for k8s, we want workload to be an empty string, so just remove this line.

lib/upgrade/etcd/conversionv1v3/workloadendpoint.go
```diff
@@ -0,0 +1,271 @@
...
+			InternalIP: ipNAT.IntIP.String(),
+			ExternalIP: ipNAT.ExtIP.String(),
+		})
+	}
+
+	return ipNATs
+}
+
+// convertProfiles updates the Kubernetes namespace portion from "k8s_ns" to "kns" for each profile.
+func convertProfiles(v1Profiles []string) []string {
```
ozdanborne: I think this function should also pass the names through `convertName`. Old profile names pass through `namespacedNameRegex` which allows the `/` character. But the new names don't allow it.

PR #701, opened 2017-11-29
Convert hostendpoints

ozdanborne:
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/numorstring/protocol.go
+++ b/lib/numorstring/protocol.go
@@ -48,6 +48,24 @@ func ProtocolFromInt(p uint8) Protocol {
 	)
 }
 
+// ProtocolV3FromProtocolV1 creates a v3 Protocol from a v1 Protocol,
+// while handling case conversion.
+func ProtocolV3FromProtocolV1(p Protocol) Protocol {
+	if p.Type == NumOrStringNum {
+		return p
+	}
+
+	for _, n := range allProtocolNames {
+		if strings.ToLower(n) == strings.ToLower(p.StrVal) {
+			return Protocol(
+				Uint8OrString{Type: NumOrStringString, StrVal: n},
+			)
+		}
+	}
+
+	return p
+}
+
 // ProtocolFromString creates a Protocol struct from a string value.
 func ProtocolFromString(p string) Protocol {
 	for _, n := range allProtocolNames {
--- a/lib/upgrade/etcd/conversionv1v3/hostendpoint.go
+++ b/lib/upgrade/etcd/conversionv1v3/hostendpoint.go
@@ -1,0 +1,134 @@
+// Copyright (c) 2017 Tigera, Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package conversionv1v3
+
+import (
+	"fmt"
+
+	log "github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	apiv1 "github.com/projectcalico/libcalico-go/lib/apis/v1"
+	"github.com/projectcalico/libcalico-go/lib/apis/v1/unversioned"
+	apiv3 "github.com/projectcalico/libcalico-go/lib/apis/v3"
+	"github.com/projectcalico/libcalico-go/lib/backend/model"
+	cnet "github.com/projectcalico/libcalico-go/lib/net"
+	"github.com/projectcalico/libcalico-go/lib/numorstring"
+)
+
+// HostEndpoint implements the Converter interface.
+type HostEndpoint struct{}
+
+// APIV1ToBackendV1 converts an APIv1 HostEndpoint structure to a KVPair containing a
+// backend HostEndpoint and HostEndpointKey.
+// This is part of the converter interface.
+func (_ HostEndpoint) APIV1ToBackendV1(a unversioned.Resource) (*model.KVPair, error) {
+	ah, ok := a.(apiv1.HostEndpoint)
+	if !ok {
+		return nil, fmt.Errorf("value is not a valid v1 HostEndpoint")
+	}
+	var ipv4Addrs []cnet.IP
+	var ipv6Addrs []cnet.IP
+	for _, ip := range ah.Spec.ExpectedIPs {
+		if ip.Version() == 4 {
+			ipv4Addrs = append(ipv4Addrs, ip)
+		} else {
+			ipv6Addrs = append(ipv6Addrs, ip)
+		}
+	}
+
+	var ports []model.EndpointPort
+	for _, port := range ah.Spec.Ports {
+		ports = append(ports, model.EndpointPort{
+			Name:     port.Name,
+			Protocol: port.Protocol,
+			Port:     port.Port,
+		})
+	}
+
+	d := model.KVPair{
+		Key: model.HostEndpointKey{
+			Hostname:   ah.Metadata.Node,
+			EndpointID: ah.Metadata.Name,
+		},
+		Value: &model.HostEndpoint{
+			Labels:            ah.Metadata.Labels,
+			Name:              ah.Spec.InterfaceName,
+			ProfileIDs:        ah.Spec.Profiles,
+			ExpectedIPv4Addrs: ipv4Addrs,
+			ExpectedIPv6Addrs: ipv6Addrs,
+			Ports:             ports,
+		},
+	}
+
+	log.WithFields(log.Fields{
+		"v1HostEndpoint": a,
+		"KVPair":         d,
+	}).Debug("Converted HostEndpoint to KVPair")
+
+	return &d, nil
+}
+
+// BackendV1ToAPIV3 converts a KVPair containing a backend HostEndpoint and HostEndpointKey
+// to an APIv3 HostEndpoint structure.
+// This is part of the Converter interface.
+func (_ HostEndpoint) BackendV1ToAPIV3(d *model.KVPair) (Resource, error) {
+	bh, ok := d.Value.(*model.HostEndpoint)
+	if !ok {
+		return nil, fmt.Errorf("value is not a valid HostEndpoint value")
+	}
+	bk, ok := d.Key.(model.HostEndpointKey)
+	if !ok {
+		return nil, fmt.Errorf("key is not a valid HostEndpoint key")
+	}
+
+	var ips []string
+	for _, ip := range bh.ExpectedIPv4Addrs {
+		ips = append(ips, ip.String())
+	}
+	for _, ip := range bh.ExpectedIPv6Addrs {
+		ips = append(ips, ip.String())
+	}
+
+	var ports []apiv3.EndpointPort
+	for _, port := range bh.Ports {
+		ports = append(ports, apiv3.EndpointPort{
+			Name:     port.Name,
+			Protocol: numorstring.ProtocolV3FromProtocolV1(port.Protocol),
+			Port:     port.Port,
+		})
+	}
+
+	ah := &apiv3.HostEndpoint{
+		ObjectMeta: v1.ObjectMeta{
+			Name:   convertName(fmt.Sprintf("%s.%s", bk.Hostname, bk.EndpointID)),
+			Labels: bh.Labels,
+		},
+		Spec: apiv3.HostEndpointSpec{
+			Node:          convertName(bk.Hostname),
+			Ports:         ports,
+			InterfaceName: bh.Name,
+			Profiles:      convertProfiles(bh.ProfileIDs),
+			ExpectedIPs:   ips,
+		},
+	}
+
+	log.WithFields(log.Fields{
+		"KVPair":         d,
+		"v3HostEndpoint": ah,
+	}).Debug("Converted KVPair to v3 Resource")
+
+	return ah, nil
+}
--- a/lib/upgrade/etcd/conversionv1v3/hostendpoint_test.go
+++ b/lib/upgrade/etcd/conversionv1v3/hostendpoint_test.go
@@ -1,0 +1,284 @@
+// Copyright (c) 2017 Tigera, Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package conversionv1v3
+
+import (
+	"fmt"
+	"testing"
+
+	. "github.com/onsi/gomega"
+	"k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	apiv1 "github.com/projectcalico/libcalico-go/lib/apis/v1"
+	"github.com/projectcalico/libcalico-go/lib/apis/v1/unversioned"
+	apiv3 "github.com/projectcalico/libcalico-go/lib/apis/v3"
+	"github.com/projectcalico/libcalico-go/lib/backend/model"
+	cnet "github.com/projectcalico/libcalico-go/lib/net"
+	"github.com/projectcalico/libcalico-go/lib/numorstring"
+)
+
+var hepTable = []struct {
+	description string
+	v1API       unversioned.Resource
+	v1KVP       *model.KVPair
+	v3API       apiv3.HostEndpoint
+}{
+	{
+		description: "Valid basic v1 hep has data moved to right place",
+		v1API: apiv1.HostEndpoint{
+			Metadata: apiv1.HostEndpointMetadata{
+				Name: "my-hep",
+				Node: "my-node",
+			},
+			Spec: apiv1.HostEndpointSpec{
+				InterfaceName: "eth3",
+			},
+		},
+		v1KVP: &model.KVPair{
+			Key: model.HostEndpointKey{
+				EndpointID: "my-hep",
+				Hostname:   "my-node",
+			},
+			Value: &model.HostEndpoint{
+				Name: "eth3",
+			},
+		},
+		v3API: apiv3.HostEndpoint{
+			ObjectMeta: v1.ObjectMeta{
+				Name: "my-node.my-hep",
+			},
+			Spec: apiv3.HostEndpointSpec{
+				InterfaceName: "eth3",
+				Node:          "my-node",
+			},
+		},
+	},
+	{
+		description: "Valid filled v1 hep has data moved to right place",
+		v1API: apiv1.HostEndpoint{
+			Metadata: apiv1.HostEndpointMetadata{
+				Name: "my-hep",
+				Node: "my-node",
+				Labels: map[string]string{
+					"foo": "bar",
+				},
+			},
+			Spec: apiv1.HostEndpointSpec{
+				ExpectedIPs: []cnet.IP{
+					cnet.MustParseIP("192.168.1.1"),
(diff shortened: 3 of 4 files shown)


lib/upgrade/etcd/conversionv1v3/hostendpoint_test.go
```diff
@@ -0,0 +1,153 @@
...
+			},
+			Value: &model.HostEndpoint{
+			},
+		},
+		v3API: apiv3.HostEndpoint{
+			ObjectMeta: v1.ObjectMeta{
+				Name: hepName,
+			},
+			Spec: apiv3.HostEndpointSpec{
+				ExpectedIPs: []string{},
```
ozdanborne: Is this ok? I'm finding its required because of how I initialize the IPs as an empty array here: https://github.com/projectcalico/libcalico-go/pull/701/files#diff-61a4c1332180370f6ebb2f38bf4d8542R86

lib/upgrade/etcd/conversionv1v3/hostendpoint_test.go
```diff
@@ -0,0 +1,153 @@
...
+				Name: hepName,
+			},
+			Spec: apiv3.HostEndpointSpec{
+				InterfaceName: ifName,
+				Node: nodeName,
+				Profiles: []string{profileName},
+				ExpectedIPs: []string{ip},
+				Ports: []apiv3.EndpointPort{
+					{
+						Protocol: numorstring.ProtocolFromStringV1(portProtocol),
```
ozdanborne: I'm guessing I need to convert from `ProtocolFromStringV1` to `ProtocolFromString`. Could use some advice here

lib/upgrade/etcd/conversionv1v3/hostendpoint_test.go
```diff
@@ -0,0 +1,153 @@
...
+	"github.com/projectcalico/libcalico-go/lib/apis/v1/unversioned"
+	apiv3 "github.com/projectcalico/libcalico-go/lib/apis/v3"
+	"github.com/projectcalico/libcalico-go/lib/backend/model"
+	//"github.com/projectcalico/libcalico-go/lib/ipip"
+	cnet "github.com/projectcalico/libcalico-go/lib/net"
+	"k8s.io/apimachinery/pkg/apis/meta/v1"
+	"github.com/projectcalico/libcalico-go/lib/numorstring"
+)
+
+const hepName = "my-hep"
```
gunjan5: why const? consts are untyped (kinda  like a parallel type system), I think it's better to use `var`

lib/upgrade/etcd/conversionv1v3/hostendpoint_test.go
gunjan5: yeah we'll need to handle that in the conversion logic, so it'll be ProtocolFromString here when that's done

lib/numorstring/protocol.go
```diff
@@ -48,6 +48,23 @@ func ProtocolFromInt(p uint8) Protocol {
...
+
+	for _, n := range allProtocolNames {
+		if strings.ToLower(n) == strings.ToLower(p.StrVal) {
+			return Protocol(
+				Uint8OrString{Type: NumOrStringString, StrVal: n},
+			)
+		}
+	}
+
+	// Unknown protocol. TODO: return error?
```
gunjan5: I think we're fine, no need to return the error, should've been caught by the v1 validator

lib/numorstring/protocol.go
```diff
@@ -48,6 +48,23 @@ func ProtocolFromInt(p uint8) Protocol {
 	)
 }
 
+func ProtocolV3FromProtocolV1(p Protocol) Protocol {
```
gunjan5: nit: comment for the exported function

lib/upgrade/etcd/conversionv1v3/hostendpoint.go
```diff
@@ -0,0 +1,120 @@
...
+
+import (
+	"k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	apiv1 "github.com/projectcalico/libcalico-go/lib/apis/v1"
+	"github.com/projectcalico/libcalico-go/lib/apis/v1/unversioned"
+	apiv3 "github.com/projectcalico/libcalico-go/lib/apis/v3"
+	"github.com/projectcalico/libcalico-go/lib/backend/model"
+	cnet "github.com/projectcalico/libcalico-go/lib/net"
+	"fmt"
```
gunjan5: before Rob says it, you should group your imports in:
1. builtins
2. 3rd party
3. projectcalico
separated by a newline

gunjan5: FYI goimports won't group them like that

lib/upgrade/etcd/conversionv1v3/hostendpoint.go
```diff
@@ -0,0 +1,120 @@
...
+	"github.com/projectcalico/libcalico-go/lib/numorstring"
+)
+
+type HostEndpoint struct{}
+
+// convertAPIToKVPair converts an API HostEndpoint structure to a KVPair containing a
+// backend HostEndpoint and HostEndpointKey.
+// This is part of the conversionHelper interface.
```
gunjan5: where is the `conversionHelper` interface?

lib/upgrade/etcd/conversionv1v3/hostendpoint.go
```diff
@@ -0,0 +1,120 @@
...
+			ipv4Addrs = append(ipv4Addrs, ip)
+		} else {
+			ipv6Addrs = append(ipv6Addrs, ip)
+		}
+	}
+
+	var ports []model.EndpointPort
+	for _, port := range ah.Spec.Ports {
+		ports = append(ports, model.EndpointPort{
+			Name:     port.Name, // name
```
gunjan5: did you mean to leave the validation tags in?

lib/upgrade/etcd/conversionv1v3/hostendpoint.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	return &d, nil
+}
+
+// convertKVPairToAPI converts a KVPair containing a backend HostEndpoint and HostEndpointKey
+// to an API HostEndpoint structure.
+// This is part of the conversionHelper interface.
+func (_ HostEndpoint) BackendV1ToAPIV3(d *model.KVPair) (Resource, error) {
+	bh := d.Value.(*model.HostEndpoint)
```
gunjan5: we should make sure the interface assertion doesn't panic, so something like:
```go
if bh, ok := d.Value.(*model.HostEndpoint); !ok {
     return nil, fmt.Errorf("value is not a valid HostEndpoint resource")
}
```

lib/upgrade/etcd/conversionv1v3/hostendpoint.go
```diff
@@ -0,0 +1,120 @@
...
+	bk := d.Key.(model.HostEndpointKey)
```
gunjan5: same here

lib/upgrade/etcd/conversionv1v3/hostendpoint.go
```diff
@@ -0,0 +1,120 @@
...
+			Protocol: numorstring.ProtocolV3FromProtocolV1(port.Protocol), // protocol X
+			Port:     port.Port,
+		})
+	}
+
+	var pIds []string
+	for _, pId := range bh.ProfileIDs {
+		pIds = append(pIds, convertName(pId))
+	}
+	v3Name := convertName(fmt.Sprintf("%s.%s", bk.Hostname, bk.EndpointID))
```
gunjan5: why even store it in a variable?

lib/upgrade/etcd/conversionv1v3/hostendpoint_test.go
```diff
@@ -0,0 +1,223 @@
...
+		},
+		v3API: apiv3.HostEndpoint{
+			ObjectMeta: v1.ObjectMeta{
+				Name: fmt.Sprintf("%s.%s", "my-node", "my-hep"),
+			},
+			Spec: apiv3.HostEndpointSpec{
+				InterfaceName: "eth3",
+				Node: "my-node",
+				Profiles: []string{"my-prof"},
+				ExpectedIPs: []string{"192.168.1.1"},
```
gunjan5: you could add an IPv6 address in this list also since you're combining v4 and v6 while converting to v3

lib/upgrade/etcd/conversionv1v3/hostendpoint.go
```diff
@@ -0,0 +1,120 @@
...
+			Labels: ah.Metadata.Labels,
+
+			Name:              ah.Spec.InterfaceName,
+			ProfileIDs:        ah.Spec.Profiles,
+			ExpectedIPv4Addrs: ipv4Addrs,
+			ExpectedIPv6Addrs: ipv6Addrs,
+			Ports:             ports,
+		},
+	}
+
```
gunjan5: also, can you add a log line at the end of both conversions so we know what it converted to and from

lib/numorstring/protocol.go
```diff
@@ -48,6 +48,23 @@ func ProtocolFromInt(p uint8) Protocol {
 	)
 }
 
+// ProtocolFromString creates a v3 Protocol from a v1 Protocol.
```
gunjan5: wrong function name

lib/upgrade/etcd/conversionv1v3/hostendpoint.go
```diff
@@ -0,0 +1,136 @@
...
+import (
+	"fmt"
+
+	apiv1 "github.com/projectcalico/libcalico-go/lib/apis/v1"
+	"github.com/projectcalico/libcalico-go/lib/apis/v1/unversioned"
+	apiv3 "github.com/projectcalico/libcalico-go/lib/apis/v3"
+	"github.com/projectcalico/libcalico-go/lib/backend/model"
+	cnet "github.com/projectcalico/libcalico-go/lib/net"
+	"github.com/projectcalico/libcalico-go/lib/numorstring"
+	log "github.com/sirupsen/logrus"
```
gunjan5: before Rob yells, quickly move the 3rd party imports after the built-ins :)

lib/upgrade/etcd/conversionv1v3/hostendpoint.go
```diff
@@ -0,0 +1,136 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type HostEndpoint struct{}
+
+// APIV1ToBackendV1 converts an APIv1 HostEndpoint structure to a KVPair containing a
+// backend HostEndpoint and HostEndpointKey.
+// This is part of the converter interface.
```
gunjan5: nit: I think this line should be for the `type HostEndpoint struct{}` since it implements the Converter interface. Don't care if this stays here tho

lib/upgrade/etcd/conversionv1v3/hostendpoint.go
```diff
@@ -0,0 +1,136 @@
...
+			ExpectedIPv6Addrs: ipv6Addrs,
+			Ports:             ports,
+		},
+	}
+
+	log.WithFields(log.Fields{
+		"Name":       ah.Spec.InterfaceName,
+		"Hostname":   ah.Metadata.Node,
+		"EndpointID": ah.Metadata.Name,
+	}).Debug("Converted HostEndpoint to KVPair")
```
gunjan5: we need the whole converted value in the debug logs to see what was the input and what was the converted output, which can be useful if the migration south

lib/upgrade/etcd/conversionv1v3/hostendpoint.go
```diff
@@ -0,0 +1,136 @@
...
+	log "github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type HostEndpoint struct{}
+
+// APIV1ToBackendV1 converts an APIv1 HostEndpoint structure to a KVPair containing a
+// backend HostEndpoint and HostEndpointKey.
+// This is part of the converter interface.
+func (h HostEndpoint) APIV1ToBackendV1(a unversioned.Resource) (*model.KVPair, error) {
```
gunjan5: why did we change this to `h` from `_`? dont see it being used anywhere in this method

lib/upgrade/etcd/conversionv1v3/hostendpoint.go
```diff
@@ -0,0 +1,136 @@
...
+// to an APIv3 HostEndpoint structure.
+// This is part of the Converter interface.
+func (_ HostEndpoint) BackendV1ToAPIV3(d *model.KVPair) (Resource, error) {
+	bh, ok := d.Value.(*model.HostEndpoint)
+	if !ok {
+		return nil, fmt.Errorf("value is not a valid HostEndpoint resource")
+	}
+	bk, ok := d.Key.(model.HostEndpointKey)
+	if !ok {
+		return nil, fmt.Errorf("key is not a valid HostEndpoint resource")
```
gunjan5: key is not a valid HostEndpoint *key*

lib/upgrade/etcd/conversionv1v3/hostendpoint.go
```diff
@@ -0,0 +1,136 @@
...
+			InterfaceName: bh.Name,
+			Profiles:      pIds,
+			ExpectedIPs:   ips,
+		},
+	}
+
+	log.WithFields(log.Fields{
+		"Name":          ah.ObjectMeta.Name,
+		"InterfaceName": ah.Spec.InterfaceName,
+	}).Debug("Converted KVPair to v3 Resource")
```
gunjan5: same here, I think we need the whole input and converted output for better debugging

lib/upgrade/etcd/conversionv1v3/hostendpoint_test.go
```diff
@@ -0,0 +1,223 @@
...
+		},
+		v3API: apiv3.HostEndpoint{
+			ObjectMeta: v1.ObjectMeta{
+				Name: fmt.Sprintf("%s.%s", "my-node", "my-hep"),
+			},
+			Spec: apiv3.HostEndpointSpec{
+				InterfaceName: "eth3",
+				Node:          "my-node",
+				Profiles:      []string{"my-prof"},
+				ExpectedIPs:   []string{"192.168.1.1"},
```
gunjan5: should add an IPv6 to this list to get some coverage to the code that combines IPv4s and v6s

lib/upgrade/etcd/conversionv1v3/hostendpoint.go
ozdanborne: Looks like goimports puts them back here. So we go against goimports with this?

gunjan5: goimport won't put them back if you leave a newline in between

lib/numorstring/protocol.go
```diff
@@ -48,6 +48,24 @@ func ProtocolFromInt(p uint8) Protocol {
 	)
 }
 
+// ProtocolV3FromProtocolV1 creates a v3 Protocol from a v1 Protocol,
```
caseydavenport: Should be fine, but looks like duplicate code from @gunjan5's PR.

https://github.com/projectcalico/libcalico-go/pull/702/files#diff-02e818a8ad4f5514992c9563a94c08b9R74

lib/upgrade/etcd/conversionv1v3/hostendpoint.go
```diff
@@ -0,0 +1,139 @@
...
+	for _, pId := range bh.ProfileIDs {
+		pIds = append(pIds, convertName(pId))
+	}
+	ah := &apiv3.HostEndpoint{
+		ObjectMeta: v1.ObjectMeta{
+			Name:   convertName(fmt.Sprintf("%s.%s", bk.Hostname, bk.EndpointID)),
+			Labels: bh.Labels,
+		},
+		Spec: apiv3.HostEndpointSpec{
+			Node:          convertName(bk.Hostname),
```
caseydavenport: @robbrockbank as per the previous discussions, do we need / have a `convertNodeName` instead?

lib/upgrade/etcd/conversionv1v3/hostendpoint.go
```diff
@@ -0,0 +1,139 @@
...
+		ports = append(ports, apiv3.EndpointPort{
+			Name:     port.Name,
+			Protocol: numorstring.ProtocolV3FromProtocolV1(port.Protocol),
+			Port:     port.Port,
+		})
+	}
+
+	var pIds []string
+	for _, pId := range bh.ProfileIDs {
+		pIds = append(pIds, convertName(pId))
```
caseydavenport: Strictly speaking, we should have a `convertProfileName`.

We need this for `WorkloadEndpoints` and it probably should be done for `HostEndpoints` as well.

The main thing is that in v1, we create per-namespace profiles called `k8s_ns.<namespace>` and in v3 the equivalent is `kns.<namespace>`.  We need to make sure we're handling that transition here.

caseydavenport: CC @robbrockbank in case it's not on your radar.

lib/upgrade/etcd/conversionv1v3/hostendpoint_test.go
```diff
@@ -0,0 +1,229 @@
...
+				Node:          "my-node",
+			},
+		},
+	},
+	{
+		description: "Valid filled v1 hep has data moved to right place",
+		v1API: apiv1.HostEndpoint{
+			Metadata: apiv1.HostEndpointMetadata{
+				Name: "my-hep",
+				Node: "my-node",
```
caseydavenport: What about labels?

lib/upgrade/etcd/conversionv1v3/hostendpoint_test.go
```diff
@@ -0,0 +1,229 @@
...
+			},
+			Spec: apiv1.HostEndpointSpec{
+				ExpectedIPs:   []cnet.IP{
+					cnet.MustParseIP("192.168.1.1"),
+					cnet.MustParseIP("fe80::"),
+				},
+				Profiles:      []string{"my-prof"},
```
caseydavenport: Would appreciate a `k8s_ns.default` profile in this list, pretty please.

lib/upgrade/etcd/conversionv1v3/hostendpoint_test.go
```diff
@@ -0,0 +1,229 @@
...
+				InterfaceName: "eth3",
+				Ports: []apiv1.EndpointPort{
+					{
```
caseydavenport: Could you add a couple more ports:
- One with only a protocol
- One with only a Port number

lib/numorstring/protocol.go
gunjan5: this one came first, I'll remove mine after this one is merged

lib/upgrade/etcd/conversionv1v3/hostendpoint_test.go
ozdanborne: Are these just supposed to check that we don't lose the data when only one field is specified?

lib/upgrade/etcd/conversionv1v3/hostendpoint.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	d := model.KVPair{
+		Key: model.HostEndpointKey{
+			Hostname:   ah.Metadata.Node,
+			EndpointID: ah.Metadata.Name,
+		},
+		Value: &model.HostEndpoint{
+			Labels: ah.Metadata.Labels,
+
```
caseydavenport: extra newline?

lib/upgrade/etcd/conversionv1v3/hostendpoint.go
caseydavenport: I think you need to use this function: https://github.com/projectcalico/libcalico-go/blob/master/lib/upgrade/etcd/conversionv1v3/workloadendpoint.go#L238 

As well as the name conversion.

lib/upgrade/etcd/conversionv1v3/hostendpoint_test.go
```diff
@@ -0,0 +1,267 @@
...
+			},
+		},
+		v3API: apiv3.HostEndpoint{
+			ObjectMeta: v1.ObjectMeta{
+				Name: fmt.Sprintf("%s.%s", "my-node", "my-hep"),
+			},
+			Spec: apiv3.HostEndpointSpec{
+				InterfaceName: "eth3",
+				Node:          "my-node",
+				Profiles:      []string{"k8s_ns.default"},
```
caseydavenport: This should expect to be `kns.default`

PR #702, opened 2017-11-29
Add v1 to v3 conversion code and tests for Policy to GNP

gunjan5:
## Description

~still working on adding some more tests.~

## Todos
- [x] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/upgrade/etcd/conversionv1v3/policy.go
+++ b/lib/upgrade/etcd/conversionv1v3/policy.go
@@ -1,0 +1,142 @@
+// Copyright (c) 2017 Tigera, Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package conversionv1v3
+
+import (
+	"fmt"
+	"strings"
+
+	log "github.com/sirupsen/logrus"
+
+	apiv1 "github.com/projectcalico/libcalico-go/lib/apis/v1"
+	"github.com/projectcalico/libcalico-go/lib/apis/v1/unversioned"
+	apiv3 "github.com/projectcalico/libcalico-go/lib/apis/v3"
+	"github.com/projectcalico/libcalico-go/lib/backend/model"
+)
+
+// Policy implements the Converter interface.
+type Policy struct{}
+
+// APIV1ToBackendV1 converts v1 Policy API to v1 Policy KVPair.
+func (_ Policy) APIV1ToBackendV1(a unversioned.Resource) (*model.KVPair, error) {
+	ap := a.(apiv1.Policy)
+
+	d := model.KVPair{
+		Key: model.PolicyKey{
+			Name: ap.Metadata.Name,
+		},
+		Value: &model.Policy{
+			Order:          ap.Spec.Order,
+			InboundRules:   rulesAPIV1ToBackend(ap.Spec.IngressRules),
+			OutboundRules:  rulesAPIV1ToBackend(ap.Spec.EgressRules),
+			Selector:       ap.Spec.Selector,
+			DoNotTrack:     ap.Spec.DoNotTrack,
+			Annotations:    ap.Metadata.Annotations,
+			PreDNAT:        ap.Spec.PreDNAT,
+			ApplyOnForward: ap.Spec.ApplyOnForward,
+			Types:          nil, // filled in below
+		},
+	}
+
+	if len(ap.Spec.Types) == 0 {
+		// Default the Types field according to what inbound and outbound rules are present
+		// in the policy.
+		if len(ap.Spec.EgressRules) == 0 {
+			// Policy has no egress rules, so apply this policy to ingress only.  (Note:
+			// intentionally including the case where the policy also has no ingress
+			// rules.)
+			d.Value.(*model.Policy).Types = []string{string(apiv1.PolicyTypeIngress)}
+		} else if len(ap.Spec.IngressRules) == 0 {
+			// Policy has egress rules but no ingress rules, so apply this policy to
+			// egress only.
+			d.Value.(*model.Policy).Types = []string{string(apiv1.PolicyTypeEgress)}
+		} else {
+			// Policy has both ingress and egress rules, so apply this policy to both
+			// ingress and egress.
+			d.Value.(*model.Policy).Types = []string{string(apiv1.PolicyTypeIngress), string(apiv1.PolicyTypeEgress)}
+		}
+	} else {
+		// Convert from the API-specified Types.
+		d.Value.(*model.Policy).Types = make([]string, len(ap.Spec.Types))
+		for i, t := range ap.Spec.Types {
+			d.Value.(*model.Policy).Types[i] = string(t)
+		}
+	}
+
+	log.WithFields(log.Fields{
+		"APIv1":  ap,
+		"KVPair": d,
+	}).Debugf("Converted Policy '%s' V1 API to V1 backend", ap.Metadata.Name)
+
+	return &d, nil
+}
+
+// BackendV1ToAPIV3 converts v1 Policy KVPair to v3 API.
+func (_ Policy) BackendV1ToAPIV3(kvp *model.KVPair) (Resource, error) {
+	bp, ok := kvp.Value.(*model.Policy)
+	if !ok {
+		return nil, fmt.Errorf("value is not a valid Policy resource")
+	}
+	bk, ok := kvp.Key.(model.PolicyKey)
+	if !ok {
+		return nil, fmt.Errorf("value is not a valid Policy resource key")
+	}
+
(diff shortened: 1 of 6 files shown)


lib/upgrade/etcd/conversionv1v3/policy.go
```diff
@@ -0,0 +1,130 @@
...
+			d.Value.(*model.Policy).Types = []string{string(apiv1.PolicyTypeIngress), string(apiv1.PolicyTypeEgress)}
+		}
+	} else {
+		// Convert from the API-specified Types.
+		d.Value.(*model.Policy).Types = make([]string, len(ap.Spec.Types))
+		for i, t := range ap.Spec.Types {
+			d.Value.(*model.Policy).Types[i] = string(t)
+		}
+	}
+
```
gunjan5: TODO: add a log line

lib/upgrade/etcd/conversionv1v3/rule.go
```diff
@@ -0,0 +1,266 @@
...
+
+// rulesV1BackendToV3API converts a Backend Rule structure slice to an API Rule structure slice.
+func rulesV1BackendToV3API(brs []model.Rule) []apiv3.Rule {
+	if brs == nil {
+		return nil
+	}
+
+	ars := make([]apiv3.Rule, len(brs))
+	for idx, br := range brs {
+		ars[idx] = ruleBackendToAPI(br)
```
caseydavenport: `rulebackendToAPIv3`?

lib/upgrade/etcd/conversionv1v3/rule.go
```diff
@@ -0,0 +1,266 @@
...
+	}
+
+	if ar.NotICMP != nil {
+		notICMPCode = ar.NotICMP.Code
+		notICMPType = ar.NotICMP.Type
+	}
+
+	if ar.Source.Net != nil || ar.Source.NotNet != nil ||
+		ar.Destination.Net != nil || ar.Destination.NotNet != nil {
+		logDeprecationOnce.Do(func() {
```
caseydavenport: Does this matter to log out? We're going to convert to the right thing anyway, so not really important, right?

lib/upgrade/etcd/conversionv1v3/rule.go
```diff
@@ -0,0 +1,266 @@
...
+	//}
+	// That rule will be converted to the following for Calico v3 API:
+	// apiv3.Rule{
+	//	Action: "allow",
+	//	Source: apiv3.EntityRule{
+	//		Selector: "(label1 == 'value1' || make == 'cake') && tag1 == ''",
+	//	},
+	//}
+	srcSelector := br.SrcSelector
+	if br.SrcTag != "" {
```
caseydavenport: This isn't quite right - try writing a test for a rule that looks like this (no selector):

```go
apiv1.rule{
  Action: "allow",
  Source: apiv1.EntityRule{Tag: "tag1"}
}
```

I think you'll find the result to have this:

```
Selector: () && tag1 == ""
```

When it should be this:

```
Selector: tag1 == ""
```

This applies for all of the following cases (dstSelector, notSrcSelector, etc) as well.

lib/upgrade/etcd/conversionv1v3/rule.go
```diff
@@ -0,0 +1,266 @@
...
+	if br.NotSrcTag != "" {
+		notSrcSelector = fmt.Sprintf("(%s) && %s == ''", br.NotSrcSelector, br.NotSrcTag)
+	}
+
+	notDstSelector := br.NotDstSelector
+	if br.NotDstTag != "" {
+		notDstSelector = fmt.Sprintf("(%s) && %s == ''", br.NotDstSelector, br.NotDstTag)
+	}
+
+	v3Protocol := numorstring.ProtocolV3FromProtocolV1(*br.Protocol)
```
caseydavenport: nil pointer dereference much?

lib/upgrade/etcd/conversionv1v3/rule.go
```diff
@@ -0,0 +1,266 @@
...
+			NotNets:     notDstNetsStr,
+			NotSelector: notDstSelector,
+			NotPorts:    br.NotDstPorts,
+		},
+	}
+}
+
+// ruleActionAPIToBackend converts the rule action field value from the API
+// value to the equivalent backend value.
+func ruleActionAPIToBackend(action string) string {
```
caseydavenport: I think you need to add in version info to all of these - is this v1API to v1Backend?

lib/upgrade/etcd/conversionv1v3/rule.go
```diff
@@ -0,0 +1,266 @@
...
+	if action == "Pass" {
```
caseydavenport: I think this needs a `strings.ToLower(action)`

lib/upgrade/etcd/conversionv1v3/rule.go
```diff
@@ -0,0 +1,266 @@
...
+		return "next-tier"
+	}
+	return action
+}
+
+// ruleActionBackendToAPI converts the rule action field value from the backend
```
caseydavenport: comment doesn't match function

lib/upgrade/etcd/conversionv1v3/rule.go
```diff
@@ -0,0 +1,266 @@
...
+		return apiv3.Pass
+	} else {
+		for _, action := range []apiv3.Action{apiv3.Allow, apiv3.Deny, apiv3.Log, apiv3.Pass} {
+			if strings.ToLower(inAction) == strings.ToLower(string(action)) {
+				return action
+			}
+		}
+	}
+
+	return apiv3.Action(inAction)
```
caseydavenport: When will we hit this return? Should it be an error instead? Or at least a big warning log?

lib/upgrade/etcd/conversionv1v3/policy.go
```diff
@@ -0,0 +1,130 @@
...
+	return &d, nil
+}
+
+// BackendV1ToAPIV3 converts v1 Policy KVPair to v3 API.
+func (_ Policy) BackendV1ToAPIV3(kvp *model.KVPair) (Resource, error) {
+	bp, ok := kvp.Value.(*model.Policy)
+	if !ok {
+		return nil, fmt.Errorf("value is not a valid Policy resource")
```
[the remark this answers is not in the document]
gunjan5: I think the calling code should log the kvp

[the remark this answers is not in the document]
gunjan5: @robbrockbank ^

lib/upgrade/etcd/conversionv1v3/policy_test.go
```diff
@@ -0,0 +1,297 @@
...
+				Name: "namyMcPolicyName",
+			},
+			Spec: apiv1.PolicySpec{
+				Order:          &order1,
+				IngressRules:   []apiv1.Rule{V1InRule1, V1InRule2},
+				EgressRules:    []apiv1.Rule{V1EgressRule1, V1EgressRule2},
+				Selector:       "thing == 'value'",
+				DoNotTrack:     true,
+				PreDNAT:        true,
+				ApplyOnForward: true,
```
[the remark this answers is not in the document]
gunjan5: I have a test case for that line 134

lib/upgrade/etcd/conversionv1v3/rule.go
```diff
@@ -0,0 +1,266 @@
...
+	apiv3 "github.com/projectcalico/libcalico-go/lib/apis/v3"
+	"github.com/projectcalico/libcalico-go/lib/backend/model"
+	"github.com/projectcalico/libcalico-go/lib/net"
+	"github.com/projectcalico/libcalico-go/lib/numorstring"
+)
+
+// rulesAPIToBackend converts an API Rule structure slice to a Backend Rule structure slice.
+func rulesAPIToBackend(ars []apiv1.Rule) []model.Rule {
+	if ars == nil {
+		return []model.Rule{}
```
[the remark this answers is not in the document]
gunjan5: this is how all our v1 backend values were stored in the backend (empty value, instead of nil), maybe it had something to do with how v1 etcd backend could unmarshal it?

lib/upgrade/etcd/conversionv1v3/rule.go
gunjan5: ok, I'll remove it

lib/upgrade/etcd/conversionv1v3/rule.go
gunjan5: yup, you're totally right. I've fixed it

lib/upgrade/etcd/conversionv1v3/rule.go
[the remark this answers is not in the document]
caseydavenport: ok, happy to keep, but keep an eye out for it.

lib/upgrade/etcd/conversionv1v3/policy.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+	bk, ok := kvp.Key.(model.PolicyKey)
+	if !ok {
+		return nil, fmt.Errorf("value is not a valid Policy resource key")
+	}
+
```
caseydavenport: I believe the plan is that we are NOT going to convert any `knp.` policies, instead relying on the new controller to do that.

@gunjan5 can you add a check here that returns an error if a `knp.` prefixed policy is provided?

caseydavenport: Ok, so the wrapper code will convert them but throw them away?

I guess it's fine - just an extra precaution.  I'm fine to leave as-is.

PR #703, opened 2017-11-29
Conversion of Bgppeer from v1 to v3

tmjd:
- [ ] I'm not sure on the conversion of IP to name.
  - Do IPv4 addresses need any expansion/padding?
  - Are any addresses stored that won't be the length of IPv4 or IPv6? If so then currently it would be turned into a hex string (with no punctuation).
- [ ] Is the name conversion the proper one? I think so, we would want to preserve any `.` in the node names I believe, is that also ok for the name of the BGPPeer also?

```release-note
None required
```

### DIFF
--- a/lib/upgrade/etcd/conversionv1v3/bgppeer.go
+++ b/lib/upgrade/etcd/conversionv1v3/bgppeer.go
@@ -1,0 +1,120 @@
+// Copyright (c) 2017 Tigera, Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package conversionv1v3
+
+import (
+	"fmt"
+
+	log "github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	apiv1 "github.com/projectcalico/libcalico-go/lib/apis/v1"
+	"github.com/projectcalico/libcalico-go/lib/apis/v1/unversioned"
+	apiv3 "github.com/projectcalico/libcalico-go/lib/apis/v3"
+	"github.com/projectcalico/libcalico-go/lib/backend/model"
+	"github.com/projectcalico/libcalico-go/lib/errors"
+	"github.com/projectcalico/libcalico-go/lib/scope"
+)
+
+// BGPPeer implements the Converter interface.
+type BGPPeer struct{}
+
+// APIV1ToBackendV1 converts v1 BGPPeer API to v1 BGPPeer KVPair.
+func (bp BGPPeer) APIV1ToBackendV1(rIn unversioned.Resource) (*model.KVPair, error) {
+	ap, ok := rIn.(apiv1.BGPPeer)
+	if !ok {
+		return nil, fmt.Errorf("Conversion to BGPPeer is not possible with %v", rIn)
+	}
+
+	if len(ap.Metadata.PeerIP.IP) == 0 {
+		return nil, fmt.Errorf("no PeerIP is set, invalid BGPPeer: %v", rIn)
+	}
+
+	k, err := bp.convertMetadataToKey(ap.Metadata)
+	if err != nil {
+		return nil, err
+	}
+
+	d := model.KVPair{
+		Key: k,
+		Value: &model.BGPPeer{
+			PeerIP: ap.Metadata.PeerIP,
+			ASNum:  ap.Spec.ASNumber,
+		},
+	}
+
+	log.WithFields(log.Fields{
+		"APIV1":  rIn,
+		"KVPair": d,
+	}).Debug("Converted BGPPeer")
+
+	return &d, nil
+}
+
+func (_ BGPPeer) convertMetadataToKey(bpm apiv1.BGPPeerMetadata) (model.Key, error) {
+	if bpm.Scope == scope.Global {
+		if bpm.Node != "" {
+			return nil, fmt.Errorf("With Global scope having a Node is invalid: %v", bpm)
+		}
+		return model.GlobalBGPPeerKey{
+			PeerIP: bpm.PeerIP,
+		}, nil
+	} else if bpm.Scope == scope.Node {
+		if bpm.Node == "" {
+			return nil, fmt.Errorf("With Node scope a Node must be defined: %v", bpm)
+		}
+		return model.NodeBGPPeerKey{
+			PeerIP:   bpm.PeerIP,
+			Nodename: bpm.Node,
+		}, nil
+	} else {
+		return nil, errors.ErrorInsufficientIdentifiers{
+			Name: "scope",
+		}
+	}
+}
+
+// BackendV1ToAPIV3 converts v1 BGPPeer KVPair to v3 API.
+func (bp BGPPeer) BackendV1ToAPIV3(kvp *model.KVPair) (Resource, error) {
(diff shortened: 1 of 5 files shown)


lib/upgrade/etcd/conversionv1v3/bgppeer.go
```diff
@@ -0,0 +1,131 @@
...
+		},
+	}
+
+	switch kvp.Key.(type) {
+	case model.GlobalBGPPeerKey:
+		r.ObjectMeta = v1.ObjectMeta{Name: bp.ipToName(peer.PeerIP.IP)}
+	case model.NodeBGPPeerKey:
+		nk := kvp.Key.(model.NodeBGPPeerKey)
+		n := convertName(nk.Nodename)
+		r.Spec.Node = n
```
gunjan5: this is in Spec, so I don't think we'll need the converted name here. We convert names for ObjectMeta name because of the stricter (k8s) regex, but not really the case for Spec.

robbrockbank: no - we should convert - we want the Node name to match the name of the Node resource.

lib/upgrade/etcd/conversionv1v3/bgppeer.go
```diff
@@ -0,0 +1,131 @@
...
+		r.ObjectMeta = v1.ObjectMeta{Name: n + "." + bp.ipToName(peer.PeerIP.IP)}
```
gunjan5: we had `Scope` as one of the keys in v1, so each resource would have `Scope` as part of its unique key. I'm wondering if ignoring the `Scope` in the name could cause a collision

robbrockbank: Should be fine - if the scope wasn't node then we won't include the node in the name (which is separated by a "." rather than the "-" separator used for the converted IP address.

lib/upgrade/etcd/conversionv1v3/bgppeer.go
```diff
@@ -0,0 +1,131 @@
...
+	default:
+		return nil, fmt.Errorf("Invalid key for BGPPeer: %v", kvp.Key)
+	}
+
+	return r, nil
+}
+
+func (_ BGPPeer) ipToName(ip net.IP) string {
+	name := ""
+	if len(ip) == net.IPv6len {
```
gunjan5: I think it's better to use `if ip.To4() != nil {` to say if it's an IPv4 or IPv6

lib/upgrade/etcd/conversionv1v3/bgppeer.go
```diff
@@ -0,0 +1,131 @@
...
+		name = fmt.Sprintf("%02x%02x", ip[0], ip[1])
+		for i := 2; i < net.IPv6len; i = i + 2 {
```
gunjan5: you can just do this:
```go
     ip6 := ip.To16()
     bytes := []string{}
     for i := 0; i < len(ip6); i += 2 {
			bytes = append(bytes, fmt.Sprintf("%.2x%.2x", ip6[i], ip6[i+1]))
	}

     name = strings.Join(bytes, "-")
```

lib/upgrade/etcd/conversionv1v3/bgppeer.go
```diff
@@ -0,0 +1,131 @@
...
+	"github.com/projectcalico/libcalico-go/lib/errors"
+	"github.com/projectcalico/libcalico-go/lib/scope"
+)
+
+// BGPPeer implements the Converter interface.
+type BGPPeer struct{}
+
+// APIV1ToBackendV1 converts v1 BGPPeer API to v1 BGPPeer KVPair.
+func (bp BGPPeer) APIV1ToBackendV1(rIn unversioned.Resource) (*model.KVPair, error) {
+	ap := rIn.(apiv1.BGPPeer)
```
gunjan5: I know I haven't done this in the IPPool example, but we should make sure the interface assertion works else it could panic. So like:
```go
ap, ok := rIn.(apiv1.BGPPeer)
if !ok {
 log.Fatalf(...)
}
```

lib/upgrade/etcd/conversionv1v3/bgppeer.go
```diff
@@ -0,0 +1,131 @@
...
+			PeerIP: ap.Metadata.PeerIP,
+			ASNum:  ap.Spec.ASNumber,
+		},
+	}
+
+	return &d, nil
+}
+
+func (_ BGPPeer) convertMetadataToKey(rmIn unversioned.ResourceMetadata) (model.Key, error) {
+	pm := rmIn.(apiv1.BGPPeerMetadata)
```
gunjan5: same here

lib/upgrade/etcd/conversionv1v3/bgppeer.go
```diff
@@ -0,0 +1,131 @@
...
+		r.Spec.Node = n
+		r.ObjectMeta = v1.ObjectMeta{Name: n + "." + bp.ipToName(peer.PeerIP.IP)}
+	default:
+		return nil, fmt.Errorf("Invalid key for BGPPeer: %v", kvp.Key)
+	}
+
+	return r, nil
+}
+
+func (_ BGPPeer) ipToName(ip net.IP) string {
```
gunjan5: this doesn't need to be a method, it can be a function. Also, I think we should move this to `names.go` so it can be used by others as well. I was going to add something similar to IPPool CIDR expansion and name conversion

lib/upgrade/etcd/conversionv1v3/bgppeer.go
tmjd: Ok, I wasn't sure if the extraction of an IPv4 from one embedded in an IPv6 address would be ok.  If that is ok then sounds good.

lib/upgrade/etcd/conversionv1v3/bgppeer.go
tmjd: Sounds good, I had thought about that but didn't add it in since it wasn't in the example.

lib/upgrade/etcd/conversionv1v3/bgppeer.go
tmjd: I'll do that, I had made it a method to make sure it didn't conflict with any other conversions.

lib/upgrade/etcd/conversionv1v3/bgppeer.go
tmjd: This combined with the suggestion of using `ip.To4()` to select how the address will be parsed, will result in any address that is not an IPv4 or IPv6 being converted to an empty string. I think this is ok because I don't think we would expect anything else but I just wanted to point out that difference to make sure you agree that will be ok.

lib/upgrade/etcd/conversionv1v3/bgppeer.go
tmjd: Did you specifically want a log.Fatal and not return an error?

gunjan5: I meant return error, sorry

lib/upgrade/etcd/conversionv1v3/bgppeer.go
```diff
@@ -0,0 +1,131 @@
...
+		Value: &model.BGPPeer{
+			PeerIP: ap.Metadata.PeerIP,
+			ASNum:  ap.Spec.ASNumber,
+		},
+	}
+
+	return &d, nil
+}
+
+func (_ BGPPeer) convertMetadataToKey(rmIn unversioned.ResourceMetadata) (model.Key, error) {
```
robbrockbank: you could just make this interface typed so you don't have to cast it - since it's only called from the one place.

lib/upgrade/etcd/conversionv1v3/bgppeer.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	d := model.KVPair{
+		Key: k,
+		Value: &model.BGPPeer{
+			PeerIP: ap.Metadata.PeerIP,
+			ASNum:  ap.Spec.ASNumber,
+		},
+	}
+
```
gunjan5: also, can you add a log line at the end of both conversions so we know what it converted to and from

PR #706, opened 2017-11-29
Fix WEP and watcher bug.

robbrockbank:
## Description
-  Fix KDD WEP name handling for Get:  Ensure WEP name is correctly formatted and includes the Pod.
-  Handle clientv3 watch channel closure by terminating the watch when the results channel is closed.

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/backend/k8s/resources/workloadendpoint.go
+++ b/lib/backend/k8s/resources/workloadendpoint.go
@@ -129,6 +129,12 @@ func (c *WorkloadEndpointClient) Get(ctx context.Context, key model.Key, revisio
 	if err != nil {
 		return nil, err
 	}
+	if wepID.Pod == "" {
+		return nil, cerrors.ErrorResourceDoesNotExist{
+			Identifier: key,
+			Err: errors.New("malformed WorkloadEndpoint name - unable to determine Pod name"),
+		}
+	}
 
 	pod, err := c.clientSet.CoreV1().Pods(k.Namespace).Get(wepID.Pod, metav1.GetOptions{ResourceVersion: revision})
 	if err != nil {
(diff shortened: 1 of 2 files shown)


lib/backend/k8s/resources/workloadendpoint.go
```diff
@@ -126,6 +126,12 @@ func (c *WorkloadEndpointClient) Get(ctx context.Context, key model.Key, revisio
 	if err != nil {
 		return nil, err
 	}
+	if wepID.Pod == "" {
```
caseydavenport: Surely this needs to be `wepIP.Pod == "" && wepID.Workload == ""`? For non-k8s the pod will be `""`

lib/clientv3/resources.go
```diff
@@ -347,7 +347,11 @@ func (w *watcher) run() {
 
 	for {
 		select {
-		case event := <-w.backend.ResultChan():
+		case event, ok := <-w.backend.ResultChan():
+			if !ok {
+				log.Info("Watcher results channel closed by remote")
```
caseydavenport: should this be debug?

lib/backend/k8s/resources/workloadendpoint.go
caseydavenport: oops, nvm. This is in the k8s backend.

lib/clientv3/resources.go
robbrockbank: Hmmm, ordinarily I'd say this feels like an info log - we wouldn't generally expect the remote to terminate the watcher suddenly.  That said, we know the CRD-backed resources terminate every minute of inactivity -- so that makes it more borderline for me.

I'll change if you have a strong preference - just let me know @caseydavenport

caseydavenport: Given this happens a few times every minute, I think it should be debug.

PR #708, opened 2017-11-29
Basic support for http rules in v3 policy.

saumoh:
## Description
Add http rules to v3 policy. Add a feature flag that goes along with it.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/apis/v3/policy.go
+++ b/lib/apis/v3/policy.go
@@ -57,6 +57,15 @@ type Rule struct {
 	Source EntityRule `json:"source,omitempty" validate:"omitempty"`
 	// Destination contains the match criteria that apply to destination entity.
 	Destination EntityRule `json:"destination,omitempty" validate:"omitempty"`
+
+	// HTTP contains match criteria that apply to HTTP requests.
+	HTTP HTTPRule `json:"http,omitempty" validate:"omitempty"`
+}
+
+type HTTPRule struct {
+	// Methods is an optional field that restricts the rule to apply only to HTTP requests that use one of the listed
+	// HTTP Methods (e.g. GET, PUT, etc.)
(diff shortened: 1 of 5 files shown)


lib/apis/v3/policy.go
```diff
@@ -57,6 +57,15 @@ type Rule struct {
 	Source EntityRule `json:"source,omitempty" validate:"omitempty"`
 	// Destination contains the match criteria that apply to destination entity.
 	Destination EntityRule `json:"destination,omitempty" validate:"omitempty"`
+
+	// HTTP contains match criteria that apply to HTTP requests.
+	HTTP HTTPRule `json:"http,omitempty" validate:"omitempty"`
```
robbrockbank: Should this be a struct pointer?   I think you'll get "http: {}" output otherwise - even if this feature is not enabled.

saumoh: @robbrockbank good point. I've made the change. thx

lib/validator/v3/alphafeaturecheck.go
```diff
@@ -39,3 +39,21 @@ func ValidateNoServiceAccountRules(ingress, egress []apiv3.Rule) error {
 
 	return nil
 }
+
+// ValidateNoHTTPRules checks if the set of rules have the
+// serviceAccount match set and if yes then it returns an error.
+func ValidateNoHTTPRules(ingress, egress []apiv3.Rule) error {
+	for _, rule := range ingress {
+		if rule.HTTP != nil {
+			return errors.New("alpha feature HTTP Rules used")
```
robbrockbank: Fine for now, but there is a special validation error type that we should be using.

PR #709, opened 2017-11-30
Switch to an annotation when storing CNI-assigned IPs to k8s pods.

fasaxc:
## Todos
- [x] Tests 
  - [x] UT
  - [x] Live test that it actually fixes the problem.
- [x] Documentation
- [x] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/backend/k8s/conversion/conversion.go
+++ b/lib/backend/k8s/conversion/conversion.go
@@ -195,6 +195,9 @@ func (c Converter) PodToWorkloadEndpoint(pod *kapiv1.Pod) (*model.KVPair, error)
 	// An IP address may not yet be assigned (or may have been removed for a Pod deletion), so
 	// handle a missing IP gracefully.
 	ipNets, err := c.GetPodIPs(pod)
+	if err != nil {
+		return nil, err
+	}
 
 	// Generate the interface name based on workload.  This must match
 	// the host-side veth configured by the CNI plugin.
--- a/lib/backend/k8s/conversion/conversion.go
+++ b/lib/backend/k8s/conversion/conversion.go
@@ -140,7 +140,28 @@ func (c Converter) IsHostNetworked(pod *kapiv1.Pod) bool {
 }
 
 func (c Converter) HasIPAddress(pod *kapiv1.Pod) bool {
-	return pod.Status.PodIP != ""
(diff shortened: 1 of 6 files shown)


lib/backend/k8s/conversion/conversion.go
```diff
@@ -173,15 +194,7 @@ func (c Converter) PodToWorkloadEndpoint(pod *kapiv1.Pod) (*model.KVPair, error)
...
-	ipNets := []string{}
-	if c.HasIPAddress(pod) {
-		_, ipNet, err := cnet.ParseCIDROrIP(pod.Status.PodIP)
-		if err != nil {
-			log.WithFields(log.Fields{"ip": pod.Status.PodIP, "pod": pod.Name}).WithError(err).Error("Failed to parse pod IP")
-			return nil, err
-		}
-		ipNets = append(ipNets, ipNet.String())
-	}
+	ipNets, err := c.GetPodIPs(pod)
```
caseydavenport: Doesn't look like we're handling this error any more.

lib/backend/k8s/k8s_test.go
```diff
@@ -1199,6 +1199,120 @@ var _ = Describe("Test Syncer API for Kubernetes backend", func() {
 		})
 	})
 
+	It("should handle patching a Pod with our IP address annotation", func() {
+		pod := &k8sapi.Pod{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "test-syncer-basic-pod",
+				Namespace: "default",
```
caseydavenport: Probably worth also having a test with pre-existing annotations, or racing requests to update annotations, to make sure we've got our patch logic right (e.g. we don't overwrite other annotations).

fasaxc: Added a test with another annotation.  I experimented a bit with races yesterday.  I found that even an `UpdateStatus()` can clobber the annotation we add if you don't include the revision.  If you include a revision then the request is rejected because the patch makes your copy of the data stale.  I worked that behaviour into the test.

Upshot is, if we race with someone using the API correctly then our patch makes their request fail.  If they're using it incorrectly and not supplying a revision then they do clobber us.

Guess we could do a test of multiple patches being applied at the same time but that really would just be testing the API server applies patches correctly.

fasaxc:
Fixes https://github.com/projectcalico/calico/issues/1405

PR #714, opened 2017-12-01
conversion v1 to v3 node

tmjd:
```release-note
None required
```

### DIFF
--- a/lib/upgrade/etcd/conversionv1v3/node.go
+++ b/lib/upgrade/etcd/conversionv1v3/node.go
@@ -1,0 +1,151 @@
+// Copyright (c) 2017 Tigera, Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package conversionv1v3
+
+import (
+	"fmt"
+
+	log "github.com/sirupsen/logrus"
+
+	apiv1 "github.com/projectcalico/libcalico-go/lib/apis/v1"
+	"github.com/projectcalico/libcalico-go/lib/apis/v1/unversioned"
+	apiv3 "github.com/projectcalico/libcalico-go/lib/apis/v3"
+	"github.com/projectcalico/libcalico-go/lib/backend/model"
+	cnet "github.com/projectcalico/libcalico-go/lib/net"
+)
+
+type Node struct{}
+
+// convertKVPairToAPI converts a KVPair containing a backend Node and NodeKey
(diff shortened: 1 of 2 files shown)


lib/upgrade/etcd/conversionv1v3/node.go
```diff
@@ -0,0 +1,151 @@
...
+		}
+
+		// If the backend has an IPv4 address then fill in the IPv4Address
+		// field.  If the IP network does not exist assume a full mask.
+		if bv.BGPIPv4Addr != nil {
+			if bv.BGPIPv4Net != nil {
+				// Stored network is normalised, so copy across the
+				// IP separately.
+				apiNode.Spec.BGP.IPv4Address = bv.BGPIPv4Net.String()
+				//apiNode.Spec.BGP.IPv4Address.IP = bv.BGPIPv4Addr.IP
```
caseydavenport: Commented out code still here.

lib/upgrade/etcd/conversionv1v3/node.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	bk, ok := d.Key.(model.NodeKey)
+	if !ok {
+		return nil, fmt.Errorf("Key is not a valid NodeKey resource: %v", d.Value)
+	}
+
+	apiNode := apiv3.NewNode()
+
+	apiNode.ObjectMeta.Name = convertName(bk.Hostname)
```
caseydavenport: We'll need to use the `convertNodeName` function when it exists.

lib/upgrade/etcd/conversionv1v3/node.go
```diff
@@ -0,0 +1,151 @@
...
+	"github.com/projectcalico/libcalico-go/lib/backend/model"
+	cnet "github.com/projectcalico/libcalico-go/lib/net"
+)
+
+type Node struct{}
+
+// convertKVPairToAPI converts a KVPair containing a backend Node and NodeKey
+// to an API Node structure.
+// This is part of the conversionHelper interface.
+func (_ Node) BackendV1ToAPIV3(d *model.KVPair) (Resource, error) {
```
caseydavenport: What's the plan for populating `Node.Spec.NodeBGPSpec.IPv4IPIPTunnelAddr`?

@tmjd @robbrockbank?

robbrockbank: That'll be retro-fitted by the upgrade script, but for simple YAML conversion it won't be possible to include that field.

tmjd: I added a comment saying that it needs to be populated external to the function.

lib/upgrade/etcd/conversionv1v3/node.go
tmjd: See if you like what I did with the alias function.

lib/upgrade/etcd/conversionv1v3/names.go
```diff
@@ -66,6 +66,11 @@ func convertNameNoDots(v1Name string) string {
 	return name
 }
 
+// Currently just an alias for convertName but it is expected to be different
```
caseydavenport: I think we probably don't want to do this aliasing - once we have a function we'll go around and find all the places we need to update.

For now, let's just use `convertName` in line.

tmjd:
I think this is ready to go now.

PR #715, opened 2017-12-01
Make name normalization a public method

robbrockbank:
## Description
Make the NameNormalization a public method and use it for the hostname normalization.
Update the name conversion to be deterministic by appending a qualifying hash.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/upgrade/etcd/conversionv1v3/names.go
+++ b/lib/upgrade/etcd/conversionv1v3/names.go
@@ -15,6 +15,8 @@
 package conversionv1v3
 
 import (
+	"crypto/sha1"
+	"encoding/hex"
 	"fmt"
 	"net"
 	"regexp"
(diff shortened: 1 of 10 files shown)


lib/names/normalize.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+var (
+	nonNameChar               = regexp.MustCompile("[^-.a-z0-9]+")
+	dotDashSeq                = regexp.MustCompile("[.-]*[.][.-]*")
+	trailingLeadingDotsDashes = regexp.MustCompile("^[.-]*(.*?)[.-]*$")
+)
+
+// Convert a name to normalized form.  This is used for conversion of os.Hostname to a
+// suitable node name, and is also used for the v2->v3 migration code.
```
caseydavenport: I think this isn't really what we want to do - users need to pass in a valid hostname, and valid hostnames don't contain `/` characters. We shouldn't try to magically convert them away, we should return an error.

lib/names/hostname.go
```diff
@@ -1,16 +1,13 @@
...
+import "os"
 
 // Hostname returns this hosts hostname, converting to lowercase so that
 // it is valid for use in the Calico API.
 func Hostname() (string, error) {
 	if h, err := os.Hostname(); err != nil {
 		return "", err
 	} else {
-		return strings.ToLower(strings.TrimSpace(h)), nil
+		return NormalizeName(h), nil
```
caseydavenport: I think this is the part I'm scared about - this lets users request invalid hostnames, and we'll then scrub them and turn them into something else.

I'd much rather error out on an invalid hostname.

lib/upgrade/etcd/conversionv1v3/names.go
```diff
@@ -15,55 +15,44 @@
...
+	// If the name is different append a qualifier.
+	return qualifiedName(v1Name, name)
+}
+
+func qualifiedName(orig, final string) string {
+	// If the name was not modified, just return the unmodified name.
+	if orig == final {
+		return orig
+	}
+	// The name was modified.  Calculate an 8-byte hex qualifier to append.
```
caseydavenport: Do we have a name character limit? 

> By convention, the names of Kubernetes resources should be up to maximum length of 253 characters

Are we not afraid of going beyond that limit? Suppose we need to handle that case _anyway_ in the migration code... but are we?

lib/names/hostname.go
```diff
@@ -1,16 +1,13 @@
...
-)
+import "os"
 
 // Hostname returns this hosts hostname, converting to lowercase so that
 // it is valid for use in the Calico API.
 func Hostname() (string, error) {
 	if h, err := os.Hostname(); err != nil {
 		return "", err
 	} else {
-		return strings.ToLower(strings.TrimSpace(h)), nil
```
caseydavenport: Wonder if this code actually needs to do smart truncation of FQDNs that are longer that the maximum name length.

lib/upgrade/etcd/conversionv1v3/names.go
robbrockbank: current plan is to run validate over the converted data.   I hand't planned on attempting to fix up over-long names, but rather reject the config and allow the user the opportunity to fix it up.

PR #718, opened 2017-12-02
Add basic selector conversion parsing

caseydavenport:
## Description

A really basic implementation of the selector conversion, to get us unblocked for now.

We may want to consider fortifying it.

Namely, I think `calico/k8s_ns == 'default'` should really resolve to `projectcalico.org/namespace == 'default' && projectcalico.org/orchestrator == 'k8s'`, but I haven't implemented that yet and it's a lot harder.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


lib/upgrade/etcd/conversionv1v3/selectors_test.go
```diff
@@ -0,0 +1,27 @@
+package conversionv1v3
```
gunjan5: need license here

PR #719, opened 2017-12-02
Add v1 to v3 conversion code and tests for Profile

gunjan5:
## Description

A very special PR. Easily hurt. Please handle with care.

## Todos
- [x] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/upgrade/etcd/conversionv1v3/profile.go
+++ b/lib/upgrade/etcd/conversionv1v3/profile.go
@@ -1,0 +1,116 @@
+// Copyright (c) 2017 Tigera, Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package conversionv1v3
+
+import (
+	"fmt"
+	"strings"
+
+	log "github.com/sirupsen/logrus"
+
+	apiv1 "github.com/projectcalico/libcalico-go/lib/apis/v1"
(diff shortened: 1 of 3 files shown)


lib/upgrade/etcd/conversionv1v3/profile.go
```diff
@@ -0,0 +1,106 @@
...
+	if !ok {
+		return nil, fmt.Errorf("value is not a valid Profile resource")
+	}
+	bk, ok := kvp.Key.(model.ProfileKey)
+	if !ok {
+		return nil, fmt.Errorf("value is not a valid Profile resource key")
+	}
+
+	ap := apiv3.NewProfile()
+	ap.Name = convertName(bk.Name)
```
caseydavenport: We have a function specifically for profile names - might need tweaking since it expects a list.

caseydavenport: https://github.com/projectcalico/libcalico-go/blob/master/lib/upgrade/etcd/conversionv1v3/workloadendpoint.go#L238:6

caseydavenport: It doesn't matter too much, given we're going to ignore anything with `k8s_ns` on the front anyway... but for correctness we should use that function.

lib/upgrade/etcd/conversionv1v3/profile_test.go
```diff
@@ -0,0 +1,159 @@
...
+			},
+		},
+		v3API: apiv3.Profile{
+			ObjectMeta: v1.ObjectMeta{
+				Name: "nameymcprofilename",
+			},
+			Spec: apiv3.ProfileSpec{
+				Ingress:       []apiv3.Rule{V3InRule1, V3InRule2},
+				Egress:        []apiv3.Rule{V3EgressRule1, V3EgressRule2},
+				LabelsToApply: map[string]string{"thingy": "val", "meep": "", "mop": ""},
```
caseydavenport: Could you add the prefix `pncs.` to one of these label keys? Just for more realistic labels.

gunjan5:
Seems like @caseydavenport is traveling today, so @robbrockbank can you take a look at the second commit to see if I've addressed Casey's comments?

PR #722, opened 2017-12-06
Add a IsClean() method to the etcdv3 datastore

robbrockbank:
## Description
Add a semi-private "IsClean()" method to the etcdv3 backend client.

If I end up implementing for KDD then I'll make it part of the backend API.

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/backend/etcdv3/etcdv3.go
+++ b/lib/backend/etcdv3/etcdv3.go
@@ -395,13 +395,28 @@ func (c *etcdV3Client) EnsureInitialized() error {
 func (c *etcdV3Client) Clean() error {
 	log.Warning("Cleaning etcdv3 datastore of all Calico data")
 	_, err := c.etcdClient.Txn(context.Background()).If().Then(
-		clientv3.OpDelete("/calico", clientv3.WithPrefix()),
+		clientv3.OpDelete("/calico/", clientv3.WithPrefix()),
 	).Commit()
 
 	if err != nil {
 		return cerrors.ErrorDatastoreError{Err: err}
 	}
 	return nil
+}
+
+// IsClean() returns true if there are no /calico/ prefixed entries in the
+// datastore.  This is not part of the exposed API, but is public to allow
+// direct consumers of the backend API to access this.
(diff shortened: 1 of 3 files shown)


lib/backend/etcdv3/etcdv3.go
```diff
@@ -395,7 +395,7 @@ func (c *etcdV3Client) EnsureInitialized() error {
 func (c *etcdV3Client) Clean() error {
 	log.Warning("Cleaning etcdv3 datastore of all Calico data")
 	_, err := c.etcdClient.Txn(context.Background()).If().Then(
-		clientv3.OpDelete("/calico", clientv3.WithPrefix()),
+		clientv3.OpDelete("/calico/", clientv3.WithPrefix()),
```
gunjan5: Why this change? It's prefix operation so should be fine, right?

lib/backend/etcdv3/etcdv3.go
```diff
@@ -404,6 +404,21 @@ func (c *etcdV3Client) Clean() error {
...
 }
 
+// IsClean() returns true if there are no /calico/ prefixed entries in the
+// datastore.  This is not part of the exposed API, but is public to allow
+// direct consumers of the backend API to access this.
+func (c *etcdV3Client) IsClean() (bool, error) {
+	log.Debug("Calling Get on etcdv3 client")
+	resp, err := c.etcdClient.Get(context.Background(), "/calico/", clientv3.WithPrefix())
+	if err != nil {
+		log.WithError(err).Info("Error returned from etcdv3 client")
```
gunjan5: Shouldn't this be an error log level or a warning at least?

lib/backend/etcdv3/etcdv3.go
robbrockbank: it's strictly more correct since this distinguishes from /calicov2/blah for example.

lib/backend/etcdv3/etcdv3.go
robbrockbank: no - that's up to the end user if they want to log it as an error

generally lower down the stack you keep your levels more as debug or info.

PR #723, opened 2017-12-06
Use %v format for non-strings (and don't have String() method)

neiljerram:
- errors
- KVPair.Key
- Identifier fields in lib/errors/errors.go


neiljerram:
This was prompted specifically by the following problem, seen in KDD/routereflector work:
```
2017-12-05 22:11:19.429 [INFO][16] watchercache.go 156: Failed to perform list of current data during resync ListRoot="/calico/ipam/v2/host/" error=operation List is not supported on { %!s(int=0)}
```

PR #724, opened 2017-12-06
bgpsyncer: only watch affinity blocks for a specific node

neiljerram:


lib/backend/syncersv1/bgpsyncer/bgpsyncer.go
```diff
@@ -57,9 +57,11 @@ func New(client api.Client, callbacks api.SyncerCallbacks, node string, watchAll
 			ListInterface:   model.ResourceListOptions{Kind: apiv3.KindBGPPeer},
 			UpdateProcessor: updateprocessors.NewBGPPeerUpdateProcessor(),
 		},
-		{
+	}
+	if node != "" {
```
robbrockbank: Maybe worth adding a comment here that we only expect node to be blank for the RR and in that case we don't need to monitor the BlockAffinities

neiljerram:
Thanks @robbrockbank I added a comment.  Have I got the details there right?

lib/backend/syncersv1/bgpsyncer/bgpsyncer.go
```diff
@@ -57,9 +57,16 @@ func New(client api.Client, callbacks api.SyncerCallbacks, node string, watchAll
 			ListInterface:   model.ResourceListOptions{Kind: apiv3.KindBGPPeer},
 			UpdateProcessor: updateprocessors.NewBGPPeerUpdateProcessor(),
 		},
-		{
+	}
+	// When this syncer is used (via confd) in calico/node, it needs to know the affinity blocks
```
robbrockbank: the affinity blocks are actually just used to set up the blackhole routes - nothing to do with IPIP.

ozdanborne: addressed

PR #725, opened 2017-12-06
Felix config intervals

### LINKED ISSUE #712: FelixConfig loses precision vs Felix's config parser
Felix supports fractional seconds/millis for `time.Duration` fields but those have been mapped through to `*int`s on the FelixConfig type.

## Expected Behavior

Should be able to set a fractional number of seconds for, say `IptablesPostWriteCheckIntervalSecs`, but it's typed as an int.

1.
2.
3.
4.

## Your Environment

* libcalico-go version:

heschlie:
## Description

Fixes #712 
Changing the type for all of the `FexixConfigurationSpec.*Secs` to be of type `*time.Duration` instead of `*int` to ensure we don't loose precision and to match what Felix uses (as we can do fractional seconds in Felix) and setting `config:"seconds"` to match what Felix expects.

## Todos
- [x] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/apis/v3/felixconfig.go
+++ b/lib/apis/v3/felixconfig.go
@@ -43,43 +43,43 @@ type FelixConfigurationSpec struct {
 	IPv6Support    *bool `json:"ipv6Support,omitempty" confignamev1:"Ipv6Support"`
 	IgnoreLooseRPF *bool `json:"ignoreLooseRPF,omitempty"`
 
-	// RouterefreshInterval is the period, in seconds, at which Felix re-checks the routes
+	// RouterefreshInterval is the period at which Felix re-checks the routes
 	// in the dataplane to ensure that no other process has accidentally broken Calico’s rules.
-	// Set to 0 to disable route refresh. [Default: 90]
-	RouteRefreshIntervalSecs *int `json:"routeRefreshIntervalSecs,omitempty" confignamev1:"RouteRefreshInterval"`
-	// IptablesRefreshInterval is the period, in seconds, at which Felix re-checks the IP sets
+	// Set to 0 to disable route refresh. [Default: 90s]
+	RouteRefreshInterval *metav1.Duration `json:"routeRefreshInterval,omitempty" configv1timescale:"seconds"`
+	// IptablesRefreshInterval is the period at which Felix re-checks the IP sets
 	// in the dataplane to ensure that no other process has accidentally broken Calico’s rules.
 	// Set to 0 to disable IP sets refresh. Note: the default for this value is lower than the
 	// other refresh intervals as a workaround for a Linux kernel bug that was fixed in kernel
 	// version 4.11. If you are using v4.11 or greater you may want to set this to, a higher value
-	// to reduce Felix CPU usage. [Default: 10]
-	IptablesRefreshIntervalSecs *int `json:"iptablesRefreshIntervalSecs,omitempty" confignamev1:"IptablesRefreshInterval"`
-	// IptablesPostWriteCheckIntervalSecs is the period, in seconds, after Felix has done a write
+	// to reduce Felix CPU usage. [Default: 10s]
+	IptablesRefreshInterval *metav1.Duration `json:"iptablesRefreshInterval,omitempty" configv1timescale:"seconds"`
+	// IptablesPostWriteCheckInterval is the period after Felix has done a write
 	// to the dataplane that it schedules an extra read back in order to check the write was not
 	// clobbered by another process. This should only occur if another application on the system
-	// doesn’t respect the iptables lock. [Default: 1]
-	IptablesPostWriteCheckIntervalSecs *int `json:"iptablesPostWriteCheckIntervalSecs,omitempty"`
+	// doesn’t respect the iptables lock. [Default: 1s]
+	IptablesPostWriteCheckInterval *metav1.Duration `json:"iptablesPostWriteCheckInterval,omitempty" configv1timescale:"seconds" confignamev1:"IptablesPostWriteCheckIntervalSecs"`
 	// IptablesLockFilePath is the location of the iptables lock file. You may need to change this
 	// if the lock file is not in its standard location (for example if you have mapped it into Felix’s
 	// container at a different path). [Default: /run/xtables.lock]
 	IptablesLockFilePath string `json:"iptablesLockFilePath,omitempty"`
-	// IptablesLockTimeoutSecs is the time, in seconds, that Felix will wait for the iptables lock,
+	// IptablesLockTimeout is the time that Felix will wait for the iptables lock,
 	// or 0, to disable. To use this feature, Felix must share the iptables lock file with all other
 	// processes that also take the lock. When running Felix inside a container, this requires the
 	// /run directory of the host to be mounted into the calico/node or calico/felix container.
-	// [Default: 0 disabled]
-	IptablesLockTimeoutSecs *int `json:"iptablesLockTimeoutSecs,omitempty"`
-	// IptablesLockProbeIntervalMillis is the time, in milliseconds, that Felix will wait between
+	// [Default: 0s disabled]
+	IptablesLockTimeout *metav1.Duration `json:"iptablesLockTimeout,omitempty" configv1timescale:"seconds" confignamev1:"IptablesLockTimeoutSecs"`
+	// IptablesLockProbeInterval is the time that Felix will wait between
 	// attempts to acquire the iptables lock if it is not available. Lower values make Felix more
-	// responsive when the lock is contended, but use more CPU. [Default: 50]
-	IptablesLockProbeIntervalMillis *int `json:"iptablesLockProbeIntervalMillis,omitempty"`
-	// IpsetsRefreshIntervalSecs is the period, in seconds, at which Felix re-checks all iptables
+	// responsive when the lock is contended, but use more CPU. [Default: 50ms]
+	IptablesLockProbeInterval *metav1.Duration `json:"iptablesLockProbeInterval,omitempty" configv1timescale:"milliseconds" confignamev1:"IptablesLockProbeIntervalMillis"`
+	// IpsetsRefreshInterval is the period at which Felix re-checks all iptables
 	// state to ensure that no other process has accidentally broken Calico’s rules. Set to 0 to
-	// disable iptables refresh. [Default: 90]
-	IpsetsRefreshIntervalSecs *int `json:"ipsetsRefreshIntervalSecs,omitempty" confignamev1:"IpsetsRefreshInterval"`
-	MaxIpsetSize              *int `json:"maxIpsetSize,omitempty"`
+	// disable iptables refresh. [Default: 90s]
+	IpsetsRefreshInterval *metav1.Duration `json:"ipsetsRefreshInterval,omitempty" configv1timescale:"seconds"`
+	MaxIpsetSize          *int `json:"maxIpsetSize,omitempty"`
 
-	NetlinkTimeoutSecs *int `json:"netlinkTimeoutSecs,omitempty"`
+	NetlinkTimeout *metav1.Duration `json:"netlinkTimeout,omitempty" configv1timescale:"seconds" confignamev1:"NetlinkTimeoutSecs"`
 
 	// MetadataAddr is the IP address or domain name of the server that can answer VM queries for
 	// cloud-init metadata. In OpenStack, this corresponds to the machine running nova-api (or in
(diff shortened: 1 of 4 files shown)


lib/backend/syncersv1/updateprocessors/configurationprocessor_test.go
```diff
@@ -206,7 +207,7 @@ var _ = Describe("Test the generic configuration update processor and the concre
 			},
 		}
 		expected := map[string]interface{}{
-			"RouteRefreshInterval":      "12345",
```
fasaxc: Is this the output that felix will see?  Sadly, felix has its own parse for these rather than parsing them as durations. If the felix value is in seconds, then it expects a float `10.5` with no `s` suffix, if the felix value is in millis, then it expects `10500.0`

heschlie: I was wondering about this, I'll see if I can get it to output floats from the converter.

lib/apis/v3/felixconfig.go
```diff
@@ -46,19 +47,19 @@ type FelixConfigurationSpec struct {
 	// RouterefreshInterval is the period, in seconds, at which Felix re-checks the routes
 	// in the dataplane to ensure that no other process has accidentally broken Calico’s rules.
 	// Set to 0 to disable route refresh. [Default: 90]
-	RouteRefreshIntervalSecs *int `json:"routeRefreshIntervalSecs,omitempty" confignamev1:"RouteRefreshInterval"`
+	RouteRefreshIntervalSecs *time.Duration `json:"routeRefreshIntervalSecs,omitempty" config:"seconds" confignamev1:"RouteRefreshInterval"`
```
robbrockbank: None of these fields should have the units suffix (e.g. Secs / Millis) - the time.Duration is already takes care of units for you.

robbrockbank: Also we should use the kubernetes definition of `Duration` in `k8s.io/apimachinery/pkg/apis/meta/v1`

lib/backend/syncersv1/updateprocessors/configurationprocessor_test.go
robbrockbank: I also think it depends on the fields - some are millis and some are s.

@fasaxc :  Are all the duration/interval fields floats, or just the secs ones?  What about the millisecond interval fields?

lib/apis/v3/felixconfig.go
fasaxc: If you change the names, you'll need to fix up the Felix UT here, which asserts that the name is either the same, or in a map of exceptions: https://github.com/projectcalico/felix/blob/master/config/config_params_test.go#L50

fasaxc: There may also be downstream STs that use the new config struct

lib/backend/syncersv1/updateprocessors/configurationprocessor_test.go
fasaxc: @robbrockbank They're all parsed as floats now.

heschlie:
@fasaxc I was pointed to the right spot to coerce these into floats for Felix, we should be set there.

lib/apis/v3/felixconfig.go
```diff
@@ -46,40 +46,40 @@ type FelixConfigurationSpec struct {
...
 	// [Default: 0 disabled]
-	IptablesLockTimeoutSecs *int `json:"iptablesLockTimeoutSecs,omitempty"`
-	// IptablesLockProbeIntervalMillis is the time, in milliseconds, that Felix will wait between
+	IptablesLockTimeout *metav1.Duration `json:"iptablesLockTimeoutSecs,omitempty" configv1timescale:"seconds" confignamev1:"IptablesLockTimeoutSecs"`
+	// IptablesLockProbeInterval is the time, in milliseconds, that Felix will wait between
 	// attempts to acquire the iptables lock if it is not available. Lower values make Felix more
 	// responsive when the lock is contended, but use more CPU. [Default: 50]
-	IptablesLockProbeIntervalMillis *int `json:"iptablesLockProbeIntervalMillis,omitempty"`
-	// IpsetsRefreshIntervalSecs is the period, in seconds, at which Felix re-checks all iptables
+	IptablesLockProbeInterval *metav1.Duration `json:"iptablesLockProbeIntervalMillis,omitempty" configv1timescale:"milliseconds"`
```
fasaxc: Needs a `confignamev1`

lib/backend/syncersv1/updateprocessors/configurationprocessor.go
```diff
@@ -230,6 +231,13 @@ func (c *configUpdateProcessor) processAddOrModified(kvp *model.KVPair) ([]*mode
 				switch vt := value.(type) {
 				case string:
 					value = vt
+				case v1.Duration:
+					switch fieldInfo.Tag.Get("configv1timescale") {
+					case "seconds":
+						value = fmt.Sprintf("%.1f", vt.Seconds())
+					case "milliseconds":
+						value = fmt.Sprintf("%.1f", float64(vt.Duration/time.Millisecond))
+					}
```
fasaxc: Bit worried that it'd be easy to leave this tag out when adding a new field.  Should we default to seconds?

lib/backend/syncersv1/updateprocessors/configurationprocessor.go
```diff
@@ -230,6 +231,13 @@ func (c *configUpdateProcessor) processAddOrModified(kvp *model.KVPair) ([]*mode
 				switch vt := value.(type) {
 				case string:
 					value = vt
+				case v1.Duration:
+					switch fieldInfo.Tag.Get("configv1timescale") {
+					case "seconds":
+						value = fmt.Sprintf("%.1f", vt.Seconds())
```
fasaxc: I think `.1` might be a bit limiting, how about `.3` so that seconds can express millisecond precision and millis can express microseconds.  That should be more than enough.

lib/backend/syncersv1/updateprocessors/configurationprocessor.go
heschlie: I think there is only one field that uses ms, so it would make some sense.

heschlie: Yea good catch.

lib/apis/v3/felixconfig.go
```diff
@@ -46,40 +46,40 @@ type FelixConfigurationSpec struct {
...
+	// IpsetsRefreshInterval is the period, in seconds, at which Felix re-checks all iptables
 	// state to ensure that no other process has accidentally broken Calico’s rules. Set to 0 to
 	// disable iptables refresh. [Default: 90]
-	IpsetsRefreshIntervalSecs *int `json:"ipsetsRefreshIntervalSecs,omitempty" confignamev1:"IpsetsRefreshInterval"`
+	IpsetsRefreshInterval *metav1.Duration `json:"ipsetsRefreshIntervalSecs,omitempty" configv1timescale:"seconds" confignamev1:"IpsetsRefreshInterval"`
```
robbrockbank: Now that the name has reverted back to the original, we no longer need the confignamev1 tag.

lib/apis/v3/felixconfig.go
```diff
@@ -135,14 +135,14 @@ type FelixConfigurationSpec struct {
...
 	IPIPMTU *int `json:"ipipMTU,omitempty" confignamev1:"IpInIpMtu"`
 
-	// ReportingIntervalSecs is the interval at which Felix reports its status into the datastore or 0 to disable.
+	// ReportingInterval is the interval at which Felix reports its status into the datastore or 0 to disable.
 	// Must be non-zero in OpenStack deployments. [Default: 30]
-	ReportingIntervalSecs *int `json:"reportingIntervalSecs,omitempty"`
-	// ReportingTTLSecs is the time-to-live setting for process-wide status reports. [Default: 90]
-	ReportingTTLSecs *int `json:"reportingTTLSecs,omitempty"`
+	ReportingInterval *metav1.Duration `json:"reportingIntervalSecs,omitempty" configv1timescale:"seconds" confignamev1:"ReportingIntervalSecs"`
+	// ReportingTTL is the time-to-live setting for process-wide status reports. [Default: 90]
```
robbrockbank: default is 90s

lib/apis/v3/felixconfig.go
```diff
@@ -46,40 +46,40 @@ type FelixConfigurationSpec struct {
...
+	IptablesLockTimeout *metav1.Duration `json:"iptablesLockTimeoutSecs,omitempty" configv1timescale:"seconds" confignamev1:"IptablesLockTimeoutSecs"`
+	// IptablesLockProbeInterval is the time, in milliseconds, that Felix will wait between
 	// attempts to acquire the iptables lock if it is not available. Lower values make Felix more
 	// responsive when the lock is contended, but use more CPU. [Default: 50]
-	IptablesLockProbeIntervalMillis *int `json:"iptablesLockProbeIntervalMillis,omitempty"`
-	// IpsetsRefreshIntervalSecs is the period, in seconds, at which Felix re-checks all iptables
+	IptablesLockProbeInterval *metav1.Duration `json:"iptablesLockProbeIntervalMillis,omitempty" configv1timescale:"milliseconds"`
+	// IpsetsRefreshInterval is the period, in seconds, at which Felix re-checks all iptables
 	// state to ensure that no other process has accidentally broken Calico’s rules. Set to 0 to
 	// disable iptables refresh. [Default: 90]
```
robbrockbank: Default: 90s

lib/apis/v3/felixconfig.go
robbrockbank: and comment should be Default: 50ms

lib/apis/v3/felixconfig.go
```diff
@@ -46,40 +46,40 @@ type FelixConfigurationSpec struct {
 	// RouterefreshInterval is the period, in seconds, at which Felix re-checks the routes
 	// in the dataplane to ensure that no other process has accidentally broken Calico’s rules.
 	// Set to 0 to disable route refresh. [Default: 90]
-	RouteRefreshIntervalSecs *int `json:"routeRefreshIntervalSecs,omitempty" confignamev1:"RouteRefreshInterval"`
+	RouteRefreshInterval *metav1.Duration `json:"routeRefreshIntervalSecs,omitempty" configv1timescale:"seconds" confignamev1:"RouteRefreshInterval"`
```
robbrockbank: Default: 90s

lib/apis/v3/felixconfig.go
```diff
@@ -46,40 +46,40 @@ type FelixConfigurationSpec struct {
...
 	// IptablesRefreshInterval is the period, in seconds, at which Felix re-checks the IP sets
 	// in the dataplane to ensure that no other process has accidentally broken Calico’s rules.
 	// Set to 0 to disable IP sets refresh. Note: the default for this value is lower than the
 	// other refresh intervals as a workaround for a Linux kernel bug that was fixed in kernel
 	// version 4.11. If you are using v4.11 or greater you may want to set this to, a higher value
 	// to reduce Felix CPU usage. [Default: 10]
-	IptablesRefreshIntervalSecs *int `json:"iptablesRefreshIntervalSecs,omitempty" confignamev1:"IptablesRefreshInterval"`
-	// IptablesPostWriteCheckIntervalSecs is the period, in seconds, after Felix has done a write
+	IptablesRefreshInterval *metav1.Duration `json:"iptablesRefreshIntervalSecs,omitempty" configv1timescale:"seconds" confignamev1:"IptablesRefreshInterval"`
```
robbrockbank: Default 10s

lib/apis/v3/felixconfig.go
```diff
@@ -46,40 +46,40 @@ type FelixConfigurationSpec struct {
...
+	// IptablesPostWriteCheckInterval is the period, in seconds, after Felix has done a write
 	// to the dataplane that it schedules an extra read back in order to check the write was not
 	// clobbered by another process. This should only occur if another application on the system
 	// doesn’t respect the iptables lock. [Default: 1]
-	IptablesPostWriteCheckIntervalSecs *int `json:"iptablesPostWriteCheckIntervalSecs,omitempty"`
+	IptablesPostWriteCheckInterval *metav1.Duration `json:"iptablesPostWriteCheckIntervalSecs,omitempty" configv1timescale:"seconds" confignamev1:"IptablesPostWriteCheckIntervalSecs"`
```
robbrockbank: Default: 1s

lib/apis/v3/felixconfig.go
```diff
@@ -135,14 +135,14 @@ type FelixConfigurationSpec struct {
 	// IPIPMTU is the MTU to set on the tunnel device. See Configuring MTU [Default: 1440]
 	IPIPMTU *int `json:"ipipMTU,omitempty" confignamev1:"IpInIpMtu"`
 
-	// ReportingIntervalSecs is the interval at which Felix reports its status into the datastore or 0 to disable.
+	// ReportingInterval is the interval at which Felix reports its status into the datastore or 0 to disable.
```
robbrockbank: Default 30s

lib/apis/v3/felixconfig.go
robbrockbank: and other defaults can be updated below too.

heschlie:
@fasaxc The build I've posted passes tests locally but Semaphore seems to be having SSL issues. This is ready for another look.

fasaxc:
Reran the tests but they hit a failure this time.  Given you're working in a different area of the codebase, I've raised an issue on the assumption that it's a flake: https://github.com/projectcalico/libcalico-go/issues/732

lib/apis/v3/felixconfig.go
```diff
@@ -45,41 +45,41 @@ type FelixConfigurationSpec struct {
 
 	// RouterefreshInterval is the period, in seconds, at which Felix re-checks the routes
 	// in the dataplane to ensure that no other process has accidentally broken Calico’s rules.
-	// Set to 0 to disable route refresh. [Default: 90]
-	RouteRefreshIntervalSecs *int `json:"routeRefreshIntervalSecs,omitempty" confignamev1:"RouteRefreshInterval"`
+	// Set to 0 to disable route refresh. [Default: 90s]
+	RouteRefreshInterval *metav1.Duration `json:"routeRefreshIntervalSecs,omitempty" configv1timescale:"seconds" confignamev1:"RouteRefreshInterval"`
```
robbrockbank: confignamev1 is not required here anymore.

PR #727, opened 2017-12-07
Error WEP conversion if k8s WorkloadID format is incorrect

robbrockbank:
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/upgrade/etcd/conversionv1v3/workloadendpoint_test.go
+++ b/lib/upgrade/etcd/conversionv1v3/workloadendpoint_test.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package conversionv1v3
+package conversionv1v3_test
 
 import (
 	"testing"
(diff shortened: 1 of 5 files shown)


lib/upgrade/etcd/conversionv1v3/workloadendpoint_test.go
```diff
@@ -343,6 +345,40 @@ func TestCanConvertV1ToV3WorkloadEndpoint(t *testing.T) {
...
+func TestBadK8sWorkloadID(t *testing.T) {
+	t.Run("Test invalid k8s workloadID (no dot in name) fails to convert", func(t *testing.T) {
+		RegisterTestingT(t)
+
+		w := conversionv1v3.WorkloadEndpoint{}
+		wepBackendV1 := &model.KVPair{
+			Key: model.WorkloadEndpointKey{
+				Hostname:       "TestNode",
+				OrchestratorID: "k8s",
+				WorkloadID:     "default/frontend-5gs43",
```
gunjan5: this example has no `.`s, but I'm guessing one with 3 or more `.`s will also fail

robbrockbank: no, >1 dot is fine.   Format is <namespace>.<pod>

Only the namespace cannot have a dot in it, but I believe the pod name can.

gunjan5: ah ok I missed the part where `SplitN` is set to only 2 matches

PR #731, opened 2017-12-11
Normalize source Net, Nets, NotNet, NotNets

gunjan5:
## Description

Destination Net/Nets/NotNet/NotNets is normalized when stored in the backend. Source Net/Nets/NotNet/NotNets should also be normalized. This PR fixes that in both v1 and in for v1->v3 upgrade converters

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/converter/rule.go
+++ b/lib/converter/rule.go
@@ -84,8 +84,8 @@ func ruleAPIToBackend(ar api.Rule) model.Rule {
 		NotICMPType: notICMPType,
 
 		SrcTag:      ar.Source.Tag,
-		SrcNet:      ar.Source.Net,
-		SrcNets:     ar.Source.Nets,
+		SrcNet:      normalizeIPNet(ar.Source.Net),
+		SrcNets:     normalizeIPNets(ar.Source.Nets),
 		SrcSelector: ar.Source.Selector,
 		SrcPorts:    ar.Source.Ports,
 		DstTag:      ar.Destination.Tag,
@@ -95,8 +95,8 @@ func ruleAPIToBackend(ar api.Rule) model.Rule {
 		DstPorts:    ar.Destination.Ports,
 
 		NotSrcTag:      ar.Source.NotTag,
-		NotSrcNet:      ar.Source.NotNet,
-		NotSrcNets:     ar.Source.NotNets,
+		NotSrcNet:      normalizeIPNet(ar.Source.NotNet),
+		NotSrcNets:     normalizeIPNets(ar.Source.NotNets),
 		NotSrcSelector: ar.Source.NotSelector,
 		NotSrcPorts:    ar.Source.NotPorts,
 		NotDstTag:      ar.Destination.NotTag,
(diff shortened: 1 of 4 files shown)


lib/converter/rule.go
```diff
@@ -84,8 +84,8 @@ func ruleAPIToBackend(ar api.Rule) model.Rule {
 		NotICMPType: notICMPType,
 
 		SrcTag:      ar.Source.Tag,
-		SrcNet:      ar.Source.Net,
-		SrcNets:     ar.Source.Nets,
+		SrcNet:      normalizeIPNet(ar.Source.Net),
```
robbrockbank: I guess it doesn't matter too much, but the converter code is now obsolete - so not really necessary to update this - guess it doesn't hurt though.

gunjan5: will still be used if we do a 2.7, right?

lib/converter/rule_test.go
```diff
@@ -30,8 +30,12 @@ var (
 	cidr2     = net.MustParseCIDR("11.0.0.1/24")
 	cidr3     = net.MustParseCIDR("12.0.0.1/24")
 	cidr4     = net.MustParseCIDR("13.0.0.1/24")
+	cidr1Net  = net.MustParseCIDR("10.0.0.0/24")
+	cidr2Net  = net.MustParseCIDR("11.0.0.0/24")
 	cidr3Net  = net.MustParseCIDR("12.0.0.0/24")
 	cidr4Net  = net.MustParseCIDR("13.0.0.0/24")
+	cidr1Norm = (&cidr1Net).Network()
+	cidr2Norm = (&cidr2Net).Network()
```
robbrockbank: Don't know why we have these cidr1-4Norm - they are identical to the cidr1-4Net.

Looks like the tests are wrong I guess.  We should be configuring the API with cidr1-4 and we should validate that the backend model has cidr1-4Net....  and cidr1-4Norm should just be removed.

gunjan5: dunno, I just copied what the other tests did for it, I can change it

lib/upgrade/etcd/conversionv1v3/rule.go
```diff
@@ -77,8 +77,8 @@ func ruleAPIToBackend(ar apiv1.Rule) model.Rule {
 		NotICMPType: notICMPType,
 
 		SrcTag:      ar.Source.Tag,
-		SrcNet:      ar.Source.Net,
-		SrcNets:     ar.Source.Nets,
+		SrcNet:      normalizeIPNet(ar.Source.Net),
```
robbrockbank: If we weren't normalizing the CIDRs here before then there is the possibility that the backend model written to the datastore might be incorrect.  If that is the case then I think we'll also need to normalize these IPNets when converting from the backend to the v3 API.

robbrockbank: Also, there are no tests for this change.

gunjan5: this code is only for the upgrade converters, so I don't see how anyone could have the wrong data in their backend yet. I guess I should change it for v1 bacendToAPI

gunjan5: updated the conversion tests (rules_def.go) to expect normalized nets

gunjan5:
once merged, this should fix parts of https://github.com/projectcalico/calicoctl/pull/1784 and https://github.com/projectcalico/calicoctl/pull/1785

lib/converter/rule.go
robbrockbank: not in master, nope

lib/upgrade/etcd/conversionv1v3/rules_def.go
```diff
@@ -33,9 +33,9 @@ var numProtocol1 = numorstring.ProtocolFromInt(240)
 var icmpType1 = 100
 var icmpCode1 = 200
 
-var cidr1StrictMaskStr = "10.0.0.0/24"
+var cidr1StrictMaskStr = "192.168.0.128/25"
```
robbrockbank: what was the purpose of this change?

gunjan5: `192.168.0.128/25` is strictly masked (normalized). We use this in Profile and Policy conversion tests where `192.168.0.129/25` is passed as APIv1 and we expect this strictly masked out from the v3 output

robbrockbank:
Squash please before merging

Issue #684, opened 2017-11-16
New lower-case name restriction may be incorrect for some uses


fasaxc:
We use the "name" validator in a lot of places in the v3 datamodel and the name now restricts to lower case:

```
v3$ ag -R 'validate.*name'
bgppeer.go
46:	Node string `json:"node,omitempty" validate:"omitempty,name"`

workloadendpoint.go
40:	Orchestrator string `json:"orchestrator,omitempty" validate:"omitempty,name"`
42:	Workload string `json:"workload,omitempty" validate:"omitempty,name"`
44:	Node string `json:"node,omitempty" validate:"omitempty,name"`
46:	ContainerID string `json:"containerID,omitempty" validate:"omitempty,name"`
48:	Pod string `json:"pod,omitempty" validate:"omitempty,name"`
50:	Endpoint string `json:"endpoint,omitempty" validate:"omitempty,name"`
68:	Profiles []string `json:"profiles,omitempty" validate:"omitempty,dive,name"`

hostendpoint.go
44:	Node string `json:"node,omitempty" validate:"omitempty,name"`
61:	Profiles []string `json:"profiles,omitempty" validate:"omitempty,dive,name"`
```

I think at least some of those fields may need a more loose validation:

- `containerID`: the CNI spec just says this will be "plain text"
- `Orchestrator` `Workload` `Pod`, `Endpoint`: not sure these are guaranteed lower-case
- `Profiles`: could be mixed case before v3

caseydavenport:
> containerID: the CNI spec just says this will be "plain text"

Yes - I think this should be more permissive. 

> Orchestrator Workload Pod, Endpoint: not sure these are guaranteed lower-case

Yep. Pod is guaranteed to be lower-case, but the others may have mixed case, though for Workload and Orchestrator we have some control over the values that get put there.

> Profiles: could be mixed case before v3

If you mean the name, then all names for all resources are lowe-case only in Calico v3.

bcreane:
Discussed with @caseydavenport and decided to prioritize mixed-case containerIDs in v3 validation. Goal is to avoid breaking k8s-friendly container orchestrations such as rkt that allow uppercase letters in containerID.

caseydavenport:
To be clear, we haven't checked what values things like rkt and cri-o use for their container IDs.

If we can confirm they pass the validation then I don't think we need to do anything.

CC @gunjan5

bcreane:
[...]

PR #733, opened 2017-12-13
Allow mixed case containerIDs

bcreane:
## Description
In order to accommodate container orchestrators such as `rkt` which allow mixed-cased container names, loosen the containerID validation to do the same.

This addresses one part (`containerID`) of https://github.com/projectcalico/libcalico-go/issues/684.

## Todos
- [X] Tests
- [ ] Documentation - cc @emanic, this may entail some doc changes
- [ ] Release note - since we haven't released v3.0 yet, I think just docs are sufficient

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/apis/v3/workloadendpoint.go
+++ b/lib/apis/v3/workloadendpoint.go
@@ -43,7 +43,7 @@ type WorkloadEndpointSpec struct {
 	// The node name identifying the Calico node instance.
 	Node string `json:"node,omitempty" validate:"omitempty,name"`
 	// The container ID.
-	ContainerID string `json:"containerID,omitempty" validate:"omitempty,name"`
+	ContainerID string `json:"containerID,omitempty" validate:"omitempty,containerID"`
 	// The Pod name.
 	Pod string `json:"pod,omitempty" validate:"omitempty,name"`
 	// The Endpoint name.
--- a/lib/names/workloadendpoint.go
+++ b/lib/names/workloadendpoint.go
@@ -36,9 +36,9 @@ import (
 // will be escaped to a double-dash (--) in the constructed name.
 //
 // List queries allow for prefix lists (for non-KDD), the client should verify that
-// then items returned in the list match the supplied identifiers using the
+// the items returned in the list match the supplied identifiers using the
 // NameMatches() method.  This is necessary because a prefix match may return endpoints
-// that do not exactly match the required identifiers.  For example , suppose you are
+// that do not exactly match the required identifiers.  For example, suppose you are
 // querying endpoints with node=node1, orch=k8s, pod=pod and endpoints is wild carded:
 // -  The name prefix would be `node1-k8s-pod-`
 // -  A list query using that prefix would also return endpoints with, for example,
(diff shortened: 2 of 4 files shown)


caseydavenport:
I think we need to fix `CalculateWorkloadEndpointName` as well: https://github.com/projectcalico/libcalico-go/pull/733/files#diff-7c3de3967905ba9e43f6a8ad8320ebfeR91

WEP names still must be lowercase, so CalculateWorkloadEndpointName needs to properly handle containerIDs that might have upper-case characters in...

gunjan5:
@bcreane I think we can use the validator.v8's build-in UUID field validator. https://godoc.org/gopkg.in/go-playground/validator.v8#hdr-Universally_Unique_Identifier_UUID (just found out about this) That might be more close to the RFC 4122 than the regex we have

gunjan5:
hmmm so the validator they use doesn't allow uppercase https://github.com/go-playground/validator/blob/v8.18.2/regexes.go#L23
`^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$` 

So does the UUID RFC spec: `6.5.4 Software generating the hexadecimal representation of a UUID shall not use upper case letters.`

Which means rkt is testing it with a regex which allows uppercase but the UUID standard doesn't 

hmmmmmmmmmmmmmmmmmmmmmmmmmmmmmm...............

bcreane:
The approach I was thinking of was to validate the containerID in `names.getSegments()` against the restrictive `names` regex (all lower-case). If it validates, then pass it through. If does not validate, then:
- take a Sha224 hash of the containerID
- truncate to 16 lower-case hexadecimal characters
- prepend "cid-" and append "-x" (the hash doesn't necessarily start/end with a letter)
- ... and use the result.

@caseydavenport , @gunjan5 , WDYT?

caseydavenport:
@bcreane I think that's OK but as discussed offline, let's see how far we can get without it and we'll address later if needed.

bcreane:
@caseydavenport - if we think k8s doesn't give us invalid containerIDs, then we can push this issue + PR out to v3.1. Is the only way to verify require us setting up a rkt container k8s cluster?

bcreane:
Reverting the hash change since we don't anticipate hitting a mixed case container id with K8s. If we hit the issue of an invalid WorkloadEndpointIdentifier w/cni due to an upper case string (i.e. <node>-cni-<containerID>-<endpoint> and containerID has upper case letters, then take a look at the attached patch as a starting point.
[wep_hash.txt](https://github.com/projectcalico/libcalico-go/files/1560390/wep_hash.txt)

bcreane:
@caseydavenport - I rebased against the latest libcalico-go (Rob's `migrator` changes). All tests pass on my laptop. Assuming semaphore fails again, would you force push the PR?

gunjan5:
@bcreane I think we also need to rev this into the cni-plugin

bcreane:
@gunjan5 - thanks for catching that. Reving libcalico-go on cni-plugin + running tests now.

gunjan5:
@bcreane just realize this now, it's not urgent but we'll also have to rev this into calicoctl because `convert` command calls validation after the conversion so old ContainerIDs might fail validation after the conversion. Rob just updated libcalico to calicoctl, im not sure if that commit picked up these changes as well

bcreane:
@gunjan5 - I did test this change on calicoctl and Rob's changes went in after mine, so he'll pick this change up.

PR #734, opened 2017-12-13
KDD handle v1 IPPool format and convert to v3

gunjan5:
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


lib/apis/v3/ippool.go
```diff
@@ -45,6 +49,10 @@ type IPPoolSpec struct {
 	NATOutgoing bool `json:"natOutgoing,omitempty"`
 	// When disabled is true, Calico IPAM will not assign addresses from this pool.
 	Disabled bool `json:"disabled,omitempty"`
+
+	// Deprecated: this field is only used for APIv1 backwards compatibility.
+	// Setting this field is not allowed, this field is for internal use only.
+	V1IPIP *apiv1.IPIPConfiguration `json:"ipip,omitempty"`
```
robbrockbank: I have a sneaky feeling kubernetes likes field names to match.  I don't think we need to rename it,.

lib/validator/v3/validator.go
```diff
@@ -233,6 +234,11 @@ func validatePortName(v *validator.Validate, topStruct reflect.Value, currentStr
 	return len(s) != 0 && len(k8svalidation.IsValidPortName(s)) == 0
 }
 
+func validateMustBeNil(v *validator.Validate, topStruct reflect.Value, currentStructOrField reflect.Value, field reflect.Value, fieldType reflect.Type, fieldKind reflect.Kind, param string) bool {
+	log.Debugf("Validate v1 IPIP. Must be nil: %s", field.String())
```
robbrockbank: This is written as a generic validator, so probably should not include the field name in the log unless you are getting it from the parameters.

gunjan5: hmm yeah I was also feeling effy about that

lib/validator/v3/validator.go
```diff
@@ -233,6 +234,11 @@ func validatePortName(v *validator.Validate, topStruct reflect.Value, currentStr
 	return len(s) != 0 && len(k8svalidation.IsValidPortName(s)) == 0
 }
 
+func validateMustBeNil(v *validator.Validate, topStruct reflect.Value, currentStructOrField reflect.Value, field reflect.Value, fieldType reflect.Type, fieldKind reflect.Kind, param string) bool {
+	log.Debugf("Validate field Must be nil: %s", field.String())
```
robbrockbank: nit: Must -> must

lib/backend/k8s/resources/customresource.go
```diff
@@ -319,7 +345,12 @@ func (c *customK8sResourceClient) Watch(ctx context.Context, list model.ListInte
 	toKVPair := func(r Resource) (*model.KVPair, error) {
 		return c.convertResourceToKVPair(r)
 	}
-	return newK8sWatcherConverter(ctx, resl.Kind+" (custom)", toKVPair, k8sWatch), nil
+
+	v1v3converter := func(r Resource) (Resource, error) {
```
robbrockbank: Urm, looking at the common threads in this code...

Looks like putting the conversion in convertResourceToKVPair will handle all of your use cases.

lib/backend/k8s/resources/watcher.go
```diff
@@ -36,28 +36,31 @@ func newK8sWatcherConverter(
 	name string,
 	converter ConvertK8sResourceToKVPair,
 	k8sWatch kwatch.Interface,
+	versionconverter VersionConverter,
```
robbrockbank: You shouldn't need to change this code I think.

Even without my suggestion above ^^^ , this helper module already includes a converter and the version conversion could have just been included in that by the consumer of this watcher helper.

lib/backend/k8s/resources/customresource.go
gunjan5: I saw that first but it feels like we're sneaking in this thing that's not obvious from reading the method name `convertResourceToKVPair` which I didn't like

gunjan5: also, `convertResourceToKVPair` is also used in Create/Update, don't we want to limit this conversion to Get/List/Watch?

lib/backend/k8s/resources/customresource.go
```diff
@@ -348,6 +357,18 @@ func (c *customK8sResourceClient) nameToKey(name string) (model.Key, error) {
...
 
 func (c *customK8sResourceClient) convertResourceToKVPair(r Resource) (*model.KVPair, error) {
+	var err error
+
+	// If the resource has a VersionConverter defined then pass the resource through
+	// the versionconverter to convert the resource version from v1 to v3.
+	// No-op for a v3 resource.
+	if c.versionconverter != nil {
+		r, err = c.versionconverter.ConvertFromK8s(r)
+		if err != nil {
```
robbrockbank: nit:  can contract the if into the assignment.

lib/backend/k8s/resources/ippool.go
```diff
@@ -41,7 +43,36 @@ func NewIPPoolClient(c *kubernetes.Clientset, r *rest.RESTClient) K8sResourceCli
...
+	if !ok {
+		return nil, fmt.Errorf("invalid type conversion")
+	}
+
+	// If IPIP field is not nil, then it means the resource has v1 IPIP data
+	// and we must convert it to v3equivalent data.
+	if ipp.Spec.IPIP != nil {
+		if !ipp.Spec.IPIP.Enabled {
+			ipp.Spec.IPIPMode = apiv3.IPIPModeNever
+		} else if ipp.Spec.IPIP.Enabled && ipp.Spec.IPIP.Mode == ipip.CrossSubnet {
```
robbrockbank: IPIP.Enabled is redundant.  You know it's true already.

lib/validator/v3/validator.go
```diff
@@ -233,6 +234,11 @@ func validatePortName(v *validator.Validate, topStruct reflect.Value, currentStr
 	return len(s) != 0 && len(k8svalidation.IsValidPortName(s)) == 0
 }
 
+func validateMustBeNil(v *validator.Validate, topStruct reflect.Value, currentStructOrField reflect.Value, field reflect.Value, fieldType reflect.Type, fieldKind reflect.Kind, param string) bool {
+	log.Debugf("Validate field must be nil: %s", field.String())
```
robbrockbank: Clearly the validator code is a very naughty boy, but we should really be using WithField() I think.

lib/backend/k8s/resources/ippool.go
gunjan5: yeah added it for the clarity, but i can remove it

lib/backend/k8s/resources/customresource.go
```diff
@@ -47,6 +47,13 @@ type customK8sResourceClient struct {
 	k8sListType         reflect.Type
 	namespaced          bool
 	resourceKind        string
+	versionconverter    VersionConverter
+}
+
+// VersionConverter converts v1 or v3 k8s resources into v3 resources.
+// For a v3 resource, it's a no-op
```
robbrockbank: Wording is a little off.  More like "For a v3 resource, the conversion should be a no-op"

lib/backend/k8s/resources/customresource.go
```diff
@@ -348,6 +357,17 @@ func (c *customK8sResourceClient) nameToKey(name string) (model.Key, error) {
 }
 
 func (c *customK8sResourceClient) convertResourceToKVPair(r Resource) (*model.KVPair, error) {
+	var err error
+
+	// If the resource has a VersionConverter defined then pass the resource through
+	// the versionconverter to convert the resource version from v1 to v3.
```
robbrockbank: versionconverter -> VersionConverter

PR #737, opened 2017-12-14
Version check added for migration

tmjd:
## Description
This adds a function to check the version in the datastore to see if it looks like a migration is still needed or has been done already.

## Todos
- [ ] Maybe make the version check configurable

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/migrate/migrate.go
+++ b/lib/migrate/migrate.go
@@ -21,6 +21,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/coreos/go-semver/semver"
 	log "github.com/sirupsen/logrus"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/uuid"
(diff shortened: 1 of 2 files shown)


lib/migrate/migrate.go
```diff
@@ -655,6 +656,62 @@ func (m *migrationHelper) queryAndConvertV1ToV3Nodes(data *MigrationData) error
...
+
+		// Migrate only if it is possible to migrate from the current version
+		if !canMigrate(v) {
+			return false, errors.New(fmt.Sprintf("Migration to v3 requires a base of Calico v2.6.4+, currently at %s", v))
+		}
+		return true, nil
+	}
+}
+
+// getV2ClusterVersion reads the CalicoVersion from the v1 client interface.
```
caseydavenport: typo: getV2ClusterVersion in comment doesn't match name.

lib/migrate/migrate.go
```diff
@@ -678,6 +735,15 @@ func convertLogLevel(logLevel string) string {
 	}
 }
 
+// NilStatusWriter implments the StatusWriterInterface for use by non-cli
```
caseydavenport: Is this supposed to be part of this PR?

lib/migrate/migrate_test.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+var _ = Describe("UT for checking the version for migration.", func() {
+
+	DescribeTable("Checking canMigrate.",
+		func(ver string, result bool) {
+			Expect(canMigrate(ver)).To(Equal(result))
+		},
+
+		Entry("Expect v2.6.4 to migrate", "v2.6.4", true),
```
caseydavenport: What about a testcase for empty string?

caseydavenport: Any maybe an auto-built git version?

Like: `v2.6.3-15-g0986c6cd`

lib/migrate/migrate.go
tmjd: I added it because I was updating my calico-node startup PR for these changes and saw the need for that and was thinking Typha could make use of it too. Want me to remove it?

PR #742, opened 2017-12-14
Version check added for migration

tmjd:
- also fixed up import and format

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/upgrade/migrator/migrate.go
+++ b/lib/upgrade/migrator/migrate.go
@@ -21,6 +21,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/coreos/go-semver/semver"
 	log "github.com/sirupsen/logrus"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/uuid"
@@ -51,6 +52,7 @@ const (
 type Interface interface {
 	ValidateConversion() (*MigrationData, error)
 	IsDestinationEmpty() (bool, error)
+	ShouldMigrate() (bool, error)
 	Migrate() (*MigrationData, error)
 	Abort() error
 	Complete() error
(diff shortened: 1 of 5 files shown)


tmjd:
I've ran `make test-containerized` and `make static-checks` locally (what Semaphore is trying to run) and they pass for me.

lib/upgrade/migrator/migrate.go
```diff
@@ -655,6 +657,62 @@ func (m *migrationHelper) queryAndConvertV1ToV3Nodes(data *MigrationData) error
 	return nil
 }
 
+// ShouldMigrate checks version information and reports if migration is needed
+// and is possible.
+func (m *migrationHelper) ShouldMigrate() (bool, error) {
+	_, err := m.clientv3.ClusterInformation().Get(context.Background(), "default", options.GetOptions{})
+	if err == nil {
+		// Nothing to do ClusterInformation exists so data has already been migrated
```
robbrockbank: Is that really a safe assertion?  The ClusterInformation is currently one of the first resources to be migrated across - it's possible that the migration could have failed part way through.

lib/upgrade/migrator/migrate.go
```diff
@@ -655,6 +657,62 @@ func (m *migrationHelper) queryAndConvertV1ToV3Nodes(data *MigrationData) error
...
+		return false, nil
+	} else if _, ok := err.(cerrors.ErrorResourceDoesNotExist); !ok {
+		// The error indicates a problem with accessing the resource
+		return false, err
+	} else {
```
robbrockbank: The way this is structured, you'll hit this branch if you have a resource does not exist error.

lib/upgrade/migrator/migrate.go
robbrockbank: IIUC I think we should be migrating if:
-  There is v3 cluster information and the cluster version is 2.x
-  There is no v3 cluster information

Checking the v1 cluster information doesn't tell us if we've already migrated - although it arguably tells us if we shouldn't migrate - e.g. because the cluster version is not set or is <2.6 (in which case they should be upgrading to the latest 2.6)

lib/upgrade/migrator/migrate.go
robbrockbank: general coding nit:  This last `else` is not required since in the previous if branches you `return` - so you can drop the else and get an extra tabs worth of room for the code.

lib/upgrade/migrator/migrate.go
caseydavenport: Yes, I think you're right.

lib/upgrade/migrator/migrate.go
caseydavenport: Don't we want to hit this for a resourceDoesNotExist error?

robbrockbank: d'oh  - for some reason I thought we then dereferenced the cluster info - but we don't ...   in my mind we did because that's what i think we need to do.  Sorry!

lib/upgrade/migrator/migrate.go
```diff
@@ -655,6 +657,74 @@ func (m *migrationHelper) queryAndConvertV1ToV3Nodes(data *MigrationData) error
...
 }
 
+// ShouldMigrate checks version information and reports if migration is needed
+// and is possible.
+func (m *migrationHelper) ShouldMigrate() (bool, error) {
+	ci, err := m.clientv3.ClusterInformation().Get(context.Background(), "default", options.GetOptions{})
+	if err == nil {
+		// Check the version to see if maybe the migration has been started
+		// but not finished.
+		if canMigrate(ci.Spec.CalicoVersion) {
```
robbrockbank: nit:  unless you're planning to add logs for each branch, this can be contracted to:

`return canMigrate(), nil`

lib/upgrade/migrator/migrate.go
```diff
@@ -655,6 +657,74 @@ func (m *migrationHelper) queryAndConvertV1ToV3Nodes(data *MigrationData) error
...
+		return false, errors.New(fmt.Sprintf("Migration to v3 requires a base of Calico v2.6.4+, currently at %s", v))
+	}
+	return true, nil
+}
+
+// getV1ClusterVersion reads the CalicoVersion from the v1 client interface.
+func (m *migrationHelper) getV1ClusterVersion() (string, error) {
+	if kv, err := m.clientv1.Get(model.GlobalConfigKey{Name: "CalicoVersion"}); err == nil {
+		return kv.Value.(string), nil
+	} else {
```
robbrockbank: nit:  else is not required here.

tmjd:
Ok I updated for the final nits @robbrockbank but I can't merge this if the tests don't pass.  I ran `make test-containerized` and `make static-checks` and they both pass on my local machine.

PR #744, opened 2017-12-15
CRUD & watcher tests for KDD IPPool migration

gunjan5:
## Description

add tests for #734 

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/clientv3/ippool_kdd_conversion_test.go
+++ b/lib/clientv3/ippool_kdd_conversion_test.go
@@ -1,0 +1,347 @@
+// Copyright (c) 2017 Tigera, Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package clientv3_test
+
+import (
+	"context"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/ginkgo/extensions/table"
+	. "github.com/onsi/gomega"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/projectcalico/libcalico-go/lib/apiconfig"
+	apiv1 "github.com/projectcalico/libcalico-go/lib/apis/v1"
+	apiv3 "github.com/projectcalico/libcalico-go/lib/apis/v3"
+	"github.com/projectcalico/libcalico-go/lib/backend"
+	"github.com/projectcalico/libcalico-go/lib/backend/model"
+	"github.com/projectcalico/libcalico-go/lib/clientv3"
+	"github.com/projectcalico/libcalico-go/lib/ipip"
+	"github.com/projectcalico/libcalico-go/lib/options"
+	"github.com/projectcalico/libcalico-go/lib/testutils"
+	"github.com/projectcalico/libcalico-go/lib/watch"
+)
+
+var _ = testutils.E2eDatastoreDescribe("IPPool KDD v1 to v3 migration tests", testutils.DatastoreK8s, func(config apiconfig.CalicoAPIConfig) {
+
+	ctx := context.Background()
+	name1 := "ippool-1"
+	name2 := "ippool-2"
+
+	spec1_v3 := apiv3.IPPoolSpec{
+		CIDR:        "1.2.3.0/24",
+		NATOutgoing: true,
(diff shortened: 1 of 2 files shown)


lib/clientv3/ippool_kdd_conversion_test.go
```diff
@@ -0,0 +1,249 @@
...
+		func(name1, name2 string, spec_v1, spec_v3 apiv3.IPPoolSpec, kvp *model.KVPair) {
+			c, err := clientv3.New(config)
+			Expect(err).NotTo(HaveOccurred())
+
+			be, err := backend.NewClient(config)
+			Expect(err).NotTo(HaveOccurred())
+			be.Clean()
+
+			By("Attempting to creating a new IPPool with the non-writable v1 IPIP field")
+			_, outError := c.IPPools().Create(ctx, &apiv3.IPPool{
```
robbrockbank: I assume here you could have just use kvp.Value.(*apiv3.IPPool) rather than needing to pass in the spec_v1

robbrockbank: it also feels more of a validation that going through the backend with the same resource works.

lib/clientv3/ippool_kdd_conversion_test.go
```diff
@@ -0,0 +1,249 @@
...
+			By("Listing all the IPPools, expecting a two results with name1/spec_v1 and name2/spec_v3")
+			outList, outError = c.IPPools().List(ctx, options.ListOptions{})
+			Expect(outError).NotTo(HaveOccurred())
+			Expect(outList.Items).To(HaveLen(2))
+			testutils.ExpectResource(&outList.Items[0], apiv3.KindIPPool, testutils.ExpectNoNamespace, name1, spec_v1)
+			testutils.ExpectResource(&outList.Items[1], apiv3.KindIPPool, testutils.ExpectNoNamespace, name2, spec_v3)
+
+		},
+
+		Entry("IPv4 IPPool CRD with v1 IPIP field and IPIP Mode CrossSubnet", name1, name2, spec1_v1, spec1_v3, kvp1_v3),
```
robbrockbank: One for IPIPEnabled and Always?
One for Disabled and Always?

lib/clientv3/ippool_kdd_conversion_test.go
```diff
@@ -0,0 +1,249 @@
...
+			Enabled: true,
+			Mode:    ipip.CrossSubnet,
+		},
+	}
+	spec1_v3 := apiv3.IPPoolSpec{
+		CIDR:        "1.2.3.0/24",
+		NATOutgoing: true,
+		IPIPMode:    apiv3.IPIPModeCrossSubnet,
+	}
+	kvp1_v3 := &model.KVPair{
```
robbrockbank: Is this really v3?

lib/clientv3/ippool_kdd_conversion_test.go
```diff
@@ -0,0 +1,249 @@
...
+	"github.com/projectcalico/libcalico-go/lib/testutils"
+	"github.com/projectcalico/libcalico-go/lib/watch"
+)
+
+var _ = testutils.E2eDatastoreDescribe("IPPool KDD v1 to v3 migration tests", testutils.DatastoreK8s, func(config apiconfig.CalicoAPIConfig) {
+
+	ctx := context.Background()
+	name1 := "ippool-1"
+	name2 := "ippool-2"
+	spec1_v1 := apiv3.IPPoolSpec{
```
robbrockbank: should spec1_v1, spec1_v3 and kvp1_v1 all refer to the same resource?

If so shouldn't kvp1_v1 just use spec1_v1 rather than redefining.  And in that case NATOutgoing is not set.
If not then a quick explanation as to the relationship between these sets is prolly worthwhile.

lib/clientv3/ippool_kdd_conversion_test.go
gunjan5: I'm calling it kvp_v3 because it's the v3 style KVPair with the whole resource being the Value, but I'm happy to change it to just `kvp`

lib/clientv3/ippool_kdd_conversion_test.go
gunjan5: what do you mean by Disabled and Always? how can that be the case? First thing we check is if it's disabled then set the mode to Never

lib/backend/syncersv1/updateprocessors/configurationprocessor.go
```diff
@@ -236,7 +236,7 @@ func (c *configUpdateProcessor) processAddOrModified(kvp *model.KVPair) ([]*mode
 					case "milliseconds":
 						ms := vt.Duration / time.Millisecond
 						nMs := vt.Duration % time.Millisecond
-						value = fmt.Sprintf("%.3f", float64(ms) + float64(nMs)/1e6)
+						value = fmt.Sprintf("%.3f", float64(ms)+float64(nMs)/1e6)
```
gunjan5: fyi: this thing was breaking the CI (static checks)

PR #747, opened 2017-12-15
Migrate v1 NatOutgoing field for KDD IPPool

gunjan5:
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/backend/k8s/resources/ippool.go
+++ b/lib/backend/k8s/resources/ippool.go
@@ -74,5 +74,11 @@ func (c IPPoolv1v3Converter) ConvertFromK8s(inRes Resource) (Resource, error) {
 		ipp.Spec.IPIP = nil
 	}
 
+	// Take a logical OR of the v1 NATOutgoing field with the v3 NATOutgoing.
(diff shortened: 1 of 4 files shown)


lib/backend/k8s/resources/ippool.go
```diff
@@ -74,5 +74,11 @@ func (c IPPoolv1v3Converter) ConvertFromK8s(inRes Resource) (Resource, error) {
 		ipp.Spec.IPIP = nil
 	}
 
+	// Take a logical OR of the v1 NATOutgoing field with the v3 NATOutgoing.
+	ipp.Spec.NATOutgoing = ipp.Spec.NATOutgoingV1 || ipp.Spec.NATOutgoing
```
robbrockbank: is `ipp.Spec.NATOutgoing |= ipp.Spec.NATOutgoingV1` valid shorthand.... i forget

robbrockbank: Should really say that since we omit the field from the JSON/YAML if set to false then if either the v1 or v3 value are true then we know NATOutgoing should be enabled.

gunjan5: nope

PR #753, opened 2017-12-18
Changing output format of felix config

heschlie:
## Description

Fixes projectcalico/felix#1656
Changing the way the configurationprocessor converts the `time.Duration`s to Strings to be more compliant with how Felix expects them to come through. With this change we get:

`0`    -> `"0"`
`0.0` -> `"0"`
`0.1` -> `"0.1"`

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


bcreane:
@heschlie - once this is merged, please make sure to update all relevant glide.lock files as well.

robbrockbank:
Any chance to get any tests that verify the conclusions? (e.g. adding a test for 0, and one with >3 decimal places)?

heschlie:
for >3 decimal places it will write it out, we could add some logic to limit it if the decimal place is long but I'm not sure we care enough?

I'll add a test for the 0 though.

PR #754, opened 2017-12-18
Fix migration of IPIP tunnel addr

robbrockbank:
## Description
We weren't updating the actual node resources with the ipip tunnel addr.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/upgrade/migrator/migrate.go
+++ b/lib/upgrade/migrator/migrate.go
@@ -388,6 +388,16 @@ func (m *migrationHelper) queryAndConvertResources() (*MigrationData, error) {
 	}
 
 	if m.clientv1.IsKDD() {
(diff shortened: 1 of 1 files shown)


lib/upgrade/migrator/migrate.go
```diff
@@ -387,6 +387,16 @@ func (m *migrationHelper) queryAndConvertResources() (*MigrationData, error) {
 		return nil, err
 	}
 
+	if m.clientv1.IsKDD() {
```
robbrockbank: The fix below iterates through all resources to update just the nodes with the IPIP tunnel address.  Moving this up here just to minimize how many resources it iterates through.  Should be minor impact though and can back this out if you'd prefer.

PR #756, opened 2017-12-18
Fix version migrate message level

tmjd:
## Description
Checks version to determine if datastore is already upgraded and emits a Debug instead of Warn message in that case.

Addresses https://github.com/projectcalico/calico/issues/1520

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/upgrade/migrator/migrate.go
+++ b/lib/upgrade/migrator/migrate.go
@@ -662,7 +662,10 @@ func (m *migrationHelper) queryAndConvertV1ToV3Nodes(data *MigrationData) error 
 func (m *migrationHelper) ShouldMigrate() (bool, error) {
 	ci, err := m.clientv3.ClusterInformation().Get(context.Background(), "default", options.GetOptions{})
 	if err == nil {
-		if canMigrate(ci.Spec.CalicoVersion) {
+		if yes, err := versionRequiresMigration(ci.Spec.CalicoVersion); err != nil {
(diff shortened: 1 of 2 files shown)


lib/upgrade/migrator/migrate.go
```diff
@@ -708,23 +714,35 @@ func (m *migrationHelper) getV1ClusterVersion() (string, error) {
 	return "", err
 }
 
-// canMigrate returns true if the given version can be upgraded, otherwise
-// returns false.
-func canMigrate(v string) bool {
+// versionRequiresMigration returns true if the given version requires
+// upgrating the data model, false otherwise. Returns an error if it was
+// not possible to parse the version, or if the version was less than the
+// minimum requires for upgrade.
```
robbrockbank: nit: requires -> required

lib/upgrade/migrator/migrate.go
```diff
@@ -691,8 +694,11 @@ func (m *migrationHelper) ShouldMigrate() (bool, error) {
 	}
 
 	// Migrate only if it is possible to migrate from the current version
-	if !canMigrate(v) {
-		log.Debugf("Migration to v3 requires a base of Calico v2.6.4+, currently at %s", v)
+	if yes, err := versionRequiresMigration(v); err != nil {
```
robbrockbank: line 688 unitialize->uninitialized

PR #760, opened 2017-12-19
Properly convert policy names for storage

caseydavenport:
## Description

Fixes https://github.com/projectcalico/calicoctl/issues/1795

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/upgrade/migrator/migrate.go
+++ b/lib/upgrade/migrator/migrate.go
@@ -874,8 +874,7 @@ func (m *migrationHelper) storeV3Resources(data *MigrationData) error {
 		// This is slightly more efficient, and cuts out some of the unneccessary additional
 		// processing. Since we applying directly to the backend we need to set the UUID
 		// and creation timestamp which is normally handled by clientv3.
-		r.GetObjectMeta().SetCreationTimestamp(metav1.Now())
-		r.GetObjectMeta().SetUID(uuid.NewUUID())
+		r = toStorage(r)
 		if err := m.applyToBackend(&model.KVPair{
 			Key:   resourceToKey(r),
 			Value: r,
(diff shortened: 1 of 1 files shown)


lib/upgrade/migrator/migrate.go
```diff
@@ -891,6 +890,32 @@ func (m *migrationHelper) storeV3Resources(data *MigrationData) error {
...
 }
 
+func toStorage(r converters.Resource) converters.Resource {
+	// Set timestamp and UID.
+	r.GetObjectMeta().SetCreationTimestamp(metav1.Now())
+	r.GetObjectMeta().SetUID(uuid.NewUUID())
+
+	// Some types require additional processing when converting to storage.
+	switch r.(type) {
+	case *apiv3.GlobalNetworkPolicy:
```
gunjan5: we can combine both cases into one with:
```go
switch r.(type) {
case *apiv3.GlobalNetworkPolicy, *apiv3.NetworkPolicy:
     r.GetObjectMeta().SetName(convertPolicyNameForStorage(r.GetObjectMeta().GetName()))
}
```

lib/upgrade/migrator/migrate.go
```diff
@@ -891,6 +890,32 @@ func (m *migrationHelper) storeV3Resources(data *MigrationData) error {
...
+		r.GetObjectMeta().SetName(convertPolicyNameForStorage(r.GetObjectMeta().GetName()))
+	case *apiv3.NetworkPolicy:
+		// For NP, we need to adjust the name before storage.
+		r.GetObjectMeta().SetName(convertPolicyNameForStorage(r.GetObjectMeta().GetName()))
+	}
+
+	return r
+}
+
+func convertPolicyNameForStorage(name string) string {
```
gunjan5: can we move this to https://github.com/projectcalico/libcalico-go/blob/master/lib/upgrade/converters/names.go#L63 with other name conversions? converters package is already imported here, so won't even need a new import

caseydavenport: @gunjan5 suppose we could - it is duplicating code, but I was trying not to make this a public function because I don't think it should be made public.

I'd rather leave as is.

PR #771, opened 2018-01-24
Fix crash on named port

caseydavenport:
## Description

xref: https://github.com/projectcalico/felix/issues/1685

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Fixed bug where Felix would crash when parsing a NetworkPolicy with a named port
```

### DIFF
--- a/lib/backend/k8s/conversion.go
+++ b/lib/backend/k8s/conversion.go
@@ -202,7 +202,9 @@ func (c Converter) NetworkPolicyToPolicy(np *extensions.NetworkPolicy) (*model.K
 	for _, r := range np.Spec.Ingress {
 		rules, err := c.k8sRuleToCalico(r.From, r.Ports, np.ObjectMeta.Namespace, true)
 		if err != nil {
-			return nil, err
+			// If we fail to parse this rule, skip it.
+			log.WithError(err).Warnf("Failed to parse rule")
+			continue
 		}
 		inboundRules = append(inboundRules, rules...)
 	}
@@ -212,7 +214,9 @@ func (c Converter) NetworkPolicyToPolicy(np *extensions.NetworkPolicy) (*model.K
 	for _, r := range np.Spec.Egress {
 		rules, err := c.k8sRuleToCalico(r.To, r.Ports, np.ObjectMeta.Namespace, false)
 		if err != nil {
-			return nil, err
+			// If we fail to parse this rule, skip it.
+			log.WithError(err).Warnf("Failed to parse rule")
+			continue
 		}
 		outboundRules = append(outboundRules, rules...)
 	}
--- a/lib/backend/k8s/conversion.go
+++ b/lib/backend/k8s/conversion.go
@@ -200,13 +200,21 @@ func (c Converter) NetworkPolicyToPolicy(np *extensions.NetworkPolicy) (*model.K
 	// Generate the inbound rules list.
 	var inboundRules []model.Rule
 	for _, r := range np.Spec.Ingress {
-		inboundRules = append(inboundRules, c.k8sRuleToCalico(r.From, r.Ports, np.ObjectMeta.Namespace, true)...)
+		rules, err := c.k8sRuleToCalico(r.From, r.Ports, np.ObjectMeta.Namespace, true)
+		if err != nil {
+			return nil, err
+		}
+		inboundRules = append(inboundRules, rules...)
 	}
 
 	// Generate the outbound rules list.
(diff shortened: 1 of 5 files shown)


lib/backend/k8s/conversion.go
```diff
@@ -445,7 +445,9 @@ func (c Converter) k8sPortToCalico(port extensions.NetworkPolicyPort) []numorstr
 	if port.Port != nil {
 		p, err := numorstring.PortFromString(port.Port.String())
 		if err != nil {
-			log.Panic("Invalid port %+v: %s", port.Port, err)
+			// If we get an invalid port, ignore it.
+			log.Errorf("Invalid port %+v: %s", port.Port, err)
+			return portList
```
redbaron: not related, but curious: why is it returning list at all? it can only have 1 element or be empty

caseydavenport: hah, yeah good spot.

I think it's a remnant of an earlier version of the code that was missed in a refactor - I believe there used to be a function that could return multiple ports.

caseydavenport:
> I think this introduces a bug "up the stack" in k8sRuleToCalico. IIUC, on line 356, you can now get protocol, dstPorts = "TCP", nil, which results in creating a Calico rule that allows all TCP.


Yes, indeed. I think the desired behavior is probably to invalidate the entire rule, or maybe even entire policy, thus not allowing any traffic. Alternative would be to still respect the selector, if present, but that feels like too much of an assumption.

caseydavenport:
@fasaxc I've updated it so that it ignores the policy entirely. I'm not 100% convinced that's the right approach just yet, so would appreciate your input :)

lib/backend/k8s/syncer.go
```diff
@@ -506,9 +506,11 @@ func (syn *kubeSyncer) readFromKubernetesAPI() {
 				continue
 			}
 			// Event is OK - parse it and send it over the channel.
-			kvp := syn.parseNetworkPolicyEvent(event)
-			latestVersions.networkPolicyVersion = kvp.Revision.(string)
-			syn.sendUpdates([]model.KVPair{*kvp}, KEY_NP)
+			kvp, err := syn.parseNetworkPolicyEvent(event)
```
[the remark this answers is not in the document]
caseydavenport: @fasaxc dang, strike two :)

As we discussed this morning, I'll update so that just the offending rule is ignored.

lib/backend/k8s/syncer.go
```diff
@@ -506,9 +506,11 @@ func (syn *kubeSyncer) readFromKubernetesAPI() {
...
+			if err == nil {
+				latestVersions.networkPolicyVersion = kvp.Revision.(string)
+				syn.sendUpdates([]model.KVPair{*kvp}, KEY_NP)
```
robbrockbank: Do you need these changes anymore?

Feels like whilst they are good we should at least handle the error case by treating as a delete.

caseydavenport: Yeah, we don't strictly need this change.

I think treating as a delete is going to be more work than it is worth given we don't have the key available here.

caseydavenport: @robbrockbank I adjusted so that so long as a KVP is returned, it will be treated as a delete.

Note that as the code is written today, this branch will never get hit because err will always be nil. This will at least work in the future if an error _is_ ever returned from this func.

lib/backend/k8s/syncer.go
```diff
@@ -668,9 +678,11 @@ func (syn *kubeSyncer) performSnapshot(versions *resourceVersions) (map[string][
...
 			versions.networkPolicyVersion = npList.ListMeta.ResourceVersion
 			for _, np := range npList.Items {
-				pol, _ := syn.converter.NetworkPolicyToPolicy(&np)
-				snap[KEY_NP] = append(snap[KEY_NP], *pol)
-				keys[KEY_NP][pol.Key.String()] = true
+				pol, err := syn.converter.NetworkPolicyToPolicy(&np)
+				if err == nil {
+					snap[KEY_NP] = append(snap[KEY_NP], *pol)
+					keys[KEY_NP][pol.Key.String()] = true
+				}
```
robbrockbank: Can we have an error log here too please since we're ignoring the error.

caseydavenport:
I closed it and then it won't let me re-open :'(

I've opened it again here with the fixes: https://github.com/projectcalico/libcalico-go/pull/774

PR #777, opened 2018-02-05
Use etcd v3.3 with keepalives

caseydavenport:
## Description

Use the newly released etcd v3.3 client code to get gRPC keepalive support, and enable keepalives on our etcd connections.

Relevant k8s PR: https://github.com/kubernetes/kubernetes/pull/58008

Fixes https://github.com/projectcalico/calico/issues/1632

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


caseydavenport:
@fasaxc I think the easiest place to FV this is in the Felix tests since the infrastructure there gives us more control over starting / stopping etcd.

caseydavenport:
As for timeouts, I think I agree but I'd rather wait for feedback from users before doing that work.

PR #781, opened 2018-02-07
Adding ability to define a floating ip for a kubernetes workload via annotation

bradbeam:
KDD implementation of https://github.com/projectcalico/cni-plugin/pull/419

Description

This PR adds the ability to define DNAT/floating IP for a kubernetes workload via annotation. Most of the plumbing was already in place, this adds the missing link.

Release Note

```
You can now define a floating IP ( DNAT ) via container annotation `cni.projectcalico.org/floatingIPs`
```

### DIFF
--- a/lib/backend/k8s/conversion/conversion.go
+++ b/lib/backend/k8s/conversion/conversion.go
@@ -17,6 +17,7 @@ package conversion
 import (
 	"crypto/sha1"
 	"encoding/hex"
+	"encoding/json"
 	"fmt"
 	"os"
 	"sort"
@@ -230,6 +231,37 @@ func (c Converter) PodToWorkloadEndpoint(pod *kapiv1.Pod) (*model.KVPair, error)
 		labels[apiv3.LabelServiceAccount] = pod.Spec.ServiceAccountName
 	}
 
+	// Pull out floating IP annotation
+	var floatingIPs []apiv3.IPNAT
+	if annotation, ok := pod.Annotations["cni.projectcalico.org/floatingIPs"]; ok {
+
+		// Parse Annotation data
+		var ips []string
+		err := json.Unmarshal([]byte(annotation), &ips)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse '%s' as JSON: %s", annotation, err)
+		}
+
+		// Get target for NAT
+		podip, podnet, err := cnet.ParseCIDROrIP(ipNets[0])
+		if err != nil {
+			return nil, fmt.Errorf("Failed to parse pod IP: %s", err)
+		}
+
+		netmask, _ := podnet.Mask.Size()
+
+		if netmask != 32 && netmask != 128 {
+			return nil, fmt.Errorf("PodIP is not a valid IP: Mask size is %d, not 32 or 128", netmask)
+		}
+
+		for _, ip := range ips {
+			floatingIPs = append(floatingIPs, apiv3.IPNAT{
+				InternalIP: podip.String(),
+				ExternalIP: ip,
+			})
+		}
+	}
+
 	// Map any named ports through.
 	var endpointPorts []apiv3.EndpointPort
 	for _, container := range pod.Spec.Containers {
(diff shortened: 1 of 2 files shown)


lib/backend/k8s/conversion/conversion.go
```diff
@@ -212,6 +213,29 @@ func (c Converter) PodToWorkloadEndpoint(pod *kapiv1.Pod) (*model.KVPair, error)
...
+	if annotation, ok := pod.Annotations["cni.projectcalico.org/floatingIPs"]; ok {
+
+		// Parse Annotation data
+		var ips []string
+		err := json.Unmarshal([]byte(annotation), &ips)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse '%s' as JSON: %s", annotation, err)
+		}
+
+		podip, _, err := cnet.ParseCIDROrIP(ipNets[0])
```
fasaxc: I think Felix further validates that this is an IP, and not a CIDR so you should validate that it is an IP, not a CIDR here (or validate that it is a full-length /32 or /128).

bradbeam: Wouldn't that always be the case with ParseCIDRorIP? It'll return net.ip / net.ipnet / err.

fasaxc: No, ParseCIDROrIP will accept `10.1.2.0/24` and return `10.1.2.0` even though there's a `/24` on there.  The `/24` will fail validation later on because Felix is expecting a bare IP address.

Also, dropping the `/24` and passing through `10.1.2.0` doesn't seem right to me.

bradbeam:
Updated

lib/backend/k8s/conversion/conversion.go
```diff
@@ -212,6 +213,37 @@ func (c Converter) PodToWorkloadEndpoint(pod *kapiv1.Pod) (*model.KVPair, error)
...
+
+		// Parse Annotation data
+		var ips []string
+		err := json.Unmarshal([]byte(annotation), &ips)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse '%s' as JSON: %s", annotation, err)
+		}
+
+		// Get target for NAT
+		podip, podnet, err := cnet.ParseCIDROrIP(ipNets[0])
```
caseydavenport: We already do some of this validation when we figure out the IP nets above: https://github.com/bradbeam/libcalico-go/blob/e8f5a298c93e95facc37ea2081ee9a1320933b19/lib/backend/k8s/conversion/conversion.go#L150:20

The one bit that function doesn't do is check the netmask. I'd suggest either dropping the netmask check since the data from the k8s API should be valid anyway, or moving the netmask check into `GetPodIPs`

caseydavenport: Something to consider - what will need to change here when multiple pod IPs are supported?

bradbeam: Not sure I have an answer for that. I'd guess it really depends on the schema for multiple interfaces, do you have any insight into what that might look like?

bradbeam: re: netmask check, it was added from the earlier review.  @fasaxc any concerns about dropping it given what casey mentioned?

caseydavenport: woops, sorry for conflicting advice! I'll defer to @fasaxc's review.

bradbeam:
Anything else that needs tweaking for this?

fasaxc:
Just waiting for some spare cycles to review it...

lib/backend/k8s/conversion/conversion.go
caseydavenport: > do you think that's a good-enough answer? Are there any other similar annotations we could take inspiration from?

I'm sure we can find some way of extending it in the future, though I don't yet know what multiple IPs might look like. There may be bits of metadata attached to them we could use. Worried that requiring knowing what the internal IP will be is not a very friendly UX since that is typically assigned to a pod at random.

bradbeam:
should be ready for rereview

PR #783, opened 2018-02-12
Validate icmp protocol

### LINKED ISSUE #657: Validator allows a profile with ICMP fields but no protocol
Originally reported here: https://github.com/projectcalico/felix/issues/1606

It causes Felix to panic because iptables doesn't support such a rule.

Hi, calico experts.
I am testing calico HostProtection.

## Expected Behavior

If policy yaml need to specify protocol, calicoctl should outputs validation error.

## Current Behavior

I created profile for hostendpoint via calicoctl.

```bash
$ cat Profile.yaml
apiVersion: v1
kind: profile
metadata:
  name: Profile
  labels:
    profile: Profile
spec:
  ingress:
  - action: allow
    icmp:
      type: 8
    source:
      nets:
      - 192.168.122.10/32
  - action: deny
    source:
  egress:
  - action: allow

$ calicoctl create -f Profile.yaml
Successfully created 1 'profile' resource(s)
```

I created hostendpoint via calicoctl.

```bash
$ cat HostProtection.yaml
- apiVersion: v1
  kind: hostEndpoint
  metadata:
    name: enp0s3
    node: HostProtection
    labels:
      role: webserver
      environment: production
  spec:
    interfaceName: enp0s3
    profiles: [Profile]
    expectedIPs: ["192.168.122.6"]

$ calicoctl create -f HostProtection.yaml
Successfully created 1 'hostEndpoint' resource(s)
```

Then i got felix error below:

```
ov 07 11:32:16 HostProtection calico-felix[1258]: ERROR table.go 778: Failed to program iptables, loading diags before panic. error=exit status 1 ipVersion=0x4 table="filter"
Nov 07 11:32:16 HostProtection calico-felix[1258]: ERROR table.go 784: Current state of iptables ipVersion=0x4 iptablesState="# Generated by iptables-save v1.6.0 on Tue Nov  7 11
Nov 07 11:32:16 HostProtection calico-felix[1258]: PANIC table.go 786: Failed to program iptables, giving up after retries error=exit status 1 ipVersion=0x4 table="filter"
Nov 07 11:32:16 HostProtection systemd[1]: calico-felix.service: Main process exited, code=exited, status=2/INVALIDARGUMENT
Nov 07 11:32:16 HostProtection systemd[1]: calico-felix.service: Unit entered failed state.
Nov 07 11:32:16 HostProtection systemd[1]: calico-felix.service: Failed with result 'exit-code'.
Nov 07 11:32:16 HostProtection systemd[1]: calico-felix.service: Service hold-off time over, scheduling restart.
Nov 07 11:32:16 HostProtection systemd[1]: Stopped Calico Felix agent.
Nov 07 11:32:16 HostProtection systemd[1]: calico-felix.service: Start request repeated too quickly.
Nov 07 11:32:16 HostProtection systemd[1]: Failed to start Calico Felix agent.
```

Calico felix agent cannot restart again.

Incidentally, when i tested yaml format below, it succeded:

```bash
cat Profile.yaml
apiVersion: v1
kind: profile
metadata:
  name: Profile
  labels:
    profile: Profile
spec:
  ingress:
  - action: allow
    protocol: icmp
    icmp:
      type: 8
    source:
      nets:
      - 192.168.122.10/32
  - action: deny
    source:
  egress:
  - action: allow
```

## Possible Solution

I think profile need protocol key in ingress policy list. Or it should validate yaml format. Is this bug?

## Your Environment

calico-felix version: 2.5
calicoctl version: 1.6.1
etcd version: 2.2.5
OS: ubuntu 16.04
Orchestrator: OpenStack Ocata

calico-felix version in protected host:2.4

```
+------------+  +-----------+  +-----------+
| OpenStack  |  | OpenStack |  | OpenStack |
| Controller |  | Compute1  |  | Compute2  |
| + etcd     |  +-----+-----+  +-----+-----+
+-----+------+        |              |
      |               |              |
------+---------------+--------------+----- 192.168.122.0/24
              |
      +-------+--------+
      | HostProtection |
      | Node           |
      +----------------+
```

ozdanborne:
## Description

Fixes #657

## Todos
- [x] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/validator/v3/validator.go
+++ b/lib/validator/v3/validator.go
@@ -77,6 +77,7 @@ var (
 	overlapsV4LinkLocal   = "IP pool range overlaps with IPv4 Link Local range 169.254.0.0/16"
 	overlapsV6LinkLocal   = "IP pool range overlaps with IPv6 Link Local range fe80::/10"
 	protocolPortsMsg      = "rules that specify ports must set protocol to TCP or UDP"
+	protocolIcmpMsg       = "rules that specify ICMP fields must set protocol to ICMP"
 
 	ipv4LinkLocalNet = net.IPNet{
 		IP:   net.ParseIP("169.254.0.0"),
@@ -692,6 +693,24 @@ func validateRule(v *validator.Validate, structLevel *validator.StructLevel) {
 		if len(rule.Destination.NotPorts) > 0 {
 			structLevel.ReportError(reflect.ValueOf(rule.Destination.NotPorts),
 				"Destination.NotPorts", "", reason(protocolPortsMsg))
+		}
+	}
+
+	icmp := numorstring.ProtocolFromString("ICMP")
+	icmpv6 := numorstring.ProtocolFromString("ICMPv6")
+	if rule.ICMP != nil && (rule.Protocol == nil || (*rule.Protocol != icmp && *rule.Protocol != icmpv6)) {
+		structLevel.ReportError(reflect.ValueOf(rule.ICMP), "ICMP", "", reason(protocolIcmpMsg))
+	}
+
+	// Check that the IPVersion of the protocol matches the IPVersion of the ICMP protocol.
+	if (rule.Protocol != nil && *rule.Protocol == icmp) || (rule.NotProtocol != nil && *rule.NotProtocol == icmp) {
+		if rule.IPVersion == nil || *rule.IPVersion != 4 {
+			structLevel.ReportError(reflect.ValueOf(rule.ICMP), "IPVersion", "", reason("must set ipversion to '4' with protocol icmp"))
+		}
+	}
+	if (rule.Protocol != nil && *rule.Protocol == icmpv6) || (rule.NotProtocol != nil && *rule.NotProtocol == icmpv6) {
+		if rule.IPVersion == nil || *rule.IPVersion != 6 {
+			structLevel.ReportError(reflect.ValueOf(rule.ICMP), "IPVersion", "", reason("must set ipversion to '6' with protocol icmpv6"))
 		}
 	}
 
--- a/lib/validator/v3/validator_test.go
+++ b/lib/validator/v3/validator_test.go
@@ -851,10 +851,65 @@ func init() {
 			}, true),
 		Entry("should accept Rule with empty dest !ports and protocol type ICMPv6",
 			api.Rule{
-				Action:   "Allow",
-				Protocol: protocolFromString("ICMPv6"),
+				Action:    "Allow",
+				IPVersion: &V6,
+				Protocol:  protocolFromString("ICMPv6"),
 				Destination: api.EntityRule{
 					NotPorts: []numorstring.Port{},
+				},
+			}, true),
+		Entry("should reject Rule with icmp fields and no protocol",
+			api.Rule{
+				Action:    "Allow",
+				IPVersion: &V4,
+				ICMP: &api.ICMPFields{
+					Type: &V0,
+				},
+			}, false),
+		Entry("should reject Rule with icmp fields and no ipversion",
+			api.Rule{
+				Action:   "Allow",
+				Protocol: protocolFromString("ICMP"),
+				ICMP: &api.ICMPFields{
+					Type: &V0,
+				},
+			}, false),
+		Entry("should reject Rule with icmpv6 fields and no ipversion",
+			api.Rule{
+				Action:   "Allow",
+				Protocol: protocolFromString("ICMPv6"),
+				ICMP: &api.ICMPFields{
+					Type: &V0,
+				},
+			}, false),
+		Entry("should reject Rule with mismatched ipversion for icmp",
+			api.Rule{
+				Action:    "Allow",
+				Protocol:  protocolFromString("ICMP"),
+				IPVersion: &V6,
+			}, false),
+		Entry("should reject Rule with mismatched ipversion for icmpv6",
+			api.Rule{
+				Action:    "Allow",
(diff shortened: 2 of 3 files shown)


lib/validator/v3/validator.go
```diff
@@ -695,6 +696,11 @@ func validateRule(v *validator.Validate, structLevel *validator.StructLevel) {
 		}
 	}
 
+	if rule.ICMP != nil && (rule.Protocol == nil || (*rule.Protocol != numorstring.ProtocolFromString("ICMP") &&
```
fasaxc: I think Felix could be tripped up by having a rule with `ipVersion=4` and `protocol=ICMPv6` too (or vice versa).  Perhaps we need to validate against that here too, or adjust the logic in felix so that it skips such inconsistent rules.

ozdanborne: A rule with mismatched ipversion and protocol is a useless rule but I don't think its really an "invalid" rule. A good example is the inverse - should we reject a rule if it states `ipVersion=6 && !protocol=icmp`? The !protocol is useless but not necessarily invalid.

fasaxc: I mean, I think that's another thing that'd crash felix right now: trying to render an IPv4 rule with ICMPv6 or vice-versa.  We should fix it here by validating against it or in Felix (by ignoring the rule on the basis that it can't possibly match anything because it specified conflicting match criteria).

fasaxc: I _think_ validating against it here is better because it's probably a bug in the user's rule so calicoctl will give them better diags than a silent ignore in Felix.

ozdanborne: addressed. PTAL @fasaxc

lib/validator/v3/validator.go
```diff
@@ -695,6 +696,32 @@ func validateRule(v *validator.Validate, structLevel *validator.StructLevel) {
...
+		*rule.Protocol != numorstring.ProtocolFromString("ICMPv6"))) {
+		structLevel.ReportError(reflect.ValueOf(rule.ICMP), "ICMP", "", reason(protocolIcmpMsg))
+	}
+
+	// Check that the IPVersion of the protocol matches the IPVersion of the rule for
+	// known protocols.
+	if rule.IPVersion != nil {
```
fasaxc: I think you should do this validation the other way around for ICMP:
```
if protocol == "ICMP" && ipVersion != 4 {
    // invalid
} else if protocol == "ICMPv6" && ipVersion != 6 {
    // invalid
}
```
(I don't think Felix would work with "ICMP" and no IP version either.)

It's sad that ICMP ended up having different protocol names for v4 and v6 but that's what we've got so if you use the v4 protocol name, your rule needs to specify IPv4 :-(

lib/validator/v3/validator.go
```diff
@@ -695,6 +696,24 @@ func validateRule(v *validator.Validate, structLevel *validator.StructLevel) {
 		}
 	}
 
+	icmp := numorstring.ProtocolFromString("ICMP")
+	icmpv6 := numorstring.ProtocolFromString("ICMP")
```
fasaxc: Copy/paste bug, should be `ICMPv6`?

lib/validator/v3/validator.go
```diff
@@ -695,6 +696,24 @@ func validateRule(v *validator.Validate, structLevel *validator.StructLevel) {
...
+	if rule.ICMP != nil && (rule.Protocol == nil || (*rule.Protocol != icmp && *rule.Protocol != icmpv6)) {
+		structLevel.ReportError(reflect.ValueOf(rule.ICMP), "ICMP", "", reason(protocolIcmpMsg))
+	}
+
+	// Check that the IPVersion of the protocol matches the IPVersion of the ICMP protocol.
+	if (rule.Protocol != nil && *rule.Protocol == icmp) || (rule.NotProtocol != nil && *rule.NotProtocol == icmp) {
+		if rule.IPVersion != nil && *rule.IPVersion != 4 {
```
fasaxc: This will let nil through, but I think we should validate that the protocol is not nil.  I.e. `if rule.IPVersion == nil || *rule.IPVersion != 4`

lib/validator/v3/validator_test.go
```diff
@@ -857,6 +857,21 @@ func init() {
...
 				},
 			}, true),
+		Entry("should reject Rule with icmp fields and no protocol",
+			api.Rule{
+				Action: "Allow",
+				ICMP: &api.ICMPFields{
+					Type: &V0,
+				},
+			}, false),
+		Entry("should allow Rule with icmp fields and protocol",
```
fasaxc: Need one for IPv6 allow case too.

ozdanborne:
Thank you @fasaxc !

lib/validator/v3/validator.go
ozdanborne: @fasaxc this change is causing problems in our downstream components. Any rule that includes ICMP rules now also needs to set IPVersion=4. Did we mean to require that?

If so, i'll probably need to update this comment:
https://github.com/projectcalico/libcalico-go/blob/master/lib/apis/v3/policy.go#L38-L40

fasaxc: Looks like Neil found some code I'd forgotten about in Felix's `ipVersionToProtoIPVersion` that does auto-set the version based on the presence of ICMP/ICMPv6 so maybe we _can_ relax this.  Sorry about that, that code wasn't where I expected it in Felix.

It does look like we need to validate against using ICMP with IPv6 and ICMPv6 with IPv4, though.

PR #784, opened 2018-02-13
Don't panic for invalid ports

ozdanborne:
## Description

Don't panic for invalid ports! 

Looking for general feedback on my first PR to remove panics in libcalico.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/backend/k8s/conversion/conversion.go
+++ b/lib/backend/k8s/conversion/conversion.go
@@ -290,13 +290,23 @@ func (c Converter) K8sNetworkPolicyToCalico(np *networkingv1.NetworkPolicy) (*mo
 	// Generate the ingress rules list.
 	var ingressRules []apiv3.Rule
 	for _, r := range np.Spec.Ingress {
-		ingressRules = append(ingressRules, c.k8sRuleToCalico(r.From, r.Ports, np.Namespace, true)...)
+		rules, err := c.k8sRuleToCalico(r.From, r.Ports, np.Namespace, true)
+		if err != nil {
+			log.WithError(err).Warn("dropping k8s rule that couldn't be converted.")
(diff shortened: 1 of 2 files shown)


lib/backend/k8s/conversion/conversion.go
```diff
@@ -446,8 +446,14 @@ func (c Converter) k8sRuleToCalico(rPeers []networkingv1.NetworkPolicyPeer, rPor
 	// TODO: This currently creates a lot of rules by making every combination of from / ports
 	// into a rule.  We can combine these so that we don't need as many rules!
 	for _, port := range ports {
+		protocol, calicoPorts, err := c.k8sPortToCalicoFields(port)
```
caseydavenport: This is subtly different from what we do in v2.6 - in v2.6 we throw away the entire rule, whereas here it looks like we're just dropping part of the rule which includes the bad port.

Not sure if there's really a significance, but we should probably behave the same.

lib/backend/k8s/conversion/conversion.go
```diff
@@ -446,8 +446,14 @@ func (c Converter) k8sRuleToCalico(rPeers []networkingv1.NetworkPolicyPeer, rPor
...
+		if err != nil {
+			// If we fail to parse this port, skip it.
+			log.WithError(err).Warn("Failed to parse k8s port. Dropping rule.")
+			return []apiv3.Rule{}
```
caseydavenport: Take a look at what we do in v2.6: https://github.com/projectcalico/libcalico-go/pull/774/files

It's similar but we return an error back one more level. I think I prefer that approach (and makes the two consistent)

PR #785, opened 2018-02-14
Fix IPAM affinity race conditions on allocation and release

### LINKED ISSUE #778: IPAM can give same affinity to multiple hosts
The IPAM affinity claiming process looks like this:

- pre-create the affinity
- try to claim the block
  - on success, return
  - on failure, try to delete the affinity (*)

If the cleanup on failure fails, it looks like we leave the datastore with an affinity that doesn't belong to the host.  

When using an affinity, it doesn't appear that we check that the block actually matches this host.

Presumably, there's also a race where another process on the host tries to use the affinity after the pre-create but before the block has actually been claimed.

## Expected Behavior

- Only one host should be able to claim an affinity.
- (When doing an affine alloc) IPs should only be allocated from blocks that really are owned by this host.

## Current Behavior

- Multiple hosts can claim an affinity.
- IPs can be allocated by all hosts who think they own the block.

## Possible Solution

- Mark the affinity as pending on creation.
- If a host sees a pending affinity when claiming an IP, check whether the block has really been claimed already.
  - if yes, update the affinity
  - if no, try to claim the block and delete affinity on failure.

1.
2.
3.
4.

## Your Environment

* libcalico-go version:

caseydavenport:
## Description

Fixes #778 

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Closes a number of race conditions and failure scenarios in IPAM block allocation and releasing.
```

```release-note
Improves log output around IPAM block allocation and releasing.
```

### DIFF
--- a/lib/ipam/ipam.go
+++ b/lib/ipam/ipam.go
@@ -103,27 +103,162 @@ func (c ipamClient) AutoAssign(ctx context.Context, args AutoAssignArgs) ([]net.
 	return v4list, v6list, nil
 }
 
+// getBlockFromAffinity returns the block referenced by the given affinity, attempting to create it if
+// it does not exist. getBlockFromAffinity will delete the provided affinity if it does not match the actual
+// affinity of the block.
+func (c ipamClient) getBlockFromAffinity(ctx context.Context, aff *model.KVPair) (*model.KVPair, error) {
+	// Parse out affinity data.
+	cidr := aff.Key.(model.BlockAffinityKey).CIDR
+	host := aff.Key.(model.BlockAffinityKey).Host
+	state := aff.Value.(*model.BlockAffinity).State
+	logCtx := log.WithFields(log.Fields{"host": host, "cidr": cidr})
+
+	// Get the block referenced by this affinity.
+	logCtx.Info("Attempting to load block")
+	b, err := c.client.Get(ctx, model.BlockKey{CIDR: cidr}, "")
+	if err != nil {
+		if _, ok := err.(cerrors.ErrorResourceDoesNotExist); ok {
+			// The block referenced by the affinity doesn't exist. Try to create it.
+			logCtx.Info("The referenced block doesn't exist, trying to create it")
+			aff.Value.(*model.BlockAffinity).State = model.StatePending
+			aff, err = c.client.Update(ctx, aff)
+			if err != nil {
+				logCtx.WithError(err).Warn("Error updating block affinity")
+				return nil, err
+			}
+			logCtx.Info("Wrote affinity as pending")
+
+			cfg, err := c.GetIPAMConfig(ctx)
+			if err != nil {
+				logCtx.WithError(err).Errorf("Error getting IPAM Config")
+				return nil, err
+			}
+
+			// Claim the block, which will also confirm the affinity.
+			logCtx.Info("Attempting to claim the block")
+			b, err := c.blockReaderWriter.claimAffineBlock(ctx, aff, *cfg)
+			if err != nil {
+				logCtx.WithError(err).Warn("Error claiming block")
+				return nil, err
+			}
+			return b, nil
+		}
+		logCtx.WithError(err).Error("Error getting block")
+		return nil, err
+	}
+
+	// If the block doesn't match the affinity, it means we've got a stale affininty hanging around.
+	// We should remove it.
+	blockAffinity := b.Value.(*model.AllocationBlock).Affinity
+	if blockAffinity == nil || *blockAffinity != fmt.Sprintf("host:%s", host) {
+		logCtx.WithField("blockAffinity", blockAffinity).Warn("Block does not match the provided affinity, deleting stale affinity")
+		_, err := c.client.Delete(ctx, aff.Key, aff.Revision)
+		if err != nil {
+			logCtx.WithError(err).Warn("Error deleting stale affinity")
+			return nil, err
+		}
+		return nil, errStaleAffinity(fmt.Sprintf("Affinity is stale: %+v", aff))
+	}
+
+	// If the block does match the affinity but the affinity has not been confirmed,
+	// try to confirm it. Treat empty string as confirmed for compatibility with older data.
+	if state != model.StateConfirmed && state != "" {
+		// Write the affinity as pending.
+		logCtx.Info("Affinity has not been confirmed - attempt to confirm it")
+		aff.Value.(*model.BlockAffinity).State = model.StatePending
+		aff, err = c.client.Update(ctx, aff)
+		if err != nil {
+			logCtx.WithError(err).Warn("Error marking affinity as pending as part of confirmation process")
+			return nil, err
+		}
+
+		// CAS the block to get a new revision and invalidate any other instances
+		// that might be trying to operate on the block.
+		logCtx.Info("Writing block to get a new revision")
+		b, err = c.client.Update(ctx, b)
+		if err != nil {
+			logCtx.WithError(err).Debug("Error writing block")
+			return nil, err
+		}
+
+		// Confirm the affinity.
+		logCtx.Info("Attempting to confirm affinity")
+		aff.Value.(*model.BlockAffinity).State = model.StateConfirmed
+		aff, err = c.client.Update(ctx, aff)
+		if err != nil {
+			logCtx.WithError(err).Debug("Error confirming affinity")
+			return nil, err
+		}
+		logCtx.Info("Affinity confirmed successfully")
+	}
+	logCtx.Info("Affinity is confirmed and block has been loaded")
+	return b, nil
+}
+
 func (c ipamClient) autoAssign(ctx context.Context, num int, handleID *string, attrs map[string]string, pools []net.IPNet, version ipVersion, host string) ([]net.IP, error) {
 
 	// Start by trying to assign from one of the host-affine blocks.  We
 	// always do strict checking at this stage, so it doesn't matter whether
 	// globally we have strict_affinity or not.
-	log.Debugf("Looking for addresses in current affine blocks for host '%s'", host)
+	logCtx := log.WithFields(log.Fields{"host": host})
+	if handleID != nil {
+		logCtx = logCtx.WithField("handle", *handleID)
+	}
+	logCtx.Info("Looking up existing affinities for host")
 	affBlocks, err := c.blockReaderWriter.getAffineBlocks(ctx, host, version, pools)
 	if err != nil {
 		return nil, err
 	}
-	log.Debugf("Found %d affine IPv%d blocks for host '%s': %v", len(affBlocks), version.Number, host, affBlocks)
+	logCtx.Debugf("Found %d affine IPv%d blocks for host: %v", len(affBlocks), version.Number, affBlocks)
 	ips := []net.IP{}
+	newIPs := []net.IP{}
+
 	for len(ips) < num {
 		if len(affBlocks) == 0 {
-			log.Infof("Ran out of existing affine blocks for host '%s'", host)
+			logCtx.Infof("Ran out of existing affine blocks for host")
 			break
 		}
 		cidr := affBlocks[0]
 		affBlocks = affBlocks[1:]
-		ips, _ = c.assignFromExistingBlock(ctx, cidr, num, handleID, attrs, host, true)
-		log.Debugf("Block '%s' provided addresses: %v", cidr.String(), ips)
+
+		// Try to assign from this block - if we hit a CAS error, we'll try this block again.
+		// For any other error, we'll break out and try the next affine block.
+		for i := 0; i < ipamEtcdRetries; i++ {
+			// Get the affinity.
+			logCtx.Infof("Trying affinity for %s", cidr)
+			aff, err := c.client.Get(ctx, model.BlockAffinityKey{Host: host, CIDR: cidr}, "")
+			if err != nil {
+				logCtx.WithError(err).Warnf("Error getting affinity")
+				break
+			}
+
+			// Get the block which is referenced by the affinity, creating it if necessary.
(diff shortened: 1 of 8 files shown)


lib/ipam/ipam_block_reader_writer.go
```diff
@@ -204,24 +229,43 @@ func (rw blockReaderWriter) claimBlockAffinity(ctx context.Context, subnet cnet.
...
+			// Some other host beat us to this block.  Cleanup and return an error.
+			_, err = rw.client.Delete(ctx, model.BlockAffinityKey{Host: host, CIDR: b.CIDR}, aff.Revision)
 			if err != nil {
-				log.Errorf("Error cleaning up block affinity: %v", err)
-				return err
+				// Failed to clean up our claim to this block.
+				log.WithError(err).Errorf("Error cleaning up block affinity after failed claim")
+				return affinityClaimedError{Block: b}
 			}
 			return affinityClaimedError{Block: b}
```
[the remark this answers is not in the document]
caseydavenport: hehe, yeah.

lib/ipam/ipam.go
```diff
@@ -291,7 +301,9 @@ func (c ipamClient) AssignIP(ctx context.Context, args AssignIPArgs) error {
 		if err != nil {
 			log.Warningf("Update failed on block %s", block.CIDR.String())
 			if args.HandleID != nil {
-				c.decrementHandle(ctx, *args.HandleID, blockCIDR, 1)
+				if err := c.decrementHandle(ctx, *args.HandleID, blockCIDR, 1); err != nil {
```
caseydavenport: We were ignoring this error before, need to think if that was right or not so leaving a comment to remind me to check.

lib/ipam/ipam.go
```diff
@@ -604,18 +669,38 @@ func (c ipamClient) RemoveIPAMHost(ctx context.Context, host string) error {
 	hostname := decideHostname(host)
 
 	// Release affinities for this host.
-	c.ReleaseHostAffinities(ctx, hostname)
+	if err := c.ReleaseHostAffinities(ctx, hostname); err != nil {
```
caseydavenport: Need to double check this too.

lib/ipam/ipam_block_reader_writer.go
```diff
@@ -160,25 +170,41 @@ func isPoolInRequestedPools(pool cnet.IPNet, requestedPools []cnet.IPNet) bool {
...
+			log.WithError(err).Error("Failed to claim affinity")
+			return err
+		}
+		log.Info("Block affinity already exists")
 
-	// Make sure hostname is not empty.
-	if host == "" {
-		log.Errorf("Hostname can't be empty")
-		return errors.New("Hostname must be sepcified to claim block affinity")
+		// Get the existing affinity in order to get revision information.
```
caseydavenport: Do we need to write it back as pending in this case? I think so.

lib/ipam/ipam_block_reader_writer.go
```diff
@@ -160,25 +178,51 @@ func isPoolInRequestedPools(pool cnet.IPNet, requestedPools []cnet.IPNet) bool {
...
+		}
+		aff.Value.(*model.BlockAffinity).State = model.StatePending
+		return rw.client.Update(ctx, aff)
 	}
+	return aff, nil
+}
+
+// claimBlockAffinity claims the provided block using the given pending affinity. If successful, it will confirm the affinity. If another host
+// steals the block, claimBlockAffinity will attempt to delete the provided pending affinity.
+func (rw blockReaderWriter) claimBlockAffinity(ctx context.Context, aff *model.KVPair, config IPAMConfig) (*model.KVPair, error) {
```
caseydavenport: I've pulled out the claiming of the affinity block into `getPendingAffinity` - calling functions now need to make sure they have the affinity in-hand in order to call this.

I've also made this return the block that it creates (or the existing one if it exists). So that the caller can use it to CAS an allocation.

lib/ipam/ipam.go
```diff
@@ -101,27 +101,139 @@ func (c ipamClient) AutoAssign(ctx context.Context, args AutoAssignArgs) ([]net.
...
+				break
+			}
+
+			// Get the block which is referenced by the affinity, creating it if necessary.
+			b, err := c.getBlockFromAffinity(ctx, aff)
+			if err != nil {
+				// Couldn't get a block for this affinity.
+				logCtx.WithError(err).Warn("Couldn't get block for affinity")
+				if _, ok := err.(cerrors.ErrorResourceUpdateConflict); ok {
+					// CAS Error - retry.
```
caseydavenport: Need to go through and debug log out all these branches.

lib/ipam/ipam.go
```diff
@@ -101,27 +101,139 @@ func (c ipamClient) AutoAssign(ctx context.Context, args AutoAssignArgs) ([]net.
...
+				logCtx.WithError(err).Warnf("Error getting affinity")
+				break
+			}
+
+			// Get the block which is referenced by the affinity, creating it if necessary.
+			b, err := c.getBlockFromAffinity(ctx, aff)
+			if err != nil {
+				// Couldn't get a block for this affinity.
+				logCtx.WithError(err).Warn("Couldn't get block for affinity")
+				if _, ok := err.(cerrors.ErrorResourceUpdateConflict); ok {
```
caseydavenport: Need to debug log in all these branches, adjust warning log so it only occurs when not a resource update conflict.

lib/ipam/ipam_block.go
```diff
@@ -82,6 +82,9 @@ func (b *allocationBlock) autoAssign(
 		// Affinity check is enabled but the host does not match - error.
 		s := fmt.Sprintf("Block affinity (%s) does not match provided (%s)", *b.Affinity, host)
 		return nil, errors.New(s)
+	} else if b.Affinity == nil {
```
caseydavenport: It's possible we hit this when trying to assign from non-affine blocks when all other blocks are full.

Should we protect against this?

lib/ipam/ipam.go
```diff
@@ -621,18 +838,38 @@ func (c ipamClient) RemoveIPAMHost(ctx context.Context, host string) error {
...
+		if err != nil {
+			log.WithError(err).Errorf("Failed to get IPAM host")
 			return err
 		}
+
+		// Remove the host tree from the datastore.
+		_, err = c.client.Delete(ctx, k, kvp.Revision)
+		if err != nil {
+			if _, ok := err.(cerrors.ErrorResourceUpdateConflict); ok {
+				// We hit a compare-and-delete error - retry.
```
fasaxc: This deletion routine feels a bit fishy; it doesn't take a revision so we could delete someone else's concurrent updates.  What happens if an affinity is created between the release above and this retry here?  Does this deletion do a prefix delete (and maybe leave a dangling block)?

lib/ipam/ipam_block_reader_writer.go
```diff
@@ -187,110 +208,146 @@ func (rw blockReaderWriter) claimBlockAffinity(ctx context.Context, subnet cnet.
...
-					continue
-				} else {
-					return err
-				}
-			}
-		}
+	// Read the model.KVPair containing the block
+	// and pull out the allocationBlock object.  We need to hold on to this
+	// so that we can pass it back to the datastore on Update.
+	obj, err := rw.client.Get(ctx, model.BlockKey{CIDR: blockCIDR}, "")
```
fasaxc: I _think_ you need to read the block before you do the update to the affinity.  Otherwise another process could slip in, reclaim the block and do an assignment _before_ you read the block here.  Then you'd go ahead and remove the affinity from the block but the CaD for the affinity would fail so we'd have a dangling affinity.

lib/ipam/ipam.go
caseydavenport: I think it does take a revision, so if a new affinity is created it should fail the deletion.

But, it will then retry and if it succeeds I think it will leave a dangling block. I think we probably need to pull the releasing of affinities into this retry loop.

So in the loop we:
- Release affinities
- Get IPAMHost
- Compare-and-delete IPAMHost

lib/ipam/ipam_block_reader_writer.go
```diff
@@ -160,25 +161,45 @@ func isPoolInRequestedPools(pool cnet.IPNet, requestedPools []cnet.IPNet) bool {
...
-	if host == "" {
-		log.Errorf("Hostname can't be empty")
-		return errors.New("Hostname must be sepcified to claim block affinity")
+		// Get the existing affinity and write it as pending.
+		aff, err = rw.client.Get(ctx, obj.Key, "")
+		if err != nil {
+			log.WithError(err).Error("Failed to get existing affinity")
+			return nil, err
+		}
+		aff.Value.(*model.BlockAffinity).State = model.StatePending
```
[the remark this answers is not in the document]
caseydavenport: Yeah I think that's reasonable.

lib/ipam/ipam_block_reader_writer.go
```diff
@@ -187,110 +208,146 @@ func (rw blockReaderWriter) claimBlockAffinity(ctx context.Context, subnet cnet.
...
+}
+
+func (rw blockReaderWriter) confirmAffinity(ctx context.Context, aff *model.KVPair) (*model.KVPair, error) {
+	host := aff.Key.(model.BlockAffinityKey).Host
+	cidr := aff.Key.(model.BlockAffinityKey).CIDR
+	log.Debugf("Confirming affinity: %s -> %s", host, cidr)
+	aff.Value.(*model.BlockAffinity).State = model.StateConfirmed
+	confirmed, err := rw.client.Update(ctx, aff)
+	if err != nil {
+		log.WithError(err).Error("Failed to confirm block affinity")
```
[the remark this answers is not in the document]
caseydavenport: I don't think it does on failure.

Right now, the calling code will see the CAS error and retry, which I think is fine. I guess checking the state here is another optimization. I'll see how it works.

caseydavenport:
@fasaxc I think I've addressed your first round of feedback, ready for another look!

lib/ipam/ipam.go
```diff
@@ -103,27 +103,162 @@ func (c ipamClient) AutoAssign(ctx context.Context, args AutoAssignArgs) ([]net.
...
+		aff, err = c.client.Update(ctx, aff)
+		if err != nil {
+			logCtx.WithError(err).Warn("Error marking affinity as pending as part of confirmation process")
+			return nil, err
+		}
+
+		// CAS the block to get a new revision and invalidate any other instances
+		// that might be trying to operate on the block.
+		logCtx.Info("Writing block to get a new revision")
+		b, err = c.client.Update(ctx, b)
```
mdianjun: Is this used for preventing other process on the same node operating this block?  Why is there any other process? Like what?

caseydavenport: @mdianjun yeah, there might be other clients trying to perform an IPAM allocation at the same time, e.g. multiple pods starting on the host at the same time. We need to make sure that only one client operates on the block at a time. If two clients conflict, then one should succeed and the other should retry.

ti-mo:
@caseydavenport Also hitting this on 2.6.7 during initial cluster bootstrap of around 40 nodes. We define an ipPool (ipip under Mesos) after all nodes have been added to the cluster, which (I would assume) significantly increases the chances of this contention causing issues, as all nodes get their blocks assigned at the same time.

```
2018-04-27T08:49:12Z public-slave-14 confd[130]: ERROR "
2018-04-27 08:49:11.878 [INFO][277] client.go 202: Loading config from environment
2018-04-27 08:49:11.895 [INFO][277] ipam.go 120: Auto-assign 1 ipv4, 0 ipv6 addrs for host 'public-slave-14'
2018-04-27 08:49:11.903 [INFO][277] ipam.go 172: Ran out of existing affine blocks for host 'public-slave-14'
2018-04-27 08:49:11.909 [INFO][277] ipam.go 195: Need to allocate 1 more addresses - allocate another block
2018-04-27 08:49:11.932 [INFO][277] ipam_block_reader_writer.go 116: Claiming a new affine block for host 'public-slave-14'
2018-04-27 08:49:11.944 [INFO][277] ipam_block_reader_writer.go 159: Host public-slave-14 claiming block affinity for 10.123.84.64/26
2018-04-27 08:49:11.994 [WARNING][277] ipam_block_reader_writer.go 187: Problem claiming block affinity:%!!(MISSING)(EXTRA errors.ErrorResourceAlreadyExists=resource already exists: BlockKey(cidr=10.123.84.64/26))
2018-04-27 08:49:12.060 [ERROR][277] ipam.go 204: Error claiming new block: 10.123.84.64/26 already claimed by %!!(MISSING)s(*string=0xc420283430)
2018-04-27 08:49:12.060 [ERROR][277] ipam.go 135: Error assigning IPV4 addresses: 10.123.84.64/26 already claimed by %!!(MISSING)s(*string=0xc420283430)
2018-04-27 08:49:12.060 [FATAL][277] allocate_ipip_addr.go 120: Unable to autoassign an address for IPIP error=10.123.84.64/26 already claimed by %!!(MISSING)s(*string=0xc420283430)
"
2018-04-27T08:49:12Z public-slave-14 confd[130]: ERROR exit status 1
```

We use `calico/node`, which doesn't seem to restart `confd` on failure. This requires us to go and manually restart the container when this occurs.

fasaxc:
@ti-mo You issue sounds different, please file a new issue.

In your case, it appears to be correctly(!) detecting that the block is already claimed but presumably, we're not retrying enough to look for a new block.

What is your pool size?

fasaxc:
Also, `confd` should be restarted on failure by the runit daemon inside `calico/node`  please file an issue in the `projectcalico/node` repo if that is not the case.

caseydavenport:
Yeah, agreed this looks like a separate issue. Let's investigate in a new thread.

PR #786, opened 2018-02-16
Pass through various fields for the policy sync API.

fasaxc:
## Description

Pass through:

- NetworkPolicy namespace.
- Original, unmodified selectors.
- Namespace selectors.
- Service account match criteria.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/backend/model/rule.go
+++ b/lib/backend/model/rule.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2016-2017 Tigera, Inc. All rights reserved.
+// Copyright (c) 2016-2018 Tigera, Inc. All rights reserved.
 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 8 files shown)


lib/backend/model/rule.go
```diff
@@ -61,6 +67,16 @@ type Rule struct {
 	NotDstNets     []*net.IPNet       `json:"!dst_nets,omitempty" validate:"omitempty"`
 	NotDstPorts    []numorstring.Port `json:"!dst_ports,omitempty" validate:"omitempty,dive"`
 
+	// These fields allow us to pass through the raw selectors from the V3 datamodel unmodified.
+	// The selectors above are formed in the update processor layer by combining the original
+	// selectors and namespace selectors into one.
```
spikecurtis: I think the Service Account selectors should be included in this block. Just like pod label selectors and namespace selectors, service account selectors are rolled into `SrcSelector` and `DstSelector`

PR #788, opened 2018-02-20
remove panic from bgp settings

ozdanborne:
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/backend/syncersv1/updateprocessors/bgpconfigprocessor.go
+++ b/lib/backend/syncersv1/updateprocessors/bgpconfigprocessor.go
@@ -15,14 +15,12 @@
 package updateprocessors
 
 import (
-	"encoding/json"
 	"reflect"
 	"strings"
 
 	apiv3 "github.com/projectcalico/libcalico-go/lib/apis/v3"
 	"github.com/projectcalico/libcalico-go/lib/backend/model"
 	"github.com/projectcalico/libcalico-go/lib/backend/watchersyncer"
-	cerrors "github.com/projectcalico/libcalico-go/lib/errors"
 )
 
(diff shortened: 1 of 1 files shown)


lib/backend/syncersv1/updateprocessors/bgpconfigprocessor.go
```diff
@@ -52,17 +50,15 @@ var logLevelToBirdLogLevel = func(value interface{}) interface{} {
 	return l
 }
 
+var nodeToNodeMeshEnabled = "{\"enabled\":true}"
+var nodeToNodeMeshDisabled = "{\"enabled\":false}"
+
 // In v1, the node mesh enabled field was wrapped up in some JSON - wrap up the value to
 // return via the syncer.
 var nodeMeshToString = func(value interface{}) interface{} {
 	enabled := value.(bool)
```
caseydavenport: This is going to panic if the provided value is not a `bool` type.

Any idea if that is possible? Might just want to do a safe type check to be sure...

ozdanborne: Looks like everywhere else we just let this potentially panic...

- https://github.com/projectcalico/libcalico-go/blob/7c60403ff17a5c4899031393c6bbc563a6db39ae/lib/backend/syncersv1/updateprocessors/clusterinfoprocessor.go#L49
- https://github.com/projectcalico/libcalico-go/blob/7c60403ff17a5c4899031393c6bbc563a6db39ae/lib/backend/syncersv1/updateprocessors/felixconfigprocessor.go#L44

Issue #795, opened 2018-02-27
"make clean && make all" fails due to deepcopy gen


ozdanborne:
I'm running `make clean && make all` and my builds are failing because `make clean` removes the deepcopy boilerplate code, but none of the other steps (`all`, `test`) recreate the deepcopy code.

## Expected Behavior

`make` should succeed after `make clean`.

## Current Behavior

`make` fails after `make clean`.

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Your Environment

* libcalico-go version:
* Operating System and version:
* Link to your project (optional):

ozdanborne:
cc: @bcreane

bcreane:
fixed in libcalico-go `master`. 

@ozdanborne - should this fix get back ported to v3.0 / v2.0 branches?

caseydavenport:
@bcreane if you could open up cherry-picks against those branches that would be great.

bcreane:
This PR isn't relevant to v2.6.

PR #796, opened 2018-02-28
Don't clean generated files

bcreane:
## Description

`make clean` no longer removes generated files. In addition, we weren't generating the deepcopy-gen associated with `upgrade/migrator/clients/v1/k8s/custom`, so I've added a target for that.

There's a new `check-gen-files` target which checks there are no un-committed changes to the generated files. This is connected to the pre-existing `static-checks` target.

Addresses: https://github.com/projectcalico/libcalico-go/issues/795.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


Makefile
```diff
@@ -29,6 +29,10 @@ DOCKER_GO_BUILD := mkdir -p .go-pkg-cache && \
 
 VENDOR_REMADE := false
 
+.PHONY: semaphore-ci
+semaphore-ci: .deepcopy_gen
```
bcreane: Oops, this should depend on test-containerized, not .deepcopy-gen

Makefile
```diff
@@ -29,13 +29,17 @@ DOCKER_GO_BUILD := mkdir -p .go-pkg-cache && \
 
 VENDOR_REMADE := false
 
+.PHONY: semaphore-ci
+semaphore-ci: test-containerized
```
tomdee: Why add a new target for this? You could just add it to static-checks

.gitignore
```diff
@@ -8,3 +8,4 @@ junit.xml
 *.swp
 .generate_exes
 .generate_files
+.deepcopy_gen
```
tomdee: I don't think we need the .deepcopy_gen file

Makefile
```diff
@@ -77,7 +81,7 @@ ut: vendor/.up-to-date
 
 .PHONY: test-containerized
 ## Run the tests in a container. Useful for CI, Mac dev.
-test-containerized: vendor/.up-to-date run-etcd run-kubernetes-master
+test-containerized: vendor/.up-to-date .deepcopy_gen run-etcd run-kubernetes-master
```
tomdee: Does this need to be here? Or should it live next to the git diff check?

Makefile
```diff
@@ -169,16 +173,19 @@ stop-etcd:
...
 	rm -rf vendor .go-pkg-cache
 
-clean-deepcopy-gen:
-	rm -f .deepcopy_gen
-	find $(TOP_SRC_DIR) -name zz_generated* -exec rm {} \;
-
+.PHONY: clean-bin
 clean-bin:
-	rm -rf $(BINDIR) .deepcopy_gen_exes
+	rm -rf $(BINDIR)
```
tomdee: nit: I don't think we really need a separate clean-bin target. Just put the `rm -rf $(BINDIR)` directly under clean:

Makefile
```diff
@@ -169,16 +173,19 @@ stop-etcd:
...
+
+.PHONY: clean-deepcopy-gen
+clean-deepcopy-gen:
```
tomdee: If this is just a developer convenience (it's not from anywhere) then you should add a comment stating that.

Makefile
```diff
@@ -215,13 +222,46 @@ $(BINDIR)/deepcopy-gen:
 	$(DOCKER_GO_BUILD) \
 		sh -c 'go build -o $@ $(LIBCALICO-GO_PKG)/vendor/k8s.io/code-generator/cmd/deepcopy-gen'
 
-# Regenerate all files if the gen exe(s) changed
-.deepcopy_gen: .deepcopy_gen_exes
-	# Generate deep copies
+./lib/upgrade/migrator/clients/v1/k8s/custom/zz_generated.deepcopy.go: \
```
tomdee: Nice. I like having the actual target file as a Makefile target.

Makefile
```diff
@@ -215,13 +222,46 @@ $(BINDIR)/deepcopy-gen:
...
+										./lib/upgrade/migrator/clients/v1/k8s/custom/doc.go \
```
tomdee: Identation seems a little off, below as well

Makefile
```diff
@@ -215,13 +222,46 @@ $(BINDIR)/deepcopy-gen:
...
+										./lib/upgrade/migrator/clients/v1/k8s/custom/globalbgpconfig.go \
```
tomdee: Can we add a wildcard rather than needing to list all the files?

Makefile
```diff
@@ -215,13 +222,46 @@ $(BINDIR)/deepcopy-gen:
...
 	$(DOCKER_GO_BUILD) \
 		sh -c '$(BINDIR)/deepcopy-gen \
 			--v 1 --logtostderr \
 			--go-header-file "./docs/boilerplate.go.txt" \
 			--input-dirs "$(LIBCALICO-GO_PKG)/lib/apis/v3" \
 			--bounding-dirs "github.com/projectcalico/libcalico-go" \
 			--output-file-base zz_generated.deepcopy'
+
+.PHONY: .deepcopy_gen
+.deepcopy_gen:	.deepcopy_gen_exes \
```
tomdee: Yep, we really don't need either of these .files. Can you get rid of them?

Makefile
tomdee: Oh, I just saw your ?ts=4 comment. Maybe it's OK, then.

bcreane:
closing + re-opening to work around semaphore blocked in "pending" state.

tomdee:
Looks awesome

PR #797, opened 2018-02-28
Add better PodCIDR handling (IPv6)

tmjd:
## Description
Handle PodCIDR when it is IPv6. Previously an IPv6 PodCIDR resulted in a panic.

## Release Note

```release-note
Handle IPv6 PodCIDR when using Kubernetes API Datastore
```

### DIFF
--- a/lib/backend/k8s/resources/node.go
+++ b/lib/backend/k8s/resources/node.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2016-2017 Tigera, Inc. All rights reserved.
+// Copyright (c) 2016-2018 Tigera, Inc. All rights reserved.
 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


lib/backend/k8s/resources/node.go
```diff
@@ -268,6 +274,10 @@ func getTunnelIp(n *kapiv1.Node) string {
 	// We need to get the IP for the podCIDR and increment it to the
 	// first IP in the CIDR.
 	tunIp := ip.To4()
+	if tunIp == nil {
+		log.WithField("podCIDR", n.Spec.PodCIDR).Errorf("Invalid podCIDR for IPv4 Tunnel")
```
caseydavenport: I don't think this is really an Error log - is should probably be Debug level, but I'd also be happy at Info.

It's not an error to have an IPv6 pod cidr, it's just that we don't want to assign a tunnel address if it's IPv6.

caseydavenport: Oh, I see. We're also doing a check up above, so this code will ideally never see an IPv6 address.

Seems like we could simplify this by only doing the check in one place. Do we need the additional `To4` check above as well as here?

tmjd: It is possible that the address could be an IPv4 embedded in an IPv6 so it could make it into the code so the `To4()` should still be done.
The nil check is not necessary currently as there shouldn't be a way to get to that point and have To4 return nil.

caseydavenport: OK how about this:

- We demote the log level here to Info and adjust the text to say `"Cannot pick an IPv4 tunnel address from the given CIDR"`

- We update the function comment to explain that if it can't pick an IPv4 address for the node, it will return an empty string.

- We leave the check above, so we shouldn't actually hit this anyway.

tmjd:
I believe this is ready to go. @caseydavenport PTAL

PR #801, opened 2018-03-01
Handle no labels on Profile conversions

tmjd:
## Description

This is fixing a panic @lwr20 observed when testing out OpenStack upgrading.

This change handles Profiles that have no Labels but do have Tags.

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/upgrade/converters/profile.go
+++ b/lib/upgrade/converters/profile.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2017 Tigera, Inc. All rights reserved.
+// Copyright (c) 2017-2018 Tigera, Inc. All rights reserved.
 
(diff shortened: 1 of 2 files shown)


lib/upgrade/converters/profile.go
```diff
@@ -85,6 +85,9 @@ func (_ Profile) BackendV1ToAPIV3(kvp *model.KVPair) (Resource, error) {
 
 	// Merge Tags and Labels into LabelsToApply.
 	combinedLabelsToApply := bp.Labels
+	if nil == combinedLabelsToApply && len(bp.Tags) != 0 {
```
robbrockbank: it's more conventional to put the variable first:  `combinedLabelsToApply == nil`

PR #804, opened 2018-03-01
OpenStack derived object migration filters

tmjd:
## Description

This adds filters to the migration to ignore the OpenStack derived WorkloadEndpoints and Profiles.

## Release Note

```release-note
Skip OpenStack derived WorkloadEndpoints and Profiles when migrating to the new data format.
```

### DIFF
--- a/lib/upgrade/migrator/migrate.go
+++ b/lib/upgrade/migrator/migrate.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2017 Tigera, Inc. All rights reserved.
+// Copyright (c) 2017-2018 Tigera, Inc. All rights reserved.
 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -351,10 +351,16 @@ var filterGNP = func(k model.Key) bool {
 	return strings.HasPrefix(gk.Name, "knp.default.")
 }
 
-// Filter to filter out K8s (namespace) backed profiles
+// Filter to filter out K8s (namespace) and OpenStack backed profiles
 var filterProfile = func(k model.Key) bool {
 	gk := k.(model.ProfileKey)
-	return strings.HasPrefix(gk.Name, "k8s_ns.")
+	return strings.HasPrefix(gk.Name, "k8s_ns.") || strings.HasPrefix(gk.Name, "openstack-sg-")
+}
+
+// Filter to filter out OpenStack backed workload endpoints
+var filterWEP = func(k model.Key) bool {
+	gk := k.(model.WorkloadEndpointKey)
+	return gk.OrchestratorID == "openstack"
 }
 
 type ic interface {
@@ -473,7 +479,7 @@ func (m *migrationHelper) queryAndConvertResources() (*MigrationData, error) {
 		m.statusBullet("handling WorkloadEndpoint resources")
 		// Query and convert the WorkloadEndpoints
 		if err := m.queryAndConvertV1ToV3Resources(
-			data, model.WorkloadEndpointListOptions{}, converters.WorkloadEndpoint{}, noFilter,
+			data, model.WorkloadEndpointListOptions{}, converters.WorkloadEndpoint{}, filterWEP,
 		); err != nil {
 			return nil, err
 		}
--- a/lib/upgrade/migrator/migrate.go
+++ b/lib/upgrade/migrator/migrate.go
@@ -360,7 +360,7 @@ var filterProfile = func(k model.Key) bool {
 // Filter to filter out OpenStack backed workload endpoints
 var filterWEP = func(k model.Key) bool {
 	gk := k.(model.WorkloadEndpointKey)
-	return gk.OrchestratorID == "openstack"
+	return gk.OrchestratorID == apiv3.OrchestratorOpenStack
 }
 
 type ic interface {
--- a/lib/upgrade/migrator/migrate_test.go
+++ b/lib/upgrade/migrator/migrate_test.go
@@ -25,6 +25,7 @@ import (
 	cerrors "github.com/projectcalico/libcalico-go/lib/errors"
 	"github.com/projectcalico/libcalico-go/lib/net"
 	"github.com/projectcalico/libcalico-go/lib/upgrade/converters"
+	"github.com/projectcalico/libcalico-go/lib/upgrade/migrator/clients"
 )
 
 var _ = Describe("UT for checking the version for migration.", func() {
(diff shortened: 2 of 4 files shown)


lib/backend/model/profile.go
```diff
@@ -27,7 +27,7 @@ import (
 )
 
 var (
-	matchProfile = regexp.MustCompile("^/?calico/v1/policy/profile/([^/]+)/(tags|rules|labels)$")
+	matchProfile = regexp.MustCompile("^/?calico/v1/policy/profile/([^/]+)(/tags|/rules|/labels)?$")
```
tmjd: I'm not sure if this should be changed. It seemed reasonable to me especially with what the ProfileListOptions seems to expect.  My motivation for the change was purely for testing reasons as it allowed what I have with the Profile test.

robbrockbank: i don't think we used the actual matched values, rather we just used this to check a match, and the two variants of this regex both have positive matched on the same thing.

I assume you must be using the match values and you wanted "/tag" instead of "tag".

It does feel a little odd returning "/tag" rather than "tag" tbh.

lib/upgrade/migrator/migrate.go
```diff
@@ -351,10 +351,16 @@ var filterGNP = func(k model.Key) bool {
...
 var filterProfile = func(k model.Key) bool {
 	gk := k.(model.ProfileKey)
-	return strings.HasPrefix(gk.Name, "k8s_ns.")
+	return strings.HasPrefix(gk.Name, "k8s_ns.") || strings.HasPrefix(gk.Name, "openstack-sg-")
+}
+
+// Filter to filter out OpenStack backed workload endpoints
+var filterWEP = func(k model.Key) bool {
+	gk := k.(model.WorkloadEndpointKey)
+	return gk.OrchestratorID == "openstack"
```
neiljerram: Use the `OrchestratorOpenStack` constant here?

lib/upgrade/migrator/migrate_test.go
```diff
@@ -57,3 +63,150 @@ var _ = Describe("UT for checking the version for migration.", func() {
...
+		// Convert the data back to a set of resources.
+		mh := &migrationHelper{clientv1: clientv1}
+		data, err := mh.queryAndConvertResources()
+		Expect(err).NotTo(HaveOccurred())
+		Expect(data.ConversionErrors).To(HaveLen(0))
+		By("Checking total conversion")
+		Expect(data.Resources).To(HaveLen(0))
+	})
+
+	It("should not filter Profiles without openstack-sg", func() {
```
neiljerram: I think it would be clearer to add "prefix" at the end of that description

lib/upgrade/migrator/migrate_test.go
```diff
@@ -57,3 +63,150 @@ var _ = Describe("UT for checking the version for migration.", func() {
 		Entry("Expect 1.2.3.4.5 to not migrate", "1.2.3.4.5", false, true),
 	)
 })
+
+type singleTypeClient struct {
```
neiljerram: Why is `singleTypeClient` needed for the Profile tests (given that `fakeClientV1` apparently works fine for the WEP tests)?

lib/upgrade/migrator/migrate_test.go
```diff
@@ -57,3 +63,150 @@ var _ = Describe("UT for checking the version for migration.", func() {
...
+		}
+		clientv1.kvps = kvps
+
+		// Convert the data back to a set of resources.
+		mh := &migrationHelper{clientv1: clientv1}
+		data, err := mh.queryAndConvertResources()
+		Expect(err).NotTo(HaveOccurred())
+		Expect(data.ConversionErrors).To(HaveLen(0))
+		By("Checking total conversion")
+		Expect(data.Resources).To(HaveLen(1))
```
neiljerram: Perhaps commonize those lines into a utility routine whose only parameter is the expected number of converted resources?

lib/upgrade/migrator/migrate_test.go
```diff
@@ -57,3 +63,150 @@ var _ = Describe("UT for checking the version for migration.", func() {
...
+	It("should not filter WorkloadEndpoints without openstack as OrchestratorID", func() {
+		clientv1 := fakeClientV1{}
+
+		kvps := []*model.KVPair{
+			&model.KVPair{
+				Key:   wk,
+				Value: &wv,
+			},
+		}
+		clientv1.kvps = kvps
```
neiljerram: Perhaps merge those two statements to `client.v1.kvps = []*model.KVPair...` ?

lib/upgrade/migrator/migrate_test.go
tmjd: The reason it does not work is because the fakeClientV1 looks through the Keys to see if they match the ListOptions Key to generate a list of the KV pairs.  To do this it takes the Keys and does a KeyToDefaultPath which for a model.ProfileKey produces `/calico/v1/policy/profile/profileName` which is, unfortunately, not matched by the Profile match regex https://github.com/projectcalico/libcalico-go/blob/master/lib/backend/model/profile.go#L30.  I'm willing to switch back to the fakeClientV1 if you can help me make it work, I did spend some time trying to use it.

lib/upgrade/migrator/migrate_test.go
tmjd: I don't understand the suggestion here, could you elaborate please?

lib/upgrade/migrator/migrate_test.go
neiljerram: Hmm, it feels to me that something is wrong here.  AFAIU, Profile is just as valid as WorkloadEndpoint, in both v1 and v3 data models - so any approach that works for WorkloadEndpoint should work for Profile as well.  I suggest checking this with Rob.

lib/upgrade/migrator/migrate_test.go
neiljerram: Bearing in mind that this set of lines is repeated 4 times, I just meant something like this:

```
func expectConvertedResources(client <whatever client's type is>, num int) {
    mh := &migrationHelper{clientv1: client}
    data, err := mh.queryAndConvertResources()
    Expect(data.ConversionErrors).To(HaveLen(0))
    By("Checking total conversion")
    Expect(data.Resources).To(HaveLen(num))
}
```

tmjd: ah ok, I was thrown off by 'only parameter'.  Thanks

tmjd:
@neiljerram I had considered and tried to make those changes to change the regex but could not come up with a satisfactory change.  I wasn't a big fan of that change myself because I don't like changing 'production' code to make a test work which is effectively what that would be doing.

PR #806, opened 2018-03-05
Fix bug where IPs were assigned from affine blocks of disabled pools.

### LINKED ISSUE #689: Addresses are allocated from disabled ipPools
When the `spec.disabled` parameter in an ipPool is set to `true`, IP addresses are still allocated from the disabled pool when using `calico-ipam`.

## Expected Behavior

Intuitively, I would expect that after disabling an ipPool w/ `calicoctl apply -f ippool.yaml` (using the yaml below), IP addresses will no longer be allocated from the pool that has been disabled.  If this is not an accurate assessment, the documentation should be updated to indicate that the disabled flag only applies to the allocation of new affine blocks.

`ippool.yaml`:

```yaml
apiVersion: v1
kind: ipPool
metadata:
  cidr: 10.125.128.0/17
spec:
  ipip:
    enabled: true
  nat-outgoing: true
  disabled: true
```

## Current Behavior

From looking at the [`autoAssign` method in `lib/client/ipam.go`](https://github.com/projectcalico/libcalico-go/blob/5cabf71a9999834727d83036e986836084e0a13a/lib/client/ipam.go#L158-L287), if a host has already been assigned a /26 block (is that what is called an "affine block?") from the disabled ipPool and the block appears first (lexographically, maybe?) in etcd and there are available IP addresses in the /26, then an IP will be allocated from the disabled pool **without ever checking if the pool is disabled**. It seems like the only place [`spec.Disabled` is checked](https://github.com/projectcalico/libcalico-go/blob/5cabf71a9999834727d83036e986836084e0a13a/lib/client/ipam.go#L254-L256) is after a number of different assignment/allocation attempts have failed.

1. Allocate an ipPool
2. Allocate a second ipPool
3. Disable the first ipPool, by setting `spec.disabled: true`
4. Request new IPs by creating new Pods: IPs will come from both pools :point_up:

## Context

We ran into this at GitHub after discovering that the wrong Pod CIDR was allocated to a production Kubernetes cluster and we needed to migrate to a new Pod CIDR without downtime. Our plan was to add a new (correct) ipPool, disable the old (errant) ipPool, then remove the old ipPool once all Pods had IPs from the correct CIDR. The idea was that Pods that had IPs from the disabled pool would still be routable, but no *new* IPs would be allocated from the disabled pool.

## Your Environment
* libcalico-go version: v1.7.1 ([via](https://github.com/projectcalico/cni-plugin/blob/c7960b84977259c8185c8eff5794ce3cfb6d073e/glide.yaml#L20-L21) cni-plugin v1.11.0)
* Operating System and version: Debian Jessie
* Kubernetes v1.6.7 using the following CNI config:

```
{
    "name": "k8s-pod-network",
    "cniVersion": "0.1.0",
    "type": "calico",
    "etcd_endpoints": "http://10.1.1.1:2379,http://10.1.1.2:2379,http://10.1.1.3:2379",
    "etcd_key_file": "",
    "etcd_cert_file": "",
    "etcd_ca_cert_file": "",
    "log_level": "debug",
    "ipam": {
        "type": "calico-ipam"
    },
    "policy": {
        "type": "k8s",
        "k8s_api_root": "https://10.2.1.1:443",
        "k8s_auth_token": SOMETOKEN
    },
    "kubernetes": {
        "kubeconfig": "/etc/cni/net.d/calico-kubeconfig"
    }
}
```

Thanks for taking a look!  I'm happy to add more context or help w/ contributing a fix.

ozdanborne:
## Description

Fixes #689 .

Of course, this change will need to be pulled into projectcalico/cni-plugin before we see any results.

## Todos
- [x] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
Fixes a bug where IPs could be assigned from disabled IP pools.
```

### DIFF
--- a/lib/ipam/ipam.go
+++ b/lib/ipam/ipam.go
@@ -195,9 +195,52 @@ func (c ipamClient) getBlockFromAffinity(ctx context.Context, aff *model.KVPair)
 	return b, nil
 }
 
-func (c ipamClient) autoAssign(ctx context.Context, num int, handleID *string, attrs map[string]string, pools []net.IPNet, version ipVersion, host string) ([]net.IP, error) {
+// determinePools compares a list of requested pools with the enabled pools
+// and returns the intersect. If any requested pool does not exist, or is not enabled, an error is returned.
+// If no pools are requested, all enabled pools are returned.
+func (c ipamClient) determinePools(requestedPools []net.IPNet, version ipVersion) ([]net.IPNet, error) {
+	enabledPools, err := c.pools.GetEnabledPools(version.Number)
+	if err != nil {
+		log.WithError(err).Errorf("Error reading configured pools")
+		return nil, err
+	}
+	log.Debugf("enabled IPPools: %v", enabledPools)
 
-	// Start by trying to assign from one of the host-affine blocks.  We
+	if len(requestedPools) > 0 {
+		log.Debugf("requested IPPools: %v", requestedPools)
+		// Build a map so we can lookup existing pools.
+		pm := map[string]bool{}
+		for _, p := range enabledPools {
+			pm[p.String()] = true
+		}
+
+		// Make sure each requested pool exists.
+		for _, rp := range requestedPools {
+			if _, ok := pm[rp.String()]; !ok {
+				// The requested pool doesn't exist.
+				return nil, fmt.Errorf("the given pool (%s) does not exist, or is not enabled", rp.IPNet.String())
+			}
+		}
+
+		return requestedPools, nil
+	}
+
+	return enabledPools, nil
+}
+
+func (c ipamClient) autoAssign(ctx context.Context, num int, handleID *string, attrs map[string]string, requestedPools []net.IPNet, version ipVersion, host string) ([]net.IP, error) {
+	// Start by sanitizing the requestedPools.
+	pools, err := c.determinePools(requestedPools, version)
+	if err != nil {
+		return nil, err
+	}
+
+	// If there are no pools, we cannot assign addresses.
+	if len(pools) == 0 {
+		return nil, errors.New("no configured Calico pools")
+	}
+
+	// First, we try to assign addresses from one of the existing host-affine blocks.  We
 	// always do strict checking at this stage, so it doesn't matter whether
 	// globally we have strict_affinity or not.
 	logCtx := log.WithFields(log.Fields{"host": host})
@@ -262,7 +305,7 @@ func (c ipamClient) autoAssign(ctx context.Context, num int, handleID *string, a
 	}
 
 	// If there are still addresses to allocate, then we've run out of
-	// blocks with affinity.  Before we can assign new blocks or assign in
+	// existing blocks with affinity to this host.  Before we can assign new blocks or assign in
 	// non-affine blocks, we need to check that our IPAM configuration
 	// allows that.
 	config, err := c.GetIPAMConfig(ctx)
--- a/lib/ipam/ipam_block_reader_writer.go
+++ b/lib/ipam/ipam_block_reader_writer.go
@@ -17,7 +17,6 @@ package ipam
 import (
 	"context"
 	"errors"
-	"fmt"
 	"hash/fnv"
 	"math/big"
 	"math/rand"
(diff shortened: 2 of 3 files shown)


lib/ipam/ipam.go
```diff
@@ -195,9 +195,52 @@ func (c ipamClient) getBlockFromAffinity(ctx context.Context, aff *model.KVPair)
...
-func (c ipamClient) autoAssign(ctx context.Context, num int, handleID *string, attrs map[string]string, pools []net.IPNet, version ipVersion, host string) ([]net.IP, error) {
+func (c ipamClient) autoAssign(ctx context.Context, num int, handleID *string, attrs map[string]string, requestedPools []net.IPNet, version ipVersion, host string) ([]net.IP, error) {
+	// Start by sanitizing the requestedPools.
+	pools := []net.IPNet{}
 
-	// Start by trying to assign from one of the host-affine blocks.  We
+	// Get all the configured pools.
+	enabledPools, err := c.pools.GetEnabledPools(version.Number)
+	if err != nil {
+		log.WithError(err).Errorf("Error reading configured pools")
```
caseydavenport: We create a context logger just a little bit below - I think you should move that above and then make sure all the logs are using `logCtx`

lib/ipam/ipam.go
```diff
@@ -195,9 +195,52 @@ func (c ipamClient) getBlockFromAffinity(ctx context.Context, aff *model.KVPair)
 	return b, nil
 }
 
-func (c ipamClient) autoAssign(ctx context.Context, num int, handleID *string, attrs map[string]string, pools []net.IPNet, version ipVersion, host string) ([]net.IP, error) {
+func (c ipamClient) autoAssign(ctx context.Context, num int, handleID *string, attrs map[string]string, requestedPools []net.IPNet, version ipVersion, host string) ([]net.IP, error) {
+	// Start by sanitizing the requestedPools.
+	pools := []net.IPNet{}
 
-	// Start by trying to assign from one of the host-affine blocks.  We
+	// Get all the configured pools.
```
caseydavenport: Probably worth encapsulating the pool determination logic into its own function for easier testing and to clean up the `autoAssign` code a bit.

lib/ipam/ipam.go
```diff
@@ -195,9 +195,52 @@ func (c ipamClient) getBlockFromAffinity(ctx context.Context, aff *model.KVPair)
...
+		for _, p := range enabledPools {
+			if isPoolInRequestedPools(p, requestedPools) {
+				pools = append(pools, p)
+			}
+		}
+	} else {
+		// If no pools were requested, default to all enabled pools.
+		pools = enabledPools
+	}
+	log.Debugf("Finding an unclaimed block from pools: %v", pools)
```
caseydavenport: This log is wrong now - we're not finding an unclaimed block yet (we might use an already claimed one)

lib/ipam/ipam.go
caseydavenport: This code then becomes:

```go
pools, err := c.determinePools(requestedPools)
if err != nil { ... }
```

lib/ipam/ipam.go
```diff
@@ -195,9 +195,52 @@ func (c ipamClient) getBlockFromAffinity(ctx context.Context, aff *model.KVPair)
...
+	pm := map[string]bool{}
+	for _, p := range enabledPools {
+		pm[p.String()] = true
+	}
+
+	// Make sure each requested pool exists.
+	for _, rp := range requestedPools {
+		if _, ok := pm[rp.String()]; !ok {
+			// The requested pool doesn't exist.
+			return nil, fmt.Errorf("the given pool (%s) does not exist, or is not enabled", rp.IPNet.String())
```
ozdanborne: @caseydavenport I'm wondering if erroring here is a bit harsh. We could just drop any pools that do not exist, just like we drop any pools that aren't enabled.

ozdanborne: Actually, I'm thinking this code is duplicate of the comparison below. Let me look into it further...

lib/ipam/ipam.go
```diff
@@ -195,9 +195,55 @@ func (c ipamClient) getBlockFromAffinity(ctx context.Context, aff *model.KVPair)
...
+		// Make sure each requested pool exists.
+		for _, rp := range requestedPools {
+			if _, ok := pm[rp.String()]; !ok {
+				// The requested pool doesn't exist.
+				return nil, fmt.Errorf("the given pool (%s) does not exist, or is not enabled", rp.IPNet.String())
+			} else {
+				pools = append(pools, rp)
+			}
+		}
+	} else {
```
caseydavenport: Just a personal preference, but I'd drop the "else" branch here and do something like this:

```
if len(requestedPools) > 0 {
  // Build a map
  ...

  // Make sure each pool exists
  ...

  // Return the requested pools.
  return requestedPools
}

return enabledPools
```

I think it makes the code a bit easier to follow.

ozdanborne: Neat. Way cleaner. Fixed

lib/ipam/ipam.go
```diff
@@ -195,9 +195,54 @@ func (c ipamClient) getBlockFromAffinity(ctx context.Context, aff *model.KVPair)
...
+		for _, rp := range requestedPools {
+			if _, ok := pm[rp.String()]; !ok {
+				// The requested pool doesn't exist.
+				return nil, fmt.Errorf("the given pool (%s) does not exist, or is not enabled", rp.IPNet.String())
+			} else {
+				pools = append(pools, rp)
+			}
+		}
+
+		return pools, nil
```
caseydavenport: One more nit here - it's bad to mix named and unnamed return values.

I'd suggest removing the named return value from above, and simply returning `requestedPools` if we don't hit an error in the loop above.

That's nice because ultimately this function will always return either the requested pools, or all enabled pools and it becomes more obvious with that change.

caseydavenport: e.g. something like this:

```go
for _, rp := range requestedPools{
  if !ok {
    ERROR!
  }
}

// Requested pools are valid
return requestedPools, nil
```

lib/ipam/ipam.go
```diff
@@ -195,9 +195,54 @@ func (c ipamClient) getBlockFromAffinity(ctx context.Context, aff *model.KVPair)
...
+		for _, p := range enabledPools {
+			pm[p.String()] = true
+		}
+
+		// Make sure each requested pool exists.
+		for _, rp := range requestedPools {
+			if _, ok := pm[rp.String()]; !ok {
+				// The requested pool doesn't exist.
+				return nil, fmt.Errorf("the given pool (%s) does not exist, or is not enabled", rp.IPNet.String())
+			} else {
```
gunjan5: nit: don't need the `else` here
(not reviewing but I was flying by and saw this)

PR #809, opened 2018-03-06
Add support for stopping watcher syncers

tomdee:
When a watchersyncer is stopped, it will emit all the cached resources
that it knew about as delete events.

### DIFF
--- a/lib/backend/api/api.go
+++ b/lib/backend/api/api.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2016-2017 Tigera, Inc. All rights reserved.
+// Copyright (c) 2016-2018 Tigera, Inc. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


robbrockbank:
Also, some of the copyrights need updating.

lib/backend/api/api.go
```diff
@@ -120,6 +120,7 @@ type Client interface {
 type Syncer interface {
 	// Starts the Syncer.  May start a background goroutine.
 	Start()
+	Stop()
```
robbrockbank: Probably should put a comment for Stop since we have one for Starts.

lib/backend/watchersyncer/watchersyncer_test.go
```diff
@@ -1107,7 +1160,7 @@ func (w *watcher) run(results <-chan api.WatchEvent) {
 				return
 			}
 
-		// Exit if we receive a done notification.
+			// Exit if we receive a done notification.
```
caseydavenport: Weird indentation here

robbrockbank:
squash please.

PR #810, opened 2018-03-08
Remove syncer() method from client interface.

tomdee:
## Description
It's no longer a part of the client, so this method can lead to circular
dependencies.

## Release Note
```release-note
Syncer() method is no longer part of the client interface.
```

### DIFF
--- a/lib/backend/k8s/k8s.go
+++ b/lib/backend/k8s/k8s.go
@@ -32,7 +32,7 @@ import (
 	cerrors "github.com/projectcalico/libcalico-go/lib/errors"
 	"github.com/projectcalico/libcalico-go/lib/net"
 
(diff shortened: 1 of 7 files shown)


lib/backend/k8s/k8s.go
```diff
@@ -32,7 +32,7 @@ import (
 	cerrors "github.com/projectcalico/libcalico-go/lib/errors"
 	"github.com/projectcalico/libcalico-go/lib/net"
 
-	"github.com/projectcalico/libcalico-go/lib/backend/syncersv1/felixsyncer"
+	//"github.com/projectcalico/libcalico-go/lib/backend/syncersv1/felixsyncer"
```
caseydavenport: @tomdee could you remove the commented out import?

Issue #716, opened 2017-12-02
k8s_test.go tries to defer It() blocks and assumes test ordering


fasaxc:
Code reading k8s_test.go, it's trying to do things like deferring `It()` calls, which makes no sense given ginkgo's approach to collecting specs. 

The correct way to run teardown code is in an AfterEach or AfterSuite block.

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Your Environment

* libcalico-go version:
* Operating System and version:
* Link to your project (optional):

caseydavenport:
@fasaxc what is the impact of this? Could it result in flaky tests?

Could you assign a priority to this issue?

fasaxc:
Depends how we run ginkgo.  If ginkgo is configured to randomise the order of these specs then the deferred `It()` may run out-of-order.  If not, then the test will do what the developer expected (run the tear down test last), but for the wrong reason.

fasaxc:
I'd put it at a P2: it's wrong and it could cause confusing hard to understand failures later but it's not causing any immediate harm.

fasaxc:
@doublek you wrote these tests, would you be able to fix them up at some point?

PR #813, opened 2018-03-09
Remove various defers from k8s_test.go

caseydavenport:
## Description

Fixes https://github.com/projectcalico/libcalico-go/issues/716

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/backend/k8s/k8s_test.go
+++ b/lib/backend/k8s/k8s_test.go
@@ -320,6 +320,41 @@ var _ = Describe("Test Syncer API for Kubernetes backend", func() {
 		go cb.ProcessUpdates()
 	})
 
+	AfterEach(func() {
+		// Clean up all Calico resources.
+		err := c.Clean()
+		Expect(err).NotTo(HaveOccurred())
+
+		// Clean up any k8s network policy left over by the test.
+		nps := networkingv1.NetworkPolicyList{}
+		err = c.clientSet.NetworkingV1().RESTClient().
+			Get().
+			Resource("networkpolicies").
+			Timeout(10 * time.Second).
+			Do().Into(&nps)
+		Expect(err).NotTo(HaveOccurred())
+
+		for _, np := range nps.Items {
+			result := c.clientSet.NetworkingV1().RESTClient().
+				Delete().
+				Resource("networkpolicies").
+				Namespace(np.Namespace).
+				Name(np.Name).
+				Timeout(10 * time.Second).
+				Do()
+			Expect(result.Error()).NotTo(HaveOccurred())
+		}
+
+		// Clean up any pods left over by the test.
+		pods, err := c.clientSet.CoreV1().Pods("").List(metav1.ListOptions{})
+		Expect(err).NotTo(HaveOccurred())
+
+		for _, p := range pods.Items {
+			err = c.clientSet.CoreV1().Pods(p.Namespace).Delete(p.Name, &metav1.DeleteOptions{})
+			Expect(err).NotTo(HaveOccurred())
+		}
+	})
+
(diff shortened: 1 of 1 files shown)


lib/backend/k8s/k8s_test.go
```diff
@@ -320,6 +320,41 @@ var _ = Describe("Test Syncer API for Kubernetes backend", func() {
...
+		// Clean up any k8s network policy left over by test.
+		nps := networkingv1.NetworkPolicyList{}
+		err := c.clientSet.NetworkingV1().RESTClient().
+			Get().
+			Resource("networkpolicies").
+			Timeout(10 * time.Second).
+			Do().Into(&nps)
+		Expect(err).NotTo(HaveOccurred())
+
+		for _, np := range nps.Items {
```
gunjan5: could we not just delete the namespace to get rid of all these things? I guess we can't delete the default namespace, but maybe if we create a namespace for every test (kinda like upstream k8s NetworkPolicy e2es) and just create the namespace in `BeforeEach` and delete it in `AfterEach`? Maybe it's too much work or we need to make sure things work in the default namespace, in which case this is fine.

lib/backend/k8s/k8s_test.go
```diff
@@ -491,39 +503,16 @@ var _ = Describe("Test Syncer API for Kubernetes backend", func() {
...
-					Timeout(10 * time.Second).
-					Do().Into(&nps)
-				Expect(err).NotTo(HaveOccurred())
-				time.Sleep(1 * time.Second)
-			}
-			panic(fmt.Sprintf("Failed to clean up policies: %+v", nps))
-		})
-	}()
+		// Delete the object.
+		result := c.clientSet.NetworkingV1().RESTClient().
```
gunjan5: bit confused, why do we need this one? Shouldn't it be taken care of in `AfterEach`? I'm sure there's a reason, but im blanking out

lib/backend/k8s/k8s_test.go
```diff
@@ -491,39 +503,16 @@ var _ = Describe("Test Syncer API for Kubernetes backend", func() {
...
+			Delete().
+			Resource("networkpolicies").
+			Namespace("default").
+			Name(np.ObjectMeta.Name).
```
gunjan5: why don't we have `Timeout(10 * time.Second).` here?

lib/backend/k8s/k8s_test.go
```diff
@@ -589,13 +578,6 @@ var _ = Describe("Test Syncer API for Kubernetes backend", func() {
 			},
 		}
 
-		// Make sure we clean up after ourselves.  We allow this to fail because
-		// part of our explicit testing below is to delete the resource.
-		defer func() {
-			gnpClient.Delete(ctx, kvp1a.Key, "")
```
gunjan5: where is the equivalent code for this (in `AfterEach`?)?

caseydavenport: This is part of the `c.Clean` call which cleans up all Calico resources.

lib/backend/k8s/k8s_test.go
caseydavenport: Yea, should just remove this actually.

lib/backend/k8s/k8s_test.go
gunjan5: ah yes! right right

lib/backend/k8s/k8s_test.go
caseydavenport: I suppose we could, but that's extra work for questionable gain. I think we should just clean up the things we know we may have created for now.

caseydavenport:
@gunjan5 all cleaned up

PR #814, opened 2018-03-09
Don't add/remove 'ossg.default.' NP prefix when to/from storage

neiljerram:
This make treatment of the 'ossg.default.' prefix consistent with that for 'knp.default.'.

### DIFF
--- a/lib/clientv3/globalnetworkpolicy.go
+++ b/lib/clientv3/globalnetworkpolicy.go
@@ -208,12 +208,20 @@ func convertPolicyNameForStorage(name string) string {
 	if strings.HasPrefix(name, "knp.") {
 		return name
 	}
+	// Similarly for "ossg."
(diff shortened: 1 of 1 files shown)


lib/clientv3/globalnetworkpolicy.go
```diff
@@ -216,6 +220,10 @@ func convertPolicyNameFromStorage(name string) string {
 	if strings.HasPrefix(name, "knp.") {
 		return name
 	}
+	// Similarly for "ossg."
+	if strings.HasPrefix(name, "ossg.") {
```
fasaxc: This function is in the _global_ network policies file but I thought we're writing namespaced network policies?

fasaxc: Looks like the functions are shared and used in both

fasaxc:
Continued in #817

PR #815, opened 2018-03-09
Remove a panic in ipam_block

### LINKED ISSUE #772: There should be no explicit panic() calls in this library
This is somewhat related to https://github.com/projectcalico/libcalico-go/issues/770, libcalico-go being at the core of calico node should never be panicing. If it ever panics, whole  calico mesh goes down without a chance of recovery, since cause of panic would likely be processes again and again.

ozdanborne:
## Description

Remove a panic in ipam_block.

Final fixes #772 

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/ipam/ipam_block.go
+++ b/lib/ipam/ipam_block.go
@@ -122,9 +122,9 @@ func (b *allocationBlock) assign(address cnet.IP, handleID *string, attrs map[st
 	}
 
 	// Convert to an ordinal.
-	ordinal := ipToOrdinal(address, *b)
-	if (ordinal < 0) || (ordinal > blockSize) {
-		return errors.New("IP address not in block")
+	ordinal, err := ipToOrdinal(address, *b)
+	if err != nil {
+		return err
 	}
 
 	// Check if already allocated.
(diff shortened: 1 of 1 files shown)


lib/ipam/ipam_block.go
```diff
@@ -416,15 +416,14 @@ func incrementIP(ip cnet.IP, increment *big.Int) cnet.IP {
...
 
-func ipToOrdinal(ip cnet.IP, b allocationBlock) int {
+func ipToOrdinal(ip cnet.IP, b allocationBlock) (int, error) {
 	ip_int := ipToInt(ip)
 	base_int := ipToInt(cnet.IP{b.CIDR.IP})
 	ord := big.NewInt(0).Sub(ip_int, base_int).Int64()
 	if ord < 0 || ord >= blockSize {
-		// IP address not in the given block.
-		log.Panicf("IP %s not in block %s", ip, b.CIDR)
+		return 0, errors.New(fmt.Sprintf("IP %s not in block %s", ip, b.CIDR))
```
ozdanborne: @caseydavenport (or @gunjan5 ? ) What would you do when you can't return nil as an integer here? Convert to a struct or pointer?

gunjan5: returning the zero value is fine, I think, if you're also returning the error with it. Calling code should be checking the error before using the int value

gunjan5: also just use `fmt.Errorf("IP %s not in block %s", ip, b.CIDR)`

caseydavenport: I'd just return the zero value, like you're doing, and then have the calling code check errors.

PR #817, opened 2018-03-12
Don't add/remove 'ossg.default.' NP prefix when to/from storage

fasaxc:
## Todos
- [x] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/clientv3/networkpolicy_e2e_test.go
+++ b/lib/clientv3/networkpolicy_e2e_test.go
@@ -20,6 +20,7 @@ import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/ginkgo/extensions/table"
 	. "github.com/onsi/gomega"
+	"github.com/projectcalico/libcalico-go/lib/backend/model"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"context"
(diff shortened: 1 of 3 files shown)


lib/clientv3/networkpolicy_e2e_test.go
```diff
@@ -468,4 +466,96 @@ var _ = testutils.E2eDatastoreDescribe("NetworkPolicy tests", testutils.Datastor
...
+			Expect(inNp.GetName()).To(Equal(name), "Create() shouldn't touch input data")
+			Expect(np.GetName()).To(Equal(name), "Create() should return the data as we'd read it")
+
+			By("Getting the right policy with either name")
+			np, err = c.NetworkPolicies().Get(ctx, namespace1, name, options.GetOptions{})
+			Expect(err).NotTo(HaveOccurred())
+			Expect(np.GetName()).To(Equal(name))
+			Expect(np.Spec).To(Equal(ingressTypesSpec1))
+
+			np, err = c.NetworkPolicies().Get(ctx, namespace1, name, options.GetOptions{})
```
neiljerram: This group of lines looks exactly the same as the previous - is that intended?

lib/clientv3/networkpolicy_e2e_test.go
```diff
@@ -468,4 +466,96 @@ var _ = testutils.E2eDatastoreDescribe("NetworkPolicy tests", testutils.Datastor
...
+		)
+	}
+	DescribeTable("name round-tripping tests",
+		func(name string) {
+			c, err := clientv3.New(config)
+			Expect(err).NotTo(HaveOccurred())
+
+			be, err := backend.NewClient(config)
+			Expect(err).NotTo(HaveOccurred())
+			be.Clean()
```
neiljerram: Is the backend client involved at all in these tests?  It's not clear to me if these tests are covering the thing that we most need, i.e. the mapping between API name and storage name for "ossg." NPs.  (But perhaps they are and I'm just not seeing it.)

lib/clientv3/networkpolicy_e2e_test.go
fasaxc: Earlier draft of the test had two names that I was expecting to alias but it turns out that the aliased name isn't allowed.  (I was expecting "foo-bar" and "default.foo-bar" to both be allowed but only the former is allowed.)

PR #818, opened 2018-03-13
Support a custom host for health server

renan:
Instead of forcing to listen on all interfaces we should give the possibility for the user to make the health server listen on a specific host/IP.

[Felix component](https://github.com/projectcalico/felix/blob/master/felix.go) should also be updated to support this flag.

## Todos
- ~[ ] Tests~
- [ ] Documentation
- [ ] Release note

### DIFF
--- a/lib/health/health.go
+++ b/lib/health/health.go
@@ -164,21 +164,25 @@ const (
 	StatusBad = 503
 )
 
-// ServeHTTP publishes the current overall liveness and readiness at http://*:PORT/liveness and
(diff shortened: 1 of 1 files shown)


renan:
This appears to have no tests tho, and my limited knowledge of Golang prevents me from writing an integration test for it :/

neiljerram:
@renan Yes, you're off the hook there, because we don't yet have any test for ServeHTTP at all. (I think I have a TODO for it somewhere.)

PR #835, opened 2018-03-30
Support watching a single node

caseydavenport:
## Description

Fixes https://github.com/projectcalico/calico/issues/1739

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/backend/k8s/k8s_test.go
+++ b/lib/backend/k8s/k8s_test.go
@@ -1769,3 +1769,65 @@ var _ = Describe("Test Syncer API for Kubernetes backend", func() {
 		})
 	})
 })
+
+var _ = Describe("Test Watch support", func() {
+	var (
+		c   *KubeClient
+		ctx context.Context
+	)
+
+	BeforeEach(func() {
+		// Create a client
+		cfg := apiconfig.CalicoAPIConfigSpec{KubeConfig: apiconfig.KubeConfig{K8sAPIEndpoint: "http://localhost:8080"}}
+		client, err := NewKubeClient(&cfg)
+		Expect(err).NotTo(HaveOccurred())
+		c = client.(*KubeClient)
+
+		ctx = context.Background()
+	})
+
+	It("Should support watching Nodes", func() {
+		By("Watching a single node", func() {
+			name := "127.0.0.1" // Node created by test/mock-node.yaml
+			watch, err := c.Watch(ctx, model.ResourceListOptions{Name: name, Kind: apiv3.KindNode}, "")
+			Expect(err).NotTo(HaveOccurred())
+
+			// We should get at least one event from the watch.
+			var receivedEvent bool
+			for i := 0; i < 10; i++ {
+				select {
+				case e := <-watch.ResultChan():
+					// Got an event. Check it's OK.
+					Expect(e.Error).NotTo(HaveOccurred())
+					Expect(e.Type).To(Equal(api.WatchAdded))
+					receivedEvent = true
+					break
+				default:
+					time.Sleep(50 * time.Millisecond)
+				}
+			}
+			Expect(receivedEvent).To(BeTrue(), "Did not receive watch event")
+		})
+
+		By("Watching all nodes", func() {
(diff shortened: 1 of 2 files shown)


shafay2000:
Thanks Casey!

lib/backend/k8s/resources/node.go
```diff
@@ -160,11 +160,16 @@ func (c *nodeClient) EnsureInitialized() error {
...
 func (c *nodeClient) Watch(ctx context.Context, list model.ListInterface, revision string) (api.WatchInterface, error) {
-	if len(list.(model.ResourceListOptions).Name) != 0 {
-		return nil, fmt.Errorf("cannot watch specific resource instance: %s", list.(model.ResourceListOptions).Name)
+	// Build watch options to pass to k8s.
+	opts := metav1.ListOptions{ResourceVersion: revision, Watch: true}
+	name := list.(model.ResourceListOptions).Name
+	if len(name) != 0 {
+		// We've been asked to watch a specific node resource.
+		log.WithField("name", name).Debug("Watching a single node")
+		opts.FieldSelector = fields.OneTermEqualSelector("metadata.name", name).String()
```
robbrockbank: hopefully a quick question.  Does the API server handle this field selector explicitly by constructing the correct watch path for etcdv3, or does this actually end up watching everything, but filter out all other nodes at the API server?

caseydavenport: I don't know the apiserver code super well, but my quick read of it led me to believe that it watches just this one specific object. It seems aware that we've requested a watch on the name of the resource, and has special case code for that.

Even if it internally watched all keys and filtered, that would still be nicer than re-querying every 5 seconds.

caseydavenport: https://github.com/kubernetes/apiserver/blob/release-1.9/pkg/registry/generic/registry/store.go#L1152-L1177

lib/backend/k8s/resources/node.go
```diff
@@ -160,11 +160,16 @@ func (c *nodeClient) EnsureInitialized() error {
 }
 
 func (c *nodeClient) Watch(ctx context.Context, list model.ListInterface, revision string) (api.WatchInterface, error) {
-	if len(list.(model.ResourceListOptions).Name) != 0 {
-		return nil, fmt.Errorf("cannot watch specific resource instance: %s", list.(model.ResourceListOptions).Name)
+	// Build watch options to pass to k8s.
+	opts := metav1.ListOptions{ResourceVersion: revision, Watch: true}
+	name := list.(model.ResourceListOptions).Name
```
tmjd: I see that we didn't check before but should we check that this conversion works?

lib/backend/k8s/k8s_test.go
```diff
@@ -1769,3 +1769,65 @@ var _ = Describe("Test Syncer API for Kubernetes backend", func() {
...
+
+	It("Should support watching Nodes", func() {
+		By("Watching a single node", func() {
+			name := "127.0.0.1" // Node created by test/mock-node.yaml
+			watch, err := c.Watch(ctx, model.ResourceListOptions{Name: name, Kind: apiv3.KindNode}, "")
+			Expect(err).NotTo(HaveOccurred())
+
+			// We should get at least one event from the watch.
+			var receivedEvent bool
+			for i := 0; i < 10; i++ {
```
tmjd: Should we make a change to the resource and see that we get that in the test, to make sure that the watch is setup correctly?

lib/backend/k8s/resources/node.go
caseydavenport: Yes, I suppose that is safest.

lib/backend/k8s/k8s_test.go
caseydavenport: I think this is enough to verify the watch is correct - we verify that we get at least one non-error event.

tmjd:
Closing and reopening to see if that makes license/cia happy.

PR #836, opened 2018-03-30
Remove feature flag check for service accounts.

saumoh:
## Description
Remove the alpha feature flag for service account rules.
Added UT's to cover the case for the default service account per namespace.

### DIFF
--- a/lib/backend/k8s/conversion/conversion.go
+++ b/lib/backend/k8s/conversion/conversion.go
@@ -46,10 +46,7 @@ const (
 )
 
 //TODO: make this private and expose a public conversion interface instead
-type Converter struct {
-	// AlphaSA set to true if the alpha feature for serviceaccounts is enabled.
-	AlphaSA bool
-}
+type Converter struct{}
 
 // VethNameForWorkload returns a deterministic veth name
 // for the given Kubernetes workload (WEP) name and namespace.
@@ -183,7 +180,7 @@ func (c Converter) PodToWorkloadEndpoint(pod *kapiv1.Pod) (*model.KVPair, error)
 	profiles = append(profiles, NamespaceProfileNamePrefix+pod.Namespace)
 
 	// Pull out the Serviceaccount based profile off the pod SA and namespace
-	if c.AlphaSA && pod.Spec.ServiceAccountName != "" {
+	if pod.Spec.ServiceAccountName != "" {
 		profiles = append(profiles, serviceAccountNameToProfileName(pod.Spec.ServiceAccountName, pod.Namespace))
 	}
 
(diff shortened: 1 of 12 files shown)


lib/backend/k8s/conversion/conversion.go
```diff
@@ -700,12 +693,9 @@ func (c Converter) SplitProfileRevision(rev string) (nsRev string, saRev string,
 		return
 	}
 
-	if c.AlphaSA == false {
-		nsRev = rev
-		return
-	}
-
 	revs := strings.Split(rev, "/")
+	// TODO(saumoh): Ask @casey: Should this be an error? or should we return nsRev, ""
```
caseydavenport: I think this is fine, since it's always passed a value as created by `JoinProfileRevisions`, which means it will always have length 2.

lib/backend/k8s/resources/profile.go
```diff
@@ -129,15 +127,11 @@ func (c *profileClient) Get(ctx context.Context, key model.Key, revision string)
 	if err != nil {
 		return nil, err
 	}
-
+	//TODO(saumoh): Convert to switch
```
caseydavenport: Planning on doing this in this PR?

lib/backend/syncersv1/updateprocessors/rules_test.go
```diff
@@ -60,10 +59,6 @@ var _ = Describe("Test the Rules Conversion Functions", func() {
 				NotNets:     []string{"192.168.40.1/32"},
 				NotSelector: "has(label1)",
 				NotPorts:    []numorstring.Port{port443},
-				ServiceAccounts: &apiv3.ServiceAccountMatch{
```
caseydavenport: Why'd you drop this bit (and the part below)?

lib/backend/k8s/resources/profile.go
saumoh: arg. yes!

lib/backend/syncersv1/updateprocessors/rules_test.go
saumoh: The service accounts pieces affected the selectors now but did not add much to this specific test cases here. Hence I took them out of here.
The impact of service accounts on the selector is explicitly captured in the UT [here](https://github.com/colabsaumoh/libcalico-go/blob/0aecad65e5dc40bb786d4ea3668f8491e8e6683e/lib/backend/syncersv1/updateprocessors/rules_test.go#L372) and [here](https://github.com/colabsaumoh/libcalico-go/blob/0aecad65e5dc40bb786d4ea3668f8491e8e6683e/lib/backend/syncersv1/updateprocessors/rules_test.go#L445)

lib/backend/k8s/resources/profile.go
```diff
@@ -130,15 +128,15 @@ func (c *profileClient) Get(ctx context.Context, key model.Key, revision string)
 		return nil, err
 	}
 
-	if strings.HasPrefix(rk.Name, conversion.NamespaceProfileNamePrefix) {
-		return c.getNamespace(ctx, rk, nsRev)
-	}
-
-	if c.AlphaSA == false {
-		return nil, fmt.Errorf("Revision %s invalid", revision)
+	prefixs := strings.SplitAfterN(rk.Name, ".", 2)
```
caseydavenport: I think this should technically be called "splits" or something like that.

PR #839, opened 2018-04-04
Remove feature flag for http and remove feature flag validation

saumoh:
## Description
- Remove feature flag check for http,
- Remove the feature flag validation code as this is no longer used,
- Remove the feature flag in the apiconfig spec.

_NOTE_: I have three distinct commits as i was not sure if we want to revert any particular change. If needed I will squash all together prior to merge.


caseydavenport:
tbh, I'm also happy to leave the mechanism in place in case we want to add more alpha features. WDYT?

saumoh:
sure. I was worried about bit-rot but removing it later is fairly easy in case we do not use it again!
Will revert the last 2 commits.

PR #841, opened 2018-04-04
Handle blank CalicoVersion in ClusterInformation which can occur during startup of Calico components

robbrockbank:
## Description
Start up of some Calico components triggers an EnsureInitialized with empty versions.  The ShouldMigrate() code which is used during KDD migration in the node and typha startup needs to handle an empty version in the cluster information.  It should treat it as if the cluster information was not present.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/upgrade/migrator/migrate.go
+++ b/lib/upgrade/migrator/migrate.go
@@ -80,10 +80,9 @@ func New(clientv3 clientv3.Interface, clientv1 clients.V1ClientInterface, status
 
 // migrationHelper implements the migrate.Interface.
 type migrationHelper struct {
-	clientv3                clientv3.Interface
-	clientv1                clients.V1ClientInterface
-	enforceEmptyDestination bool
-	statusWriter            StatusWriterInterface
+	clientv3     clientv3.Interface
+	clientv1     clients.V1ClientInterface
+	statusWriter StatusWriterInterface
 }
 
 // Error types encountered during validation and migration.
@@ -368,7 +367,7 @@ type ic interface {
 }
 
 func (m *migrationHelper) v3DatastoreIsClean() (bool, error) {
-	bc := m.clientv3.(backend).Backend()
+	bc := m.clientv3.(backendClientAccessor).Backend()
 	if i, ok := bc.(ic); ok {
 		return i.IsClean()
 	}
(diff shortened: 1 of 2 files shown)


lib/upgrade/migrator/migrate.go
```diff
@@ -1066,15 +1071,15 @@ func (m *migrationHelper) migrateIPAMData() error {
 	return nil
 }
 
-// backend is an interface used to access the backend client from the main clientv3.
-type backend interface {
+// backendClientAccessor is an interface used to access the backend client from the main clientv3.
+type backendClientAccessor interface {
```
robbrockbank: just a rename to avoid clashes with backend package.

lib/upgrade/migrator/migrate.go
```diff
@@ -676,18 +675,23 @@ func (m *migrationHelper) queryAndConvertV1ToV3Nodes(data *MigrationData) error
 // If neither the v1 nor v3 versions are present, then no migration is required.
 func (m *migrationHelper) ShouldMigrate() (bool, error) {
 	ci, err := m.clientv3.ClusterInformation().Get(context.Background(), "default", options.GetOptions{})
-	if err == nil {
+	if err == nil && ci.Spec.CalicoVersion != "" {
```
robbrockbank: I did the logic this way to reduce the complexity of the diff.

lib/upgrade/migrator/migrate.go
```diff
@@ -676,18 +675,25 @@ func (m *migrationHelper) queryAndConvertV1ToV3Nodes(data *MigrationData) error
 // If neither the v1 nor v3 versions are present, then no migration is required.
 func (m *migrationHelper) ShouldMigrate() (bool, error) {
 	ci, err := m.clientv3.ClusterInformation().Get(context.Background(), "default", options.GetOptions{})
-	if err == nil {
+    if err == nil && ci.Spec.CalicoVersion != ""{
```
robbrockbank: I did the logic this way to reduce the complexity of the diff.  I coulda done:

```
if err == nil {
  if CalicoVersion != "" {
    ... check if requires migration ...
  }
  ... treat as non-existent ...
}
```

but everything is shifted over and the diff is harder to read.

Can change though if you'd prefer.

robbrockbank: I did the logic this way to reduce the complexity of the diff.  I coulda done:

```
if err == nil {
  if CalicoVersion != "" {
    ... check if requires migration ...
  }
  ... treat as non-existent ...
}
```

but everything is shifted over and the diff is harder to read.

Can change though if you'd prefer.

lib/upgrade/migrator/migrate.go
```diff
@@ -676,18 +675,25 @@ func (m *migrationHelper) queryAndConvertV1ToV3Nodes(data *MigrationData) error
 // If neither the v1 nor v3 versions are present, then no migration is required.
 func (m *migrationHelper) ShouldMigrate() (bool, error) {
 	ci, err := m.clientv3.ClusterInformation().Get(context.Background(), "default", options.GetOptions{})
-	if err == nil {
+	if err == nil && ci.Spec.CalicoVersion == "" {
```
robbrockbank: I did the logic this way to reduce the complexity of the diff.  I coulda done:

```
if err == nil {
  if CalicoVersion != "" {
    ... check if requires migration ...
  }
  ... treat as non-existent ...
}
```

but everything is shifted over and the diff is harder to read.

Can change though if you'd prefer.

PR #858, opened 2018-04-24
Initial commit for http paths match rules

saumoh:
## Description
Add support for HTTP path matching. The path match may be either `exact` or `prefix`. If multiple matches are specified then they are OR'd together.
Sample
```
kind: GlobalNetworkPolicy
metadata:
  name: database
spec:
  selector: app == 'database'
  ingress:
    - action: Allow
      source:
        serviceAccounts:
          names: ["summary"]
      http:
        paths:
         - exact: /v2/keys
         - prefix: /v2/keys/accounts
  egress:
    - action: Allow

---
```

NOTE: @caseydavenport feel free to add other reviewers.


lib/apis/v3/policy.go
```diff
@@ -59,13 +64,25 @@ type Rule struct {
...
+	Methods []string `json:"methods,omitempty" validate:"omitempty,httpMethods"`
+	// Paths is an optional field that restricts the rule to apply to HTTP requests that use one of the listed
+	// HTTP Paths. Paths may be either of the form:
+	// exact: <path> or
+	// prefix: <path-prefix>
+	// e.g:
+	// - exact: /foo
+	// - prefix: /bar
+	// Multiple paths are OR'd together.
+	Paths []map[string]string `json:"paths,omitempty" validate:"omitempty,httpPaths"`
```
fasaxc: Why is this a generic `map[string]string` instead of defining a `HTTPPathMatch` struct?  You could structure it as 
```
type HTTPPathMatch struct {
    Path string
    MatchType string  // one of "prefix" "exact", defaults to exact
}
```

lib/validator/v3/validator.go
```diff
@@ -427,6 +429,49 @@ func validateCIDR(v *validator.Validate, topStruct reflect.Value, currentStructO
...
 }
 
+// validateHTTPMethods validate if the HTTP Method match clauses are valid.
+func validateHTTPMethods(v *validator.Validate, topStruct reflect.Value, currentStructOrField reflect.Value, field reflect.Value, fieldType reflect.Type, fieldKind reflect.Kind, param string) bool {
+	validMethods := []string{"GET", "PUT", "POST", "HEAD", "DELETE", "OPTIONS", "CONNECT", "*"}
+	methods := field.Interface().([]string)
+	for _, method := range methods {
+		match := false
+		for _, v := range validMethods {
+			if strings.Compare(strings.ToUpper(method), v) == 0 {
```
fasaxc: Should we allow lower case here?  I thought we required the correct casing elsewhere, and it means that Felix/Dikastes need to be tolerant to different casing.

lib/validator/v3/validator.go
```diff
@@ -427,6 +429,49 @@ func validateCIDR(v *validator.Validate, topStruct reflect.Value, currentStructO
 	return err == nil
 }
 
+// validateHTTPMethods validate if the HTTP Method match clauses are valid.
+func validateHTTPMethods(v *validator.Validate, topStruct reflect.Value, currentStructOrField reflect.Value, field reflect.Value, fieldType reflect.Type, fieldKind reflect.Kind, param string) bool {
+	validMethods := []string{"GET", "PUT", "POST", "HEAD", "DELETE", "OPTIONS", "CONNECT", "*"}
```
fasaxc: We have a set library, it might give a cleaner way to express this.
```
var validMethods = set.From("GET", "PUT", "POST", "HEAD", "DELETE", "OPTIONS", "CONNECT", "*")

func validateHTTPMethods(v *validator.Validate, topStruct reflect.Value, currentStructOrField reflect.Value, field reflect.Value, fieldType reflect.Type, fieldKind reflect.Kind, param string) bool {
    ...
    for _, method := range methods {
        valid := validMethods.Contains(method)
        if !valid { ... }
    }
    ....
}
```

lib/validator/v3/validator.go
```diff
@@ -427,6 +429,49 @@ func validateCIDR(v *validator.Validate, topStruct reflect.Value, currentStructO
...
+		}
+		for k, v := range path {
+			switch api.HTTPPathMatchType(strings.ToLower(k)) {
+			case api.HTTPPatchMatchPrefix, api.HTTPPathMatchExact:
+				//no op
+			default:
+				return false
+			}
+			// Checks from https://tools.ietf.org/html/rfc3986#page-22
+			if strings.HasPrefix(v, "/") == false ||
```
fasaxc: Prefer `!strings.HasPrefix()` over comparing to `false`.

lib/validator/v3/validator.go
```diff
@@ -427,6 +429,49 @@ func validateCIDR(v *validator.Validate, topStruct reflect.Value, currentStructO
 	return err == nil
 }
 
+// validateHTTPMethods validate if the HTTP Method match clauses are valid.
+func validateHTTPMethods(v *validator.Validate, topStruct reflect.Value, currentStructOrField reflect.Value, field reflect.Value, fieldType reflect.Type, fieldKind reflect.Kind, param string) bool {
```
fasaxc: These functions need to use structlevel.ReportError to tell the user what is wrong.

lib/validator/v3/validator.go
saumoh: yes. will change to address this.

fasaxc:
@saumoh I noticed a new commit but it doesn't seem to have addressed all my comments.  Please let me know when it's ready for another look.

lib/apis/v3/policy.go
caseydavenport: Even if we don't change the format, it should be its own struct I think. e.g.

```
// Only one of Prefix or Exact can be specified (enforced by validation)
type HTTPMatch struct {
    Prefix string
    Exact string
}
```

saumoh: @caseydavenport since we desire the `oneof` (either prefix or exact) semantics. It would have to be an interface, I think.
Which is fine but then afaik the un-marshaling of a struct `HTTPMatch` which encloses an interface would have to be done manually by overriding the `json.UnmarshalJSON`.
 
For example:
```
type HTTPMatch struct {
   Methods []string `json:"methods,omitempty"`
   Paths []HTTPPath `json:"paths,omitempty"`
}

type HTTPPath interface {
  GetPath() string
}

type HTTPPathExact struct {
  Exact string `json:"exact"`
}

func (e *HTTPPathExact) GetPath() string { return e.Exact }

type HTTPPathPrefix struct {
  Prefix string `json:"prefix"`
}

func (e *HTTPPathPrefix) GetPath() string { return e.Prefix }

// HTTPMatch is enclosing struct for both HTTPMethod and HTTPPath
func (p *HTTPMatch) UnmarshalJSON(data []byte) error {
  var objMap map[string]*json.RawMessage

  err := json.Unmarshal(data, &objMap)
  if err != nil {
    fmt.Println("err1")
    return err
  }
....

  var rawPaths []*json.RawMessage
  err = json.Unmarshal(*objMap["paths"], &rawPaths)
  if err != nil {
    return err
  }

  f.Paths = make([]HTTPPath, len(rawPaths))

  var pathType map[string]string
  for index,  rawPath := range rawPaths {
    err := json.Unmarshal(*rawPath, &pathType)
    if err != nil {
      return err
    }
    fmt.Printf("Data: %#v\n", pathType)

    for k, v := range pathType {
      switch strings.ToLower(k) {
        case "exact":
          f.Paths[index] = &HTTPPathExact{Exact: v}
        case "prefix":
          f.Paths[index] = &HTTPPathPrefix{Prefix: v}
        default:
          return errors.New("Unsupported type for HTTPPath")

```

I feel that using an interface abstraction we are not gaining much and @fasaxc suggestion would be good middle ground between an implementation that is too loose (with a raw `map[string]string`) and having an API with a struct. Unless we feel that by having users specifying `MatchType` could be error prone.
wdyt?

caseydavenport: I don't understand why - just have both fields present in the same struct, and use validation to make sure the user never specified both of them. That is way less complicated than using interfaces. It's a common pattern in k8s APIs.

saumoh: I see what you mean! and will do so. 
@fasaxc Will make this changes and ping u when it's ready.

lib/validator/v3/validator.go
saumoh: k. Will remove this restriction.

lib/backend/model/rule.go
```diff
@@ -82,7 +83,8 @@ type Rule struct {
 }
 
 type HTTPMatch struct {
-	Methods []string `json:"methods,omitempty" validate:"omitempty"`
+	Methods []string         `json:"methods,omitempty" validate:"omitempty"`
+	Paths   []apiv3.HTTPPath `json:"paths,omitempty" validate:"omitempty"`
```
fasaxc: Had to pause and think about whether it's ok to import apiv3 from the model package.  I _think_ it is but we need to be careful about introducing cyclic dependencies so typically we don't import the different API levels packages from each other.

PR #859, opened 2018-04-26
Integration upgrade tweaks

tmjd:
## Description

    Made clearing V1 Ready flag atomic
    - Required adding Update to the migration V1 client
      (copied from old client code)
    - Added function to check if the migration is in progress

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/upgrade/migrator/clients/clients.go
+++ b/lib/upgrade/migrator/clients/clients.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2017 Tigera, Inc. All rights reserved.
+// Copyright (c) 2017-2018 Tigera, Inc. All rights reserved.
 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -41,6 +41,7 @@ const (
 
 type V1ClientInterface interface {
 	Apply(d *model.KVPair) (*model.KVPair, error)
+	Update(d *model.KVPair) (*model.KVPair, error)
 	Get(k model.Key) (*model.KVPair, error)
 	List(l model.ListInterface) ([]*model.KVPair, error)
 	IsKDD() bool
--- a/lib/upgrade/migrator/clients/v1/etcdv2/etcdv2.go
+++ b/lib/upgrade/migrator/clients/v1/etcdv2/etcdv2.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2016-2017 Tigera, Inc. All rights reserved.
+// Copyright (c) 2016-2018 Tigera, Inc. All rights reserved.
 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -16,6 +16,7 @@ package etcdv2
 
 import (
 	goerrors "errors"
+	"strconv"
 	"strings"
 
 	"time"
@@ -87,6 +88,22 @@ func NewEtcdClient(config *apiv1.EtcdConfig) (*EtcdClient, error) {
 	keys := etcd.NewKeysAPI(client)
 
 	return &EtcdClient{etcdClient: client, etcdKeysAPI: keys}, nil
+}
+
+// Update an existing entry in the datastore.  This errors if the entry does
+// not exist.
+func (c *EtcdClient) Update(d *model.KVPair) (*model.KVPair, error) {
+	// If the request includes a revision, set it as the etcd previous index.
+	options := etcd.SetOptions{PrevExist: etcd.PrevExist}
+	if len(d.Revision) != 0 {
+		var err error
+		if options.PrevIndex, err = strconv.ParseUint(d.Revision, 10, 64); err != nil {
+			return nil, err
+		}
+		log.Debugf("Performing CAS against etcd index: %v\n", options.PrevIndex)
+	}
+
+	return c.set(d, &options)
 }
 
 // Set an existing entry in the datastore.  This ignores whether an entry already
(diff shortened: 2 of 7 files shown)


lib/upgrade/migrator/clients/v1/etcdv2/etcdv2.go
```diff
@@ -34,6 +35,8 @@ var (
 	etcdGetOpts          = &etcd.GetOptions{Quorum: true}
 	etcdListOpts         = &etcd.GetOptions{Quorum: true, Recursive: true, Sort: true}
 	etcdListChildrenOpts = &etcd.GetOptions{Quorum: true, Recursive: false, Sort: true}
+	etcdCreateOpts       = &etcd.SetOptions{PrevExist: etcd.PrevNoExist}
+	etcdDeleteEmptyOpts  = &etcd.DeleteOptions{Recursive: false, Dir: true}
```
mgleung: What happens if we were to delete a node without deleting its children? Do the children get orphaned or is there some other way to access the children after its parent node has been deleted?

tmjd: TBH I didn't really consider the options (other than that the Create will error if it already exists). I copied the Delete from https://github.com/projectcalico/libcalico-go/blob/master/lib/backend/etcd/etcd.go#L177, and the etcdDeleteEmptyOpts is needed for that one.  I think the Delete could be cut down a lot but I went with just copying something, that I assume worked, instead of trying to write it myself.

mgleung: Looks good, just looked it up, it deletes only if the directory at the path is empty.

lib/upgrade/migrator/clients/v1/etcdv2/etcdv2.go
```diff
@@ -89,12 +92,54 @@ func NewEtcdClient(config *apiv1.EtcdConfig) (*EtcdClient, error) {
...
+	return c.set(d, etcdCreateOpts)
+}
+
 // Set an existing entry in the datastore.  This ignores whether an entry already
 // exists.
 func (c *EtcdClient) Apply(d *model.KVPair) (*model.KVPair, error) {
 	return c.set(d, etcdApplyOpts)
 }
 
+// Delete an entry in the datastore.  This errors if the entry does not exists.
```
mgleung: nit: `exist`

lib/upgrade/migrator/clients/v1/k8s/k8s.go
```diff
@@ -213,6 +213,14 @@ func (c *KubeClient) Apply(d *model.KVPair) (*model.KVPair, error) {
 	}
 }
 
+func (c *KubeClient) CmpAndSwap(k model.Key, oldValue, newValue interface{}) (*model.KVPair, error) {
+	log.Warn("Attempt to 'CmpAnsSwap' using kubernetes backend is not supported.")
```
mgleung: nit: `CmpAndSwap`

lib/upgrade/migrator/clients/clients.go
```diff
@@ -41,6 +41,7 @@ const (
 
 type V1ClientInterface interface {
 	Apply(d *model.KVPair) (*model.KVPair, error)
+	CmpAndSwap(k model.Key, oldValue, newValue interface{}) (*model.KVPair, error)
```
caseydavenport: This is a weird interface - why not just do `Update(kvpair)`? The datastore can do the compare and swap for you, so not sure why you need the old and new value here?

tmjd: We only want to update if it has the oldValue. We want this to fail if it is not the specific oldValue being provided.  I'm not sure how we achieve that with `Update(kvpair)`.

caseydavenport: The KVPair has a revision number which can be used for this.

Update is a CAS operation, whereas create isn't.

tmjd: Sorry I didn't realize you were talking about an already existing function (in the previous) client or I would have went and looked at that.

Do you think the Migration code is safe to assume that the Ready flag is present? If not then I'll have to add back in the Create function to the migration clientV1 interface.

caseydavenport: eh, didn't realize all of that code has been ripped out now... 

> Do you think the Migration code is safe to assume that the Ready flag is present?

Do you? I sorta think it might be, but I'm not close enough to the migration to know for sure. Maybe Rob would know.

Alternatively, can't we store this state as an annotation on e.g. the k8s kube-system namespace or something and clear it when we're done? Just another thought (feel free to ignore if it's not any easier)

tmjd: Yeah I think it is safe to expect that it is there. It has been created by EnsureInitialized since v2.4.1 (and possible earlier).

lib/upgrade/migrator/clients/v1/k8s/k8s.go
```diff
@@ -203,6 +203,16 @@ func buildCRDClientV1(cfg rest.Config) (*rest.RESTClient, error) {
 	return cli, nil
 }
 
+// Update an existing entry in the datastore.  This errors if the entry does
+// not exist. (Not implemented for KDD.)
+func (c *KubeClient) Update(d *model.KVPair) (*model.KVPair, error) {
+	log.Warn("Attempt to 'CmpAndSwap' using kubernetes backend is not supported.")
```
caseydavenport: typo: CmpAndSwap

PR #861, opened 2018-05-01
[WIP] [Prototype] More flexible route reflector support

caseydavenport:
## Description

This is a prototype of something I'm investigating now - not necessarily a thread that will come to fruition.

Goal is to provide more flexible native support for route reflector configuration via calicoctl. It does this by:
- Allowing configuration of `routeReflectorClusterID` on the Node object, which turns that node into a RR.
- Adding label selectors to BGP Peer objects to make peering configuration easier.

It relies on corresponding changes in a few places:
- calico/node - https://github.com/projectcalico/node/pull/5
- calicoctl - 
- confd - https://github.com/projectcalico/confd/pull/116

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/apis/v3/node.go
+++ b/lib/apis/v3/node.go
@@ -68,7 +68,7 @@ type NodeBGPSpec struct {
 	// IPv4IPIPTunnelAddr is the IPv4 address of the IP in IP tunnel.
 	IPv4IPIPTunnelAddr string `json:"ipv4IPIPTunnelAddr,omitempty" validate:"omitempty,ipv4"`
 	// RouteReflectorClusterID enables this node as a route reflector within the given cluster.
-	RouteReflectorClusterID string `json:"routeReflectorClusterID,omitempty" validate:"omitempty"`
+	RouteReflectorClusterID string `json:"routeReflectorClusterID,omitempty" validate:"omitempty,ipv4"`
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
--- a/lib/clientv3/node_e2e_test.go
+++ b/lib/clientv3/node_e2e_test.go
@@ -47,7 +47,7 @@ var _ = testutils.E2eDatastoreDescribe("Node tests", testutils.DatastoreEtcdV3, 
 	spec1 := apiv3.NodeSpec{
 		BGP: &apiv3.NodeBGPSpec{
 			IPv4Address:             "1.2.3.4",
-			RouteReflectorClusterID: "cluster1",
+			RouteReflectorClusterID: "10.0.0.0",
 		},
 		OrchRefs: []apiv3.OrchRef{
 			{


lib/backend/model/bgppeer.go
```diff
@@ -32,6 +32,31 @@ var (
 	typeBGPPeer        = reflect.TypeOf(BGPPeer{})
 )
 
+type SelectorBGPPeerKey struct {
+	Name string `json:"-" validate:"omitempty"`
+}
+
+func (key SelectorBGPPeerKey) defaultPath() (string, error) {
+	e := fmt.Sprintf("/calico/bgp/v1/global/peer_sel/%s", key.Name)
```
neiljerram: Is this intentionally still in the v1 space?  (That feels surprising to me.)

lib/backend/syncersv1/updateprocessors/bgppeerprocessor.go
```diff
@@ -42,31 +45,45 @@ func convertBGPPeerV2ToV1(kvp *model.KVPair) (*model.KVPair, error) {
...
 
-	// Correct data types.  Handle the conversion.  Start with the v1 key.  The PeerIP and
-	// the Node are now in the Spec - if a Node is not specified then this is a global
-	// peer.
+	// Correct data types. Handle the conversion.
 	ip := cnet.ParseIP(v3res.Spec.PeerIP)
-	if ip == nil {
-		return nil, errors.New("PeerIP is not assigned or is malformed")
+	if ip == nil && v3res.Spec.PeerSelector == "" {
+		return nil, errors.New("Peer IP or selector is not assigned or is malformed")
```
neiljerram: It would be nicer to have a precise error here; one of:
- Either Peer IP or Peer Selector must be specified.
- Peer IP is malformed
- Peer Selector is malformed

lib/backend/syncersv1/updateprocessors/bgppeerprocessor.go
```diff
@@ -42,31 +45,45 @@ func convertBGPPeerV2ToV1(kvp *model.KVPair) (*model.KVPair, error) {
...
+		// This peer selects a specific node. Use an exact selector
+		// to represent this.
+		selector = fmt.Sprintf("projectcalico.org/node == '%s'", v3res.Spec.Node)
+		log.Infof("Converted node-specific BGPPeer to selector - %s", selector)
+	}
+
+	val := &model.BGPPeer{
+		Name:         v3res.Name,
+		ASNum:        v3res.Spec.ASNumber,
+		NodeSelector: v3res.Spec.NodeSelector,
```
neiljerram: Guess this should be `selector`?  (Otherwise `selector` is unused, I think.)

lib/backend/syncersv1/updateprocessors/bgppeerprocessor.go
```diff
@@ -42,31 +45,45 @@ func convertBGPPeerV2ToV1(kvp *model.KVPair) (*model.KVPair, error) {
...
+
+	// Determine the node selector to use. We convert old-style "Global" and
+	// per node peers into selector peers for simplicity.
+	selector := v3res.Spec.NodeSelector
+	if v3res.Spec.Node == "" && selector == "" {
+		// This is an old-style "Global" peer, which means
+		// select all nodes.
+		selector = "all()"
+		log.Infof("Converted global BGPPeer to selector - %s", selector)
+	} else if v3res.Spec.Node != "" {
```
neiljerram: `&& selector == ""` again?

lib/apis/v3/node.go
```diff
@@ -67,6 +67,9 @@ type NodeBGPSpec struct {
 	IPv6Address string `json:"ipv6Address,omitempty" validate:"omitempty,cidrv6"`
 	// IPv4IPIPTunnelAddr is the IPv4 address of the IP in IP tunnel.
 	IPv4IPIPTunnelAddr string `json:"ipv4IPIPTunnelAddr,omitempty" validate:"omitempty,ipv4"`
+	// RouteReflectorClusterID enables this node as a route reflector, automatically peering with
+	// other nodes in the same cluster.
+	RouteReflectorClusterID string `json:"routeReflectorClusterID,omitempty" validate:"omitempty"`
```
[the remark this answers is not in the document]
caseydavenport: Yea, that comment is wrong (from an earlier version of this experiment).

I do expect that it will work for a calico/node instance to perform local programming and act as a RR. Seems to work from my playing around.

lib/backend/model/bgppeer.go
caseydavenport: Yeah, it's because confd only speaks the "v1" api over the syncer, so we need to convert the v3 stuff into something it understands (even though this obviously wasn't a part of the original v1 API and will never show up in etcd with a /v1 key)

PR #863, opened 2018-05-12
Allow configuration of veth prefix in KDD mode

caseydavenport:
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/backend/k8s/conversion/conversion.go
+++ b/lib/backend/k8s/conversion/conversion.go
@@ -18,6 +18,7 @@ import (
 	"crypto/sha1"
 	"encoding/hex"
 	"fmt"
+	"os"
 	"sort"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


fasaxc:
Looks like a bit of a hack TBH! 

Specific concern: in general, felix might not be configured through env vars so that variable might not be set when libcalico-go is being used in Felix.  Maybe we can guarantee that it will be set that way in the particular scenario that we care about?  In that case, I'd be tempted to introduce a dedicated env var instead, just to keep the separation a bit cleaner `CALICO_IFACE_PREFIX_OVERRIDE=eni`

I think the _right_ thing to do would be to pass a prefix into that function and fix up all the layers of calling code but maybe this is urgent enough to require a quick fix?

caseydavenport:
@fasaxc yes, it is definitely a "hack", but a calculated one.

> Specific concern: in general, felix might not be configured through env vars so that variable might not be set when libcalico-go is being used in Felix. Maybe we can guarantee that it will be set that way in the particular scenario that we care about?

Yep, we can definitely make sure it is set that way in the scenario this PR is meant to address. 

TBH I wouldn't be super happy with a separate env var. It just feels like 99% of the time this is what the user is going to want, and the other 1% of the time it will still work for them. Adding two knobs here feels like extra operation complexity.

We can still clean this up by allowing this to be set via felix config in the future without breaking compatibility.

fasaxc:
But don't we need this set for more components that Felix too?  `calicoctl` `kube-controllers` etc?

PR #867, opened 2018-05-21
Support pod and namespace selectors in K8s policy

spikecurtis:
Signed-off-by: Spike Curtis <spike@tigera.io>

## Description
Upcoming K8s release will allow both pod and namespace selectors on a policy rule https://github.com/kubernetes/kubernetes/pull/60452

This updates our conversion from k8s NetworkPolicy to Calico NetworkPolicy to correctly render this API change.  Note that Calico v3 API already supports this construction, so the only change is how we render K8s NetworkPolicy in the Calico API.

## Todos
- [X] Tests
- [X] Documentation - none required.
- [X] Release note

## Release Note

```release-note
For updated Kubernetes clusters that allow it, you may include both a pod and namespace selector on a NetworkPolicyPeer.
```


shafay2000:
Thanks Casey!

PR #868, opened 2018-05-24
bugfix: IPAM block affinity leaks after upgrade

gunjan5:
## Description

fixes https://github.com/projectcalico/calico/issues/1956

## Todos
- [x] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
bugfix: IPAM doesn't use old IP blocks after upgrade to Calico 3.1
```

### DIFF
--- a/lib/backend/model/keys.go
+++ b/lib/backend/model/keys.go
@@ -15,12 +15,12 @@
 package model
 
 import (
+	"bytes"
 	"encoding/json"
-	"reflect"
-	"strings"
-
 	"fmt"
 	net2 "net"
+	"reflect"
+	"strings"
 	"time"
 
 	"github.com/projectcalico/libcalico-go/lib/net"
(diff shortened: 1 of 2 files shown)


lib/backend/model/keys.go
```diff
@@ -286,8 +286,22 @@ func ParseValue(key Key, rawData []byte) (interface{}, error) {
...
+		// This is a special case to address backwards compatibility from the time when we had no state information as block affinity value.
+		// example:
+		// Key: "/calico/ipam/v2/host/myhost.io/ipv4/block/172.29.82.0-26"
+		// Value: ""
+		// In 3.0.7 we added block affinity state as the value, so old "" value is no longer a valid JSON, so for that
+		// particular case we replace the "" with a "{}" so it can be parsed and we don't leak blocks after upgrade to Calico 3.0.7
+		// See: https://github.com/projectcalico/calico/issues/1956
+		if bytes.Equal(rawData, []byte(``)) && valueType == typeBlockAff {
+			rawData = []byte(`{}`)
+			goto retryWithEmptyJSON
```
caseydavenport: Rather than use a `goto` here, can't we just call `json.Unmarshal` here? I think that would be simpler and nicer.

gunjan5: I thought of it but then thought it's more duplicate code, but I also don't like `goto` very much, so I'm fine either way. I'll change it

Issue #820, opened 2018-03-13
Malformed ip pool name causes crash


fasaxc:
Seen on Calico 2.6, if you have an IP pool in etcd that has an unexpected name format then we fail to parse it with a panic.

Example key: `/calico/v1/ipam/v4/pool/577559` (taken from an old OpenStack scale testing script).

## Expected Behavior

Warning log and drop.

## Current Behavior

```
panic: invalid CIDR address: 577559

goroutine 71 [running]:
github.com/projectcalico/felix/vendor/github.com/projectcalico/libcalico-go/lib/backend/model.KeyFromDefaultPath(0xc420321940, 0x1e, 0xc420627b30, 0x1)
	/go/src/github.com/projectcalico/felix/vendor/github.com/projectcalico/libcalico-go/lib/backend/model/keys.go:209 +0x139c
github.com/projectcalico/felix/vendor/github.com/projectcalico/libcalico-go/lib/backend/etcd.(*etcdSyncer).sendUpdate(0xc4201a3b60, 0xc420321940, 0x1e, 0xc420042640, 0x3e, 0x124, 0x1)
	/go/src/github.com/projectcalico/felix/vendor/github.com/projectcalico/libcalico-go/lib/backend/etcd/syncer.go:525 +0xf7
github.com/projectcalico/felix/vendor/github.com/projectcalico/libcalico-go/lib/backend/etcd.(*etcdSyncer).mergeUpdates(0xc4201a3b60, 0xc42004f0e0, 0xc42004f080, 0xc42004f140)
	/go/src/github.com/projectcalico/felix/vendor/github.com/projectcalico/libcalico-go/lib/backend/etcd/syncer.go:456 +0x15d1
created by github.com/projectcalico/felix/vendor/github.com/projectcalico/libcalico-go/lib/backend/etcd.(*etcdSyncer).Start
	/go/src/github.com/projectcalico/felix/vendor/github.com/projectcalico/libcalico-go/lib/backend/etcd/syncer.go:158 +0x192
```

## Possible Solution

Handle the error, log and drop instead of panicking.  Audit other key parsing functions for same.

## Steps to Reproduce (for bugs)

1. Manually insert an IP pool with above key
2. Start felix (probably other components affected)
3.
4.

## Your Environment

* libcalico-go version: Something around edcbcf64348004bfeb0dc3882cd2790743ec7728 (using felix from the calico-2.6 PPA)
* Operating System and version: Ubuntu 14.04
[...]

PR #875, opened 2018-05-31
malformed IPPool panic removal

dmmcquay:
## Description

Remove panic when getting key of type IPPool when CIDR parse fails. 

fixes https://github.com/projectcalico/libcalico-go/issues/820

## Todos
- [x] Tests

## Release Note

```release-note
None required
```

Signed-off-by: derek mcquay <derek@tigera.io>

### DIFF
--- a/lib/backend/model/keys.go
+++ b/lib/backend/model/keys.go
@@ -237,9 +237,10 @@ func KeyFromDefaultPath(path string) Key {
 		cidr := strings.Replace(mungedCIDR, "-", "/", 1)
 		_, c, err := net.ParseCIDR(cidr)
 		if err != nil {
-			panic(err)
+			log.WithError(err).Warningf("Failed to parse CIDR %s", cidr)
+		} else {
+			return IPPoolKey{CIDR: *c}
 		}
-		return IPPoolKey{CIDR: *c}
 	} else if m := matchGlobalConfig.FindStringSubmatch(path); m != nil {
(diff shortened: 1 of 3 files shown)


lib/backend/model/keys.go
```diff
@@ -237,9 +237,10 @@ func KeyFromDefaultPath(path string) Key {
 		cidr := strings.Replace(mungedCIDR, "-", "/", 1)
 		_, c, err := net.ParseCIDR(cidr)
 		if err != nil {
-			panic(err)
+			log.Warningf("Failed to parse CIDR %s: %v", cidr, err)
```
bcreane: suggest `log.WithError(err).Warningf()`

robbrockbank:
You'll actually need this change in the appropriate Calico v2.x branch of libcalico-go.  As it stands, I don't think this will actually impact any code that is actually executed in the master branch (we have a lot of old code that we really should have removed by now).  I think it's fine having this change in master too, but just wanted to highlight this.

I believe an additional change is required in the `KeyFromDefaultPath()` method in the `model/ippool.go` file.  At the moment we are ignoring errors returned from the `ParseCIDR` call.  This code is executed when we list IPPools, and *is* used in the master codebase in the v2->v3 upgrade code.

dmmcquay:
@robbrockbank I asked @bcreane about which branch I should create this PR against and he mentioned that pushing to master and then cherry-picking the changes to the needed branches. Does that still seem like a valid idea given what these changes touch?

Issue #535, opened 2017-09-25
Return K8s permission error message to caller.


spikecurtis:
## Expected Behavior

When K8s API server denies a request, it includes an error message saying the user and the resource that was attempted to be accessed.  It would be really handy to return this message to the caller.  

## Current Behavior

However, in converting to a Calico error, we just return a generic "connection unauthorized" string.

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Context

I'm trying to debug RBAC permissions on my cluster.  It took me a couple hours to track things down to the fact that `libcalico-go` tries to pull from the `extensions` API group, not the `networking` API group.

## Your Environment

* libcalico-go version:
* Operating System and version:
* Link to your project (optional):

PR #878, opened 2018-06-07
Include descriptive error details for calico errors

dmmcquay:
## Description

Add more information to calico errors particularly when converting from another type (e.g. Kubernetes error). 

An example is when K8s API server denies a request, it includes an error message saying the user and the resource that was attempted to be accessed. This information will be included and more descriptive instead of just "connection is unauthorized"

fixes https://github.com/projectcalico/libcalico-go/issues/535

## Release Note

```release-note
None required
```

Signed-off-by: derek mcquay <derek@tigera.io>


tmjd:
Awesome :+1: thanks

PR #880, opened 2018-06-07
Clean up Makefile

tomdee:
Add ci target and reorder into common format


Makefile
```diff
@@ -27,8 +30,81 @@ DOCKER_GO_BUILD := mkdir -p .go-pkg-cache && \
...
+	rm -f ./lib/upgrade/migrator/clients/v1/k8s/custom/zz_generated.deepcopy.go
+
+###############################################################################
+# Building the binary
+###############################################################################
+.PHONY: update-vendor
+# Update the pins in glide.lock to reflect the updated glide.yaml.
+# Note: no dependency on glide.yaml so we don't automatically update glide.lock without
+# developer intervention.
+update-vendor glide.lock:
```
caseydavenport: I don't remember seeing an `update-vendor` target in the other PRs - do we want / need this?

Makefile
```diff
@@ -49,45 +125,32 @@ check-format: vendor/.up-to-date
...
+.PHONY: ut-native
+## Run the UTs natively.  This requires a local etcd and local kubernetes master to be running.
+ut-native: vendor/.up-to-date
 	./run-uts
 
-.PHONY: test-containerized
+.PHONY:ut
 ## Run the tests in a container. Useful for CI, Mac dev.
-test-containerized: vendor/.up-to-date run-etcd run-kubernetes-master
+ut: vendor/.up-to-date run-etcd run-kubernetes-master
```
caseydavenport: This makes me indescribably happy.

caseydavenport: Admittedly, I don't like `vendor/.up-to-date` as a target, since it means a manually performed `glide up` will get overwritten.

Might just have to be a live with... but it's a bit painful.

Makefile
tomdee: As discussed, we think not.

Makefile
tomdee: Got rid if the .up-to-date business

PR #886, opened 2018-06-12
support setting passive mode in bgp peer

dqminh:
For https://github.com/projectcalico/calico/issues/1603

This allow use to set `passive: true` in bgp peer spec so we can leverage that in calico/node to set proper bird config.

I think this should touches all the required place here. Then we would want to modify calico/node for confd config and calicoctl for updating new fields.

cc @caseydavenport

### DIFF
--- a/lib/apis/v3/bgppeer.go
+++ b/lib/apis/v3/bgppeer.go
@@ -48,6 +48,8 @@ type BGPPeerSpec struct {
 	PeerIP string `json:"peerIP" validate:"omitempty,ip"`
 	// The AS Number of the peer.
 	ASNumber numorstring.ASNumber `json:"asNumber"`
+	// Passive specifies whether the peer is in passive mode
+	Passive bool `json:"passive,omitempty"`
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
--- a/lib/apis/v3/zz_generated.deepcopy.go
+++ b/lib/apis/v3/zz_generated.deepcopy.go
@@ -19,13 +19,12 @@
 package v3
 
 import (
-	reflect "reflect"
-
 	apis_v1 "github.com/projectcalico/libcalico-go/lib/apis/v1"
 	numorstring "github.com/projectcalico/libcalico-go/lib/numorstring"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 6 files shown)


caseydavenport:
@dqminh thanks for this pr!

I think there's a bit more to do here. Namely, [update the model.BGPeer to include this field (this is what confd sees)](https://github.com/projectcalico/libcalico-go/blob/master/lib/backend/syncersv1/updateprocessors/bgppeerprocessor.go#L66-L69)

Is there a corresponding PR for the BIRD templates? It would be good to review them together since they rely on each other.

Also, looks like the bot would like you to sign the CLA :)

mattalberts:
@caseydavenport I've gone ahead and updated the PR with a change to ```model.BGPPeer```.

Would it be worth modifying ```api.BGPPeer``` as well to allow filling in the full ```model.BGPPeer``` struct?

As a potentially unrelated side-note,  how does back-porting a change work? We'll be migrating to 2.6.x; I just want to make certain this change would land there as well.

caseydavenport:
> As a potentially unrelated side-note, how does back-porting a change work? We'll be migrating to 2.6.x; I just want to make certain this change would land there as well.

So, we'll need to work through this one. Strictly speaking, this is an API change and so would require a minor bump rather than a patch release (e.g v2.7.0, or v3.2.0). I don't think we have any plans to cut a v2.7.0 release.

I'll need to discuss internally and get back to you, but in the absence of that I imagine this would land in v3.2.0.

caseydavenport:
Looks like the fomatting check is failing, proabably need to run `goimports` over the new code. 

You'll also need to run `make gen-files` to update some autogenerated code.

mattalberts:
@caseydavenport sorry about that!  I've let go imports fix the formatting (shame shame shame).  I've also re-run gen-files; though, that only netted a change to deepcopy (the import order).

caseydavenport:
This looks good to me.

Before we merge I'd like to be able to review it side-by-side with the corresponding `confd` template changes to make this work. Is there a PR for that as well?

mattalberts:
There isn't one yet ... but ... I can make one! :)

caseydavenport:
Any update on this guy?

caseydavenport:
@mattalberts @dqminh this has gotten out of date and needs a rebase.

I'm going to close for now since there hasn't been any recent activity, but do ping me if you find time to work on this and we can reopen - I'd quite like to see it go in!

PR #894, opened 2018-06-15
Fix formatting of gen-files

tomdee:


bcreane:
This will get lost next someone runs `make gen-files`. What about adding a `gofmt` call to that target so the change persists across updates?

PR #900, opened 2018-07-03
Bump client-go version to v7.0.0

song-jiang:
## Description

Bump client-go version to v7.0.0 to support exec-based client auth provider.
Versions of k8s.io/api and k8s.io/apimachinery are set by client-go.
k8s.io/code-generator version set to last commit of release-1.10.
k8s.io/gengo updated to latest version from master.

## Todos
- [x] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


glide.yaml
```diff
@@ -29,15 +29,14 @@ import:
...
 - package: k8s.io/apimachinery
-  version: release-1.8
 - package: k8s.io/api
-  version: release-1.8
 - package: k8s.io/client-go
-  version: 82aa063804cf055e16e8911250f888bc216e8b61
+  version: v7.0.0
 - package: k8s.io/code-generator
-  version: release-1.8
+  version: 9de8e796a74d16d2a285165727d04c185ebca6dc
```
caseydavenport: How did we pick these pins here for k8s.io/code-generator and k8s.io/gengo? Can we let these float as well or do they need to be pinned?

If they're pinned for a specific reason, could you add a comment explaining why?

song-jiang: I am not confident to leave k8s.io/code-generator float because unlike `api` or `apimachinery` there is no obvious dependency information how it is set by client-go.  

For k8s.io/gengo, I believe we just need to update to latest version.

song-jiang:
Added comment and squash.

PR #905, opened 2018-07-11
Update to latest client-go release and remove all glide warnings

tomdee:


glide.yaml
```diff
@@ -1,51 +1,81 @@
...
 - package: github.com/projectcalico/go-yaml-wrapper
 - package: github.com/prometheus/client_golang
-  version: ^0.9.0-pre1
   subpackages:
   - prometheus
+# 1.1.0+ needed since our code expects the v4 uuid creation function to return a single value.
 - package: github.com/satori/go.uuid
   version: ^1.1.0
 - package: github.com/sirupsen/logrus
-  version: ^1.0.3
```
caseydavenport: I believe we had pinned this because later versions were not compatible with our log formatting code.

Maybe it's OK if all importing codebases pin this, but worth considering.

tomdee: I've restored the logrus pin and added a comment

PR #906, opened 2018-07-12
Pin logrus and update comment

tomdee:


robbrockbank:
[WARN]	Lock file may be out of date. Hash check of YAML failed. You may need to run 'update'

:-)

tomdee:
Ooops! Yes, will fix up tm

PR #907, opened 2018-07-12
Makefile: Add a new target to check downstream projects

tomdee:
Cheks out downstream projects, updates their glide pin and tries to
build them

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,6 +44,7 @@ clean:
 	find . -name '*.coverprofile' -type f -delete
 	rm -rf vendor .go-pkg-cache
 	rm -rf $(BINDIR)
+	rm -rf checkouts
 
 .PHONY: clean-gen-files
 ## Convenience target for devs to remove generated files and related utilities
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -237,8 +238,27 @@ stop-etcd:
 ###############################################################################
 .PHONY: ci
 ## Run what CI runs
-ci: clean check-glide-warnings static-checks test
-
+ci: clean check-glide-warnings static-checks test build-libcalico-users
+
+# The list of repos that use libcalico (that support being tests in this way)
```
bcreane: "tested" reads better than "tests"

tomdee:
Yes, if any fail to build then libcalico-go CI will fail. So if someone submits a PR that stops another project from compiling then they'll need admin override (to ignore CI) to merge it.

I'm now in two minds about merging this - it slows down the CI quite a bit. On the other hand, it's very easy to remove so we could easily try it out and then back it out if it's too much of a burden

bcreane:
Let's try it out - if it's too burdensome, then, as you say, we can take it out.

PR #911, opened 2018-08-02
Add FelixConfiguration for external nodes cidr list.

saumoh:
Add FelixConfiguration to set a list of CIDR's which may be used to originate tunneled traffic.

### DIFF
--- a/lib/apis/v3/felixconfig.go
+++ b/lib/apis/v3/felixconfig.go
@@ -198,7 +198,11 @@ type FelixConfigurationSpec struct {
 
 	// NATPortRange specifies the range of ports that is used for port mapping when doing outgoing NAT. When unset the default behavior of the
 	// network stack is used.
-	NATPortRange *numorstring.Port `json: natPortRange,omitempty`
+	NATPortRange *numorstring.Port `json:"natPortRange,omitempty"`
+
+	// ExternalNodesCIDRList is a list of CIDR's of external-non-calico-nodes which may source tunnel traffic and have
+	// the tunneled traffic be accepted at calico nodes.
+	ExternalNodesCIDRList *[]string `json:"externalNodesList,omitempty"`
 
 	DebugMemoryProfilePath          string           `json:"debugMemoryProfilePath,omitempty"`
 	DebugDisableLogDropping         *bool            `json:"debugDisableLogDropping,omitempty"`
--- a/lib/validator/v3/validator.go
+++ b/lib/validator/v3/validator.go
@@ -545,6 +545,21 @@ func validateFelixConfigSpec(v *validator.Validate, structLevel *validator.Struc
 			}
 		}
 	}
+
+	// Validate that the externalNodesCIDRList is composed of valid cidr's.
+	if c.ExternalNodesCIDRList != nil {
+		for _, cidr := range *c.ExternalNodesCIDRList {
+			log.Debugf("Cidr is: %s", cidr)
+			ip, _, err := cnet.ParseCIDROrIP(cidr)
+			if err != nil {
+				structLevel.ReportError(reflect.ValueOf(cidr),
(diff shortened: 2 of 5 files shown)


robbrockbank:
You will also need to run the auto generator to update the deep copy gen file.

Should be a Makefile target for that.

lib/apis/v3/felixconfig.go
```diff
@@ -198,7 +198,11 @@ type FelixConfigurationSpec struct {
 
 	// NATPortRange specifies the range of ports that is used for port mapping when doing outgoing NAT. When unset the default behavior of the
 	// network stack is used.
-	NATPortRange *numorstring.Port `json: natPortRange,omitempty`
+	NATPortRange *numorstring.Port `json:"natPortRange,omitempty"`
+
+	// ExternalNodesCidrList is a list of CIDR's of external-non_calico-nodes which may source tunnel traffic and have
+	// the tunneled traffic be accepted at calico nodes.
+	ExternalNodesCidrList *[]string `json:"externalNodesList,omitempty"`
```
robbrockbank: Shouldn't this be `[]string` rather than `*[]string`

lib/apis/v3/felixconfig.go
```diff
@@ -198,7 +198,11 @@ type FelixConfigurationSpec struct {
 
 	// NATPortRange specifies the range of ports that is used for port mapping when doing outgoing NAT. When unset the default behavior of the
 	// network stack is used.
-	NATPortRange *numorstring.Port `json: natPortRange,omitempty`
+	NATPortRange *numorstring.Port `json:"natPortRange,omitempty"`
+
+	// ExternalNodesCidrList is a list of CIDR's of external-non_calico-nodes which may source tunnel traffic and have
```
robbrockbank: `non_calico` -> `non-calico` ?

fasaxc: Think this should be `ExternalNodesCIDRList` to match Go's normal naming scheme.

lib/validator/v3/validator_test.go
```diff
@@ -481,6 +481,10 @@ func init() {
 		Entry("should reject a named port KubeNodePortRanges value", api.FelixConfigurationSpec{KubeNodePortRanges: &[]numorstring.Port{
 			numorstring.NamedPort("testport"),
 		}}, false),
+		Entry("should accept a valid list of ExternalNodesCidrList", api.FelixConfigurationSpec{ExternalNodesCidrList: &[]string{"1.1.1.1", "1.1.1.2/32", "1.1.3.0/23", "abcd::1", "abef::2/128"}},
```
fasaxc: We shouldn't accept IPv6 CIDRs here because Felix will fail to put those in the all-hosts IP set.  IPIP is an IPv4 protocol anyway, at least the variant that we support is.

saumoh: sgtm. i'll make this change.

saumoh:
@robbrockbank @fasaxc Thanks for taking a look at this. I've incorporated both of your feedbacks.
The semantic of `ptr to slice` is used in quite a few places in `FelixConfiguration` and we decided to kept it the same for consistency. Also it helps to differentiate between the `nil` value and not set at all. The patch in the `validator` uses that.

lib/validator/v3/validator.go
```diff
@@ -545,6 +545,25 @@ func validateFelixConfigSpec(v *validator.Validate, structLevel *validator.Struc
 			}
 		}
 	}
+
+	// Validate that the externalNodesCIDRList is composed of valid cidr's.
+	if c.ExternalNodesCIDRList != nil {
+		if len(*c.ExternalNodesCIDRList) == 0 {
+			structLevel.ReportError(reflect.ValueOf(c.ExternalNodesCIDRList),
```
fasaxc: Do we error on empty slices elsewhere?  Feels like an empty slice should mean "don't add any external CIDRs"

PR #913, opened 2018-08-08
Make the list of string comma delimited.

saumoh:
To make the list of strings easily to parse and human friendly, render it as a comma delimited list in the backend.

cc: @fasaxc

### DIFF
--- a/lib/backend/syncersv1/updateprocessors/configurationprocessor.go
+++ b/lib/backend/syncersv1/updateprocessors/configurationprocessor.go
@@ -240,6 +240,9 @@ func (c *configUpdateProcessor) processAddOrModified(kvp *model.KVPair) ([]*mode
 					default:
 						value = fmt.Sprintf("%v", vt.Seconds())
 					}
+				case []string:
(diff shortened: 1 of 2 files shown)


lib/backend/syncersv1/updateprocessors/configurationprocessor.go
```diff
@@ -240,6 +240,9 @@ func (c *configUpdateProcessor) processAddOrModified(kvp *model.KVPair) ([]*mode
 					default:
 						value = fmt.Sprintf("%v", vt.Seconds())
 					}
+				case []string:
+					// Make a list of strings comma delimited
+					value = strings.Join(vt, ", ")
```
robbrockbank: nit:  do you really need a trailing space?  this is just for internal message flow so prolly not necessary.  don't really mind though provided felix can parse it ok.

saumoh: ya. not really needed and parsing in felix would be fine. will nuke it.

PR #919, opened 2018-09-06
Changes needed for BGP->confd sync via Typha

neiljerram:
It needs to be here after all, because used by both confd and typha.

### DIFF
--- a/lib/backend/syncersv1/bgpsyncer/bgpsyncer.go
+++ b/lib/backend/syncersv1/bgpsyncer/bgpsyncer.go
@@ -1,0 +1,51 @@
+// Copyright (c) 2017-2018 Tigera, Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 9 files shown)


lib/backend/syncersv1/bgpsyncer/bgpsyncer.go
```diff
@@ -0,0 +1,51 @@
...
+		},
+		{
+			ListInterface:   model.ResourceListOptions{Kind: apiv3.KindBGPConfiguration},
+			UpdateProcessor: updateprocessors.NewBGPConfigUpdateProcessor(),
+		},
+		{
+			ListInterface: model.ResourceListOptions{Kind: apiv3.KindNode},
+		},
+		{
+			ListInterface: model.ResourceListOptions{Kind: apiv3.KindBGPPeer},
```
caseydavenport: Do we want to include the updateprocessors for these guys?

neiljerram: No; it's a key part of the overall design that confd sees the v3 Nodes and Peers.

neiljerram: (Nodes, because confd needs node labels.  Peers, to handle the selector fields and fill in reverse peerings.)

PR #926, opened 2018-09-12
Build release-v3.2 branches of libcalico users, not master

neiljerram:


tomdee:
LGTM - though another option would be to just remove `libcalico-users` as a dependency from `ci`

tomdee:
Hmmm, actually it isn't passing CI so I suspect there's something else going on

neiljerram:
@tomdee 
> though another option would be to just remove `libcalico-users` as a dependency from `ci`

Do you mean just on this branch?

neiljerram:
@tomdee So the problem is that the regex for picking out the old version from glide.yaml
```
        export OLD_VER=$$(grep --after 50 libcalico-go glide.yaml |grep --max-count=1 --only-matching --perl-regexp "version:\s*\K[\.0-9a-z]+") ;\
```
doesn't get the whole of the `release-v3.2` version as exists e.g. in cni-plugin's glide.yaml on the release-v3.2 branch:
```
- package: github.com/projectcalico/libcalico-go
  version: release-v3.2
```

I think I'd suggest
- making the regex more general, such that the bit inside the `[...]` matches anything that is not whitespace
- probably applying this change across all repos and branches.

WDYT?

PR #930, opened 2018-09-25
Cleanup IPAM non-affine check with no IP pools

caseydavenport:
## Description

We had a redundant check for whether or not IP pools were defined.

Also adds a test to make sure we don't allocate IPs from other pools if one is specified during an auto-assign. 

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/ipam/ipam.go
+++ b/lib/ipam/ipam.go
@@ -404,15 +404,6 @@ func (c ipamClient) autoAssign(ctx context.Context, num int, handleID *string, a
 	rem := num - len(ips)
 	if config.StrictAffinity != true && rem != 0 {
 		logCtx.Infof("Attempting to assign %d more addresses from non-affine blocks", rem)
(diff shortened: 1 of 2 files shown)


lib/ipam/ipam.go
```diff
@@ -446,6 +437,9 @@ func (c ipamClient) autoAssign(ctx context.Context, num int, handleID *string, a
 						logCtx.WithError(err).Warningf("Failed to assign IPs from non-affine block in pool %s", p.String())
 						break
 					}
+					if len(newIPs) == 0 {
+						break
```
tomdee: Would it be worth adding a log line here?

caseydavenport: I had one but removed it because we already generate a log line when there are no IPs.

PR #931, opened 2018-09-26
Allow the blockSize to be configured for Calico IPAM

tomdee:
Allow the blockSize to be configured for Calico IPAM.
    
This removes the current fixed block size of /26 (for ipv4) or /122 (for IPv6).
The previous fixed values, now just become defaults.
    
The block size can be set per-pool, but once set it can't be changed.
```release-note
New Feature: Allow the blockSize to be configured for Calico IPAM.
```

### DIFF
--- a/lib/clientv3/ippool.go
+++ b/lib/clientv3/ippool.go
@@ -91,7 +91,7 @@ func (r ipPools) Update(ctx context.Context, res *apiv3.IPPool, opts options.Set
 		return nil, err
 	}
 
-	// Get the existing settings, so that we can validate the CIDR has not changed.
+	// Get the existing settings, so that we can validate the CIDR and block size have not changed.
 	old, err := r.Get(ctx, res.Name, options.GetOptions{})
 	if err != nil {
 		return nil, err
@@ -189,8 +189,10 @@ func (r ipPools) Delete(ctx context.Context, name string, opts options.DeleteOpt
 func (r ipPools) Get(ctx context.Context, name string, opts options.GetOptions) (*apiv3.IPPool, error) {
 	out, err := r.client.resources.Get(ctx, opts, apiv3.KindIPPool, noNamespace, name)
 	if out != nil {
+		convertIpPoolFromStorage(out.(*apiv3.IPPool))
 		return out.(*apiv3.IPPool), err
 	}
+
 	return nil, err
 }
 
@@ -200,7 +202,38 @@ func (r ipPools) List(ctx context.Context, opts options.ListOptions) (*apiv3.IPP
 	if err := r.client.resources.List(ctx, opts, apiv3.KindIPPool, apiv3.KindIPPoolList, res); err != nil {
 		return nil, err
 	}
+
+	// Default values when reading from backend.
+	for _, pool := range res.Items {
+		convertIpPoolFromStorage(&pool)
+	}
+
 	return res, nil
+}
+
+// Default pool values when reading from storage
+func convertIpPoolFromStorage(pool *apiv3.IPPool) error {
+	// Default the blockSize if it wasn't previously set
+	if pool.Spec.BlockSize == 0 {
+		// Get the IP address of the CIDR to find the IP version
+		ipAddr, _, err := cnet.ParseCIDR(pool.Spec.CIDR)
+		if err != nil {
+			return cerrors.ErrorValidation{
+				ErroredFields: []cerrors.ErroredField{{
+					Name:   "IPPool.Spec.CIDR",
+					Reason: "IPPool CIDR must be a valid subnet",
+					Value:  pool.Spec.CIDR,
+				}},
+			}
+		}
+
+		if ipAddr.Version() == 4 {
+			pool.Spec.BlockSize = 26
+		} else {
+			pool.Spec.BlockSize = 122
+		}
+	}
+	return nil
 }
 
 // Watch returns a watch.Interface that watches the IPPools that match the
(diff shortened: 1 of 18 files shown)


lib/clientv3/ippool.go
```diff
@@ -251,6 +284,57 @@ func (r ipPools) validateAndSetDefaults(ctx context.Context, new, old *apiv3.IPP
...
+	if old != nil && old.Spec.BlockSize != new.Spec.BlockSize {
+		errFields = append(errFields, cerrors.ErroredField{
+			Name:   "IPPool.Spec.BlockSize",
+			Reason: "IPPool BlockSize cannot be modified",
+			Value:  new.Spec.BlockSize,
+		})
+	}
+
+	// Default the blockSize
+	if new.Spec.BlockSize == 0 {
```
caseydavenport: Should we perform this defaulting before the assertion right above?

For example, if someone applies a pool with `blockSize` unset, and the one in the datastore is set to `/26`, then we don't need to throw a validation error there.

lib/clientv3/ippool.go
```diff
@@ -251,6 +284,57 @@ func (r ipPools) validateAndSetDefaults(ctx context.Context, new, old *apiv3.IPP
...
+		} else {
+			new.Spec.BlockSize = 122
+		}
+	}
+
+	if ipAddr.Version() == 4 {
+		if new.Spec.BlockSize > 32 || new.Spec.BlockSize < 20 {
+			errFields = append(errFields, cerrors.ErroredField{
+				Name:   "IPPool.Spec.BlockSize",
+				Reason: "IPPool BlockSize must be between 20 and 32",
```
caseydavenport: Might want to say "IPv4 block size must be..."

lib/clientv3/ippool_e2e_test.go
```diff
@@ -34,28 +34,32 @@ import (
 	"github.com/projectcalico/libcalico-go/lib/watch"
 )
 
-var _ = testutils.E2eDatastoreDescribe("IPPool tests", testutils.DatastoreAll, func(config apiconfig.CalicoAPIConfig) {
+var _ = testutils.E2eDatastoreDescribe("IPPool tests", testutils.DatastoreEtcdV3, func(config apiconfig.CalicoAPIConfig) {
```
caseydavenport: Hm, I guess there is now some IPAM code that gets hit by these tests?

I think we should be aiming to keep this set to `DatastoreAll` if possible. Can we pull out the bits that only work for etcd into a separate set of tests?

I guess CRUD on IPPools now lists IPAM blocks, which makes this fail for KDD? If so, we should track adding this back somewhere as part of implementing KDD IPAM so we don't lose these.

lib/ipam/ipam.go
```diff
@@ -773,8 +792,15 @@ func (c ipamClient) ClaimAffinity(ctx context.Context, cidr net.IPNet, host stri
 // its affinity will not be released and no error will be returned.
 // If an empty string is passed as the host, then the hostname is automatically detected.
 func (c ipamClient) ReleaseAffinity(ctx context.Context, cidr net.IPNet, host string) error {
+	// Verify the requested CIDR falls within a configured pool.
+	pool := c.blockReaderWriter.withinConfiguredPools(net.IP{IP: cidr.IP})
```
caseydavenport: I think we should rename this function from `withinConfiguredPools` to something like `getPoolForIP` or something like that to match its new behavior.

lib/ipam/ipam.go
```diff
@@ -1165,13 +1193,15 @@ func (c ipamClient) decrementHandle(ctx context.Context, handleID string, blockC
 // GetAssignmentAttributes returns the attributes stored with the given IP address
 // upon assignment.
 func (c ipamClient) GetAssignmentAttributes(ctx context.Context, addr net.IP) (map[string]string, error) {
-	blockCIDR := getBlockCIDRForAddress(addr)
-	obj, err := c.client.Get(ctx, model.BlockKey{blockCIDR}, "")
+	pool := c.blockReaderWriter.withinConfiguredPools(addr)
```
caseydavenport: Don't we need to check if the pool is nil?

lib/ipam/ipam_block.go
```diff
@@ -21,56 +21,49 @@ import (
...
-	for i := 0; i < blockSize; i++ {
+	for i := 0; i < numAddresses; i++ {
 		b.Unallocated[i] = i
 	}
 
-	return allocationBlock{&b}
+	return allocationBlock{&b, numAddresses}
+}
+
+func getNumAddresses(cidr cnet.IPNet, blockSize int) int {
```
caseydavenport: I think this guy could use a comment.

lib/ipam/ipam_block.go
```diff
@@ -21,56 +21,49 @@ import (
...
-
 // Wrap the backend AllocationBlock struct so that we can
 // attach methods to it.
 type allocationBlock struct {
 	*model.AllocationBlock
+	numAddresses int
 }
 
-func newBlock(cidr cnet.IPNet) allocationBlock {
+func newBlock(cidr cnet.IPNet, blockSize int) allocationBlock {
```
caseydavenport: I'm a bit confused why we need both `cidr` and `blockSize` here, especially since it looks like all the calling code calculates `blockSize` using `cidr.Mask.Size()`

lib/ipam/ipam_block_reader_writer.go
```diff
@@ -63,7 +64,8 @@ func (rw blockReaderWriter) getAffineBlocks(ctx context.Context, host string, ve
 			ids = append(ids, k.CIDR)
 		} else {
 			for _, pool := range pools {
-				if pool.Contains(k.CIDR.IPNet.IP) {
+				poolNet := cnet.MustParseCIDR(pool.Spec.CIDR)
```
caseydavenport: I think we want to avoid using `MustParseCIDR` in production code - IIRC it was a testing tool?

If it fails to parse it will panic. It shouldn't ever happen, since we validate the CIDR, but I think still best to handle this gracefully and avoid panics in libcalico.

lib/ipam/ipam_block_reader_writer.go
```diff
@@ -440,3 +451,28 @@ func randomBlockGenerator(pool cnet.IPNet, hostName string) func() *cnet.IPNet {
...
+	for _, o := range datastoreObjs.KVPairs {
+		k := o.Key.(model.BlockKey)
+		if k.CIDR.IPNet.Contains(ip.IP) {
+			log.Debugf("Found IP %s in block %s", ip.String(), k.String())
+			return &k.CIDR, nil
+		}
+	}
+
+	// No blocks found.
+	//return cnet.IPNet{}, errors.New(fmt.Sprintf("no blocks found containing IP %s", ip.String()))
```
caseydavenport: Can drop this commented line

lib/ipam/ipam_block.go
```diff
@@ -21,56 +21,34 @@ import (
...
-	TotalBits:         128,
-	BlockPrefixLength: 122,
-	BlockPrefixMask:   net.CIDRMask(122, 128),
-}
-
 // Wrap the backend AllocationBlock struct so that we can
 // attach methods to it.
 type allocationBlock struct {
 	*model.AllocationBlock
+	numAddresses int
```
caseydavenport: Not a big deal, but couldn't we do this without storing this information in a new field?

e.g. just call `numAddresses(..)` on the block cidr whenever we need this information or expose a function that can calculate this on the fly? Doesn't seem too computationally complex.

lib/ipam/ipam_block_reader_writer.go
```diff
@@ -377,18 +380,31 @@ func blockGenerator(pool cnet.IPNet) func() *cnet.IPNet {
 // Returns a generator that, when called, returns a random
 // block from the given pool.  When there are no blocks left,
 // the it returns nil.
-func randomBlockGenerator(pool cnet.IPNet, hostName string) func() *cnet.IPNet {
+func randomBlockGenerator(ipPool *v3.IPPool, hostName string) func() *cnet.IPNet {
+	_, pool, err := cnet.ParseCIDR(ipPool.Spec.CIDR)
+	if err != nil {
```
caseydavenport: ick, I guess we should log an error in this case so if somehow we do get here we'll see it in the logs.

PR #932, opened 2018-10-02
Fix node labeling

caseydavenport:
## Description

Was seeing that deleting a node label wasn't working. Turns out this is why!

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/clientv3/node_e2e_test.go
+++ b/lib/clientv3/node_e2e_test.go
@@ -39,7 +39,49 @@ import (
 	"github.com/projectcalico/libcalico-go/lib/watch"
 )
 
-var _ = testutils.E2eDatastoreDescribe("Node tests", testutils.DatastoreEtcdV3, func(config apiconfig.CalicoAPIConfig) {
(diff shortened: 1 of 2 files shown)


lib/clientv3/node_e2e_test.go
```diff
@@ -221,6 +221,47 @@ var _ = testutils.E2eDatastoreDescribe("Node tests", testutils.DatastoreEtcdV3,
 			bconfig, err := c.BGPConfigurations().Get(ctx, nodeConfigName, options.GetOptions{})
 			Expect(bconfig).Should(BeNil())
 		})
+
+	})
+
+	var _ = testutils.E2eDatastoreDescribe("Node tests (kdd)", testutils.DatastoreK8s, func(config apiconfig.CalicoAPIConfig) {
```
neiljerram: Indentation looks wrong here.  Is it because this is positioned _inside_ the `E2eDatastoreDescribe` for "Node tests (etcdv3)"?  Surely that's wrong, if so?

caseydavenport: Yes... good spot!

I suppose I trust `goimports` a bit too much :)

PR #935, opened 2018-10-05
IPPool Create: Error if blocks conflict

tomdee:
- [ ] Tests
- [ ] Documentation
- [ ] Release note

```release-note
None required
```

When IPPools are deleted, blocks are left in place. Now that IPPools can
be created with different blockSizes, it's important that an
IPPool can't be created that would conflict with existing blocks that
use a different blockSize. If this was allowed, then overlapping blocks
would be created and IP addresses could be double assigned.

When creating an IPPool, check the list of existing blocks that are
covered by the pool being created. If they have a different block size
then throw an error.

### DIFF
--- a/lib/clientv3/ippool.go
+++ b/lib/clientv3/ippool.go
@@ -20,14 +20,14 @@ import (
 	"net"
 	"time"
 
-	log "github.com/sirupsen/logrus"
-
 	apiv3 "github.com/projectcalico/libcalico-go/lib/apis/v3"
+	"github.com/projectcalico/libcalico-go/lib/backend/model"
 	cerrors "github.com/projectcalico/libcalico-go/lib/errors"
 	cnet "github.com/projectcalico/libcalico-go/lib/net"
 	"github.com/projectcalico/libcalico-go/lib/options"
 	validator "github.com/projectcalico/libcalico-go/lib/validator/v3"
 	"github.com/projectcalico/libcalico-go/lib/watch"
+	log "github.com/sirupsen/logrus"
 )
 
 // IPPoolInterface has methods to work with IPPool resources.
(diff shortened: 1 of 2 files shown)


lib/clientv3/ippool.go
```diff
@@ -63,9 +63,49 @@ func (r ipPools) Create(ctx context.Context, res *apiv3.IPPool, opts options.Set
...
+
+	blocks, err := r.client.backend.List(ctx, model.BlockListOptions{IPVersion: ipVersion}, "")
+	// If the error is OperationUnsupported then skip the blockSize check - IPAM is not supported on KDD.
+	if _, ok := err.(cerrors.ErrorOperationNotSupported); !ok && err != nil {
+		return nil, err
+	} else if err != nil {
+		for _, b := range blocks.KVPairs {
+			k := b.Key.(model.BlockKey)
+			ones, _ := k.CIDR.Mask.Size()
+			if ones != poolBlockSize {
```
caseydavenport: Could we combine these two if statements so we don't need to nest so far?

lib/clientv3/ippool.go
```diff
@@ -63,9 +63,49 @@ func (r ipPools) Create(ctx context.Context, res *apiv3.IPPool, opts options.Set
...
+	}
+
+	ipVersion := 4;
+	if poolIP.To4() == nil {
+		ipVersion = 6
+	}
+
+	blocks, err := r.client.backend.List(ctx, model.BlockListOptions{IPVersion: ipVersion}, "")
+	// If the error is OperationUnsupported then skip the blockSize check - IPAM is not supported on KDD.
+	if _, ok := err.(cerrors.ErrorOperationNotSupported); !ok && err != nil {
```
caseydavenport: Does this work when err is an OperationNotSupported? Won't we go into the next branch and try to access a nil pointer?

caseydavenport: Yup I think that looks good now.

lib/clientv3/ippool.go
```diff
@@ -63,9 +63,47 @@ func (r ipPools) Create(ctx context.Context, res *apiv3.IPPool, opts options.Set
...
+	if poolIP.To4() == nil {
+		ipVersion = 6
+	}
+
+	blocks, err := r.client.backend.List(ctx, model.BlockListOptions{IPVersion: ipVersion}, "")
+	if _, ok := err.(cerrors.ErrorOperationNotSupported); !ok && err != nil {
+		// There was an error and it wasn't OperationNotSupported - return it.
+		return nil, err
+	} else if err == nil {
+		// SKip the block check if the error is OperationUnsupported - IPAM is not supported on KDD.
```
caseydavenport: nit: SKip

lib/clientv3/ippool.go
```diff
@@ -63,9 +63,48 @@ func (r ipPools) Create(ctx context.Context, res *apiv3.IPPool, opts options.Set
...
+		// Skip the block check if the error is OperationUnsupported - IPAM is not supported on KDD.
+		for _, b := range blocks.KVPairs {
+			k := b.Key.(model.BlockKey)
+			ones, _ := k.CIDR.Mask.Size()
+			// Check if this block has a different size to the pool, and that it overlaps with the pool.
+			if ones != poolBlockSize && k.CIDR.IsNetOverlap(*poolCIDR) {
+				return nil, cerrors.ErrorValidation{
+					ErroredFields: []cerrors.ErroredField{{
+						Name:   "IPPool.Spec.BlockSize",
+						Reason: "IPPool BlocksSize conflicts with existing allocations that use a different blockSize",
```
caseydavenport: BlocksSize -> blockSize (to be consistent with name as present in the user-facing yaml/json)

PR #936, opened 2018-10-09
Add blocksize tests and simple benchmark

tomdee:
- [ ] Tests
- [ ] Documentation
- [ ] Release note

```release-note
None required
```


lib/ipam/randomblock_test.go
```diff
@@ -33,6 +33,10 @@ var _ = Describe("Random Block Generator", func() {
 
 		Entry("IPv4 CIDR", "10.10.0.0/24", 26),
 		Entry("IPv6 CIDR", "fd80:24e2:f998:72d6::/120", 122),
+		Entry("IPv4 CIDR", "10.10.0.0/16", 20),
```
caseydavenport: Should probably have unique descriptions for all of these tests.

lib/ipam/ipam_test.go
```diff
@@ -90,14 +100,92 @@ type testArgsClaimAff struct {
...
-	// Tests that need to ensure a clean datastore should invokke Clean() on the datastore at the start of the
+	// Tests that need to ensure a clean datastore should invoke Clean() on the datastore at the start of the
 	// tests.
 	bc, err := backend.NewClient(config)
 	if err != nil {
 		panic(err)
 	}
 	ic := NewIPAMClient(bc, ipPools)
 
+	Context("With a clean datastore", func() {
```
caseydavenport: I think the context here should be something that describes the tests - e.g. "Measuring allocation performance"

lib/clientv3/ippool_e2e_test.go
```diff
@@ -705,6 +705,98 @@ var _ = testutils.E2eDatastoreDescribe("IPPool tests", testutils.DatastoreAll, f
...
+					CIDR:      "1.2.3.0/24",
+					BlockSize: 33,
+				},
+			}, options.SetOptions{})
+			Expect(err).To(HaveOccurred())
+			Expect(err).To(BeAssignableToTypeOf(errors.ErrorValidation{}))
+			Expect(err.Error()).To(ContainSubstring("block size must be between"))
+		})
+
+		It("should prevent the creation of a pool with an identical or overlapping CIDR using block sizes", func() {
```
caseydavenport: Do we have a test that checks the mismatched block size logic?

e.g.
- Create a pool, blocksize X
- Allocate an IP (and thus a block) from that pool
- Delete the first pool.
- Recreate the pool with blocksize Y, see that it fails.
- Release the IP allocation
- Recreate the pool with blocksize Y, see that it succeeds.

tomdee: I thought I already covered that - https://github.com/projectcalico/libcalico-go/pull/935/files#diff-0912dca51c2f14012b137b66c0b4babdR727

caseydavenport: Yep, looks like it - I just couldn't remember if it was already there or not.

PR #940, opened 2018-10-15
Update data model for HostEndpoints with interfaceName *

neiljerram:
Part of https://github.com/projectcalico/calico/pull/2232

### DIFF
--- a/lib/clientv3/common_test.go
+++ b/lib/clientv3/common_test.go
@@ -31,6 +31,8 @@ import (
 	"github.com/projectcalico/libcalico-go/lib/testutils"
 )
 
+var MatchResource = testutils.Resource
+
 var _ = testutils.E2eDatastoreDescribe("Common resource tests", testutils.DatastoreAll, func(config apiconfig.CalicoAPIConfig) {
 	Describe("Common resource tests", func() {
 		It("Should return the data that was stored in the datastore, even if it was changed", func() {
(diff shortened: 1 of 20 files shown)


lib/clientv3/hostendpoint_e2e_test.go
```diff
@@ -146,9 +151,13 @@ var _ = testutils.E2eDatastoreDescribe("HostEndpoint tests", testutils.Datastore
 			outList, outError = c.HostEndpoints().List(ctx, options.ListOptions{})
 			Expect(outError).NotTo(HaveOccurred())
 			Expect(outList.Items).To(HaveLen(2))
-			testutils.ExpectResource(&outList.Items[0], apiv3.KindHostEndpoint, testutils.ExpectNoNamespace, name1, spec1)
-			testutils.ExpectResource(&outList.Items[1], apiv3.KindHostEndpoint, testutils.ExpectNoNamespace, name2, spec2)
-
+			if name1 < name2 {
```
caseydavenport: Seems like it would be much better to have an ExpectResourceList() assertion which asserts that the given resource is within the given list.

WDYT? This approach feels very brittle to me.

caseydavenport: Alternatively, we could just swap the order of the arguments in the Entry() so that name1 is always < name2?

That's still brittle though.

neiljerram: It has snowballed a bit, but I think I've improved the situation by using ConsistOf and a gomega matcher....

lib/clientv3/common_test.go
```diff
@@ -31,6 +31,8 @@ import (
 	"github.com/projectcalico/libcalico-go/lib/testutils"
 )
 
+var MatchResource = testutils.Resource
```
caseydavenport: Isn't this already defined in testutils?

neiljerram: Yes, but that's in a different package (testutils, as opposed to clientv3_test).  The reason for it here is to avoid all the test code having to say `Expect(x).To(testutils.MatchResource(...))`.

PR #945, opened 2018-10-24
Fix #944: Set NodeNetworkUnavailable flag to False on k8s

### LINKED ISSUE #944: Needs to clear NodeNetworkUnavailable flag on Kubernetes
Same issue : https://github.com/projectcalico/canal/issues/74

## Expected Behavior
Calico has to update the `NetworkUnavailable` status on node

## Current Behavior
`NetworkUnavailable` stay to `True` and you can't use the node

## Possible Solution
Calico has to update the `NetworkUnavailable` status on node

Use calico on GCE with `kubelet --cloudprovider=gce`

## Context
https://github.com/kubernetes/kubernetes/issues/44254
https://kubernetes.io/docs/concepts/architecture/nodes/#condition
https://github.com/weaveworks/weave/pull/3307

aarnaud:
Ref issue : https://github.com/projectcalico/libcalico-go/issues/944

Calico has to update the `NetworkUnavailable` status on node

We have to add new authorization to calico-node
```
kind: ClusterRole
apiVersion: rbac.authorization.k8s.io/v1beta1
metadata:
  name: calico-node
rules:
  - apiGroups: [""]
    resources:
      - nodes/status
    verbs:
      - patch
      - update
```

## Context
https://github.com/kubernetes/kubernetes/issues/44254
https://kubernetes.io/docs/concepts/architecture/nodes/#condition
https://github.com/weaveworks/weave/pull/3307

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -1,5 +1,5 @@
-hash: 15945432714a12a6889c810b136591540f5f089a0893f33040f6c5df49bc3a82
-updated: 2018-07-19T15:44:36.618139978-07:00
+hash: a8427dbe46eb3e380d06794674aa5ced0377e4dc9a4501ba11e88fe863d24249
+updated: 2018-10-24T10:19:38.572931402-04:00
 imports:
 - name: cloud.google.com/go
   version: 3b1ae45394a234c385be014e9a488f2bb6eef821
@@ -18,7 +18,7 @@ imports:
   subpackages:
   - quantile
 - name: github.com/coreos/etcd
-  version: 33245c6b5b49130ca99280408fadfab01aac0e48
+  version: 420a452267a7ce45b3fcbed04d54030d69964fc1
   subpackages:
   - auth/authpb
   - client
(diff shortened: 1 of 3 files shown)


caseydavenport:
@aarnaud thanks for the PR!

I think there are a few things we need to consider while implementing this:
- Sometimes, Calico is running in "policy only" mode, and relying on something else (e.g. GCE routes) to do routing. In that case, it shouldn't set NetworkUnavailable to false, since it's not doing the networking.
- This code path only gets hit when running in k8s API datastore mode. Some users do use etcd as their datastore when running in k8s, so we probably want to move this function into a common codepath.
- We don't want to execute this code when we're running on a non-k8s node. 

I think we can address these by moving the code in this PR into the node startup logic here: https://github.com/projectcalico/node/blob/master/pkg/startup/startup.go

Then, we'll have all the information available as to whether or not we want to set this to false.

aarnaud:
Thanks @caseydavenport, I hesitated to fix that in startup.go, i will do that

aarnaud:
@caseydavenport I close this PR because the patch was move in https://github.com/projectcalico/node/pull/89

PR #948, opened 2018-10-31
[wip] Interim BGPRoute data model

neiljerram:


lib/apis/v3/bgproute.go
```diff
@@ -0,0 +1,73 @@
...
+// BGPRouteList contains a list of BGPRoute resources.
+type BGPRouteList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata"`
+	Items           []BGPRoute `json:"items"`
+}
+
+// NewBGPRoute creates a new (zeroed) BGPRoute struct with the TypeMetadata initialised to the current
+// version.
+func NewBGPRoute() *BGPRoute {
```
briansan: i'm not sure what the design guidelines are for creating new CRDs but would it make sense to accept `(nodeName, cidr string)` as args? that way the caller of this function could construct a complete object

lib/apis/v3/bgproute.go
```diff
@@ -0,0 +1,73 @@
...
+		TypeMeta: metav1.TypeMeta{
+			Kind:       KindBGPRoute,
+			APIVersion: GroupVersionCurrent,
+		},
+	}
+}
+
+// NewBGPRouteList creates a new (zeroed) BGPRouteList struct with the TypeMetadata initialised to the current
+// version.
+func NewBGPRouteList() *BGPRouteList {
```
briansan: similarly accept `(routes... *BGPRoute)` as variable args into a list to return a fully constructed `BGPRouteList` object?

neiljerram:
Decided that we don't need this for now.

PR #956, opened 2018-11-04
Automatic update of libcalico-go to 489a4d4fcdf275caca1edf59e95054b6e8601555

doublek:


doublek:
Script bug/error. No need to update libcalico-go for a libcalico-go update.

PR #969, opened 2018-11-11
Cleanup tests

caseydavenport:
## Description

Had a couple of hours on a plane so thought I'd make some improvements here.

PR turned out quite a bit longer than I had hoped... but I think it has some good stuff in it.

- Split out tests which require a datastore into an "fv" target, so its easier to run a "fast" subset when developing.
- Standardize on Ginkgo - there were a few tests not using it.
- Fixup a number of suites which didn't have logging configured properly.
- Fix some copy/paste errors I spotted where report files had the same name across suites. 

`make ut` now takes about 20 seconds on my machine.
`make fv` still takes > 5min.

`make test` runs both `ut` and `fv`. 

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ default: build
 all: build
 
 ## Run the tests
-test: ut
+test: ut fv
 
 # Define some constants
 #######################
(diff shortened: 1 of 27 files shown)


caseydavenport:
> Are you sure you want to standardise on Ginkgo? I've been allowing go test tests to go in and I think we've worked out that they're compatible. I'd like to use more go test tests over time; ginkgo has too steep a learning curve. OTOH, it'a lot of work to got from ginkgo to go test everywhere :-(

It seems to be that we've already standardized on Ginkgo, based on the number of tests we have written in it. I think it's valuable to use a single framework - not against making an official switch off of Ginkgo, but for now I think it's our default, and I'd rather not have two test frameworks where we could have one.

Issue #505, opened 2017-09-01
Use "patch" instead of "update"


caseydavenport:
Calico should be using PATCH to update k8s resources so it doesn't conflict or write over other data in the k8s API. 

Specifically, it should do this for the following resources that currently use update:
- [ ] Nodes/status (instead of current Update on Nodes)
- [x] Pods/status

I've seen this manifest on a 2k node cluster when calico is racing with something else when trying to update node resources. Looks like this:

```
$ k logs calico-node-r143m -n kube-system -c calico-node -p
Checking datastore connection
Datastore connection verified
ERROR: Unable to set node resource configuration: update conflict: 'Node(name=gke-casey-2k-test-default-pool-e6a5c621-479x)'
Terminating
time="2017-09-01T17:22:15Z" level=error msg="Failed to apply object: too many retries" Key=Node(name=gke-casey-2k-test-default-pool-e6a5c621-479x)
Calico node failed to start
```

Causes a number of restarts for some pods:

```
calico-node-zw81w   2/2       Running   4          2h
calico-node-zw8z9   2/2       Running   3          2h
calico-node-zwr08   2/2       Running   0          2h
calico-node-zx7rh   2/2       Running   0          2h
calico-node-zx9fx   2/2       Running   0          2h
calico-node-zxmhx   2/2       Running   0          2h
calico-node-zz1l4   2/2       Running   0          2h
calico-node-zz1t3   2/2       Running   1          2h
calico-node-zz40j   2/2       Running   2          2h
calico-node-zz5l0   2/2       Running   0          2h
calico-node-zzm0x   2/2       Running   1          2h
calico-node-zzmv6   2/2       Running   4          2h
```

## Expected Behavior

Calico should start successfully without needing to restart and should not write over data in the k8s API that it doesn't own.

## Current Behavior

Calico needs to restart a few times before it succeeds in updating node config, and might potentially write over data it doesn't own.

## Possible Solution

I think we need to be using PATCH for operations on the node + pods/status in KDD.

## Context

I've seen the conflict above on really big clusters, or clusters where there are lots of things acting on nodes, but it could happen on smaller ones as well if the timing is right.

caseydavenport:
Probably don't want to use patch for nodes, since we still want CAS operations.
Best just to use the status subresource but still use Update()

caseydavenport:
FWIW I've run tests recently that no longer hit the conflict, but we still should drop our privileges so we only need status access.

PR #971, opened 2018-11-12
Switching node to use UpdateStatus instead of Update

insequent:
## Description

Changing nodeClient.Update() to use K8s' client `UpdateStatus` instead of `Update`. This is meant to help with concurrency issues we've seen in the past. 

## Todos
- [ ] Tests
- [x] Documentation
- [ ] Release note

## Release Note
```release-note
Switching node resource to use UpdateStatus instead of Update when updating K8s
```

### DIFF
--- a/lib/backend/k8s/resources/node.go
+++ b/lib/backend/k8s/resources/node.go
@@ -60,6 +60,7 @@ func (c *nodeClient) Create(ctx context.Context, kvp *model.KVPair) (*model.KVPa
 }
 
 func (c *nodeClient) Update(ctx context.Context, kvp *model.KVPair) (*model.KVPair, error) {
+	log.Debug("Received Update request on Node type")
 	// Get a current copy of the node to fill in fields we don't track.
 	oldNode, err := c.clientSet.CoreV1().Nodes().Get(kvp.Key.(model.ResourceKey).Name, metav1.GetOptions{})
 	if err != nil {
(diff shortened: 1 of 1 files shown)


insequent:
In reference to https://github.com/projectcalico/libcalico-go/issues/505

lib/backend/k8s/resources/node.go
```diff
@@ -60,7 +60,13 @@ func (c *nodeClient) Create(ctx context.Context, kvp *model.KVPair) (*model.KVPa
 }
 
 func (c *nodeClient) Update(ctx context.Context, kvp *model.KVPair) (*model.KVPair, error) {
-	// Get a current copy of the node to fill in fields we don't track.
+	log.Debug("Received Update request on Node type")
+	// We switched to using UpdateStatus here to avoid conflicts in write
```
caseydavenport: I think the idea here is to reduce privileges primarily.

lib/backend/k8s/resources/node.go
```diff
@@ -60,7 +60,13 @@ func (c *nodeClient) Create(ctx context.Context, kvp *model.KVPair) (*model.KVPa
...
+	// https://github.com/kubernetes/kubernetes/issues/39113
+	return c.UpdateStatus(ctx, kvp)
+}
+
+func (c *nodeClient) UpdateStatus(ctx context.Context, kvp *model.KVPair) (*model.KVPair, error) {
```
caseydavenport: I don't think we need a separate function here for `UpdateStatus` - especially not a public one.

Let's just keep this as one function - `Update()` and simply change the API call we're making to k8s.

PR #974, opened 2018-11-16
Add IPAM arg to set a limit on the number of auto-assigned blocks.

fasaxc:
## Description

Need this for a CNI plugin workaround that I'm working on.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/ipam/ipam.go
+++ b/lib/ipam/ipam.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2016-2017 Tigera, Inc. All rights reserved.
+// Copyright (c) 2016-2018 Tigera, Inc. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


lib/ipam/ipam_test.go
```diff
@@ -513,6 +514,9 @@ var _ = testutils.E2eDatastoreDescribe("IPAM tests", testutils.DatastoreEtcdV3,
 				Hostname:  host,
 				IPv4Pools: []cnet.IPNet{fromPool},
 			}
+			if strings.Contains(host, "block-limit") {
```
caseydavenport: Couldn't we just add a new argument to this?

Might be useful to be able to test block limits other than 1 or 0.

lib/ipam/ipam.go
```diff
@@ -320,6 +329,12 @@ func (c ipamClient) autoAssign(ctx context.Context, num int, handleID *string, a
 		rem := num - len(ips)
 		retries := ipamEtcdRetries
 		for rem > 0 && retries > 0 {
+			if maxNumBlocks > 0 && numBlocksOwned >= maxNumBlocks {
+				log.WithError(err).Errorf("Unable to allocate a new block, hit blocks per host limit (%v)",
```
caseydavenport: log.WithError(err) here isn't quite right, right? Since we're not handling an error returned from elsewhere.

Also, would be nice to include numBlocksOwned as well, just in case we somehow get a weird situation where we own say 5 blocks but the limit is 1.

caseydavenport: Could do:

```
fields := log.Fields{"limit": maxNumBlocks, "owned": numBlocksOwned}
log.WithFields(fields).WithError(ErrBlockLimit).Errorf("unable to allocate block")
```

PR #977, opened 2018-11-21
Add WHAT and GINKGO_FOCUS to test commands.

fasaxc:
## Description

Makes it possible to do `make WHAT=./lib/ipam` and `make GINKGO_FOCUS=foo`

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -144,6 +144,9 @@ check-glide-warnings:
 ut-cover: vendor
 	./run-uts
 
+WHAT?=.
+GINKGO_FOCUS?=.*
+
 .PHONY:ut
 ## Run the fast set of unit tests in a container.
 ut: vendor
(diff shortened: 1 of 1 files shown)


fasaxc:
Ran some more tests locally and checked semaphore ran the right baseline tests.

caseydavenport:
Hey @fasaxc @neiljerram  - I think we had rolled out standardization which agreed upon `GINKGO_ARGS` being the means to allow passing arguments to GINKGO.

There are a few things about this PR that worry me:

- It breaks standardization with other repositories.
- It provides knobs which duplicate functionality with existing options. 

Is there something we can do to address those and also meet the needs this PR is trying to address? What _are_ the needs this PR is trying to address?

fasaxc:
There's already a focus in the base command so you can't pass another one.

caseydavenport:
@fasaxc huh, when I was playing with this it seemed like `ginkgo` supported multiple focus arguments, though it looks like I interpreted that incorrectly. When specifying multiple `-focus` arguments, it prefers the last one specified, rather than doing an AND combination. 

Given that, it seems the `GINKGO_FOCUS` is maybe warranted (though overriding the default also seems like it would be fine in most cases). 

What about the `WHAT` argument? Can't we just do something like this to get the same effect without needing a new variable?

```
make fv GINKGO_ARGS="./lib/ipam"
```

fasaxc:
I think the base command had a path in it so you ended up running the test
you wanted, and all the others.

caseydavenport:
Yeah, it had a hardcoded `.` in it before. We could add that `.` to `GINKGO_ARGS` but then you'd need to remember to add it back in when overriding, which is annoying.

I like the idea of adding a `WHAT` option but we need to roll it out across all the core repositories for consistency. I don't want fragmented arguments across the different repos.

PR #978, opened 2018-11-21
Fix up generated file dependencies.

fasaxc:
## Description

- make gen-files a real PHONY so it forces a rebuild
- make other targets depend on files
- fix git diff check to only check generated files.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,13 +46,6 @@ clean:
 	rm -rf $(BINDIR)
 	rm -rf checkouts
 
-.PHONY: clean-gen-files
-## Convenience target for devs to remove generated files and related utilities
-clean-gen-files:
-	rm -f $(BINDIR)/deepcopy-gen
-	rm -f ./lib/apis/v3/zz_generated.deepcopy.go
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -63,17 +63,20 @@ vendor: glide.lock
...
-## Build generated-go utilities (e.g. deepcopy_gen) and generated files
-gen-files: $(BINDIR)/deepcopy-gen \
-           ./lib/apis/v3/zz_generated.deepcopy.go \
+GENERATED_FILES:=./lib/apis/v3/zz_generated.deepcopy.go \
            ./lib/upgrade/migrator/clients/v1/k8s/custom/zz_generated.deepcopy.go
 
+.PHONY: gen-files
+## Build generated-go utilities (e.g. deepcopy-gen) and generated files
+gen-files:
+	$(MAKE) clean-gen-files
```
neiljerram: Suggest at least checking that `clean-gen-files` is good, and possibly inlining the `rm` here, as `make clean-gen-files` shouldn't be a useful thing anymore.

fasaxc: done

PR #986, opened 2018-12-04
WIP: implement method to diff labels

briansan:
## Description
- `ApplyKddMeta` implements logic to determine whether or not labels from the input `ObjectMeta` into the `Node` object.
-- stores the input `ObjectMeta.ResourceVersion` as `Annotations["kdd.resourceVersion"]` in calico node
-- prefix all labels from the k8s Node with `kdd.`

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note
```
todo
```

### DIFF
--- a/lib/apis/v3/node.go
+++ b/lib/apis/v3/node.go
@@ -37,6 +37,42 @@ type Node struct {
 	Spec NodeSpec `json:"spec,omitempty"`
 }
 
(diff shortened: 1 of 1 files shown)


briansan:
note that this approach to the solution will probably also require a change to our selector logic/process

PR #987, opened 2018-12-05
Log out in FV test when we recive an unexpected number of events

caseydavenport:
## Description

In this build: https://semaphoreci.com/calico/libcalico-go/branches/master/builds/1175

We see this output:

```
2018-12-05 04:13:14.560 [INFO][1635] resources.go 223: Expected: EventType:ADDED; Kind:projectcalico.org/v3, Kind=BGPPeer; Name:bgppeer-2; Namespace:

2018-12-05 04:13:14.560 [INFO][1635] resources.go 238: Actual:   EventType:ADDED; Kind:projectcalico.org/v3, Kind=BGPPeer; Name:bgppeer-2; Namespace:

2018-12-05 04:13:14.560 [INFO][1635] resources.go 223: Expected: EventType:DELETED; Kind:projectcalico.org/v3, Kind=BGPPeer; Name:bgppeer-1; Namespace:

2018-12-05 04:13:14.560 [INFO][1635] resources.go 238: Actual:   EventType:ADDED; Kind:projectcalico.org/v3, Kind=BGPPeer; Name:bgppeer-2; Namespace:
```

Which seems to indicate we've gotten two ADDED events for the same thing, instead of an ADD and a DELETE.

Not sure what's going on, but this PR adds some diags that might be useful. 

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/testutils/resources.go
+++ b/lib/testutils/resources.go
@@ -199,11 +199,32 @@ func (t *testResourceWatcher) expectEvents(kind string, anyOrder bool, expectedE
 	// events, so protect against that scenario - we'll check later once we've
 	// constructed useful diagnostics.
 	var actualEvents []watch.Event
-	if len(t.events) >= len(expectedEvents) {
-		actualEvents = t.events[:len(expectedEvents)]
+	log.Infof("Received %s events, expected %d", len(t.events), len(expectedEvents))
+	if len(t.events) != len(expectedEvents) {
+		// Log out the events we received before failing the test.
+		log.Errorf("Number of received events does not match expected.")
+		for _, e := range t.events {
+			var o runtime.Object
+			if e.Type == watch.Deleted {
+				o = e.Previous
+			} else {
+				o = e.Object
+			}
+			log.Infof(
+				"Received event: EventType:%s; Kind:%s; Name:%s; Namespace:%s",
+				e.Type,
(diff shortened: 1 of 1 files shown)


lib/testutils/resources.go
```diff
@@ -199,8 +199,29 @@ func (t *testResourceWatcher) expectEvents(kind string, anyOrder bool, expectedE
 	// events, so protect against that scenario - we'll check later once we've
 	// constructed useful diagnostics.
 	var actualEvents []watch.Event
+	log.Infof("Received %s events, expected %d", len(t.events), len(expectedEvents))
+	// CASEY: TODO - Can we assert this generally?
+	Expect(len(t.events)).To(Equal(len(expectedEvents)), "Wrong number of events received")
```
caseydavenport: Tests seem to pass with this assertion, so might be worth adding it in. WDYT?

neiljerram: If we are keeping this, we should do it in a way that means we'll still benefit from the other added diags below.  (As it stands, I believe the added code below will never be executed.)

neiljerram: I'm afraid I don't know if it is correct to assert this generally.  I suggest trying to check out by reviewing where and how `expectEvents` is called.

PR #988, opened 2018-12-07
Force resync on too short watches: workaround for typha issue #227

ikatson:
This forces a resync if the watch was too short - this is the same heuristic used by k8s/client-go
to make watches more robust.

The actual issue is filed in typha as it was first observed there: https://github.com/projectcalico/typha/issues/227

This helps solve the problem reliably. The code makes "resync" behavior more aggressive, but there's nothing breaking here, so hopefully can get away without writing tests / learning how to use the calico testing harness.

### DIFF
--- a/lib/backend/watchersyncer/watchercache.go
+++ b/lib/backend/watchersyncer/watchercache.go
@@ -38,6 +38,7 @@ type watcherCache struct {
 	logger               *logrus.Entry
 	client               api.Client
 	watch                api.WatchInterface
+	watchStart           time.Time
 	resources            map[string]cacheEntry
 	oldResources         map[string]cacheEntry
 	results              chan<- interface{}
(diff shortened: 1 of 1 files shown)


ikatson:
I wanted to close this PR a couple days back, as it turned out to be mostly an openshift specific issue.

But i closed the actual openshift issue instead of it by mistake openshift/origin#21636

As @caseydavenport didn't close this though but instead asked for a review, maybe it's still worth considering merging in, as it definitely helps.

I'll leave it up to you guys to decide if to close it or merge, thanks!

caseydavenport:
@ikatson yeah, we may just close given the above, but thought it worth considering if this was a valid thing to do anyway, just as a precaution.

I'm a bit hesitant though, simply because increasing the chances of performing a full resync can cause performance issues if they occur too frequently.

Issue #989, opened 2018-12-10
IPAM Block does not protect against duplicates in Unallocated list


jsok:
[`AllocationBlock`](https://github.com/projectcalico/libcalico-go/blob/release-v3.4/lib/backend/model/block.go#L91) is a simple list and the code always assumes the values are unique.

## Expected Behavior

1. An assigned IP should not be assigned twice!
1. Whenever releasing IPs, the supplied list of IPs should be de-duped and handled defensively.

## Current Behavior

`IPAM.ReleaseIPs()` happily accepts a list of IPs that contains duplicates. This will leave duplicates in `Unallocated` which then creates more havoc, in particular **assigning already assigned IPs**.

Reproduced the behaviour in an e2e test: https://github.com/projectcalico/libcalico-go/pull/990

## Possible Solution

Validate ReleaseIPs input and also add checks to [`ipam_block.go`'s `release()` function](https://github.com/projectcalico/libcalico-go/blob/release-v3.4/lib/ipam/ipam_block.go#L203-L206) to stop duplicates being added. Ideally use `lib/set` to avoid bugs like this occurring again.

## Steps to Reproduce (for bugs)

Call `ReleaseIPs([]cnet.IP{ip, ip}` with a duplicate entry in the list, then keeping assigning IPs and you will eventually get `ip` assigned twice.

Have now observed multiple instances of doubly-assigned IPs in running production cluster.

Doing a quick audit reveals an IP block with duplicates in the unallocated list, but have observed other IP pools with the same issue:

```
for subnet in $(etcdctl ls --sort /calico/ipam/v2/assignment/ipv4/block)
do
    etcdctl get $subnet | jq 'if (.unallocated | length) != (.unallocated | unique | length) then "duplicate found \(.cidr)" else empty end'
done
```

## Context

A running container has it's IP assigned to another container. Whichever exits first will release that IP leaving the other with an IP that's been released. Rinse-and-repeat.

## Your Environment

Calico 2.6.12, running Docker with libnetwork on Mesos.

* libcalico-go version: 1.7.3
* Operating System and version: Ubuntu 18.04 bionic
* Link to your project (optional):

jsok:
@caseydavenport happy to work on a fix if you have any guidance, otherwise I have a solution I can suggest via PR?

caseydavenport:
[...]

PR #990, opened 2018-12-10
Reproduction of double-releaseIP bug

jsok:
## Description

Setup a scenario in which an IP is assigned, but it's ordinal appears twice in the AllocationBlock's Unallocated list making it possible to assign the IP twice.

Have branched of 2.6 release which affects my environment, but reading the 3.x code suggests this bug will still be present in newer releases too.

Proof-of-concept for #989 

## Todos
- [X] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

TBA

### DIFF
--- a/lib/client/ipam_block.go
+++ b/lib/client/ipam_block.go
@@ -156,9 +156,16 @@ func (b *allocationBlock) release(addresses []cnet.IP) ([]cnet.IP, map[string]in
 	delRefCounts := map[int]int{}
 	attrsToDelete := []int{}
 
+	// De-duplicate addresses to ensure reference counting is correcet
+	uniqueAddresses := make(map[string]struct{})
+	for _, ip := range addresses {
+		uniqueAddresses[ip.IP.String()] = struct{}{}
+	}
 	// Determine the ordinals that need to be released and the
 	// attributes that need to be cleaned up.
-	for _, ip := range addresses {
+	for ipStr, _ := range uniqueAddresses {
+		ip := cnet.MustParseIP(ipStr)
+
 		// Convert to an ordinal.
 		ordinal := ipToOrdinal(ip, *b)
 		if (ordinal < 0) || (ordinal > blockSize) {
--- a/lib/client/ipam_test.go
+++ b/lib/client/ipam_test.go
@@ -105,6 +105,68 @@ type testArgsClaimAff struct {
 
 var _ = testutils.E2eDatastoreDescribe("IPAM tests", testutils.DatastoreEtcdV2, func(config api.CalicoAPIConfig) {
 
+	FDescribe("IPAM ReleaseIPs with duplicate IP can result in assigning an IP multiple times", func() {
+		c := testutils.CreateCleanClient(config)
+		ic := setupIPAMClient(c, true)
+
+		// The IP which we will cause a multiple assignment of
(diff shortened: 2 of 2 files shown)


lib/client/ipam_test.go
```diff
@@ -105,6 +105,68 @@ type testArgsClaimAff struct {
...
+
+			It("Should allocate most of the pool", func() {
+				Expect(bulkAssignErr).NotTo(HaveOccurred())
+				Expect(ips).NotTo(ContainElement(assignedIP))
+				Expect(len(ips)).To(Equal(255))
+			})
+
+			_, releaseErr := ic.ReleaseIPs([]cnet.IP{
+				cnet.IP{assignedIP},
+				cnet.IP{assignedIP},
```
jsok: This is the essence of the issue - ReleaseIPs does not sanitize the IP list given nor does AllocationBlock's release() method.

lib/client/ipam_block.go
```diff
@@ -209,10 +210,16 @@ func (b *allocationBlock) release(addresses []cnet.IP) ([]cnet.IP, map[string]in
 	}
 
 	// Release requested addresses.
-	for _, ordinal := range ordinals {
+	uniqueOrdinals := set.FromArray(ordinals)
```
caseydavenport: If I understand this right, I think we need to de-duplicate the addresses a bit further up in this function.

Namely, the loop above where we iterate over the addresses does reference counting which won't be accurate in the case where duplicate addresses are provided, since the reference will be double counted in the loop above. 

Essentially this line: https://github.com/projectcalico/libcalico-go/pull/990/files#diff-1c8ffa16b869346e7e10b73ea36a9f42R180

Since `attrIdx` is tied directly to the ordinal being iterated over.

jsok: Yes good point, will dedupe the addresses as the first step.

jsok:
@caseydavenport took a bit to figure it out (de-duping a slice of structs ain't that straight forward!) but it's ready for review again now. Thanks for your patience.

lib/client/ipam_block.go
```diff
@@ -158,19 +159,31 @@ func (b *allocationBlock) release(addresses []cnet.IP) ([]cnet.IP, map[string]in
 
 	// Determine the ordinals that need to be released and the
 	// attributes that need to be cleaned up.
+	uniqueAddresses := set.New()
 	for _, ip := range addresses {
+		uniqueAddresses.Add(&ip)
```
caseydavenport: Hm, I'm not sure that using a pointer here will do the right thing.

For example, we could have two separate structures with the same value which wouldn't count as duplicates in this approach IIUC. 

I think we might need to de-dupe with a map, using a key which is a proxy for the value of the structure, which is really what we want to be unique.

e.g.

```
deduped := map[string]net.IP{}
for _, ip := range addresses {
    deduped[ip.String()] = ip
}

for _, ip := range deduped {
    // Existing logic
    ... 
}
```

WDYT?

jsok: Yeah I think you’re correct. I actually implemented a String() -> MustParseIP() version but thought it would be more efficient with pointers. Will push the suggested version tonight.

jsok:
@caseydavenport change pushed 👍

jsok:
@caseydavenport see #1013 for port to `master` branch.

PR #991, opened 2018-12-10
wip: add NodeSelector to IPPools and filter accordingly in ipam

briansan:
## Description
- adds the NodeSelector field to IPPools resource
- updates `ipamClient.determinePools` code to apply selector logic from IPPool NodeSelector on the Node labels

- needs feedback: 
```
node, err := c.client.Get(ctx, model.NodeKey{Hostname: host}, "")
```
doesn't seem to work with the following error:
```
2018-12-10 21:46:11.719 [DEBUG][81] etcdv3.go 306: Processing Get request model-etcdKey=Node(name=single-host) rev=""
2018-12-10 21:46:11.719 [ERROR][81] etcdv3.go 310: Unable to convert model.Key to an etcdv3 etcdKey model-etcdKey=Node(name=single-host) rev=""
2018-12-10 21:46:11.719 [ERROR][81] ipam.go 255: failed to get node for host error=Node is a composite type, so not handled with a single path host="single-host"
```

please advise @robbrockbank @caseydavenport

### DIFF
--- a/lib/apis/v3/ippool.go
+++ b/lib/apis/v3/ippool.go
@@ -53,6 +53,9 @@ type IPPoolSpec struct {
 	// The block size to use for IP address assignments from this pool. Defaults to 26 for IPv4 and 112 for IPv6.
 	BlockSize int `json:"blockSize,omitempty"`
 
+	// Allows IPPool to allocate for a specific node by label selector.
+	NodeSelector string `json:"nodeSelector,omitempty" validate:"omitempty,selector"`
+
 	// Deprecated: this field is only used for APIv1 backwards compatibility.
 	// Setting this field is not allowed, this field is for internal use only.
 	IPIP *apiv1.IPIPConfiguration `json:"ipip,omitempty" validate:"omitempty,mustBeNil"`
--- a/lib/clientv3/ippool_e2e_test.go
+++ b/lib/clientv3/ippool_e2e_test.go
@@ -831,8 +831,13 @@ var _ = testutils.E2eDatastoreDescribe("IPPool tests (etcd only)", testutils.Dat
 			}, options.SetOptions{})
 			Expect(err).NotTo(HaveOccurred())
 
+			// Create test node
+			host := "host-test"
+			_, err = c.Nodes().Create(ctx, &apiv3.Node{ObjectMeta: metav1.ObjectMeta{Name: host}}, options.SetOptions{})
+			Expect(err).NotTo(HaveOccurred())
+
 			// Allocate an IP so that a block is allocated
-			assigned, _, err := c.IPAM().AutoAssign(ctx, ipam.AutoAssignArgs{Num4: 1})
+			assigned, _, err := c.IPAM().AutoAssign(ctx, ipam.AutoAssignArgs{Num4: 1, Hostname: host})
 			Expect(err).NotTo(HaveOccurred())
 			Expect(assigned).To(HaveLen(1))
 
@@ -898,8 +903,12 @@ var _ = testutils.E2eDatastoreDescribe("IPPool tests (etcd only)", testutils.Dat
 			}, options.SetOptions{})
 			Expect(err).NotTo(HaveOccurred())
 
+			host := "host-test"
+			_, err = c.Nodes().Create(ctx, &apiv3.Node{ObjectMeta: metav1.ObjectMeta{Name: host}}, options.SetOptions{})
+			Expect(err).NotTo(HaveOccurred())
+
 			// Allocate an IP so that a block is allocated
-			assigned, _, err := c.IPAM().AutoAssign(ctx, ipam.AutoAssignArgs{Num4: 1})
+			assigned, _, err := c.IPAM().AutoAssign(ctx, ipam.AutoAssignArgs{Num4: 1, Hostname: host})
 			Expect(err).NotTo(HaveOccurred())
 			Expect(assigned).To(HaveLen(1))
 
@@ -915,7 +924,7 @@ var _ = testutils.E2eDatastoreDescribe("IPPool tests (etcd only)", testutils.Dat
 
 			// Allocate an IP so that a block is allocated
 			pool2 := []cnet.IPNet{cnet.MustParseCIDR(p2.Spec.CIDR)}
-			assigned, _, err = c.IPAM().AutoAssign(ctx, ipam.AutoAssignArgs{IPv4Pools: pool2, Num4: 1})
+			assigned, _, err = c.IPAM().AutoAssign(ctx, ipam.AutoAssignArgs{IPv4Pools: pool2, Num4: 1, Hostname: host})
 			Expect(err).NotTo(HaveOccurred())
 			Expect(assigned).To(HaveLen(1))
 
--- a/lib/clientv3/ippool_e2e_test.go
+++ b/lib/clientv3/ippool_e2e_test.go
@@ -47,10 +47,11 @@ var _ = testutils.E2eDatastoreDescribe("IPPool tests", testutils.DatastoreAll, f
 		BlockSize: 26,
 	}
 	spec1_2 := apiv3.IPPoolSpec{
-		CIDR:        "1.2.3.0/24",
-		NATOutgoing: true,
-		IPIPMode:    apiv3.IPIPModeNever,
-		BlockSize:   26,
+		CIDR:         "1.2.3.0/24",
+		NATOutgoing:  true,
+		IPIPMode:     apiv3.IPIPModeNever,
+		BlockSize:    26,
+		NodeSelector: `foo == "bar"`,
 	}
 	spec2 := apiv3.IPPoolSpec{
 		CIDR:        "2001::/120",
--- a/lib/ipam/ipam.go
+++ b/lib/ipam/ipam.go
@@ -27,6 +27,7 @@ import (
 	cerrors "github.com/projectcalico/libcalico-go/lib/errors"
 	"github.com/projectcalico/libcalico-go/lib/names"
 	"github.com/projectcalico/libcalico-go/lib/net"
+	"github.com/projectcalico/libcalico-go/lib/selector"
 )
 
 const (
@@ -205,7 +206,7 @@ func (c ipamClient) getBlockFromAffinity(ctx context.Context, aff *model.KVPair)
 // determinePools compares a list of requested pools with the enabled pools
 // and returns the intersect. If any requested pool does not exist, or is not enabled, an error is returned.
 // If no pools are requested, all enabled pools are returned.
-func (c ipamClient) determinePools(requestedPoolNets []net.IPNet, version int) ([]v3.IPPool, error) {
+func (c ipamClient) determinePools(requestedPoolNets []net.IPNet, version int, node *model.Node) ([]v3.IPPool, error) {
 	enabledPools, err := c.pools.GetEnabledPools(version)
 	if err != nil {
 		log.WithError(err).Errorf("Error reading configured pools")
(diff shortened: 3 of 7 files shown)


lib/apis/v3/ippool.go
```diff
@@ -53,6 +53,9 @@ type IPPoolSpec struct {
 	// The block size to use for IP address assignments from this pool. Defaults to 26 for IPv4 and 112 for IPv6.
 	BlockSize int `json:"blockSize,omitempty"`
 
+	// Allows IPPool to allocate for a specific node by label selector.
```
caseydavenport: We should update the clientv3 e2e tests to ensure setting / getting this field is done correctly.

lib/ipam/ipam.go
```diff
@@ -226,10 +227,19 @@ func (c ipamClient) determinePools(requestedPoolNets []net.IPNet, version int) (
 
 		// Make sure each requested pool exists
 		for _, rp := range requestedPoolNets {
-			if pool, ok := pm[rp.String()]; !ok {
+			pool, ok := pm[rp.String()]
+			if !ok {
 				// The requested pool doesn't exist.
 				return nil, fmt.Errorf("the given pool (%s) does not exist, or is not enabled", rp.IPNet.String())
-			} else {
+			} else if len(pool.Spec.NodeSelector) == 0 || node == nil {
```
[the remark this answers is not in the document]
briansan: unfortunately computers only understand hope as 011010000110111101110000011001010001010...

But in all seriousness, it seems like a pretty safe assumption to make that the ipam code wont execute before `calico/node.Startup` has run as kubectl shouldn't be trying to schedule pods on a node that doesn't have cni setup. However, I wonder if this is guaranteed to be the case 100% of the time which is where my skepticism comes in...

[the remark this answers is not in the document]
briansan: tl;dr i agree we should return an error just to be safe

lib/ipam/ipam.go
```diff
@@ -226,10 +227,22 @@ func (c ipamClient) determinePools(requestedPoolNets []net.IPNet, version int) (
...
-			if pool, ok := pm[rp.String()]; !ok {
+			pool, ok := pm[rp.String()]
+			if !ok {
 				// The requested pool doesn't exist.
 				return nil, fmt.Errorf("the given pool (%s) does not exist, or is not enabled", rp.IPNet.String())
-			} else {
+			} else if len(pool.Spec.NodeSelector) == 0 {
+				// No nodeSelector specified
+				enabledPools = append(enabledPools, pool)
+			} else if node, ok := nodeKV.Value.(*model.Node); !ok {
```
caseydavenport: I don't think this check here is useful.

We really have two cases:
- A nodeSelector is provided, in which case the node _needs_ to be passed.
- A nodeSelector is not provided, in which case we may want to allow the node to be nil (but not aware of a use-case for that). 

In practice, an IP allocation should always occur attached to a particular Calico node, so the node should never actually be nil. So, I think we can just drop this case, and return an error when we do a `Nodes().Get()` in the calling function.

lib/ipam/ipam.go
```diff
@@ -239,8 +252,15 @@ func (c ipamClient) determinePools(requestedPoolNets []net.IPNet, version int) (
 }
 
 func (c ipamClient) autoAssign(ctx context.Context, num int, handleID *string, attrs map[string]string, requestedPools []net.IPNet, version int, host string, maxNumBlocks int) ([]net.IP, error) {
+	// Retrieve node for given hostname to use for ip pool node selection
+	node, err := c.client.Get(ctx, model.ResourceKey{Kind: v3.KindNode, Name: host}, "")
+	if err != nil {
+		log.WithError(err).WithField("host", host).Warn("failed to get node for host")
```
caseydavenport: Should upgrade this log message to be error level

caseydavenport: Also, the IPAM code calls this "host" but everywhere else we refer to this as "node" (it's an historical thing that the IPAM code still calls this host).

So, let's just use the terminology "node" instead of "host" in all of the logging to be consistent.

lib/ipam/ipam_test.go
```diff
@@ -401,6 +427,34 @@ var _ = testutils.E2eDatastoreDescribe("IPAM tests", testutils.DatastoreEtcdV3,
...
+			applyPool(pool2.String(), true, "")
+
+			v4, _, outErr := ic.AutoAssign(context.Background(), AutoAssignArgs{
+				Num4:     100,
+				Num6:     0,
+				Hostname: host,
+			})
+			Expect(outErr).NotTo(HaveOccurred())
+
+			// Expect all the IPs to be in pool2.
```
caseydavenport: I think the comment is wrong

lib/ipam/ipam.go
caseydavenport: Might be good to just update this branch to return an error - essentially, it the IPPool has a node selector, then it is an error for the given nodeKV to be nil.

lib/ipam/ipam.go
```diff
@@ -226,10 +227,22 @@ func (c ipamClient) determinePools(requestedPoolNets []net.IPNet, version int) (
 
 		// Make sure each requested pool exists
 		for _, rp := range requestedPoolNets {
```
caseydavenport: So I think there's an issue here, in that this loop only gets executed if `requestedPoolNets > 0`.

Otherwise, we just use all enabled IP pools. I think we need to rework that in order to meet the desired behavior.

If no requestedPoolNets are provided, we still want to filter out only the pools which select this node.

lib/ipam/ipam_block_reader_writer_test.go
```diff
@@ -270,6 +275,7 @@ var _ = testutils.E2eDatastoreDescribe("IPAM affine block allocation tests", tes
 						defer GinkgoRecover()
 
 						testhost := "single-host"
+						applyNode(bc, testhost, nil)
```
caseydavenport: Same here - we're going to apply the node 4 times since we're inside a loop. Let's do this outside of the loop.

lib/ipam/ipam_test.go
```diff
@@ -401,6 +427,34 @@ var _ = testutils.E2eDatastoreDescribe("IPAM tests", testutils.DatastoreEtcdV3,
...
+		It("should favor node ip pool with nodeSelector foo=bar", func() {
+			host := "host"
+			pool1 := cnet.MustParseNetwork("10.0.0.0/24")
+			pool2 := cnet.MustParseNetwork("20.0.0.0/24")
+
+			bc.Clean()
+			deleteAllPools()
+
+			applyNode(bc, host, nil)
+			applyPool(pool1.String(), true, `foo == "bar"`)
```
caseydavenport: I suspect if you applied the selector to the other pool, then this test would start failing (because of the bug mentioned above w.r.t determining pools).

To test this, let's enhance this testcase to test allocations on two nodes - one which is matched by pool1, and one for pool2.

caseydavenport: (e.g., add selector `foo != bar`)

lib/clientv3/ippool_e2e_test.go
```diff
@@ -831,8 +831,13 @@ var _ = testutils.E2eDatastoreDescribe("IPPool tests (etcd only)", testutils.Dat
 			}, options.SetOptions{})
 			Expect(err).NotTo(HaveOccurred())
 
+			// Create test node
```
caseydavenport: See this test: https://github.com/projectcalico/libcalico-go/blob/82d36d69d1fc0c6317b261c58bdcb7948766723f/lib/clientv3/ippool_e2e_test.go#L167

I think we should enhance it so that `spec1_2` includes a nodeselector.

The test includes an assertion that you can write to datastore, read it back, and the values are the same.

briansan: addressed in 9fe3116

lib/ipam/ipam_test.go
briansan: addressed 9fe3116

lib/ipam/ipam_block_reader_writer_test.go
briansan: addressed 9fe3116

lib/ipam/ipam.go
briansan: addressed 9fe3116

lib/ipam/ipam_block_reader_writer_test.go
```diff
@@ -184,6 +186,9 @@ var _ = testutils.E2eDatastoreDescribe("IPAM affine block allocation tests", tes
 						defer GinkgoRecover()
 
 						testhost := fmt.Sprintf("host-%d", j)
+						applyNode(bc, testhost, nil)
```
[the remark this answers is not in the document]
briansan: The one below is definitely misplaced but I think this one is supposed to be in the loop. Reason being that we seem to executing the test against a different node on each iteration.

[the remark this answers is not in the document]
caseydavenport: yeah, you're right.

lib/ipam/ipam.go
```diff
@@ -205,42 +206,76 @@ func (c ipamClient) getBlockFromAffinity(ctx context.Context, aff *model.KVPair)
 // determinePools compares a list of requested pools with the enabled pools
 // and returns the intersect. If any requested pool does not exist, or is not enabled, an error is returned.
 // If no pools are requested, all enabled pools are returned.
-func (c ipamClient) determinePools(requestedPoolNets []net.IPNet, version int) ([]v3.IPPool, error) {
+func (c ipamClient) determinePools(requestedPoolNets []net.IPNet, version int, node *v3.Node) ([]v3.IPPool, error) {
+	// Nil check input node object.
```
caseydavenport: So, if you're going to assert this is never nil, then why make it a pointer in the first place?

Strictly speaking, this function only needs a node passed in if an enabled IP pool has a node selector.

lib/ipam/ipam.go
```diff
@@ -205,42 +206,76 @@ func (c ipamClient) getBlockFromAffinity(ctx context.Context, aff *model.KVPair)
...
+	// Retrieve node for given hostname to use for ip pool node selection
+	node, err := c.client.Get(ctx, model.ResourceKey{Kind: v3.KindNode, Name: host}, "")
+	if err != nil {
+		log.WithError(err).WithField("node", host).Error("failed to get node for host")
+		return nil, err
+	}
+
 	// Start by sanitizing the requestedPools.
-	pools, err := c.determinePools(requestedPools, version)
+	pools, err := c.determinePools(requestedPools, version, node.Value.(*v3.Node))
```
caseydavenport: We should probably do a proper type check for this so we don't panic if this fails.

lib/ipam/ipam.go
```diff
@@ -205,42 +206,76 @@ func (c ipamClient) getBlockFromAffinity(ctx context.Context, aff *model.KVPair)
 // determinePools compares a list of requested pools with the enabled pools
 // and returns the intersect. If any requested pool does not exist, or is not enabled, an error is returned.
 // If no pools are requested, all enabled pools are returned.
-func (c ipamClient) determinePools(requestedPoolNets []net.IPNet, version int) ([]v3.IPPool, error) {
+func (c ipamClient) determinePools(requestedPoolNets []net.IPNet, version int, node *v3.Node) ([]v3.IPPool, error) {
```
caseydavenport: I think this function has become complicated enough that it should have its own set of unit tests to assert correct behavior. A few I think we should have:

- Default case (returns all enabled pools, filters out disabled pools)
- Pass in a few requested pools, should do the right thing.
- Don't pass in requested pools, some enabled pools have nodeselectors which match the node, some don't
- Pass in requested pools, some requested pools have nodeselectors, some don't.
- some other cases, probably.

lib/ipam/ipam.go
briansan: I agree that the node _should_ only be passed in if an enabled IP pool has a node selector, but the fact that this method actually fetches the pool makes this a chicken/egg problem. Would prefer to provide the egg, even if we don't end up using it..

briansan: It looks like calico Nodes are created as a pointer to the Node object. Not sure how this is stored/retrieved by the backing datastore though...
https://github.com/projectcalico/node/blob/master/pkg/startup/startup.go#L345

caseydavenport:
Closing in favor of https://github.com/projectcalico/libcalico-go/pull/993

PR #992, opened 2018-12-13
Return CIDR instead of only IPs from calico IPAM.

kirandivekar:
This change is useful for rest of the apps to
know about the CIDR mask in addition to IP
address assigned

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


fasaxc:
Fixed the conflicts under #998

PR #994, opened 2018-12-20
Support status reporting across multiple OpenStack regions

neiljerram:
## Release Note

```release-note
To do
```

### DIFF
--- a/lib/backend/model/statusreports.go
+++ b/lib/backend/model/statusreports.go
@@ -67,7 +67,11 @@ type ActiveStatusReportListOptions struct {
 }
 
 func (options ActiveStatusReportListOptions) defaultPathRoot() string {
-	k := "/calico/felix/v2/" + options.RegionString + "/host"
+	k := "/calico/felix/v2/"
+	if options.RegionString == "" {
+		return k
+	}
+	k = k + options.RegionString + "/host"
 	if options.Hostname == "" {
 		return k
 	}
@@ -133,7 +137,11 @@ type LastStatusReportListOptions struct {
 }
 
 func (options LastStatusReportListOptions) defaultPathRoot() string {
-	k := "/calico/felix/v2/" + options.RegionString + "/host"
+	k := "/calico/felix/v2/"
+	if options.RegionString == "" {
+		return k
+	}
+	k = k + options.RegionString + "/host"
 	if options.Hostname == "" {
 		return k
 	}
--- a/lib/backend/model/statusreports.go
+++ b/lib/backend/model/statusreports.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"reflect"
 	"regexp"
+	"strings"
 
 	"github.com/projectcalico/libcalico-go/lib/errors"
 	log "github.com/sirupsen/logrus"
@@ -44,6 +45,9 @@ func (key ActiveStatusReportKey) defaultDeletePath() (string, error) {
 	}
 	if key.RegionString == "" {
 		return "", errors.ErrorInsufficientIdentifiers{Name: "regionString"}
+	}
+	if strings.Contains(key.RegionString, "/") {
+		return "", ErrorSlashInRegionString(key.RegionString)
 	}
 	e := fmt.Sprintf("/calico/felix/v2/%s/host/%s/status", key.RegionString, key.Hostname)
 	return e, nil
@@ -114,6 +118,9 @@ func (key LastStatusReportKey) defaultDeletePath() (string, error) {
 	}
 	if key.RegionString == "" {
 		return "", errors.ErrorInsufficientIdentifiers{Name: "regionString"}
+	}
+	if strings.Contains(key.RegionString, "/") {
+		return "", ErrorSlashInRegionString(key.RegionString)
 	}
 	e := fmt.Sprintf("/calico/felix/v2/%s/host/%s/last_reported_status", key.RegionString, key.Hostname)
 	return e, nil
--- a/lib/backend/model/statusreports.go
+++ b/lib/backend/model/statusreports.go
@@ -84,6 +84,10 @@ func (options ActiveStatusReportListOptions) KeyFromDefaultPath(ekey string) Key
 		log.WithError(err).Debugf("Bad region in path %s", ekey)
 		return nil
 	}
+	if options.Region != "" && region != options.Region {
+		log.Debugf("Didn't match region %s != %s", options.Region, region)
+		return nil
+	}
 	name := r[0][2]
 	if options.Hostname != "" && name != options.Hostname {
 		log.Debugf("Didn't match name %s != %s", options.Hostname, name)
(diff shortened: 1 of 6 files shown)


lib/backend/model/statusreports.go
```diff
@@ -59,10 +60,11 @@ func (key ActiveStatusReportKey) String() string {
 
 type ActiveStatusReportListOptions struct {
 	Hostname string
+	Region   string
 }
 
 func (options ActiveStatusReportListOptions) defaultPathRoot() string {
-	k := "/calico/felix/v1/host"
+	k := "/calico/felix/v2/" + regionString(options.Region) + "/host"
```
fasaxc: As discussed, code does what felix needs but it's breaking the normal convention in this API (where empty string means "don't filter on this field").

neiljerram: Fixed in next commit by changing the `Region` fields to `RegionString`.  A region _string_ is never validly empty.

lib/backend/model/statusreports.go
```diff
@@ -24,13 +24,14 @@ import (
...
-	matchLastStatusReport   = regexp.MustCompile("^/?calico/felix/v1/host/([^/]+)/last_reported_status")
+	matchActiveStatusReport = regexp.MustCompile("^/?calico/felix/v2/([^/]+)/host/([^/]+)/status$")
+	matchLastStatusReport   = regexp.MustCompile("^/?calico/felix/v2/([^/]+)/host/([^/]+)/last_reported_status")
 	typeStatusReport        = reflect.TypeOf(StatusReport{})
 )
 
 type ActiveStatusReportKey struct {
-	Hostname string `json:"-" validate:"required,hostname"`
+	Hostname     string `json:"-" validate:"required,hostname"`
+	RegionString string
```
fasaxc: Missing the json/validate tags?

lib/backend/model/statusreports.go
```diff
@@ -77,16 +82,22 @@ func (options ActiveStatusReportListOptions) KeyFromDefaultPath(ekey string) Key
...
 		return nil
 	}
-	return ActiveStatusReportKey{Hostname: name}
+	return ActiveStatusReportKey{Hostname: name, RegionString: regionString}
 }
 
 type LastStatusReportKey struct {
-	Hostname string `json:"-" validate:"required,hostname"`
+	Hostname     string `json:"-" validate:"required,hostname"`
+	RegionString string
```
fasaxc: Missing tags?

lib/backend/model/workloadendpointstatus.go
```diff
@@ -87,10 +93,11 @@ type WorkloadEndpointStatusListOptions struct {
 	OrchestratorID string
 	WorkloadID     string
 	EndpointID     string
+	RegionString   string
 }
 
 func (options WorkloadEndpointStatusListOptions) defaultPathRoot() string {
-	k := "/calico/felix/v1/host"
+	k := "/calico/felix/v2/" + options.RegionString + "/host"
```
fasaxc: I think if the region string is empty, this should return `"/calico/felix/v2/"`

lib/backend/model/statusreports.go
fasaxc: We should probably validate that it doesn't have a `/` in it if nothing else.

fasaxc: Just a thought: could you make RegionString into a typedef?  I.e. `type RegionString string`  Then the user would have to go through the RegionString function.

lib/backend/model/workloadendpointstatus.go
```diff
@@ -26,14 +26,15 @@ import (
...
-	matchWorkloadEndpointStatus = regexp.MustCompile("^/?calico/felix/v1/host/([^/]+)/workload/([^/]+)/([^/]+)/endpoint/([^/]+)$")
+	matchWorkloadEndpointStatus = regexp.MustCompile("^/?calico/felix/v2/([^/]+)/host/([^/]+)/workload/([^/]+)/([^/]+)/endpoint/([^/]+)$")
 )
 
 type WorkloadEndpointStatusKey struct {
 	Hostname       string `json:"-"`
 	OrchestratorID string `json:"-"`
 	WorkloadID     string `json:"-"`
 	EndpointID     string `json:"-"`
+	RegionString   string
```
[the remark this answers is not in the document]
neiljerram: I don't know.  I suspect it was just cargo culting, and thought that any real need should be revealed by tests, and so left it off the new field here.

(The json and validate tags are used for validating some v1 structures - https://github.com/projectcalico/libcalico-go/blob/master/lib/validator/v1/validator.go#L91 - but that never applies to Key and Options structures.)

lib/backend/model/statusreports.go
neiljerram: Yes, that's a good idea, and I could make the RegionString function do escaping.

lib/backend/model/statusreports.go
neiljerram: (As above)

lib/backend/model/statusreports.go
neiljerram: Actually no, that requires too much change everywhere, and is inconsistent which all the other string-like fields that are just 'string'.  Let's just do the validation.

lib/backend/model/workloadendpointstatus.go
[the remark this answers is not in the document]
fasaxc: > As a special case, if the field tag is "-", the field is always omitted. Note that a field with name "-" can still be generated using the tag "-,".

So, why are we omitting those fields from JSON?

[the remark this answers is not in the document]
fasaxc: https://golang.org/pkg/encoding/json/

lib/backend/model/statusreports.go
fasaxc: OK, can live with that; should there be a `validate` tag, then?

lib/backend/model/statusreports.go
fasaxc: I don't follow, shouldn't this have a validate tag even if the json tag isn't relevant?

lib/backend/model/statusreports.go
```diff
@@ -97,7 +116,13 @@ func (key LastStatusReportKey) defaultDeletePath() (string, error) {
 	if key.Hostname == "" {
 		return "", errors.ErrorInsufficientIdentifiers{Name: "hostname"}
 	}
-	e := fmt.Sprintf("/calico/felix/v1/host/%s/last_reported_status", key.Hostname)
+	if key.RegionString == "" {
+		return "", errors.ErrorInsufficientIdentifiers{Name: "regionString"}
+	}
+	if strings.Contains(key.RegionString, "/") {
```
fasaxc: Why implement this validation differently from every other key, which seems to use the validate tag?

PR #997, opened 2019-01-08
encapsulate node selector logic to ippool resource

briansan:
## Description
Implements `DoesMatchNode` method on `IPPool` resource to modularize the node selection logic.

## Todos
- [x] Tests
- [x] Documentation

### DIFF
--- a/lib/apis/v3/ippool.go
+++ b/lib/apis/v3/ippool.go
@@ -66,9 +66,9 @@ type IPPoolSpec struct {
 	NATOutgoingV1 bool `json:"nat-outgoing,omitempty" validate:"omitempty,mustBeFalse"`
 }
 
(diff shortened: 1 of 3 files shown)


lib/apis/v3/ippool.go
```diff
@@ -65,6 +66,22 @@ type IPPoolSpec struct {
 	NATOutgoingV1 bool `json:"nat-outgoing,omitempty" validate:"omitempty,mustBeFalse"`
 }
 
+// DoesMatchNode determines whether or not the IPPool's nodeSelector
+// matches the labels on the given node.
+func (pool IPPool) DoesMatchNode(n Node) (bool, error) {
```
caseydavenport: nit on naming: I'd call this "SelectsNode" rather than "MatchesNode", since "Selects" is the verb that selectors perform.

PR #999, opened 2019-01-09
Release affinities when IP pools no longer select the node

briansan:
- adds more context to the user when pod fails to create container due to ip pool exhaustion
- pulls out the code for releasing a block when the pool no longer matches the node into a separate function

### DIFF
--- a/lib/ipam/ipam.go
+++ b/lib/ipam/ipam.go
@@ -28,7 +28,6 @@ import (
 	"github.com/projectcalico/libcalico-go/lib/names"
 	"github.com/projectcalico/libcalico-go/lib/net"
 	cnet "github.com/projectcalico/libcalico-go/lib/net"
-	"github.com/projectcalico/libcalico-go/lib/selector"
 )
 
 const (
@@ -289,7 +288,7 @@ func (c ipamClient) autoAssign(ctx context.Context, num int, handleID *string, a
 
 	// If there are no pools, we cannot assign addresses.
 	if len(pools) == 0 {
-		return nil, errors.New("no configured Calico pools")
+		return nil, fmt.Errorf("no configured Calico pools for node %v", host)
 	}
 
 	// First, we try to assign addresses from one of the existing host-affine blocks.  We
@@ -665,6 +664,8 @@ func (c ipamClient) releaseIPsFromBlock(ctx context.Context, ips []net.IP, block
 	logCtx := log.WithField("cidr", blockCIDR)
 	for i := 0; i < ipamEtcdRetries; i++ {
 		logCtx.Info("Getting block so we can release IPs")
+
+		// Get allocation block for cidr.
 		obj, err := c.client.Get(ctx, model.BlockKey{CIDR: blockCIDR}, "")
 		if err != nil {
 			if _, ok := err.(cerrors.ErrorResourceDoesNotExist); ok {
@@ -676,67 +677,32 @@ func (c ipamClient) releaseIPsFromBlock(ctx context.Context, ips []net.IP, block
 			}
 		}
 
-		// Block exists - get the allocationBlock from the KVPair.
-		b := allocationBlock{obj.Value.(*model.AllocationBlock)}
+		// Determine whether or not the block's pool still matches the node.
+		released, err := c.ensureConsistentAffinity(ctx, obj.Value.(*model.AllocationBlock))
+		if err != nil {
+			if updateErr, ok := err.(cerrors.ErrorResourceUpdateConflict); ok {
+				logCtx.WithError(updateErr).Debug("Failed to ensure consistent affinity")
+				continue
+			}
+			logCtx.WithError(err).Error("Error ensuring consistent affinity")
+			return nil, err
+		}
 
-		// Retrieve node for this allocation. We do this so we can clean up affinity for blocks
-		// which should no longer be affine to this host.
-		host := getHostAffinity(b.AllocationBlock)
-		logCtx.Debugf("block affinity: %s", host)
-		if host != "" {
-			// Get the corresponding node object.
-			logCtx.Debugf("Looking up node for host affinity %s", host)
-			node, err := c.client.Get(ctx, model.ResourceKey{Kind: v3.KindNode, Name: host}, "")
+		// Releasing the affinity modifies the block, requery so the code below
+		// can release the IP addresses without an update conflict.
+		if released {
+			obj, err = c.client.Get(ctx, model.BlockKey{CIDR: blockCIDR}, "")
 			if err != nil {
-				if _, ok := err.(cerrors.ErrorResourceDoesNotExist); !ok {
-					logCtx.WithError(err).WithField("node", host).Error("failed to get node for host")
-					return nil, err
+				if _, ok := err.(cerrors.ErrorResourceDoesNotExist); ok {
+					// The block does not exist - all addresses must be unassigned.
+					return ips, nil
 				}
-				// We still want to be able to release IPs if the node doesn't exist.
-				logCtx.Info("Node doesn't exist, releasing IPs anyway")
-			} else {
-				// Make sure the returned value is a valid node.
-				v3n, ok := node.Value.(*v3.Node)
-				if !ok {
-					return nil, fmt.Errorf("Datastore returned malformed node object")
-				}
-
-				// If the IP pool which owns this block no longer selects this node,
-				// we should release the block's affinity to this node so it can be
-				// used elsewhere.
-				pool := c.blockReaderWriter.getPoolForIP(cnet.IP{blockCIDR.IP})
-				if pool == nil {
-					// No IP pool owns this block.
-				} else if sel, err := selector.Parse(pool.Spec.NodeSelector); err != nil {
-					// Invalid selector syntax.
-					logCtx.WithError(err).WithField("selector", pool.Spec.NodeSelector).Error("failed to parse selector")
-					return nil, err
-				} else if !sel.Evaluate(v3n.Labels) {
-					// Pool does not match this node's label, release this block's affinity.
-					if err := c.ReleaseAffinity(ctx, blockCIDR, host); err != nil {
-						return nil, err
-					}
-
-					// Since this updates the block, we need to requery it so the code below
-					// can release the IP addresses without an update conflict.
-					obj, err = c.client.Get(ctx, model.BlockKey{CIDR: blockCIDR}, "")
-					if err != nil {
-						if _, ok := err.(cerrors.ErrorResourceDoesNotExist); ok {
-							// The block does not exist - all addresses must be unassigned.
-							return ips, nil
-						} else {
-							// Unexpected error reading block.
-							return nil, err
-						}
-					}
-
-					// Block exists - get the allocationBlock from the KVPair.
-					b = allocationBlock{obj.Value.(*model.AllocationBlock)}
-				}
+				return nil, err
 			}
 		}
 
 		// Release the IPs.
+		b := allocationBlock{obj.Value.(*model.AllocationBlock)}
 		unallocated, handles, err2 := b.release(ips)
 		if err2 != nil {
 			return nil, err2
@@ -1162,6 +1128,31 @@ func (c ipamClient) releaseByHandle(ctx context.Context, handleID string, blockC
 				return err
 			}
 		}
+
+		// Determine whether or not the block's pool still matches the node.
+		released, err := c.ensureConsistentAffinity(ctx, obj.Value.(*model.AllocationBlock))
+		if err != nil {
+			if updateErr, ok := err.(cerrors.ErrorResourceUpdateConflict); ok {
+				logCtx.WithError(updateErr).Debug("Failed to ensure consistent affinity")
+				continue
+			}
+			logCtx.WithError(err).Error("Error ensuring consistent affinity")
+			return err
+		}
+
+		// Since this updates the block, we need to requery it so the code below
+		// can release the IP addresses without an update conflict.
+		if released {
+			obj, err = c.client.Get(ctx, model.BlockKey{CIDR: blockCIDR}, "")
+			if err != nil {
+				if _, ok := err.(cerrors.ErrorResourceDoesNotExist); ok {
+					// The block does not exist - all addresses must be unassigned.
+					return nil
+				}
+				return err
+			}
+		}
+
 		block := allocationBlock{obj.Value.(*model.AllocationBlock)}
 		num := block.releaseByHandle(handleID)
 		if num == 0 {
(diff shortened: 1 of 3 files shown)


lib/ipam/ipam.go
```diff
@@ -661,82 +667,21 @@ func (c ipamClient) ReleaseIPs(ctx context.Context, ips []net.IP) ([]net.IP, err
 	return unallocated, nil
 }
 
-func (c ipamClient) releaseIPsFromBlock(ctx context.Context, ips []net.IP, blockCIDR net.IPNet) ([]net.IP, error) {
+func (c ipamClient) releaseIPsFromBlock(ctx context.Context, ips []net.IP, blockCIDR net.IPNet, node v3.Node) ([]net.IP, error) {
 	logCtx := log.WithField("cidr", blockCIDR)
 	for i := 0; i < ipamEtcdRetries; i++ {
 		logCtx.Info("Getting block so we can release IPs")
-		obj, err := c.client.Get(ctx, model.BlockKey{CIDR: blockCIDR}, "")
+		obj, err := c.getBlockAndReleaseIfNotSelectsNode(ctx, node, blockCIDR)
```
caseydavenport: I'd rather not combine so much into one function here - let's keep the original Get() here and just encapsulate the "maybeReleaseAffinity()" part. WDYT?

lib/ipam/ipam.go
```diff
@@ -676,67 +677,27 @@ func (c ipamClient) releaseIPsFromBlock(ctx context.Context, ips []net.IP, block
 			}
 		}
 
-		// Block exists - get the allocationBlock from the KVPair.
-		b := allocationBlock{obj.Value.(*model.AllocationBlock)}
+		// Determine whether or not the block's pool still matches the node.
+		released, err := c.maybeReleaseAffinity(ctx, obj.Value.(*model.AllocationBlock))
+		if err != nil {
```
caseydavenport: I think we need to handle the resource update conflict error type here - if we hit a conflict, we should start at the beginning of the loop again.

lib/ipam/ipam.go
```diff
@@ -1383,6 +1344,62 @@ func (c ipamClient) convertBackendToIPAMConfig(cfg *model.IPAMConfig) *IPAMConfi
 	}
 }
 
+// maybeReleaseAffinity retrieves the pool and node for the given block and determines
+// if the pool still selects node. If it no longer matches, it will release the block
+// affinity for that node.
+// Returns a bool indicating if the block was released.
```
caseydavenport: block -> block affinity.

lib/ipam/ipam.go
```diff
@@ -1383,6 +1344,62 @@ func (c ipamClient) convertBackendToIPAMConfig(cfg *model.IPAMConfig) *IPAMConfi
...
+func (c ipamClient) maybeReleaseAffinity(ctx context.Context, b *model.AllocationBlock) (bool, error) {
```
caseydavenport: I know I suggested this name, but I think a more accurate name would be `ensureConsistentAffinity` or something like that. This function is meant to release affinity for a specific scenario, so I'd like to make that clear in the name.

lib/ipam/ipam.go
```diff
@@ -1383,6 +1344,62 @@ func (c ipamClient) convertBackendToIPAMConfig(cfg *model.IPAMConfig) *IPAMConfi
...
+	logCtx := log.WithField("cidr", b.CIDR)
+	// Retrieve node for this allocation. We do this so we can clean up affinity for blocks
```
caseydavenport: nit: leave a newline before comments

lib/ipam/ipam.go
```diff
@@ -1383,6 +1344,62 @@ func (c ipamClient) convertBackendToIPAMConfig(cfg *model.IPAMConfig) *IPAMConfi
...
+	// which should no longer be affine to this host.
+	host := getHostAffinity(b)
+	logCtx.Debugf("block affinity: %s", host)
+
+	// Exit function is host is empty.
```
caseydavenport: Should have a more detailed comment here if we're going to have one. As it is, the comment is pretty self-explanatory based on the following two lines of code, making it a bit redundant :)

lib/ipam/ipam.go
```diff
@@ -1383,6 +1344,62 @@ func (c ipamClient) convertBackendToIPAMConfig(cfg *model.IPAMConfig) *IPAMConfi
...
+// maybeReleaseAffinity retrieves the pool and node for the given block and determines
+// if the pool still selects node. If it no longer matches, it will release the block
+// affinity for that node.
+// Returns a bool indicating if the block was released.
+func (c ipamClient) maybeReleaseAffinity(ctx context.Context, b *model.AllocationBlock) (bool, error) {
+	logCtx := log.WithField("cidr", b.CIDR)
+	// Retrieve node for this allocation. We do this so we can clean up affinity for blocks
+	// which should no longer be affine to this host.
+	host := getHostAffinity(b)
+	logCtx.Debugf("block affinity: %s", host)
```
caseydavenport: Instead of logging this out, let's add it to the context logger.

It would be nice to have as context for other logs we emit later on.

lib/ipam/ipam.go
```diff
@@ -1383,6 +1344,62 @@ func (c ipamClient) convertBackendToIPAMConfig(cfg *model.IPAMConfig) *IPAMConfi
...
+	if !ok {
+		return false, fmt.Errorf("Datastore returned malformed node object")
+	}
+
+	if pool := c.blockReaderWriter.getPoolForIP(cnet.IP{b.CIDR.IPNet.IP}); pool == nil {
+		// No IP pool owns this block, return.
+		return false, nil
+	} else if sel, err := pool.SelectsNode(*v3n); err != nil {
+		// Invalid selector syntax.
+		logCtx.WithError(err).WithField("selector", pool.Spec.NodeSelector).Error("failed to parse selector")
```
caseydavenport: nit: we don't know for sure this is an invalid selector. Should say "error checking if pool selects node"

lib/ipam/ipam.go
```diff
@@ -1383,6 +1344,62 @@ func (c ipamClient) convertBackendToIPAMConfig(cfg *model.IPAMConfig) *IPAMConfi
...
+	}
+
+	// Make sure the returned value is a valid node.
+	v3n, ok := node.Value.(*v3.Node)
+	if !ok {
+		return false, fmt.Errorf("Datastore returned malformed node object")
+	}
+
+	if pool := c.blockReaderWriter.getPoolForIP(cnet.IP{b.CIDR.IPNet.IP}); pool == nil {
+		// No IP pool owns this block, return.
```
caseydavenport: Probably worth a debug log here.

lib/ipam/ipam.go
```diff
@@ -1383,6 +1344,62 @@ func (c ipamClient) convertBackendToIPAMConfig(cfg *model.IPAMConfig) *IPAMConfi
...
+func (c ipamClient) maybeReleaseAffinity(ctx context.Context, b *model.AllocationBlock) (bool, error) {
+	logCtx := log.WithField("cidr", b.CIDR)
+	// Retrieve node for this allocation. We do this so we can clean up affinity for blocks
+	// which should no longer be affine to this host.
+	host := getHostAffinity(b)
+	logCtx.Debugf("block affinity: %s", host)
+
+	// Exit function is host is empty.
+	if host == "" {
+		return false, nil
```
caseydavenport: Probably worth a debug log here - e.g. "Block already has no affinity"

lib/ipam/ipam.go
```diff
@@ -1383,6 +1344,62 @@ func (c ipamClient) convertBackendToIPAMConfig(cfg *model.IPAMConfig) *IPAMConfi
...
+	}
+
+	// If the IP pool which owns this block no longer selects this node,
+	// we should release the block's affinity to this node so it can be
+	// used elsewhere.
+	logCtx.Debugf("Looking up node for host affinity %s", host)
```
caseydavenport: I'd clarify we're looking up the node _labels_ here - it will help whoever is reading these logs to keep the context of what's going on.

lib/ipam/ipam.go
```diff
@@ -676,67 +677,27 @@ func (c ipamClient) releaseIPsFromBlock(ctx context.Context, ips []net.IP, block
...
 
-		// Retrieve node for this allocation. We do this so we can clean up affinity for blocks
-		// which should no longer be affine to this host.
-		host := getHostAffinity(b.AllocationBlock)
-		logCtx.Debugf("block affinity: %s", host)
-		if host != "" {
-			// Get the corresponding node object.
-			logCtx.Debugf("Looking up node for host affinity %s", host)
-			node, err := c.client.Get(ctx, model.ResourceKey{Kind: v3.KindNode, Name: host}, "")
+		// Since this updates the block, we need to requery it so the code below
```
caseydavenport: I'd tweak this comment slightly now to say "Releasing the affinity modifies the block, so we need to re-query it if it was released..."

lib/ipam/ipam.go
```diff
@@ -1383,6 +1344,62 @@ func (c ipamClient) convertBackendToIPAMConfig(cfg *model.IPAMConfig) *IPAMConfi
...
+
+	if pool := c.blockReaderWriter.getPoolForIP(cnet.IP{b.CIDR.IPNet.IP}); pool == nil {
+		// No IP pool owns this block, return.
+		return false, nil
+	} else if sel, err := pool.SelectsNode(*v3n); err != nil {
+		// Invalid selector syntax.
+		logCtx.WithError(err).WithField("selector", pool.Spec.NodeSelector).Error("failed to parse selector")
+		return false, err
+	} else if sel {
+		// Pool selects node, no change.
```
caseydavenport: Probably worth a debug log here too

lib/ipam/ipam.go
```diff
@@ -1383,6 +1349,69 @@ func (c ipamClient) convertBackendToIPAMConfig(cfg *model.IPAMConfig) *IPAMConfi
...
+	if !ok {
+		return false, fmt.Errorf("Datastore returned malformed node object")
+	}
+
+	// Fetch the pool for the given CIDR and check if it selects the node.
+	if pool := c.blockReaderWriter.getPoolForIP(cnet.IP{b.CIDR.IPNet.IP}); pool == nil {
+		logCtx.Debug("no pools own this block")
+		return false, nil
+	} else {
+		logCtx = logCtx.WithField("pool", pool)
```
caseydavenport: I don't think we want to log the ensure IP pool in the context logger. We can do it once as a debug log, or maybe just log out the CIDR and selector since that's really what we care about?

lib/ipam/ipam.go
```diff
@@ -1383,6 +1349,69 @@ func (c ipamClient) convertBackendToIPAMConfig(cfg *model.IPAMConfig) *IPAMConfi
...
+		if sel, err := pool.SelectsNode(*v3n); err != nil {
+			logCtx.WithError(err).Error("Failed to determine node selection")
+			return false, err
+		} else if sel {
+			logCtx.Debug("Pools selects node, no change")
+			return false, nil
+		} else {
```
caseydavenport: No need for an else block here just for a log

lib/ipam/ipam.go
caseydavenport: Probably just use the existing `logCtx` and write out something like this:

```
logCtx.Debugf("Block belongs with IPPool %s with nodeSelector %s", ...)
```

lib/ipam/ipam.go
```diff
@@ -1383,6 +1349,69 @@ func (c ipamClient) convertBackendToIPAMConfig(cfg *model.IPAMConfig) *IPAMConfi
...
+	v3n, ok := node.Value.(*v3.Node)
+	if !ok {
+		return false, fmt.Errorf("Datastore returned malformed node object")
+	}
+
+	// Fetch the pool for the given CIDR and check if it selects the node.
+	if pool := c.blockReaderWriter.getPoolForIP(cnet.IP{b.CIDR.IPNet.IP}); pool == nil {
+		logCtx.Debug("no pools own this block")
+		return false, nil
+	} else {
```
caseydavenport: I think this else is redundant - if we enter the fist if statement, we'll return, so we can simplify the logic here by removing the "else"

lib/ipam/ipam.go
```diff
@@ -1383,6 +1349,69 @@ func (c ipamClient) convertBackendToIPAMConfig(cfg *model.IPAMConfig) *IPAMConfi
...
+		logCtx = logCtx.WithField("pool", pool)
+		if sel, err := pool.SelectsNode(*v3n); err != nil {
+			logCtx.WithError(err).Error("Failed to determine node selection")
+			return false, err
+		} else if sel {
+			logCtx.Debug("Pools selects node, no change")
```
caseydavenport: "Pools selects" -> "Pool selects"

lib/ipam/ipam.go
```diff
@@ -1383,6 +1349,69 @@ func (c ipamClient) convertBackendToIPAMConfig(cfg *model.IPAMConfig) *IPAMConfi
...
+			return false, nil
+		} else {
+			logCtx.Debug("Pools no longer selects node, releasing block affinity")
```
caseydavenport: "Pools no longer" -> "Pool no longer"

lib/ipam/ipam.go
```diff
@@ -1383,6 +1349,69 @@ func (c ipamClient) convertBackendToIPAMConfig(cfg *model.IPAMConfig) *IPAMConfi
 	}
 }
 
+// ensureConsistentAffinity retrieves the pool and node for the given block and determines
+// if the pool still selects node. If it no longer matches, it will release the block
+// affinity for that node.
+// Returns a bool indicating if the block affinity was released.
+func (c ipamClient) ensureConsistentAffinity(ctx context.Context, b *model.AllocationBlock) (bool, error) {
```
caseydavenport: Potential optimization here would be to alter the signature like so:


```
func (c ipamClient) ensureConsistentAffinity(ctx context.Context, b *model.AllocationBlock) (*model.AllocationBlock, error) {
```

If we return a *model.AllocationBlock, we can return the input `b` when we haven't modified the block, and return the re-queried block if we need to (i.e., we've released the affinity). WDYT? Would remove the need for the calling code to do that.

lib/ipam/ipam.go
```diff
@@ -1383,6 +1349,69 @@ func (c ipamClient) convertBackendToIPAMConfig(cfg *model.IPAMConfig) *IPAMConfi
...
+		} else if sel {
+			logCtx.Debug("Pools selects node, no change")
+			return false, nil
+		} else {
+			logCtx.Debug("Pools no longer selects node, releasing block affinity")
+		}
+	}
+
+	// Pool does not match this node's label, release this block's affinity.
+	if err := c.ReleaseAffinity(ctx, b.CIDR, host); err != nil {
```
caseydavenport: I think we might be doing too much by calling `c.ReleaseAffinity`.

We could optimize this by copying this logic: https://github.com/projectcalico/libcalico-go/blob/061482258438a87d00299fd86c5780b445997fb7/lib/ipam/ipam.go#L898-L911

`c.ReleaseAffinity` some some stuff we don't need to - e.g., it re-queries the IP pools and checks that the CIDR is within a pool. It also checks the hostname, and implements retries, all of which we're already doing here.

lib/ipam/ipam.go
briansan: Yeah, logging out the entire IP pool is overkill. The CIDR context is already added at the beginning of the function so I'll just add the selector.

lib/ipam/ipam.go
briansan: Yeah, I was thinking of doing something like that. Not sure why I ultimately chose to do it this way...

briansan: Now I remember...The re-querying logic is no longer in this function as requested by https://github.com/projectcalico/libcalico-go/pull/999#discussion_r246602316. Should we bring the re-querying back in?

lib/ipam/ipam.go
briansan: Good call, but I'm not a big fan of copy/pasta when we can avoid it. It looks like this code is also duplicated here:

https://github.com/projectcalico/libcalico-go/blob/061482258438a87d00299fd86c5780b445997fb7/lib/ipam/ipam.go#L968

I'll try to break it out as its own function first and if that doesn't work, then I'll start boiling some water 😁

lib/ipam/ipam.go
caseydavenport: I think that comment was in reference to something slightly different, but I'm happy to apply the same logic here for now. Let's carry on for now.

lib/ipam/ipam.go
```diff
@@ -1383,6 +1343,91 @@ func (c ipamClient) convertBackendToIPAMConfig(cfg *model.IPAMConfig) *IPAMConfi
...
+		logCtx.Debug("No pools own this block")
+		return false, nil
+	} else if sel, err := pool.SelectsNode(*v3n); err != nil {
+		logCtx.WithField("selector", pool.Spec.NodeSelector).WithError(err).Error("Failed to determine node selection")
+		return false, err
+	} else if sel {
+		logCtx.Debug("Pool selects node, no change")
+		return false, nil
+	} else {
+		logCtx.WithField("selector", pool.Spec.NodeSelector).Debug("Pool no longer selects node, releasing block affinity")
```
caseydavenport: In this case I think moving this log statement outside of the else and putting it right above the call the `c.releaseBlockAffinity` would reduce the number of lines of code, and move the log statement closer to the thing it is actually describing. We can do that here since all of the other if / else if blocks have returns in them.

caseydavenport: bah, again it's cuz we're using the pool here. For some reason I can't see that ever.

lib/ipam/ipam.go
```diff
@@ -1162,6 +1098,30 @@ func (c ipamClient) releaseByHandle(ctx context.Context, handleID string, blockC
 				return err
 			}
 		}
+
+		// Determine whether or not the block's pool still matches the node.
+		released, err := c.ensureConsistentAffinity(ctx, obj.Value.(*model.AllocationBlock))
+		if err != nil {
+			if updateErr, ok := err.(cerrors.ErrorResourceUpdateConflict); ok {
+				logCtx.WithError(updateErr).Warningf("Failed to update block - retry #%d", i)
```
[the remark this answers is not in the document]
caseydavenport: I'd also log this probably belongs at Debug level - I think this is a mainline case that we'll expect to hit pretty regularly.

lib/ipam/ipam.go
```diff
@@ -1383,6 +1343,91 @@ func (c ipamClient) convertBackendToIPAMConfig(cfg *model.IPAMConfig) *IPAMConfi
...
+	// we should release the block's affinity to this node so it can be
+	// used elsewhere.
+	logCtx.Debugf("Looking up node labels for host affinity")
+	node, err := c.client.Get(ctx, model.ResourceKey{Kind: v3.KindNode, Name: host}, "")
+	if err != nil {
+		if _, ok := err.(cerrors.ErrorResourceDoesNotExist); !ok {
+			logCtx.WithError(err).WithField("node", host).Error("Failed to get node for host")
+			return false, err
+		}
+		logCtx.Info("Node doesn't exist, releasing IPs anyway")
```
caseydavenport: just spotted this, but we can't know for sure if this function will only be called in the "release IPs" context - we might want to do this elsewhere, so I might tweak this to say "Node doesn't exist, no need to release affinity".

Fine to do this in a follow-on PR. Actually, we shouldn't ever really hit this branch because if the node doesn't exist, we should have already released the affinity, I think. But, best to be safe!

PR #1002, opened 2019-01-15
IPAM: ensure that empty blocks affinities when ip pool deselects the node

briansan:
## Description
- adds a return value for `getAffineBlocks` to return allocation blocks that don't belong in the set of passed in ip pools
- adds a parameter to `releaseBlockAffinity` to indicate whether or not non-empty blocks should be released
- moves the call to `ensureConsistentAffinity` to occur after the IP has been released to fix regression issues due to previous change

## Todos
- [x] Tests
- [x] Documentation

### DIFF
--- a/lib/ipam/ipam.go
+++ b/lib/ipam/ipam.go
@@ -207,12 +207,13 @@ func (c ipamClient) getBlockFromAffinity(ctx context.Context, aff *model.KVPair)
 // If any requested pool does not exist, or is not enabled, an error is returned.
 // If no pools are requested, all enabled pools are returned.
 // Also applies selector logic on node labels to determine if the pool is a match.
-func (c ipamClient) determinePools(requestedPoolNets []net.IPNet, version int, node v3.Node) ([]v3.IPPool, error) {
+// Returns the set of matching pools as well as the full set of ip pools.
+func (c ipamClient) determinePools(requestedPoolNets []net.IPNet, version int, node v3.Node) (matchingPools, enabledPools []v3.IPPool, err error) {
 	// Get all the enabled IP pools from the datastore.
-	enabledPools, err := c.pools.GetEnabledPools(version)
+	enabledPools, err = c.pools.GetEnabledPools(version)
 	if err != nil {
 		log.WithError(err).Errorf("Error getting IP pools")
-		return nil, err
+		return
 	}
 	log.Debugf("enabled pools: %v", enabledPools)
 	log.Debugf("requested pools: %v", requestedPoolNets)
@@ -229,7 +230,8 @@ func (c ipamClient) determinePools(requestedPoolNets []net.IPNet, version int, n
 	for _, rp := range requestedPoolNets {
 		if pool, ok := pm[rp.String()]; !ok {
 			// The requested pool doesn't exist.
-			return nil, fmt.Errorf("the given pool (%s) does not exist, or is not enabled", rp.IPNet.String())
+			err = fmt.Errorf("the given pool (%s) does not exist, or is not enabled", rp.IPNet.String())
+			return
 		} else {
 			log.Debugf("Requested IP pool is ok to use: %s", pool.Name)
 			requestedPools = append(requestedPools, pool)
@@ -241,18 +243,19 @@ func (c ipamClient) determinePools(requestedPoolNets []net.IPNet, version int, n
 	// node selectors have not always existed.
 	if len(requestedPools) > 0 {
 		log.Debugf("Using the requested IP pools")
-		return requestedPools, nil
+		matchingPools = requestedPools
+		return
 	}
 
 	// At this point, we've determined the set of enabled IP pools which are valid for use.
 	// We only want to use IP pools which actually match this node, so do a filter based on
 	// selector.
-	matchingPools := []v3.IPPool{}
 	for _, pool := range enabledPools {
-		matches, err := pool.SelectsNode(node)
+		var matches bool
+		matches, err = pool.SelectsNode(node)
 		if err != nil {
 			log.WithError(err).WithField("pool", pool).Error("failed to determine if node matches pool")
-			return nil, err
+			return
 		}
 		if !matches {
 			// Do not consider pool enabled if the nodeSelector doesn't match the node's labels.
@@ -263,7 +266,7 @@ func (c ipamClient) determinePools(requestedPoolNets []net.IPNet, version int, n
 		matchingPools = append(matchingPools, pool)
 	}
 
-	return matchingPools, nil
+	return
 }
 
 func (c ipamClient) autoAssign(ctx context.Context, num int, handleID *string, attrs map[string]string, requestedPools []net.IPNet, version int, host string, maxNumBlocks int) ([]net.IPNet, error) {
@@ -281,7 +284,7 @@ func (c ipamClient) autoAssign(ctx context.Context, num int, handleID *string, a
 	}
 
 	// Determine the correct set of IP pools to use for this request.
-	pools, err := c.determinePools(requestedPools, version, *v3n)
+	pools, allPools, err := c.determinePools(requestedPools, version, *v3n)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


lib/ipam/ipam.go
```diff
@@ -299,10 +299,25 @@ func (c ipamClient) autoAssign(ctx context.Context, num int, handleID *string, a
 		logCtx = logCtx.WithField("handle", *handleID)
 	}
 	logCtx.Info("Looking up existing affinities for host")
-	affBlocks, err := c.blockReaderWriter.getAffineBlocks(ctx, host, version, pools)
+	affBlocks, affBlocksToRelease, err := c.blockReaderWriter.getAffineBlocks(ctx, host, version, pools)
```
caseydavenport: I think we need to check the IP pool here to make sure the `affBlockToRelease` isn't in a pool that selects this node. 

Take this example:
- There are two pools which select this node - `pool1`, `pool2`
- There are two pools which do _not_ select this node - `pool3`, `pool4`
- I pass in an explicit pool which doesn't select this node - `pool3`.

We don't want to release any blocks that are affine to this node that are within `pool1` or `pool2`. But, we do want to release any empty blocks within `pool4`.

caseydavenport: Might need to augment `determinePools` or something to return both the pools to use for the allocation, as well as all pools, so that we can match against other pools which may not have been provided on this particular request.

lib/ipam/ipam.go
```diff
@@ -744,6 +735,16 @@ func (c ipamClient) releaseIPsFromBlock(ctx context.Context, ips []net.IP, block
 		for handleID, amount := range handles {
 			c.decrementHandle(ctx, handleID, blockCIDR, amount)
 		}
+
+		// Determine whether or not the block's pool still matches the node.
+		if _, err := c.ensureConsistentAffinity(ctx, obj.Value.(*model.AllocationBlock)); err != nil {
+			if updateErr, ok := err.(cerrors.ErrorResourceUpdateConflict); ok {
```
caseydavenport: Hm, I think this is probably not going to work the way we expect any more now that it is moved.

Let's say we hit a conflict error - when we retry above, the block will have already released the IPs and so we'll probably hit line 701 and return without reaching this code a second time.

Similarly, I think if we hit some other error, we'll return an error to the caller indicating we couldn't release the IPs when in fact we _did_ actually release the IPs.

We probably need to rethink what the right thing to do on failure is. WDYT?

lib/ipam/ipam_block_reader_writer.go
```diff
@@ -284,6 +292,11 @@ func (rw blockReaderWriter) releaseBlockAffinity(ctx context.Context, host strin
 		return errBlockClaimConflict{Block: b}
 	}
 
+	// Don't release block affinity if we require it to be empty and it's not empty.
+	if requireEmpty && !b.empty() {
+		return nil
```
caseydavenport: Let's add a debug log here so we know when we're hitting this branch.

lib/ipam/ipam_test.go
```diff
@@ -678,6 +678,88 @@ var _ = testutils.E2eDatastoreDescribe("IPAM tests", testutils.DatastoreEtcdV3,
...
+			Expect(err).NotTo(HaveOccurred())
+
+			// Change the selector for the IP pool so that it no longer matches node1.
+			applyPool(pool1.String(), true, `foo != "bar"`)
+
+			// The block should still have an affinity to this host.
+			Expect(len(getAffineBlocks(bc, host))).To(Equal(1))
+
+			// The block should still have affinity.
+			opts := model.BlockListOptions{IPVersion: 4}
```
caseydavenport: This bit of code actually isn't doing what the comment says - it's querying blocks rather than affinities, and asserting that a single block exists but not doing any assertion on the affinity.

lib/ipam/ipam_block_reader_writer.go
briansan: addressed in a808da3

lib/ipam/ipam_test.go
briansan: addressed in a808da3

lib/ipam/ipam.go
```diff
@@ -299,10 +301,43 @@ func (c ipamClient) autoAssign(ctx context.Context, num int, handleID *string, a
...
-	affBlocks, err := c.blockReaderWriter.getAffineBlocks(ctx, host, version, pools)
+	affBlocks, affBlocksToRelease, err := c.blockReaderWriter.getAffineBlocks(ctx, host, version, pools)
 	if err != nil {
 		return nil, err
 	}
+
+	// Release any blocks still affine to this host but no longer selected by the larger ip pool.
+	for _, block := range affBlocksToRelease {
+		blockSelectsNode := false
+		for _, pool := range allPools {
```
caseydavenport: What do you think about making use of this helper function here: https://github.com/projectcalico/libcalico-go/blob/26206752595de5a06705389a8302de48219016dc/lib/ipam/ipam_block_reader_writer.go#L352:29

It would want to be modified a bit so you can pass in the list of pools, to prevent having to re-query for each, but I think it would help us commonize this logic a bit.

i.e.,

```
p := c.blockReaderWriter.getPoolForIP(allPools, block)
if p != nil && !p.SelectsNode(node) {
    // Release the block affinity
}
```

lib/ipam/ipam_test.go
```diff
@@ -678,6 +678,158 @@ var _ = testutils.E2eDatastoreDescribe("IPAM tests", testutils.DatastoreEtcdV3,
...
+			out, err = bc.List(context.Background(), opts, "")
+			Expect(err).NotTo(HaveOccurred())
+			Expect(len(out.KVPairs)).To(Equal(1))
+
+		})
+
+		// Create one ip pool, call AutoAssign, call ReleaseIPs,
+		// create another ip pool, call AutoAssign explicitly passing the second pool,
+		// ensure that the block affinity from the first ip pool is not released.
+		It("should not release blocks when the ips within are released but still selects the node while a different pool is explicitly requested", func() {
```
caseydavenport: Nice.

lib/ipam/ipam.go
```diff
@@ -1378,7 +1385,7 @@ func (c ipamClient) convertBackendToIPAMConfig(cfg *model.IPAMConfig) *IPAMConfi
 // if the pool still selects node. If it no longer matches, it will release the block
 // affinity for that node.
 // Returns a bool indicating if the block affinity was released.
-func (c ipamClient) ensureConsistentAffinity(ctx context.Context, b *model.AllocationBlock) (bool, error) {
+func (c ipamClient) ensureConsistentAffinity(ctx context.Context, b *model.AllocationBlock) error {
```
caseydavenport: Good spot - this is a nice cleanup.

lib/ipam/ipam.go
```diff
@@ -207,12 +207,12 @@ func (c ipamClient) getBlockFromAffinity(ctx context.Context, aff *model.KVPair)
 // If any requested pool does not exist, or is not enabled, an error is returned.
 // If no pools are requested, all enabled pools are returned.
 // Also applies selector logic on node labels to determine if the pool is a match.
```
caseydavenport: We should update the comment here to indicate it also returns the full set of enabled IP pools

lib/ipam/ipam.go
```diff
@@ -299,10 +301,44 @@ func (c ipamClient) autoAssign(ctx context.Context, num int, handleID *string, a
 		logCtx = logCtx.WithField("handle", *handleID)
 	}
 	logCtx.Info("Looking up existing affinities for host")
-	affBlocks, err := c.blockReaderWriter.getAffineBlocks(ctx, host, version, pools)
+	affBlocks, affBlocksToRelease, err := c.blockReaderWriter.getAffineBlocks(ctx, host, version, pools)
 	if err != nil {
 		return nil, err
 	}
+
+	// Release any blocks still affine to this host but no longer selected by the larger ip pool.
```
caseydavenport: I think we should also mention that we only do this for blocks which have no IPs allocated, since that's an important qualification here.

caseydavenport: Also, "selected by the larger IP pool" is probably more accurately stated as "no longer part of an IP pool which selects this node".

lib/ipam/ipam.go
```diff
@@ -299,10 +301,44 @@ func (c ipamClient) autoAssign(ctx context.Context, num int, handleID *string, a
...
+		if err != nil {
+			logCtx.WithError(err).WithField("pool", pool).Error("Failed to determine if node matches pool, skipping")
+			continue
+		}
+		if blockSelectsNode {
+			logCtx.Debug("Block's pool still selects node, refusing to remove affinity")
+			continue
+		}
+
+		// If we made it this far, relase the block affinity.
```
caseydavenport: "release the block affinity, requiring it to be empty"

lib/ipam/ipam.go
```diff
@@ -299,10 +301,44 @@ func (c ipamClient) autoAssign(ctx context.Context, num int, handleID *string, a
...
+		if err = c.blockReaderWriter.releaseBlockAffinity(ctx, host, block, true); err != nil {
+			if _, ok := err.(errBlockClaimConflict); ok {
+				// Not claimed by this host - ignore.
+			} else if _, ok := err.(cerrors.ErrorResourceDoesNotExist); ok {
+				// Block does not exist - ignore.
+			} else {
+				return nil, err
```
caseydavenport: Just had a thought, but I wonder if we need to retry here on conflict errors, or if it's OK to simply ignore errors.

Think of this scenario:
- We list all the blocks we think we should release
- We go to release a block, but in the meantime someone else has allocated an IP address from that block.
- We fail with an update conflict error

In that scenario, we no longer want to release the block since it isn't empty any more. If we retry the `releaseBlockAffinity`, it will query the node, determine that it isn't empty, and carry-on as expected. The current logic won't retry, and instead will return an error, failing the allocation. WDYT?

lib/ipam/ipam.go
```diff
@@ -1198,8 +1192,17 @@ func (c ipamClient) releaseByHandle(ctx context.Context, handleID string, blockC
 			}
 			logCtx.Info("Successfully released IPs from block")
 		}
-
 		c.decrementHandle(ctx, handleID, blockCIDR, num)
+
+		// Determine whether or not the block's pool still matches the node.
+		if err = c.ensureConsistentAffinity(ctx, block.AllocationBlock); err != nil {
+			if updateErr, ok := err.(cerrors.ErrorResourceUpdateConflict); ok {
```
caseydavenport: I think we want to do the same here as we did in `ReleaseIP`

PR #1003, opened 2019-01-16
Let Calico Nodes inherit labels from Kubernetes Nodes

fasaxc:
## Description

- Copy Kubernetes labels into the Calico Node object.
- Keep track of the Kubernetes labels and old Calico labels for round-trip purposes.

## Todos
- [x] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/backend/k8s/resources/node.go
+++ b/lib/backend/k8s/resources/node.go
@@ -16,6 +16,7 @@ package resources
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 
@@ -34,10 +35,12 @@ import (
 )
 
 const (
-	nodeBgpIpv4AddrAnnotation = "projectcalico.org/IPv4Address"
-	nodeBgpIpv6AddrAnnotation = "projectcalico.org/IPv6Address"
-	nodeBgpAsnAnnotation      = "projectcalico.org/ASNumber"
-	nodeBgpCIDAnnotation      = "projectcalico.org/RouteReflectorClusterID"
+	nodeBgpIpv4AddrAnnotation   = "projectcalico.org/IPv4Address"
+	nodeBgpIpv6AddrAnnotation   = "projectcalico.org/IPv6Address"
+	nodeBgpAsnAnnotation        = "projectcalico.org/ASNumber"
+	nodeBgpCIDAnnotation        = "projectcalico.org/RouteReflectorClusterID"
+	nodeK8sLabelAnnotation      = "projectcalico.org/kube-labels"
+	nodeShadowedLabelAnnotation = "projectcalico.org/shadowed-labels"
 )
 
 func NewNodeClient(c *kubernetes.Clientset) K8sResourceClient {
@@ -196,6 +199,13 @@ func K8sNodeToCalico(k8sNode *kapiv1.Node) (*model.KVPair, error) {
 	calicoNode.ObjectMeta.Name = k8sNode.Name
 	SetCalicoMetadataFromK8sAnnotations(calicoNode, k8sNode)
 
+	// Calico Nodes inherit labels from Kubernetes nodes, do that merge.
+	err := mergeCalicoAndK8sLabels(calicoNode, k8sNode)
+	if err != nil {
+		log.WithError(err).Error("Failed to merge Calico and Kubernetes labels.")
+		return nil, err
+	}
+
 	// Extract the BGP configuration stored in the annotations.
 	bgpSpec := &apiv3.NodeBGPSpec{}
 	annotations := k8sNode.ObjectMeta.Annotations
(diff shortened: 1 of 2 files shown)


lib/backend/k8s/resources/node.go
```diff
@@ -282,6 +300,113 @@ func mergeCalicoNodeIntoK8sNode(calicoNode *apiv3.Node, k8sNode *kapiv1.Node) (*
...
+	// Now, copy the Kubernetes Node labels over.  Note: this may overwrite Calico labels of the same name, but that's
+	// consistent with the kube-controllers behavior.
+	var shadowedLabels map[string]string
+	for k, v := range k8sNode.Labels {
+		if calicoNode.Labels == nil {
+			calicoNode.Labels = map[string]string{}
+		}
+		if calValue, ok := calicoNode.Labels[k]; ok {
+			if shadowedLabels == nil {
+				shadowedLabels = map[string]string{}
```
briansan: why not just initialize `shadowedLabels` on L311?

lib/backend/k8s/resources/node.go
```diff
@@ -282,6 +300,113 @@ func mergeCalicoNodeIntoK8sNode(calicoNode *apiv3.Node, k8sNode *kapiv1.Node) (*
...
+// calicoNode (which were loaded from our annotation).  Kubernetes labels take precedence.  To make the operation
+// reversible (so that we can support write back of a Calico node that was read from Kubernetes), we also store the
+// complete set of Kubernetes labels in an annotation along with any Calico labels that were shadowed by Kubernetes
+// labels.
+func mergeCalicoAndK8sLabels(calicoNode *apiv3.Node, k8sNode *kapiv1.Node) error {
+	// Now, copy the Kubernetes Node labels over.  Note: this may overwrite Calico labels of the same name, but that's
+	// consistent with the kube-controllers behavior.
+	var shadowedLabels map[string]string
+	for k, v := range k8sNode.Labels {
+		if calicoNode.Labels == nil {
```
briansan: nit: could we do this lazy instantiation outside of the for loop?

lib/backend/k8s/resources/node.go
```diff
@@ -282,6 +300,113 @@ func mergeCalicoNodeIntoK8sNode(calicoNode *apiv3.Node, k8sNode *kapiv1.Node) (*
...
+		if rawShadowedLabels != "" {
+			if err := json.Unmarshal([]byte(rawShadowedLabels), &shadowedLabels); err != nil {
+				log.WithError(err).Error("Failed to unmarshal Calico labels from " +
+					nodeShadowedLabelAnnotation + " annotation")
+				return nil, err
+			}
+		}
+
+		// Now remove any labels that match the k8s ones.
+		if log.GetLevel() >= log.DebugLevel {
```
briansan: is this if statement necessary?

briansan:
This solution looks good for resolving the Calico Node labels with the Kubernetes Node labels via annotations, but I was wondering how any changes within the labels are going to be persisted? It seems like these conversions are done in memory?

lib/backend/k8s/resources/node.go
```diff
@@ -282,6 +300,113 @@ func mergeCalicoNodeIntoK8sNode(calicoNode *apiv3.Node, k8sNode *kapiv1.Node) (*
...
+		calicoNode.Annotations[nodeShadowedLabelAnnotation] = string(bytes)
+	}
+	return nil
+}
+
+// restoreCalicoLabels tries to undo the transformation done by mergeCalicoLabels.  It returns a copy of the node with
+// Kubernetes labels removed from the Labels field and restores any shadowed Calico labels.  It also filters the
+// bookkeeping annotations out of the copy.
+func restoreCalicoLabels(calicoNode *apiv3.Node) (*apiv3.Node, error) {
+	calicoNode = calicoNode.DeepCopy()
```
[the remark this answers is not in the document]
fasaxc: That would do a shallow copy of the Node, not a deep copy.  A shallow copy copies all the fields of the struct by value.  For simple types such as ints that works fine but for reference types (such as maps) it would only copy the reference, so we'd end up sharing the labels between the original and copied object.

In an earlier iteration of the code, I did use a shallow copy and then I manually copied the labels when needed; however, it added complexity and I found that I wrote more bugs than Id like in such a short piece of code so I switched to a  deep copy operation.  It probably does perform slightly worse but it's a lot simpler.

lib/backend/k8s/resources/node.go
fasaxc: No, if you see that pattern, it's a performance optimisation.  WithFields can be expensive if it's done in a loop (it has to copy the log entry and create a new fields dict).  Probably borderline here but this function could be used in a migration or a list operation for 10k nodes with 10 labels each so I thought it might be worthwhile (and it's trivial to do).

lib/backend/k8s/resources/node.go
```diff
@@ -282,6 +300,113 @@ func mergeCalicoNodeIntoK8sNode(calicoNode *apiv3.Node, k8sNode *kapiv1.Node) (*
...
+
+		// Now remove any labels that match the k8s ones.
+		if log.GetLevel() >= log.DebugLevel {
+			log.WithFields(log.Fields{"k8s": k8sLabels, "shadowed": shadowedLabels}).Debug("Loaded label annotations")
+		}
+		for k, k8sVal := range k8sLabels {
+			if calVal, ok := calicoNode.Labels[k]; ok && calVal == k8sVal {
+				// Value is still set and it matches the value from k8s.
+				if shadowedVal, ok := shadowedLabels[k]; ok {
+					// The k8s value shadowed an old Calico label, restore the Calico value.
```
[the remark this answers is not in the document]
fasaxc: You're right.  Let's be consistent (don't know why I had in mind that kube-controllers wouldn't do that).

fasaxc:
> This solution looks good for resolving the Calico Node labels with the Kubernetes Node labels via annotations, but I was wondering how any changes within the labels are going to be persisted? It seems like these conversions are done in memory?

@briansan the Calico Node resource is backed by the k8s Node resource.  So, if you do a get for a Calico Node, it actually gets the k8s Node and then runs this logic on it to copy the k8s data tos a synthesised Calico Node.  When you do a write to a Calico Node object, it does the reverse, copying the Calico data to the k8s Node and then it does the write to the k8s Node.

caseydavenport:
Once merged could you open a cherry-pick onto v3.5?

PR #1009, opened 2019-01-22
Add OpenstackRegion to FelixConfiguration

neiljerram:


fasaxc:
Looks like deep-copy doesn't change for adding a new string field.

PR #816, opened 2018-03-09
[Prototype] Add support for Calico IPAM to Kubernetes datastore

caseydavenport:
## Description

Something I was hacking on to see how difficult it would be. Still has a ways to go.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

PR #1010, opened 2019-02-04
Support for Calico IPAM in KDD mode

caseydavenport:
## Description

Updated version of https://github.com/projectcalico/libcalico-go/pull/816 which brings the branch up to date with master and gets more tests passing. 

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Calico IPAM can now be used with the Kubernetes API datastore
```


lib/backend/k8s/resources/affinityblock.go
```diff
@@ -1,174 +0,0 @@
-// Copyright (c) 2016-2017 Tigera, Inc. All rights reserved.
```
caseydavenport: We can't just delete this outright if we still want to support `host-local` IPAM as well.

We will need some way to trigger the code to use this representation of affinity for host-local, and the datastore-driven one when using calico-ipam.

lib/backend/k8s/resources/client.go
```diff
@@ -45,7 +46,7 @@ type K8sResourceClient interface {
 	// Delete removes the object specified by the KVPair.  If the KVPair
 	// contains revision information, the delete only succeeds if the
 	// revision is still current.
-	Delete(ctx context.Context, key model.Key, revision string) (*model.KVPair, error)
+	Delete(ctx context.Context, key model.Key, revision string, uid *types.UID) (*model.KVPair, error)
```
caseydavenport: It would be nice to remove the need for passing the `uid` explicitly here. It should probably be passed as part of the key, and respected on get/list/update/delete.

lib/backend/k8s/resources/ipam_affinity.go
```diff
@@ -0,0 +1,206 @@
...
+		},
+		Value: &apiv3.BlockAffinity{
+			TypeMeta: metav1.TypeMeta{
+				Kind:       apiv3.KindBlockAffinity,
+				APIVersion: "crd.projectcalico.org/v1",
+			},
+			ObjectMeta: metav1.ObjectMeta{
+				Name:            name,
+				ResourceVersion: kvpv1.Revision,
+				Annotations: map[string]string{
```
caseydavenport: Not sure why we're storing this in annotations. Guess it's not really a problem, but we could store in the spec if we wanted.

lib/backend/k8s/resources/ipam_affinity.go
```diff
@@ -0,0 +1,206 @@
...
+		Revision: kvpv3.Revision,
+	}
+}
+
+func v3Fields(k model.Key) (name, cidr, host string) {
+	host = k.(model.BlockAffinityKey).Host
+	cidr = fmt.Sprintf("%s", k.(model.BlockAffinityKey).CIDR)
+	h := sha1.New()
+	h.Write([]byte(fmt.Sprintf("%s+%s", host, cidr)))
+	name = fmt.Sprintf("%s-%sc", host, hex.EncodeToString(h.Sum(nil))[:11])
```
caseydavenport: We need to agree a design for the name calculation - this is a hack.

lib/backend/k8s/resources/ipam_affinity.go
```diff
@@ -0,0 +1,206 @@
...
+type affinityBlockClient struct {
+	rc customK8sResourceClient
+}
+
+func toV1(kvpv3 *model.KVPair) *model.KVPair {
+	cidrStr := kvpv3.Value.(*apiv3.BlockAffinity).Annotations["projectcalico.org/cidr"]
+	host := kvpv3.Value.(*apiv3.BlockAffinity).Annotations["projectcalico.org/host"]
+	_, cidr, err := net.ParseCIDR(cidrStr)
+	if err != nil {
+		panic(err)
```
caseydavenport: Need to do better than panic here. We're missing the right error handling paths in this code.

lib/backend/k8s/resources/ipam_affinity.go
```diff
@@ -0,0 +1,206 @@
...
+				State: string(state),
+			},
+		},
+		Revision: kvpv1.Revision,
+	}
+}
+
+func (c *affinityBlockClient) Create(ctx context.Context, kvp *model.KVPair) (*model.KVPair, error) {
+	nkvp := toV3(kvp)
+	//nkvp.Value.(Resource).GetObjectMeta().SetFinalizers([]string{"ipam.projectcalico.org"})
```
caseydavenport: Commented code to be removed.

lib/backend/k8s/resources/ipam_affinity.go
```diff
@@ -0,0 +1,206 @@
...
+			},
+			Spec: apiv3.BlockAffinitySpec{
+				State: string(state),
+			},
+		},
+		Revision: kvpv1.Revision,
+	}
+}
+
+func (c *affinityBlockClient) Create(ctx context.Context, kvp *model.KVPair) (*model.KVPair, error) {
```
caseydavenport: This `affinityBlockClieny` exists solely to do v1<->v3 and conversion, since the CRDs are in "v3" format and the IPAM code uses models.

We should clean this up, either by:
- Updating the IPAM code to use v3 style resources, and adjusting etcdv3 to to the conversion.
- Enhancing `customK8sResourceClient` to do the conversion for us, removing the need for so much boilerplate. 
- ???

lib/backend/k8s/resources/ipam_block.go
```diff
@@ -0,0 +1,246 @@
...
+// Implements the api.Client interface for IPAMBlocks.
+type ipamBlockClient struct {
+	rc customK8sResourceClient
+}
+
+func (c ipamBlockClient) toV1(kvpv3 *model.KVPair) *model.KVPair {
+	cidrStr := kvpv3.Value.(*apiv3.IPAMBlock).Annotations["projectcalico.org/cidr"]
+	_, cidr, err := net.ParseCIDR(cidrStr)
+	if err != nil {
+		panic(err)
```
caseydavenport: More panic.

lib/backend/k8s/resources/ipam_block.go
```diff
@@ -0,0 +1,246 @@
...
+			Attributes:     attrs,
+			Deleting:       ab.Spec.Deleting,
+		},
+		Revision: kvpv3.Revision,
+	}
+}
+
+func (c ipamBlockClient) v3Fields(k model.Key) (name, cidr string) {
+	cidr = fmt.Sprintf("%s", k.(model.BlockKey).CIDR)
+	name = strings.Replace(cidr, ".", "-", -1)
```
caseydavenport: Also here - need to agree the name calculation. This is a hack.

lib/backend/k8s/resources/ipam_config.go
```diff
@@ -0,0 +1,97 @@
...
+	return &ipamConfigClient{}
+}
+
+// Implements the api.Client interface for AffinityBlocks.
+type ipamConfigClient struct {
+	rc customK8sResourceClient
+}
+
+func (c *ipamConfigClient) Create(ctx context.Context, kvp *model.KVPair) (*model.KVPair, error) {
+	log.Warn("Operation Create is not supported on AffinityBlock type")
```
caseydavenport: Logs are wrong.

lib/backend/k8s/resources/ipam_config.go
```diff
@@ -0,0 +1,97 @@
...
+	return nil, cerrors.ErrorOperationNotSupported{
+		Identifier: kvp.Key,
+		Operation:  "Create",
+	}
+}
+
+func (c *ipamConfigClient) Update(ctx context.Context, kvp *model.KVPair) (*model.KVPair, error) {
```
caseydavenport: We might need to support Update.

lib/backend/k8s/resources/ipam_handle.go
```diff
@@ -0,0 +1,178 @@
...
+		Value: &model.IPAMHandle{
+			HandleID: handle,
+			Block:    block,
+		},
+		Revision: kvpv3.Revision,
+	}
+}
+
+func (c ipamHandleClient) v3Fields(k model.Key) string {
+	return strings.ToLower(k.(model.IPAMHandleKey).HandleID)
```
caseydavenport: Needs design agreement.

lib/backend/k8s/resources/ipam_affinity.go
```diff
@@ -0,0 +1,206 @@
...
+
+	return &affinityBlockClient{rc: rc}
+}
+
+// Implements the api.Client interface for AffinityBlocks.
+type affinityBlockClient struct {
+	rc customK8sResourceClient
+}
+
+func toV1(kvpv3 *model.KVPair) *model.KVPair {
```
briansan: Shouldn't this (as well as L87) have a receiver, i.e. `func (c affinityBlockClient) toV1(...)`?

briansan: I guess it technically doesn't have to but otherwise, the name of this function without a receiver seems pretty ambiguous

lib/backend/k8s/resources/resources.go
```diff
@@ -144,6 +144,8 @@ func ConvertCalicoResourceToK8sResource(resIn Resource) (Resource, error) {
 	meta.Namespace = rom.GetNamespace()
 	meta.ResourceVersion = rom.GetResourceVersion()
 	meta.Labels = rom.GetLabels()
+	meta.UID = rom.GetUID()
+	meta.Finalizers = rom.GetFinalizers()
```
caseydavenport: Don't actually need the finalizers any more - we can delete.

lib/backend/model/block_affinity.go
```diff
@@ -36,11 +37,13 @@ const (
 	StateConfirmed       BlockAffinityState = "confirmed"
 	StatePending         BlockAffinityState = "pending"
 	StatePendingDeletion BlockAffinityState = "pendingDeletion"
+	StateDeleted         BlockAffinityState = "deleted"
```
caseydavenport: Probably should be `StateBlockDeleted`

lib/ipam/ipam_block_reader_writer.go
```diff
@@ -347,6 +345,76 @@ func (rw blockReaderWriter) releaseBlockAffinity(ctx context.Context, host strin
 	return nil
 }
 
+func (rw blockReaderWriter) queryAffinity(ctx context.Context, k model.Key, revision string) (*model.KVPair, error) {
```
caseydavenport: Need comments for these functions.

lib/ipam/ipam_block_reader_writer.go
```diff
@@ -347,6 +345,76 @@ func (rw blockReaderWriter) releaseBlockAffinity(ctx context.Context, host strin
...
+	return rw.client.Get(ctx, k, revision)
+}
+
+func (rw blockReaderWriter) updateAffinity(ctx context.Context, aff *model.KVPair) (*model.KVPair, error) {
+	return rw.client.Update(ctx, aff)
+}
+
+func (rw blockReaderWriter) deleteAffinity(ctx context.Context, aff *model.KVPair) error {
+	// Mark the affinity as block deleted.
+	aff.Value.(*model.BlockAffinity).State = model.StateDeleted
```
caseydavenport: We only need to do this for KDD mode. Would be nice to remove the need for the extra write in etcd, potentially.

lib/ipam/ipam_block_reader_writer.go
```diff
@@ -347,6 +345,76 @@ func (rw blockReaderWriter) releaseBlockAffinity(ctx context.Context, host strin
...
+	if b.Value.(*model.AllocationBlock).Deleting {
+		// If the block is in the process of being deleted, we should return
+		// a conflict error and retry.
+		return nil, cerrors.ErrorResourceUpdateConflict{Err: fmt.Errorf("Resource being deleted"), Identifier: b.Key}
+	}
+
+	return rw.client.Update(ctx, b)
+}
+
+func (rw blockReaderWriter) deleteBlock(ctx context.Context, b *model.KVPair) error {
```
caseydavenport: We need to implement a `queryBlock` here as well (similar to query affinity above). 

And, we need to do all of these for IPAM handle as well.

lib/ipam/ipam_test.go
```diff
@@ -1304,18 +1367,45 @@ func applyPoolWithBlockSize(cidr string, enabled bool, nodeSelector string, bloc
...
+		n.Name = host
+		n.Labels = labels
+
+		// Create/Update the node
+		newNode, err := kc.CoreV1().Nodes().Create(&n)
+		if err != nil && kerrors.IsAlreadyExists(err) {
+			oldNode, _ := kc.CoreV1().Nodes().Get(host, metav1.GetOptions{})
+			oldNode.Labels = labels
+
+			newNode, _ = kc.CoreV1().Nodes().Update(oldNode)
```
caseydavenport: We should at least log the error here to aid in debugging.

lib/ipam/ipam_test.go
```diff
@@ -1304,18 +1367,45 @@ func applyPoolWithBlockSize(cidr string, enabled bool, nodeSelector string, bloc
...
+				APIVersion: "v1",
+			},
+		}
+		n.Name = host
+		n.Labels = labels
+
+		// Create/Update the node
+		newNode, err := kc.CoreV1().Nodes().Create(&n)
+		if err != nil && kerrors.IsAlreadyExists(err) {
+			oldNode, _ := kc.CoreV1().Nodes().Get(host, metav1.GetOptions{})
```
caseydavenport: Should log this error too

lib/backend/k8s/k8s.go
```diff
@@ -291,6 +311,40 @@ func (c *KubeClient) Clean() error {
...
+		log.WithField("Kind", balo).Warning("Failed to list resources")
+	} else {
+		for _, r := range rs.KVPairs {
+			if _, err = c.Delete(ctx, r.Key, r.Revision); err != nil {
+				log.WithField("Key", r.Key).Warning("Failed to delete entry from KDD")
+			}
+		}
+	}
+
+	halo := model.BlockAffinityListOptions{}
```
briansan: This doesn't seem right...I think it's supposed to be `IPAMBlockListOptions?`

lib/ipam/ipam_block_reader_writer.go
```diff
@@ -347,6 +345,76 @@ func (rw blockReaderWriter) releaseBlockAffinity(ctx context.Context, host strin
...
+		return nil, err
+	}
+
+	if aff.Value.(*model.BlockAffinity).State != model.StateDeleted {
+		// We're all good - return the afifnity.
+		return aff, nil
+	}
+
+	// If the affinity is marked as deleted, we need to delete it.
+	if err := rw.deleteAffinity(ctx, aff); err != nil {
```
briansan: Doesn't it seem a little precarious to be deleting a resource on a query operation?

briansan: Wouldn't it be better just to return a `DoesNotExist` error and trust that whomever marked the affinity for deletion will follow through with it?

lib/backend/k8s/resources/client.go
caseydavenport: Eh, maybe not so bad - this is only added to the k8s ResourceClient, not the main backend Client interface, so it's potentially in the right place.

The `model.Key` interface is shared across backends, and not all of them will have a UID concept, so it probably doesn't belong on the `model.Key` interface.

lib/ipam/ipam_block_reader_writer.go
caseydavenport: I'm not sure what you mean by precarious in this case :D 

We actually can't rely on this being deleted by the code that marked it for deletion - it's totally possible that the other process successfully marked it as deleted, but then failed to delete the resource for some reason, meaning we need to clean up after it the next time this data is accessed.

lib/apis/v3/blockaffinity.go
```diff
@@ -0,0 +1,73 @@
+// Copyright (c) 2017 Tigera, Inc. All rights reserved.
```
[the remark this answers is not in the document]
briansan: submitted https://github.com/projectcalico/libcalico-go/pull/1011 to update the copyright headers throughout

lib/backend/k8s/resources/ipam_affinity.go
```diff
@@ -0,0 +1,247 @@
...
+		// If the name is too long, we need to shorten it.
+		// Remove enough characters to get it below the 253 character limit,
+		// as well as 11 characters to add a hash which helps with uniqueness,
+		// and two characters for the `-` separators between clauses.
+		name = fmt.Sprintf("%s-%s", host[:252-len(cidrstr)-13], cidrstr)
+
+		// Add a hash to help with uniqueness.
+		// Kubernetes requires all names to end with an alphabetic character, so
+		// append a 'c' to the end to ensure we always meet this requirement.
+		h := sha1.New()
```
fasaxc: sha1 is known-broken; best to use sha-256 if only because it means that grepping the codebase for sha1 will turn up no hits, making it easier to know that we're not using it where we shouldn't.  (Unlikely that it's breakage affects this use case!)

lib/backend/k8s/resources/ipam_affinity.go
```diff
@@ -0,0 +1,247 @@
...
+				// Matches the given host and IP version.
+				kvpl.KVPairs = append(kvpl.KVPairs, v1kvp)
+			}
+		}
+	}
+	return kvpl, nil
+}
+
+func (c *affinityBlockClient) Watch(ctx context.Context, list model.ListInterface, revision string) (api.WatchInterface, error) {
+	log.Warn("Operation Watch is not supported on AffinityBlock type")
```
fasaxc: Don't we need to watch blocks in confd?

lib/backend/k8s/resources/ipam_block.go
fasaxc: Don't we already have a function that converts a CIDR to a k8s name (for storing IPPools?)

lib/backend/k8s/resources/ipam_block.go
```diff
@@ -0,0 +1,268 @@
...
+	name = strings.Replace(name, "/", "-", -1)
+	return
+}
+
+func (c ipamBlockClient) toV3(kvpv1 *model.KVPair) *model.KVPair {
+	name, cidr := c.v3Fields(kvpv1.Key)
+
+	ab := kvpv1.Value.(*model.AllocationBlock)
+
+	// The CRD needs ints, not *int, so convert here.
```
fasaxc: For my education, what does happen to `*int`?

lib/backend/model/block.go
```diff
@@ -31,7 +32,8 @@ var (
 )
 
 type BlockKey struct {
-	CIDR net.IPNet `json:"-" validate:"required,name"`
+	CIDR net.IPNet  `json:"-" validate:"required,name"`
+	UID  *types.UID `json:"-"`
```
fasaxc: Should this be part of the key or should it be part of the KVPair?

lib/backend/k8s/resources/ipam_affinity.go
caseydavenport: I think we only watch block affinities in confd.

lib/backend/k8s/resources/ipam_block.go
caseydavenport: Ooooh, this might have been fixed since I prototyped this.  Here was the original issue: https://github.com/kubernetes/gengo/issues/11

lib/backend/model/block.go
caseydavenport: Good question - I was originally imagining this as part of the identifier for a resource, which is why I put it in the Key.

But, I guess it's analogous to the revision, which isn't included in the key. I could see it going either way.

lib/backend/k8s/resources/ipam_block.go
caseydavenport: Yep, looks like it basically does the same thing as this. I've updated the code to use that now.

lib/backend/k8s/resources/ipam_affinity.go
caseydavenport: For now, I suggest we allow a bit of duplication to reduce the need for additional interfaces in the backend.... I think the code will be more readable this way.

lib/backend/model/block.go
caseydavenport: Ah, the other issue is that `Delete` doesn't take a KVPair - it just takes a Key.

I guess we could update everything to pass around KVPs without Values, even for delete?

lib/backend/k8s/resources/ipam_affinity.go
caseydavenport: woops, these are affinities!

lib/backend/k8s/resources/ipam_block.go
fasaxc: That issue is for `*[]int`, not `[]*int`?

lib/backend/model/block.go
fasaxc: Heh, even the comment on Delete says it takes a KVPair :-)  

Maybe add a `DeleteKVPair()` method that takes revision/UID from the KVP instead of changing all callers?  I do feel like the UID is more akin to the revision than the key.  Adding it to the key breaks assumptions about the key being 1-1 with the name of the object, which might have unintended consequences.

caseydavenport: Yep, that sounds good to me. Fixed in latest commit.

lib/backend/k8s/resources/ipam_block.go
caseydavenport: Bah, you're right.

Let me just try using pointers and see what happens.. it has been a while since I prototyped this.

caseydavenport: Huh, this looks like it's working now.

I swore this was broken when I did this way back. Guess it's been fixed!

lib/backend/k8s/k8s.go
```diff
@@ -111,6 +111,8 @@ func NewKubeClient(ca *apiconfig.CalicoAPIConfigSpec) (api.Client, error) {
 	}
 
 	// Create the clientset
+	config.QPS = float32(500)
```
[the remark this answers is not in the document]
fasaxc: I think the default is something like 10 / 5.  Wonder if we can get away with something that's within an order of magnitude of that

lib/backend/k8s/k8s.go
```diff
@@ -416,6 +461,20 @@ func (c *KubeClient) Apply(ctx context.Context, kvp *model.KVPair) (*model.KVPai
 	return updated, nil
 }
 
+// Delete an entry in the datastore. This is a no-op when using the k8s backend.
```
fasaxc: Out of date comment?

lib/backend/k8s/resources/networkpolicy.go
```diff
@@ -128,8 +129,14 @@ func (c *networkPolicyClient) Apply(ctx context.Context, kvp *model.KVPair) (*mo
 		Operation:  "Apply",
 	}
 }
+func (c *networkPolicyClient) DeleteKVP(ctx context.Context, kvp *model.KVPair) (*model.KVPair, error) {
```
fasaxc: Easier to just call through to delete rather than leaving these unsupported?

lib/backend/k8s/k8s.go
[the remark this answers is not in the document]
caseydavenport: I've tried with slightly lower values - QPS=50 and Burst = 100, and some of the "Measurement" FVs in this repo start to fail.

It's perhaps not a huge deal, the tests were defined somewhat arbitrarily - how much do we care about a 1s allocation   time vs a 2s time?

lib/backend/k8s/resources/networkpolicy.go
caseydavenport: Yep, makes sense - some of them actually are unsupported though.

caseydavenport: Ah, I guess you mean since it will return an unsupported call anyway. sgtm.

lib/backend/k8s/resources/ipam_affinity.go
```diff
@@ -0,0 +1,269 @@
...
+	if len(name) >= 253 {
+		// If the name is too long, we need to shorten it.
+		// Remove enough characters to get it below the 253 character limit,
+		// as well as 11 characters to add a hash which helps with uniqueness,
+		// and two characters for the `-` separators between clauses.
+		name = fmt.Sprintf("%s-%s", host[:252-len(cidrname)-13], cidrname)
+
+		// Add a hash to help with uniqueness.
+		// Kubernetes requires all names to end with an alphabetic character, so
+		// append a 'c' to the end to ensure we always meet this requirement.
```
fasaxc: Where does this req come from?  I'm sure I've created pods with names like `busybox-1` before, would be nice to avoid having to add an extra random char to the end if we can

lib/backend/k8s/resources/ipam_affinity.go
```diff
@@ -0,0 +1,269 @@
...
+	}
+	v1kvp, err := c.toV1(kvp)
+	if err != nil {
+		return nil, err
+	}
+	return v1kvp, nil
+}
+
+func (c *affinityBlockClient) Delete(ctx context.Context, key model.Key, revision string, uid *types.UID) (*model.KVPair, error) {
+	log.Warn("Operation Delete is not supported on BlockAffinity type - use DeleteKVP")
```
fasaxc: Worth a comment saying `For correctness, deleting a block requires the UID.`.  Should avoid someone adding int he "missing" method during maintenance.

lib/backend/model/block.go
```diff
@@ -90,13 +90,22 @@ type AllocationBlock struct {
 	Allocations    []*int                `json:"allocations"`
 	Unallocated    []int                 `json:"unallocated"`
 	Attributes     []AllocationAttribute `json:"attributes"`
+	Deleting       bool                  `json:"deleting"`
```
fasaxc: Sorry for the late feedback.  Is it worth trying to sync the different approaches to this flag across the different resources?  I.e. here we use a flag called Deleting, but the Affinity uses a State field?

lib/backend/model/block.go
```diff
@@ -90,13 +90,22 @@ type AllocationBlock struct {
...
 
 	// HostAffinity is deprecated in favor of Affinity.
 	// This is only to keep compatiblity with existing deployments.
 	// The data format should be `Affinity: host:hostname` (not `hostAffinity: hostname`).
 	HostAffinity *string `json:"hostAffinity,omitempty"`
 }
 
+func (b *AllocationBlock) MarkDeleted() {
+	b.Deleting = true
```
fasaxc: Difference between Deleting and Deleted seems a bit off.  I have a mild preference for standardising on "Deleted"

lib/backend/model/deletion.go
```diff
@@ -0,0 +1,6 @@
+package model
```
fasaxc: Please add copyright to new files

lib/ipam/ipam_block.go
```diff
@@ -166,49 +166,59 @@ func (b *allocationBlock) release(addresses []cnet.IP) ([]cnet.IP, map[string]in
...
+	log.Debugf("Releasing addresses from block: %v", uniqueAddresses)
 	for ipStr, _ := range uniqueAddresses {
 		ip := cnet.MustParseIP(ipStr)
-
 		// Convert to an ordinal.
 		ordinal, err := ipToOrdinal(ip, *b)
 		if err != nil {
 			return nil, nil, err
 		}
+		log.Debugf("Address %s is ordinal %d", ip, ordinal)
```
fasaxc: +1 for having some debug trace in this logic!

lib/ipam/ipam_block_reader_writer_test.go
```diff
@@ -813,3 +951,70 @@ var _ = testutils.E2eDatastoreDescribe("IPAM affine block allocation tests", tes
...
+		var err error
+		bc, err = backend.NewClient(config)
+		Expect(err).NotTo(HaveOccurred())
+		bc.Clean()
+		ctx = context.Background()
+		_, net, err = cnet.ParseCIDR("10.0.0.0/26")
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	It("should respect Kubernetes UID for affinities", func() {
```
fasaxc: Should we have one for handles?

lib/backend/k8s/resources/ipam_affinity.go
caseydavenport: Yeah, good spot. That was something from the original prototype which may have been misguided.

lib/ipam/ipam_block_reader_writer_test.go
briansan: We decided it wouldn't be necessary to include a race condition test for handles since we are using the same Delete flag approach for both resources so would follow that just testing affinities is sufficient to cover the race condition for both. If we were to change the approach for deleting handles, then it would be necessary to add in a separate test case; but for now, I think it's worth it to take out the technical loan for the time.

briansan:
I agree that hiding the logic within KDD implementation of DeleteKVP would be worth it. It would avoid having our etcd implementation go through these unnecessary checks since it is not vulnerable to this potential race. Seems like a pretty small change.. @caseydavenport WDYT

caseydavenport:
@briansan yep, I've made that change locally. Will push soon.

lib/ipam/ipam_block_reader_writer_test.go
caseydavenport: @briansan this is just the UID tests, not the race condition tests.

I've added a test for handles respecting UID as well.

caseydavenport:
@fasaxc OK, I think it should all be addressed in the latest commit.

Note: I've bumped the "Burst" back up on the k8s client, since for some /32 blocksize allocations we require lots of datastore requests to find an unclaimed block, and to find an empty block. I've left the QPS at the default 5 though, as it seems to work fine. 

I'd like to optimize that further later, but it needs more design and I'd rather not hold this PR up further on it.

PR #1013, opened 2019-02-05
IPAM.ReleaseIPs de-dupes when unallocating

### LINKED ISSUE #989: IPAM Block does not protect against duplicates in Unallocated list

jsok:
## Description

IPAM.ReleaseIPs() happily accepts a list of IPs that contains
duplicates. This will leave duplicates in Unallocated which then creates
more havoc, in particular assigning already assigned IPs.

De-dupe the list of addresses before calculating ordinals to be
unalloacted.

Fixes #989

Port of #990 to `master` branch.

## Todos
- [x] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

TBA

### DIFF
--- a/lib/ipam/ipam_test.go
+++ b/lib/ipam/ipam_test.go
@@ -1186,7 +1186,7 @@ var _ = testutils.E2eDatastoreDescribe("IPAM tests", testutils.DatastoreEtcdV3, 
 		// - Assign should not return no error.
 		// - ReleaseIPs should return a slice with one (unallocatedIPs) and no error.
 		Entry("Assign 1 IPv4 address with AssignIP then try to release 2 IPs (assign one and release it)", net.IP{}, true, []string{"192.168.1.0/24", "fd80:24e2:f998:72d6::/120"}, net.ParseIP("192.168.1.0"), 0, []cnet.IP{}, nil),
-		Entry("Assign 1 IPv4 address with AssignIP then try to release 2 IPs (release a second one)", net.ParseIP("192.168.1.1"), false, []string{"192.168.1.0/24", "fd80:24e2:f998:72d6::/120"}, net.IP{}, 0, []cnet.IP{{net.ParseIP("192.168.1.1")}}, nil),
+		Entry("Assign 1 IPv4 address with AssignIP then try to release 2 IPs (release a second one)", net.ParseIP("192.168.1.1"), false, []string{"192.168.1.0/24", "fd80:24e2:f998:72d6::/120"}, net.IP{}, 0, []cnet.IP{cnet.MustParseIP("192.168.1.1")}, nil),
 	)
 
 	DescribeTable("ClaimAffinity: claim IPNet vs actual number of blocks claimed",
--- a/lib/ipam/ipam_test.go
+++ b/lib/ipam/ipam_test.go
@@ -211,6 +211,59 @@ var _ = testutils.E2eDatastoreDescribe("IPAM tests", testutils.DatastoreEtcdV3, 
 		})
 	})
 
+	Describe("IPAM ReleaseIPs with duplicates in the request should be safe", func() {
+		host := "host-A"
+		pool1 := cnet.MustParseNetwork("10.0.0.0/26")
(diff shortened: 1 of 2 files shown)


lib/ipam/ipam_test.go
```diff
@@ -1133,7 +1186,7 @@ var _ = testutils.E2eDatastoreDescribe("IPAM tests", testutils.DatastoreEtcdV3,
 		// - Assign should not return no error.
 		// - ReleaseIPs should return a slice with one (unallocatedIPs) and no error.
 		Entry("Assign 1 IPv4 address with AssignIP then try to release 2 IPs (assign one and release it)", net.IP{}, true, []string{"192.168.1.0/24", "fd80:24e2:f998:72d6::/120"}, net.ParseIP("192.168.1.0"), 0, []cnet.IP{}, nil),
-		Entry("Assign 1 IPv4 address with AssignIP then try to release 2 IPs (release a second one)", net.ParseIP("192.168.1.1"), false, []string{"192.168.1.0/24", "fd80:24e2:f998:72d6::/120"}, net.IP{}, 0, []cnet.IP{{net.ParseIP("192.168.1.1")}}, nil),
+		Entry("Assign 1 IPv4 address with AssignIP then try to release 2 IPs (release a second one)", net.ParseIP("192.168.1.1"), false, []string{"192.168.1.0/24", "fd80:24e2:f998:72d6::/120"}, net.IP{}, 0, []cnet.IP{cnet.MustParseIP("192.168.1.1")}, nil),
```
jsok: Getting failure:

```

  Expected

      <[]net.IP | len:1, cap:1>: [

          {IP: [192, 168, 1, 1]},

      ]

  to equal

      <[]net.IP | len:1, cap:1>: [

          {

              IP: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 255, 255, 192, 168, 1, 1],

          },

      ]



  /go/src/github.com/projectcalico/libcalico-go/lib/ipam/ipam_test.go:1157
```

This is presumably because gomega `Equal` matcher is using `reflect.DeepEqual` rather than `net.IP.Equal`. Now that the de-duping going from IP -> String -> IP the embedded IPv4/v6 struct is getting mixed up in the tests :(

PR #1021, opened 2019-02-14
add support for IPIP tunnel address on node in kdd mode

briansan:
## Description
- adds a new annotation field for the Node resource to optionally store the IPIP tunnel address

## Todos
- [X] Tests


lib/backend/k8s/resources/node.go
```diff
@@ -198,6 +199,15 @@ func (c *nodeClient) Watch(ctx context.Context, list model.ListInterface, revisi
 
 // K8sNodeToCalico converts a Kubernetes format node, with Calico annotations, to a Calico Node.
 func K8sNodeToCalico(k8sNode *kapiv1.Node) (*model.KVPair, error) {
+	// Validate the pod CIDR field
+	if k8sNode.Spec.PodCIDR != "" {
```
caseydavenport: Do we need this validation if we're not using the field any more?

PR #1022, opened 2019-02-14
IptablesNATOutgoingInterfaceFilter config param added

roshanirathi:
This PR is to implement a IptablesNATOutgoingInterfaceFilter config param to solve a Docker masq issue.

### DIFF
--- a/lib/apis/v3/felixconfig.go
+++ b/lib/apis/v3/felixconfig.go
@@ -215,7 +215,7 @@ type FelixConfigurationSpec struct {
 	DebugSimulateCalcGraphHangAfter *metav1.Duration `json:"debugSimulateCalcGraphHangAfter,omitempty" configv1timescale:"seconds"`
 	DebugSimulateDataplaneHangAfter *metav1.Duration `json:"debugSimulateDataplaneHangAfter,omitempty" configv1timescale:"seconds"`
 
-	IptablesNATOutgoingInterfaceFilter string `json:"iptablesNATOutgoingInterfaceFilter,omitempty" validate:"omitempty,natParam"`
+	IptablesNATOutgoingInterfaceFilter string `json:"iptablesNATOutgoingInterfaceFilter,omitempty" validate:"omitempty,ifaceFilter"`
 }
 
 // ProtoPort is combination of protocol and port, both must be specified.
--- a/lib/apis/v3/felixconfig.go
+++ b/lib/apis/v3/felixconfig.go
@@ -215,7 +215,7 @@ type FelixConfigurationSpec struct {
 	DebugSimulateCalcGraphHangAfter *metav1.Duration `json:"debugSimulateCalcGraphHangAfter,omitempty" configv1timescale:"seconds"`
 	DebugSimulateDataplaneHangAfter *metav1.Duration `json:"debugSimulateDataplaneHangAfter,omitempty" configv1timescale:"seconds"`
 
-	IptablesNATOutgoingInterfaceFilter *bool `json:"iptablesNATOutgoingInterfaceFilter,omitempty"`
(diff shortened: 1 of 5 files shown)


lib/apis/v3/felixconfig.go
```diff
@@ -214,6 +214,8 @@ type FelixConfigurationSpec struct {
 	DebugDisableLogDropping         *bool            `json:"debugDisableLogDropping,omitempty"`
 	DebugSimulateCalcGraphHangAfter *metav1.Duration `json:"debugSimulateCalcGraphHangAfter,omitempty" configv1timescale:"seconds"`
 	DebugSimulateDataplaneHangAfter *metav1.Duration `json:"debugSimulateDataplaneHangAfter,omitempty" configv1timescale:"seconds"`
+
+	IptablesNATOutgoingInterfaceFilter *bool `json:"iptablesNATOutgoingInterfaceFilter,omitempty"`
```
fasaxc: As per Felix change, this should be a string.

Also, we should validate this value in `validateFelixConfigSpec` in `validator.go` using same regex as used in Felix (and add UT in `validator_test.go`).

roshanirathi: ack

lib/validator/v3/validator.go
```diff
@@ -230,6 +232,12 @@ func validateInterface(v *validator.Validate, topStruct reflect.Value, currentSt
 	return s == "*" || interfaceRegex.MatchString(s)
 }
 
+func validateNATParam(v *validator.Validate, topStruct reflect.Value, currentStructOrField reflect.Value, field reflect.Value, fieldType reflect.Type, fieldKind reflect.Kind, param string) bool {
+	s := field.String()
+	log.Debugf("Validate NAT Parameter: %s", s)
+	return s == "*" || natParamRegex.MatchString(s)
```
fasaxc: I don't think `*` should be valid here.  `*` is a special-case value for the interface name of a host endpoint so we don't want to copy that here.

lib/apis/v3/felixconfig.go
```diff
@@ -214,6 +214,8 @@ type FelixConfigurationSpec struct {
 	DebugDisableLogDropping         *bool            `json:"debugDisableLogDropping,omitempty"`
 	DebugSimulateCalcGraphHangAfter *metav1.Duration `json:"debugSimulateCalcGraphHangAfter,omitempty" configv1timescale:"seconds"`
 	DebugSimulateDataplaneHangAfter *metav1.Duration `json:"debugSimulateDataplaneHangAfter,omitempty" configv1timescale:"seconds"`
+
+	IptablesNATOutgoingInterfaceFilter string `json:"iptablesNATOutgoingInterfaceFilter,omitempty" validate:"omitempty,natParam"`
```
fasaxc: Suggest you change `natParam` to `ifaceFilter` since we could use that validation to validate other similar fields in future.

lib/validator/v3/validator.go
roshanirathi: ack

lib/apis/v3/felixconfig.go
roshanirathi: ack

lib/validator/v3/validator.go
```diff
@@ -230,6 +232,12 @@ func validateInterface(v *validator.Validate, topStruct reflect.Value, currentSt
 	return s == "*" || interfaceRegex.MatchString(s)
 }
 
+func validateNATParam(v *validator.Validate, topStruct reflect.Value, currentStructOrField reflect.Value, field reflect.Value, fieldType reflect.Type, fieldKind reflect.Kind, param string) bool {
```
fasaxc: Nit: should be `validateIfaceFilterParam` now.

PR #1026, opened 2019-02-15
Handle errors querying IP pools in IPAM code.

caseydavenport:
We were just ignoring errors here before.

- [ ] Tests
- [ ] Documentation
- [ ] Release note

```release-note
None required
```

### DIFF
--- a/lib/ipam/ipam.go
+++ b/lib/ipam/ipam.go
@@ -315,7 +315,11 @@ func (c ipamClient) autoAssign(ctx context.Context, num int, handleID *string, a
 	// Release any emptied blocks still affine to this host but no longer part of an IP Pool which selects this node.
 	for _, block := range affBlocksToRelease {
 		// Determine the pool for each block.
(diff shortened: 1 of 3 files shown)


lib/ipam/ipam.go
```diff
@@ -946,6 +962,7 @@ func (c ipamClient) ReleaseAffinity(ctx context.Context, cidr net.IPNet, host st
 // to the given host.  If an empty string is passed as the host,
 // then the hostname is automatically detected.
 func (c ipamClient) ReleaseHostAffinities(ctx context.Context, host string, mustBeEmpty bool) error {
+	log.Debugf("Releasing affinities for host %s. MustBeEmpty? %b", host, mustBeEmpty)
```
fasaxc: `s/%b/%t/`  (or just use `%v` unless formatting is super-critical)

fasaxc:
SGTM.  Looks like the CI failure was due to a linter check.

PR #1036, opened 2019-02-21
Add contains/startswith/endswith selector  operators.

fasaxc:
## Description

Add `contains`, `starts with` and `ends with` operators to selector syntax.

## Todos
- [x] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Selector syntax now supports operators for substring matching: "contains", "starts with" and "ends with".
```


neiljerram:
FYI there's also a release note on the draft doc PR that I raised at https://github.com/projectcalico/calico/pull/2410.  I think the release note here is good, so should probably delete the one at #2410, if #2410 is used at all.

fasaxc:
Looks like I hit an unrelated flake.

PR #1040, opened 2019-03-04
Exclude completed pods from being workload endpoints.

fasaxc:
## Description

When a pod finishes, 

- the kubelet calls the CNI plugin to remove the pod from the network 
- this returns the IP to the IPAM pool
- however, the Pod resource can stay around for a long time and the Pod resource keeps its IP
- this was resulting in Felix thinking that there are multiple workload endpoints with the same IP.

Exclude completed pods from those that are considered to be valid workload endpoints.

Fixes https://github.com/projectcalico/felix/issues/1966.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/backend/k8s/conversion/conversion_test.go
+++ b/lib/backend/k8s/conversion/conversion_test.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2016-2017 Tigera, Inc. All rights reserved.
+// Copyright (c) 2016-2019 Tigera, Inc. All rights reserved.
 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


lib/backend/k8s/conversion/conversion_test.go
```diff
@@ -381,7 +381,37 @@ var _ = Describe("Test Pod conversion", func() {
...
+				Status: kapiv1.PodStatus{
+					PodIP: "192.168.0.1",
+					Phase: podPhase,
+				},
+			}
+			Expect(c.IsValidCalicoWorkloadEndpoint(&pod)).To(Equal(expectedResult))
+		},
+		Entry("Pending", kapiv1.PodPending, true),
+		Entry("Running", kapiv1.PodRunning, true),
+		Entry("Unknown", kapiv1.PodUnknown, true),
```
neiljerram: Worth a quick comment to say what Unknown means, I think.

PR #1047, opened 2019-03-05
Treat finished pods as having no IPs

fasaxc:
## Description

- Workload endpoints with no IP get filtered out int he syncer and result in a deletion to Felix.
- They are presented as is on the v3 get/list/watch API, resulting in correct API update type behaviour (although it's a bit odd to have a wep with no IP it's already possible so it feels less risky).

## Todos
- [x] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/backend/k8s/k8s_test.go
+++ b/lib/backend/k8s/k8s_test.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2016-2017 Tigera, Inc. All rights reserved.
+// Copyright (c) 2016-2019 Tigera, Inc. All rights reserved.
 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


lib/backend/k8s/k8s_test.go
```diff
@@ -1297,6 +1310,131 @@ var _ = testutils.E2eDatastoreDescribe("Test Syncer API for Kubernetes backend",
 		})
 	})
 
+	// There are several states that we consider "finished", run the test for each one.
+	for _, finishPhase := range []k8sapi.PodPhase{k8sapi.PodSucceeded, k8sapi.PodFailed} {
```
tmjd: I'm not sure if it is really necessary but the "finished" check actually has 3 phases that it considers finished, with the 3rd being `podCompleted` though as commented is not part of the K8s API. I think we should include that in the tests especially since it is not actually from the K8s API.

PR #1048, opened 2019-03-05
Ensure protocol is TCP is HTTP rule is given

jpalardy:
## Description

This PR adds a check to make sure that combination of rules make sense together. Specifically, if an HTTP-level rule is given, the protocol has to be TCP.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Validate that protocol is set correctly if an HTTPMatch is specified in a NetworkPolicy
```

### DIFF
--- a/lib/validator/v3/validator.go
+++ b/lib/validator/v3/validator.go
@@ -72,6 +72,7 @@ var (
 	overlapsV6LinkLocal   = "IP pool range overlaps with IPv6 Link Local range fe80::/10"
 	protocolPortsMsg      = "rules that specify ports must set protocol to TCP or UDP"
 	protocolIcmpMsg       = "rules that specify ICMP fields must set protocol to ICMP"
+	protocolAndHTTPMsg    = "rules that specify HTTP fields must set protocol to TCP"
 
 	ipv4LinkLocalNet = net.IPNet{
 		IP:   net.ParseIP("169.254.0.0"),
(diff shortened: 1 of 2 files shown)


lib/validator/v3/validator.go
```diff
@@ -72,6 +72,7 @@ var (
 	overlapsV6LinkLocal   = "IP pool range overlaps with IPv6 Link Local range fe80::/10"
 	protocolPortsMsg      = "rules that specify ports must set protocol to TCP or UDP"
 	protocolIcmpMsg       = "rules that specify ICMP fields must set protocol to ICMP"
+	protocolAndHTTPMsg    = "rules that specify HTTP fields must set protocol to TCP"
```
jpalardy: the wording might need some work…

lib/validator/v3/validator_test.go
```diff
@@ -17,13 +17,13 @@ package v3_test
 import (
 	. "github.com/onsi/ginkgo/extensions/table"
 	. "github.com/onsi/gomega"
-	"k8s.io/apimachinery/pkg/apis/meta/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
jpalardy: this and next were added by `goimports`…

caseydavenport: Yeah, I've noticed it has started doing this recently. /shrug.

lib/validator/v3/validator.go
caseydavenport: I think this looks alright to me - it matches the format of the other strings.

Maybe "can only set protocol to TCP" since "must" is a little bit strong?

or 

"rules that specify HTTP fields must set protocol to TCP or empty" or something like that?

jpalardy: good good… I like the 2nd one better, I think.

jpalardy: Updated.

PR #1055, opened 2019-03-06
Add NetworkSet resource

lmm:
## Description

Add NetworkSet, which is the namespaced version of GlobalNetworkSet.

Unit tests pending.

## Todos
- [x] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
Add NetworkSet resource. NetworkSets are the namespaced-equivalent of GlobalNetworkSet.
```


lib/clientv3/networkset.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+// networkSets implements NetworkSetInterface
+type networkSets struct {
+	client client
+}
+
+// Create takes the representation of a NetworkSet and creates it.  Returns the stored
+// representation of the NetworkSet, and an error, if there is any.
+func (r networkSets) Create(ctx context.Context, res *apiv3.NetworkSet, opts options.SetOptions) (*apiv3.NetworkSet, error) {
```
caseydavenport: I think we'll want to add some logic like we do for workload endpoints: https://github.com/projectcalico/libcalico-go/blob/master/lib/clientv3/workloadendpoint.go#L58

This ensures that we always set the `projectcalico.org/namespace` label, allowing it to be selected by NetworkPolicy objects.

Note this logic, which enforces that NetworkPolicy selectors are always limited to a particular namespace by adding an additional selector clause: https://github.com/projectcalico/libcalico-go/blob/master/lib/backend/syncersv1/updateprocessors/networkpolicyprocessor.go#L52-L59

We do a similar, though substantially more complicated operation for the rule selectors as well to ensure they are namespaced correctly: https://github.com/projectcalico/libcalico-go/blob/master/lib/backend/syncersv1/updateprocessors/rules.go#L108-L127

lib/validator/v3/validator.go
```diff
@@ -954,6 +955,23 @@ func validateNetworkPolicy(structLevel validator.StructLevel) {
 	}
 }
 
+func validateNetworkSet(structLevel validator.StructLevel) {
```
caseydavenport: Is this going to work if someone does a get, modify, update type set of operations?

Won't the GET return an object with a namespace label on it? Maybe we just need to filter out that label on GET.

caseydavenport: Or alternatively only return an error if it doesn't match the Namespace field? WDYT?

lmm: As discussed, we'll leave this check in and simplify the insertion of the `projectcalico.org/namespace` label.

PR #1058, opened 2019-03-12
Add watch kdd

jpalardy:
This PR adds support for named `Watch` for calico resources.

I'm open to suggestions on structure or style.

## Todos
- [x] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/backend/k8s/k8s_test.go
+++ b/lib/backend/k8s/k8s_test.go
@@ -240,6 +240,26 @@ func (c cb) ExpectDeleted(kvps []model.KVPair) {
 	}
 }
 
+// ExpectAddedEvent checks for an api.WatchAdded coming down a specific chan
+// this is used in several tests below
+func ExpectAddedEvent(events <-chan api.WatchEvent) *api.WatchEvent {
+	var receivedEvent *api.WatchEvent
+	for i := 0; i < 10; i++ {
+		select {
+		case e := <-events:
+			// Got an event. Check it's OK.
+			Expect(e.Error).NotTo(HaveOccurred())
+			Expect(e.Type).To(Equal(api.WatchAdded))
+			receivedEvent = &e
+			break
+		default:
+			time.Sleep(50 * time.Millisecond)
+		}
+	}
+	Expect(receivedEvent).NotTo(BeNil(), "Did not receive watch event")
+	return receivedEvent
+}
+
 // GetSyncerValueFunc returns a function that can be used to query the value of
 // an entry in our syncer state store.  It's useful for performing "Eventually" testing.
 //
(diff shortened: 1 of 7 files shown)


lib/backend/k8s/k8s_test.go
```diff
@@ -1967,6 +1987,154 @@ var _ = testutils.E2eDatastoreDescribe("Test Watch support", testutils.Datastore
...
+			name := "default"
+			watch, err := c.Watch(ctx, model.ResourceListOptions{Name: name, Kind: apiv3.KindProfile}, "")
+			Expect(err).NotTo(HaveOccurred())
+			event := ExpectAddedEvent(watch.ResultChan())
+			Expect(event.New.Key.String()).To(Equal("Profile(kns.default)"))
+		})
+		By("Watching all Profiles", func() {
+			watch, err := c.Watch(ctx, model.ResourceListOptions{Kind: apiv3.KindProfile}, "")
+			Expect(err).NotTo(HaveOccurred())
+			ExpectAddedEvent(watch.ResultChan())
```
caseydavenport: We generate one of these per namespace, so might want to verify we get a second one here too.

lib/backend/k8s/k8s_test.go
```diff
@@ -1967,6 +1987,154 @@ var _ = testutils.E2eDatastoreDescribe("Test Watch support", testutils.Datastore
...
+		createTestNetworkPolicy := func() {
+			np := networkingv1.NetworkPolicy{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "test-watch-net-policy",
+				},
+			}
+			res := c.ClientSet.NetworkingV1().RESTClient().Post().Resource("networkpolicies").Namespace("default").Body(&np).Do()
+			Expect(res.Error()).NotTo(HaveOccurred())
+		}
+		deleteAllNetworkPolicies := func() {
```
caseydavenport: Ginkgo supports an `AfterEach()` option which might be well-suited for this. It ensures you run this after each test, even if the test itself fails.

The way it is currently written, an `Expect` might fail and then this function won't be executed.

lib/backend/k8s/k8s_test.go
```diff
@@ -1967,6 +1987,154 @@ var _ = testutils.E2eDatastoreDescribe("Test Watch support", testutils.Datastore
...
+			err := c.Clean()
+			Expect(err).NotTo(HaveOccurred())
+			nps := networkingv1.NetworkPolicyList{}
+			err = c.ClientSet.NetworkingV1().RESTClient().Get().Resource("networkpolicies").Timeout(10 * time.Second).Do().Into(&nps)
```
caseydavenport: Shouldn't need to do the call to `RESTClient()` here.

I think `NetworkingV1().NetworkPolicies().List(metav1.ListOptions{})` should do the trick (it does all the restclient stuff for you).

caseydavenport: Same for the other calls through to `RESTClient`

lib/backend/k8s/k8s_test.go
```diff
@@ -1967,6 +1987,154 @@ var _ = testutils.E2eDatastoreDescribe("Test Watch support", testutils.Datastore
...
+		})
+		By("Watching all Profiles", func() {
+			watch, err := c.Watch(ctx, model.ResourceListOptions{Kind: apiv3.KindProfile}, "")
+			Expect(err).NotTo(HaveOccurred())
+			ExpectAddedEvent(watch.ResultChan())
+		})
+	})
+
+	It("Should support watching NetworkPolicies", func() {
+		createTestNetworkPolicy := func() {
```
caseydavenport: Might be nice to have this function take a name, so we can create multiple distinct policies. WDYT?

If we can create two policies, then we can assert we only get one event when watching a particular name.

lib/backend/k8s/k8s_test.go
```diff
@@ -1967,6 +1987,154 @@ var _ = testutils.E2eDatastoreDescribe("Test Watch support", testutils.Datastore
...
+				},
+			}
+			_, err := c.Create(ctx, pool)
+			Expect(err).NotTo(HaveOccurred())
+		}
+		deleteTestIPPool := func() {
+			_, err := c.Delete(ctx, model.ResourceKey{Name: "test-ippool", Kind: apiv3.KindIPPool}, "")
+			Expect(err).NotTo(HaveOccurred())
+		}
+		By("Watching a single NetworkPolicy", func() {
```
caseydavenport: copy/paste error - still says NetworkPolicy :)

lib/backend/k8s/k8s_test.go
```diff
@@ -1967,6 +1987,154 @@ var _ = testutils.E2eDatastoreDescribe("Test Watch support", testutils.Datastore
 		ctx = context.Background()
 	})
 
+	It("Should support watching Profiles", func() {
+		By("Watching a single Profile", func() {
+			name := "default"
```
caseydavenport: I think we should have two tests here:
- Watch a single profile backed by a Namespace (`name := kns.default`)
- Watch a single profile backed by a serviceaccount (`name := ksa.namespace-2.default`)

lib/backend/k8s/k8s_test.go
jpalardy: Ok, I'm redoing this with `Describe` and `It` (instead of `By`) so I can use `Before/AfterEach`

jpalardy:
Added commits which addresses most conversations above.

jpalardy:
That last test failure brought up a good point: if profiles watch both namespaces and serviceAccounts, what does it mean to watch a specific name if it can point to 2 resources?

(in practice, you'll get events for both)

what was failing:
- existing "default" namespace
- a serviceAccount named "default" under one of the non-default namespaces

lib/backend/k8s/k8s_test.go
```diff
@@ -1967,6 +1987,222 @@ var _ = testutils.E2eDatastoreDescribe("Test Watch support", testutils.Datastore
...
+			deleteAllServiceAccounts()
+		})
+		It("supports watching a specific profile (from namespace)", func() {
+			watch, err := c.Watch(ctx, model.ResourceListOptions{Name: "namespace-1", Kind: apiv3.KindProfile}, "")
+			Expect(err).NotTo(HaveOccurred())
+			event := ExpectAddedEvent(watch.ResultChan())
+			Expect(event.New.Key.String()).To(Equal("Profile(kns.namespace-1)"))
+		})
+		It("supports watching a specific profile (from serviceAccount)", func() {
+			watch, err := c.Watch(ctx, model.ResourceListOptions{Name: "test-sa-1", Kind: apiv3.KindProfile}, "")
```
caseydavenport: I think in this case, we're passing in the wrong thing for the name. Basically, the name we receive on the output event should match the name we pass in.

In this case, that means the watch needs to parse the name and decide which of the two objects it actually is meant to watch (or both, if no name is present).

I think the tests need to be updated to pass in `kns.namespace-1` and `ksa.default.test-sa-1`, and then it should be more apparent what needs to be done here.

Sorry I didn't spot this earlier, but the names we pass in on this API are the _calico_ resource names, not the names of the backing Kubernetes objects.

lib/backend/k8s/k8s_test.go
```diff
@@ -2010,13 +2010,13 @@ var _ = testutils.E2eDatastoreDescribe("Test Watch support", testutils.Datastore
...
-			watch, err := c.Watch(ctx, model.ResourceListOptions{Name: "namespace-1", Kind: apiv3.KindProfile}, "")
+			watch, err := c.Watch(ctx, model.ResourceListOptions{Name: "kns.default", Kind: apiv3.KindProfile}, "")
 			Expect(err).NotTo(HaveOccurred())
 			event := ExpectAddedEvent(watch.ResultChan())
-			Expect(event.New.Key.String()).To(Equal("Profile(kns.namespace-1)"))
+			Expect(event.New.Key.String()).To(Equal("Profile(kns.default)"))
 		})
 		It("supports watching a specific profile (from serviceAccount)", func() {
-			watch, err := c.Watch(ctx, model.ResourceListOptions{Name: "test-sa-1", Kind: apiv3.KindProfile}, "")
+			watch, err := c.Watch(ctx, model.ResourceListOptions{Name: "ksa.test-sa-1", Kind: apiv3.KindProfile}, "")
```
caseydavenport: Should be "ksa.default.test-sa-1"

lib/backend/k8s/resources/profile.go
```diff
@@ -229,10 +230,21 @@ func (c *profileClient) Watch(ctx context.Context, list model.ListInterface, rev
...
 		return nil, fmt.Errorf("ListInterface is not a ResourceListOptions: %s", list)
 	}
+	namePrefix := ""
 	if len(rlo.Name) != 0 {
 		// We've been asked to watch a specific profile.
 		log.WithField("name", rlo.Name).Debug("Watching a single profile")
-		opts.FieldSelector = fields.OneTermEqualSelector("metadata.name", rlo.Name).String()
+		if strings.HasPrefix(rlo.Name, "kns.") {
+			namePrefix = "kns"
+			shortName := strings.TrimPrefix(rlo.Name, "kns.")
```
caseydavenport: Let's use the existing conversion functions I linked you for these so we're note duplicating work.

lib/backend/k8s/resources/watcher.go
```diff
@@ -46,7 +46,9 @@ func newK8sWatcherConverter(
 		cancel:     cancel,
 		resultChan: make(chan api.WatchEvent, resultsBufSize),
 	}
-	go wc.processK8sEvents()
+	if wc.k8sWatch != nil {
```
caseydavenport: Hm, this one feels a little suspect to me - shouldn't the calling code simply not call this function in this case?

Not sure the full story but it feels like this might belong outside of this function.

lib/backend/k8s/k8s_test.go
jpalardy: ok, that's what I thought 👍

lib/backend/k8s/resources/profile.go
jpalardy: good stuff!

lib/backend/k8s/resources/watcher.go
jpalardy: this one is a long-ish story… we'll sync tomorrow.

lib/backend/k8s/k8s_test.go
```diff
@@ -1967,6 +1987,227 @@ var _ = testutils.E2eDatastoreDescribe("Test Watch support", testutils.Datastore
...
+		}
+		BeforeEach(func() {
+			createTestNetworkPolicy("test-net-policy-1")
+			createTestNetworkPolicy("test-net-policy-2")
+		})
+		AfterEach(func() {
+			deleteAllNetworkPolicies()
+		})
+		It("supports watching a specific networkpolicy", func() {
+			watch, err := c.Watch(ctx, model.ResourceListOptions{Name: "test-net-policy-1", Kind: apiv3.KindNetworkPolicy}, "")
```
caseydavenport: Grr, I think we missed this one too - note the mismatch between the name we're passing in and the name we're receiving on the watch. They should be the same, just like for Profiles..

lib/backend/k8s/resources/profile.go
```diff
@@ -222,9 +224,40 @@ func (c *profileClient) EnsureInitialized() error {
...
+			namePrefix = conversion.NamespaceProfileNamePrefix
+			var err error
+			ns, err = c.ProfileNameToNamespace(rlo.Name)
+			if err != nil {
+				return nil, err
+			}
+			opts.FieldSelector = fields.OneTermEqualSelector("metadata.name", ns).String()
+		} else if strings.HasPrefix(rlo.Name, conversion.ServiceAccountProfileNamePrefix) {
+			namePrefix = conversion.ServiceAccountProfileNamePrefix
+			var err error
```
caseydavenport: nit: Let's just pre-declare 'err' outside of this block (up where we declare everything else) so we don't need to do it twice down here.

lib/backend/k8s/resources/profile.go
```diff
@@ -237,40 +270,44 @@ func (c *profileClient) Watch(ctx context.Context, list model.ListInterface, rev
...
+		nsWatch, err = c.clientSet.CoreV1().Namespaces().Watch(opts)
+		if err != nil {
+			return nil, K8sErrorToCalico(err, list)
+		}
 	}
 	converter := func(r Resource) (*model.KVPair, error) {
 		k8sNamespace, ok := r.(*kapiv1.Namespace)
 		if !ok {
-			return nil, errors.New("profile conversion with incorrect k8s resource type")
+			return nil, errors.New("Profile conversion with incorrect k8s resource type")
```
caseydavenport: In golang, errors typically always start with a lowercase so that they can be more easily chained. Probably OK here since we're referring to a specific type, but thought it worth calling out.

lib/backend/k8s/resources/profile.go
```diff
@@ -237,40 +270,44 @@ func (c *profileClient) Watch(ctx context.Context, list model.ListInterface, rev
 		"saRev": saRev,
 	}).Debug("Watching two resources at individual revisions")
```
caseydavenport: nit: this log isn't necessarily right any more if a name is specified.

lib/backend/k8s/resources/profile.go
jpalardy: I was trying to be consistent with the rest of the file. Specifically with [line 304](https://github.com/projectcalico/libcalico-go/blob/c9b134e0e76ac14882aefcd701d979043956bcb4/lib/backend/k8s/resources/profile.go#L304) with the same error message just a few lines below.

lib/backend/k8s/resources/workloadendpoint.go
```diff
@@ -219,12 +220,20 @@ func (c *WorkloadEndpointClient) EnsureInitialized() error {
...
+	// Build watch options to pass to k8s.
+	opts := metav1.ListOptions{ResourceVersion: revision, Watch: true}
+	rlo, ok := list.(model.ResourceListOptions)
+	if !ok {
+		return nil, fmt.Errorf("ListInterface is not a ResourceListOptions: %s", list)
+	}
+	if len(rlo.Name) != 0 {
+		// We've been asked to watch a specific workloadendpoint
+		log.WithField("name", rlo.Name).Debug("Watching a single workloadendpoint")
+		opts.FieldSelector = fields.OneTermEqualSelector("metadata.name", rlo.Name).String()
```
caseydavenport: One more here: https://github.com/projectcalico/libcalico-go/blob/master/lib/backend/k8s/conversion/conversion.go#L76-L78

lib/backend/k8s/k8s_test.go
```diff
@@ -1967,6 +1987,227 @@ var _ = testutils.E2eDatastoreDescribe("Test Watch support", testutils.Datastore
...
+		}
+		BeforeEach(func() {
+			createTestPod("test-pod-1")
+			createTestPod("test-pod-2")
+		})
+		AfterEach(func() {
+			deleteAllPods()
+		})
+		It("supports watching a specific workloadEndpoint", func() {
+			watch, err := c.Watch(ctx, model.ResourceListOptions{Name: "default/127.0.0.1-k8s-test--pod--1-eth0)", Kind: apiv3.KindWorkloadEndpoint}, "")
```
caseydavenport: Ah, in this case `default/` is not part of the pod name - it's the pod's namespace.

lib/backend/k8s/resources/networkpolicy.go
```diff
@@ -289,9 +288,23 @@ func (c *networkPolicyClient) EnsureInitialized() error {
...
-	resl := list.(model.ResourceListOptions)
-	if len(resl.Name) != 0 {
-		return nil, fmt.Errorf("cannot watch specific resource instance: %s", list.(model.ResourceListOptions).Name)
+	// Build watch options to pass to k8s.
+	opts := metav1.ListOptions{Watch: true}
+	rlo, ok := list.(model.ResourceListOptions)
+	if !ok {
+		return nil, fmt.Errorf("ListInterface is not a ResourceListOptions: %s", list)
+	}
+	if len(rlo.Name) != 0 {
```
caseydavenport: Hm, I think we only need to parse the policy name if this is a `kns.` prefixed policy, since those are backed by k8s.

There are two watches in this func, so need to handle both cases and only start one of the two watches if a name is passed (like with profiles)

lib/backend/k8s/resources/profile.go
```diff
@@ -222,55 +224,101 @@ func (c *profileClient) EnsureInitialized() error {
...
+		} else {
+			return nil, fmt.Errorf("Unsupported prefix for resource name: %s", rlo.Name)
+		}
 	}
 
 	nsRev, saRev, err := c.SplitProfileRevision(revision)
 	if err != nil {
 		return nil, err
 	}
+	{
```
caseydavenport: Hm, this is a bit complicated just for a log line... I was thinking more just move the log lines right next to the point where we create the watcher - at that point we have the knowledge about whether we're starting a particular watch.

No need to maintain a single log event for this.

lib/backend/k8s/resources/profile.go
```diff
@@ -222,55 +224,101 @@ func (c *profileClient) EnsureInitialized() error {
...
-		"saRev": saRev,
-	}).Debug("Watching two resources at individual revisions")
-
-	nsWatch, err := c.clientSet.CoreV1().Namespaces().Watch(metav1.ListOptions{ResourceVersion: nsRev})
-	if err != nil {
-		return nil, K8sErrorToCalico(err, list)
+	opts.ResourceVersion = nsRev
+	var nsWatch kwatch.Interface = kwatch.NewFake()
+	if namePrefix == "" || namePrefix == conversion.NamespaceProfileNamePrefix {
+		nsWatch, err = c.clientSet.CoreV1().Namespaces().Watch(opts)
```
caseydavenport: This is where I was thinking of putting the log stuff from above

lib/backend/k8s/k8s_test.go
```diff
@@ -1967,6 +1987,227 @@ var _ = testutils.E2eDatastoreDescribe("Test Watch support", testutils.Datastore
...
+		}
+		BeforeEach(func() {
+			createTestPod("test-pod-1")
+			createTestPod("test-pod-2")
+		})
+		AfterEach(func() {
+			deleteAllPods()
+		})
+		It("supports watching a specific workloadEndpoint", func() {
+			watch, err := c.Watch(ctx, model.ResourceListOptions{Name: "127.0.0.1-k8s-test--pod--1-eth0", Kind: apiv3.KindWorkloadEndpoint}, "")
```
caseydavenport: Hm, I'd expect we need to pass in a Namespace for all of the namespaced objects.

caseydavenport: Ah, it's in the default namespace so `""` works fine. nvm.

lib/backend/k8s/resources/workloadendpoint.go
```diff
@@ -219,12 +220,27 @@ func (c *WorkloadEndpointClient) EnsureInitialized() error {
...
-		return nil, fmt.Errorf("cannot watch specific resource instance: %s", list.(model.ResourceListOptions).Name)
+	// Build watch options to pass to k8s.
+	opts := metav1.ListOptions{ResourceVersion: revision, Watch: true}
+	rlo, ok := list.(model.ResourceListOptions)
+	if !ok {
+		return nil, fmt.Errorf("ListInterface is not a ResourceListOptions: %s", list)
+	}
+	if len(rlo.Name) != 0 {
+		if len(rlo.Namespace) == 0 {
+			return nil, errors.New("Cannot watch a specific WorkloadEndpoint without a namespace")
```
jpalardy: Make error start with lowercase?

caseydavenport: Definitely :D

Issue #478, opened 2017-07-21
Persistent Syncer connection failures not exposed for readiness reporting


fasaxc:
If the KDD Syncer's watch ends with a persistent network error then it just keeps retrying (with warning logs) but it doesn't tell its client that it's having trouble.  Ideally, consistent failures would be exposed in some way, so that we can feed them into the healthcheck and report Felix/Typha as not ready.

## Possible Solution
- Could have the Syncer optionally register with the healthcheck
- Could have the Syncer spot consistent failures and either resync or flag it using the status API.

## Steps to Reproduce (for bugs)

1. Let syncer come into sync
2. kill the API server
3. Syncer logs warnings but keeps retrying
4.

## Context
This was desired behaviour for the Felix/Typha readiness feature

## TODO

- [ ] resolve in libcalico-go
- [ ] add k8sfv tests

caseydavenport:
@fasaxc does this affect KDD and etcd or just KDD?

Could this just be triggered by the Syncer indicating "WaitForDatastore"?

fasaxc:
Yeah, that sounds like a good option.

caseydavenport:
Went to have a look a reproducing this today and saw something perhaps interesting (or maybe I'm just tired?)

I noticed that following the repro steps of killing the apiserver _did _ result in Felix going to not-ready in KDD mode. Looking closer though, it only seems to do so after receiving an update over the syncer that says the IPIP tunnel address has been deleted, triggering Felix to restart.

```
2019-03-08 03:10:11.737 [INFO][42] resource.go 260: Target config /etc/calico/confd/config/bird_ipam.cfg has been updated
2019-03-08 03:10:11.730 [INFO][3095] watchercache.go 242: Failed to create watcher ListRoot="/calico/resources/v3/projectcalico.org/nodes" error=Get https://10.96.0.1:443/api/v1/nodes?resourceVersion=5344&watch=true: dial tcp 10.96.0.1:443: connect: connection refused performFullResync=false
2019-03-08 03:10:11.730 [INFO][3095] watchercache.go 242: Failed to create watcher ListRoot="/calico/resources/v3/projectcalico.org/networkpolicies" error=Get https://10.96.0.1:443/apis/networking.k8s.io/v1/networkpolicies?resourceVersion=4985&watch=true: dial tcp 10.96.0.1:443: connect: connection refused performFullResync=false
[...]

PR #1063, opened 2019-03-20
Send WaitForDatastore if there's a watch error

lmm:
## Description

This updates the syncer to be able to report `WaitForDatastore` if any of its `watcherCache`s are not in-sync. Previously, the `watcherSyncer` state only went from `WaitForDatastore` -> `ResyncInProgress` -> `InSync`. This PR adds another state change: `InSync` -> `WaitForDatastore`.

Fixes https://github.com/projectcalico/libcalico-go/issues/478

This depends on a change to confd to handle multiple `InSync` statuses: https://github.com/projectcalico/confd/pull/230

## Todos
- [x] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Improve `calico/node` readiness probe reporting when watches on the datastore error out.
```

### DIFF
--- a/lib/backend/watchersyncer/watchercache.go
+++ b/lib/backend/watchersyncer/watchercache.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2018 Tigera, Inc. All rights reserved.
+// Copyright (c) 2017-2019 Tigera, Inc. All rights reserved.
 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -42,6 +42,7 @@ type watcherCache struct {
 	oldResources         map[string]cacheEntry
 	results              chan<- interface{}
 	hasSynced            bool
+	errors               int
 	resourceType         ResourceType
 	currentWatchRevision string
 }
@@ -49,6 +50,7 @@ type watcherCache struct {
 var (
 	ListRetryInterval = 1000 * time.Millisecond
 	WatchPollInterval = 5000 * time.Millisecond
+	ErrorThreshold    = 15
 )
 
 // cacheEntry is an entry in our cache.  It groups the a key with the last known
@@ -112,6 +114,7 @@ mainLoop:
 				// Handle a WatchError.  First determine if the error type indicates that the
 				// watch has closed, and if so we'll need to resync and create a new watcher.
 				wc.results <- event.Error
+				wc.onError()
 				if e, ok := event.Error.(cerrors.ErrorWatchTerminated); ok {
 					wc.logger.Debug("Received watch terminated error - recreate watcher")
 					if !e.ClosedByRemote {
@@ -121,6 +124,10 @@ mainLoop:
 						wc.logger.Debug("Watch was not closed by remote - full resync required")
 						wc.currentWatchRevision = ""
 					}
+					wc.resyncAndCreateWatcher(ctx)
+				} else if wc.errors > ErrorThreshold {
+					// For any other error, trigger a full resync if we're past the error threshold.
+					wc.currentWatchRevision = ""
 					wc.resyncAndCreateWatcher(ctx)
 				}
 			default:
@@ -186,6 +193,7 @@ func (wc *watcherCache) resyncAndCreateWatcher(ctx context.Context) {
 			if err != nil {
 				// Failed to perform the list.  Pause briefly (so we don't tight loop) and retry.
 				wc.logger.WithError(err).Info("Failed to perform list of current data during resync")
+				wc.onError()
 				select {
 				case <-time.After(ListRetryInterval):
 					continue
@@ -240,6 +248,7 @@ func (wc *watcherCache) resyncAndCreateWatcher(ctx context.Context) {
 			//      watch retry.  This would require some care to ensure the correct errors are captured
 			//      for the different datastore drivers.
 			wc.logger.WithError(err).WithField("performFullResync", performFullResync).Info("Failed to create watcher")
+			wc.onError()
 			performFullResync = true
 			continue
 		}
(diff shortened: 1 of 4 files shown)


lib/backend/watchersyncer/watchersyncer_test.go
```diff
@@ -131,21 +131,30 @@ var _ = Describe("Test the backend datastore multi-watch syncer", func() {
 		By("Driving a bunch of List complete, Watch fail events for 2/3 resource types")
 		expectedDuration := watchersyncer.WatchPollInterval * 2
 		minDuration := 70 * expectedDuration / 100
-		maxDuration := 130 * expectedDuration / 100
+		maxDuration := 140 * expectedDuration / 100
```
lmm: This section is a bit slower now with all of the `ExpectStatusUpdate` calls

lib/backend/watchersyncer/watchersyncer_test.go
```diff
@@ -131,21 +131,30 @@ var _ = Describe("Test the backend datastore multi-watch syncer", func() {
...
+		before := time.Now()
```
lmm: Move this up since at the original location all the events will have been handled while expecting the status updates.

lib/backend/watchersyncer/watchercache.go
```diff
@@ -405,9 +408,11 @@ func (wc *watcherCache) markAsValid(resourceKey string) {
 	}
 }
 
-// sendWaitForDatastore signals to the syncer that this watcherCache is not in sync.  See finishResync()
+// sendWaitForDatastore signals to the syncer that this watcherCache is not in-sync if the number of consecutive errors
+// exceeds the error threshold.  See finishResync() for how the watcherCache goes back to in-sync.
 func (wc *watcherCache) sendWaitForDatastore() {
-	if wc.hasSynced {
+	wc.errors++
+	if wc.hasSynced && wc.errors > ErrorThreshold {
```
jpalardy: should that be `>=`

lib/backend/watchersyncer/watchercache.go
```diff
@@ -401,3 +407,13 @@ func (wc *watcherCache) markAsValid(resourceKey string) {
 		}
 	}
 }
+
+// sendWaitForDatastore signals to the syncer that this watcherCache is not in-sync if the number of consecutive errors
+// exceeds the error threshold.  See finishResync() for how the watcherCache goes back to in-sync.
+func (wc *watcherCache) sendWaitForDatastore() {
```
caseydavenport: nit: if this is called `sendWaitForDatastore()` I'd expect it to send a wait for datastore event every time it is called.

I think it would be good to rename this to something like `onError()` or something, which makes it more clear _when_ this function is meant to be called (when we get an error).

lib/backend/watchersyncer/watchercache.go
caseydavenport: Don't think it makes a huge difference - 3 is sort of a magic number. Could just as well be 4.

lib/backend/watchersyncer/watchersyncer_test.go
```diff
@@ -98,6 +98,51 @@ var _ = Describe("Test the backend datastore multi-watch syncer", func() {
...
+		rs.clientListResponse(r1, genError)
+		rs.ExpectStatusUnchanged()
+		rs.clientListResponse(r1, genError)
+		rs.ExpectStatusUpdate(api.WaitForDatastore)
+
+		// Now get back in-sync and assert that multiple errors up to the threshold does not update status.
+		rs.clientListResponse(r1, emptyList)
+		rs.ExpectStatusUpdate(api.ResyncInProgress)
+		rs.ExpectStatusUpdate(api.InSync)
+		rs.clientWatchResponse(r1, genError)
```
caseydavenport: Might be nice to encode this into a loop based off `ErrorThreshold` so that we don't need to modify this test if we modify that threshold.

lib/backend/watchersyncer/watchersyncer_test.go
```diff
@@ -127,25 +172,36 @@ var _ = Describe("Test the backend datastore multi-watch syncer", func() {
...
 		By("Driving a bunch of List complete, Watch fail events for 2/3 resource types")
 		expectedDuration := watchersyncer.WatchPollInterval * 2
 		minDuration := 70 * expectedDuration / 100
-		maxDuration := 130 * expectedDuration / 100
+		maxDuration := 150 * expectedDuration / 100
+		before := time.Now()
 		rs.clientListResponse(r1, emptyList)
 		rs.ExpectStatusUpdate(api.ResyncInProgress)
 		rs.clientWatchResponse(r1, genError)
+		rs.ExpectStatusUnchanged()
```
caseydavenport: So I understand, were these test changes required? Or were they added to test the new functionality?

It seems like this might be covered in the new tests you added above, so wondering what this change is for.

lib/backend/watchersyncer/watchercache.go
caseydavenport: Looks like our retry interval is 1s right now, meaning that:
- one watch failure + one list failure
- one list failure
- one list failure

Will take 3 seconds total to hit 4 retries and trigger this behavior. I think we probably want to bump this number up - thinking something more like 15 seconds (retries) should count as a persistent error.

lib/backend/watchersyncer/watchersyncer_test.go
lmm: The `ExpectStatusUpdate` ones are required for any status change but you're right the `ExpectStatusUnchanged` calls aren't required and aren't needed here as that's tested above.

lib/backend/watchersyncer/watchercache.go
```diff
@@ -122,6 +125,12 @@ mainLoop:
 						wc.currentWatchRevision = ""
 					}
 					wc.resyncAndCreateWatcher(ctx)
+				} else {
```
caseydavenport: nit: should just do an `else if` here rather than splitting onto two lines.

caseydavenport:
@lmm could you squash your commits? Then I'll merge. Thanks!

Issue #973, opened 2018-11-16
Hard to diagnose if when health checks fail


fasaxc:
The health aggregator is quite chatty about the current state, logging `summary=&health.HealthReport{Live:false, Ready:false}` whenever the health is queried.  However, it doesn't log which component is reporting unhealthly.

## Expected Behavior

Think the ideal behaviour would be:

- Always log any unhealthy components when the status is calculated.
- Only log the status itself if it has changed since the last calculation.

## Current Behavior

- Status is logged but per-component status is only logged at debug level.

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Your Environment

* libcalico-go version: 
* Operating System and version:
* Link to your project (optional):

caseydavenport:
We get _some_ output from `kubectl describe`, but for Felix that output is typically just "It failed" (the BIRD output is usually much more detailed and useful to the user).

Felix usually spits out something like this:
```
  Warning  Unhealthy              10s   kubelet, casey-crc-kadm-ms  Readiness probe failed: calico/node is not ready: felix is not ready: readiness probe reporting 503
```

It would be awesome if we could expose this sort of information through kubectl describe as well as part of this.

PR #1064, opened 2019-03-25
Improve health reporting

lmm:
## Description

Fixes https://github.com/projectcalico/libcalico-go/issues/973

The `HealthAggregator` now only logs the overall health report if it has changed and unhealthy reporters are logged at WARN.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Health aggregator now only logs if overall health status has changed. Details of unhealthy reporters are now logged.
```

### DIFF
--- a/lib/health/health.go
+++ b/lib/health/health.go
@@ -30,6 +30,9 @@ type HealthReport struct {
 }
 
 type reporterState struct {
+	// The reporter's name.
+	name string
+
 	// The health indicators that this reporter reports.
 	reports HealthReport
 
@@ -43,7 +46,7 @@ type reporterState struct {
 	timestamp time.Time
 }
 
-// TimedOut checks whether the reporterState is due for another report. This is the case when
+// TimedOut checks whether the reporter is due for another report. This is the case when
 // the reports are configured to expire and the time since the last report exceeds the report timeout duration.
 func (r *reporterState) TimedOut() bool {
 	return r.timeout != 0 && time.Since(r.timestamp) > r.timeout
@@ -79,6 +82,7 @@ func (aggregator *HealthAggregator) RegisterReporter(name string, reports *Healt
 	aggregator.mutex.Lock()
 	defer aggregator.mutex.Unlock()
 	aggregator.reporters[name] = &reporterState{
+		name:      name,
 		reports:   *reports,
 		timeout:   timeout,
 		latest:    HealthReport{Live: true},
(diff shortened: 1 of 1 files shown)


lib/health/health.go
```diff
@@ -122,32 +126,53 @@ func (aggregator *HealthAggregator) Summary() *HealthReport {
...
+				logEntry.Debug("Healthy reporter")
+			})
+		}
 	}
 
-	log.WithField("summary", summary).Info("Overall health")
+	// Summary status has changed so update previous status and log.
+	if summary.Live != aggregator.lastReport.Live && summary.Ready != aggregator.lastReport.Ready {
+		aggregator.lastReport = summary
+		log.WithField("lastSummary", summary).Info("Overall health")
```
fasaxc: Now you've put this behind a check, how about changing the log to say "Overall health status changed"

lib/health/health.go
```diff
@@ -122,32 +126,53 @@ func (aggregator *HealthAggregator) Summary() *HealthReport {
...
 			summary.Ready = false
 		}
+
+		logEntry := log.WithFields(log.Fields{
+			"name":           name,
+			"reporter-state": reporter,
+		})
+
+		if reporter.reports.Live && notReportedLive || reporter.reports.Ready && notReportedReady {
+			logFuncs = append(logFuncs, func() {
```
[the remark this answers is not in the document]
lmm: I think it'll be nicer to split up the failing liveness vs readiness checks as you're hinting here too.

lib/health/health.go
```diff
@@ -122,32 +132,59 @@ func (aggregator *HealthAggregator) Summary() *HealthReport {
...
-			(reporter.timeout != 0 && time.Since(reporter.timestamp) > reporter.timeout)) {
+		stillReady := reporter.latest.Ready && !reporter.TimedOut()
+		if summary.Ready && reporter.reports.Ready && !stillReady {
 			summary.Ready = false
 		}
+
+		switch {
+		case reporter.reports.Live && !stillLive:
+			failedLivenessChecks = append(failedLivenessChecks, name)
+		case reporter.reports.Ready && !stillReady:
```
fasaxc: This case can only be triggered if the one above is not so a reporter that is neither live nor ready will only show up as not live.  I guess that's OK since liveness is more severe than readiness but for debug purposes, might be better to have both bits of information.

lib/health/health.go
```diff
@@ -122,32 +132,59 @@ func (aggregator *HealthAggregator) Summary() *HealthReport {
...
+				"reporter-state": reporter,
+			}).Debug("Reporter is healthy")
+		}
 	}
 
-	log.WithField("summary", summary).Info("Overall health")
+	// Summary status has changed so update previous status and log.
+	if summary.Live != aggregator.lastReport.Live || summary.Ready != aggregator.lastReport.Ready {
+		aggregator.lastReport = summary
+		log.WithField("lastSummary", summary).Info("Overall health status changed")
```
fasaxc: It hink you're logging out the current summary, here, "lastSummary" hints that it's the previous value.

lib/health/health.go
```diff
@@ -122,32 +132,59 @@ func (aggregator *HealthAggregator) Summary() *HealthReport {
...
+
+		for _, name := range failedLivenessChecks {
```
fasaxc: I was thinking of logging a single log with a list of reporter names without the detailed state of each reporter.  I guess logging out each reporter gives us a lot more to go on, though so let's keep your approach of one log per reporter.  If you're going to do one log per reporter though, I'd suggest making   `failedLivenessChecks` a `[]*reporterState` so that you can just append the reporter instead of its name.  They you won't need to look it up again.

PR #1065, opened 2019-03-26
Datamodel for VXLAN encapsulation support 

caseydavenport:
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
VXLAN encapsulation can now be configured per IP pool
```

### DIFF
--- a/lib/apis/v3/node.go
+++ b/lib/apis/v3/node.go
@@ -42,6 +42,9 @@ type NodeSpec struct {
 	// BGP configuration for this node.
 	BGP *NodeBGPSpec `json:"bgp,omitempty" validate:"omitempty"`
 
+	// IPv4VXLANTunnelAddr is the IPv4 address of the VXLAN tunnel.
+	IPv4VXLANTunnelAddr string `json:"ipv4VXLANTunnelAddr,omitempty" validate:"omitempty,ipv4"`
+
 	// OrchRefs for this node.
 	OrchRefs []OrchRef `json:"orchRefs,omitempty" validate:"omitempty"`
 }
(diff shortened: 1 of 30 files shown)


lib/validator/v3/validator.go
```diff
@@ -123,6 +124,7 @@ func init() {
 	registerFieldValidator("labels", validateLabels)
 	registerFieldValidator("ipVersion", validateIPVersion)
 	registerFieldValidator("ipIpMode", validateIPIPMode)
+	registerFieldValidator("vxlanMode", validateVXLANMode)
```
fasaxc: Maybe add a validator test?

lib/apis/v3/node.go
```diff
@@ -67,6 +67,8 @@ type NodeBGPSpec struct {
 	IPv6Address string `json:"ipv6Address,omitempty" validate:"omitempty,cidrv6"`
 	// IPv4IPIPTunnelAddr is the IPv4 address of the IP in IP tunnel.
 	IPv4IPIPTunnelAddr string `json:"ipv4IPIPTunnelAddr,omitempty" validate:"omitempty,ipv4"`
+	// IPv4IPIPTunnelAddr is the IPv4 address of the VXLAN tunnel.
+	IPv4VXLANTunnelAddr string `json:"ipv4VXLANTunnelAddr,omitempty" validate:"omitempty,ipv4"`
```
lmm: You missed paste update here 😄

caseydavenport:
OK - I think this is good to go.

There is an outstanding question about whether we need an additional field for the node IP address that isn't a subfield of the BGP struct, since VXLAN requires the node IP but doesn't use BGP.

I think we can merge as-is though, and then deliberate on whether we want to do anything about that afterwards. 

One option might to add a new top-level field and deprecate the BGP one, and then switching clients over to using the new field for both BGP and VXLAN, falling back on the BGP field if the top-level isn't present.

PR #1070, opened 2019-04-15
Repro failure to watch block.

fasaxc:
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

PR #1071, opened 2019-04-15
Fix calculation of deafult path for IPAM blocks

caseydavenport:
- [ ] Tests
- [ ] Documentation
- [ ] Release note

```release-note
None required
```


caseydavenport:
Includes tests from https://github.com/projectcalico/libcalico-go/pull/1070

fasaxc:
Think I find the root cause of this issue; I've put up a more generic fix: #1072

PR #1072, opened 2019-04-16
Fix watching of IPAM blocks

fasaxc:
## Description

etcd's list and watch methods were using different logic for calculating the key.  The list routine was tolerant to a `/` at the end of the default path but the watch routine was not.  Make the logic common and handle an extra `/` in both cases.  Improve logging to log out the actual key that we watch on rather than the key that we're about to add a `/` to!

## Todos
- [x] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


song-jiang:
@caseydavenport would be a better person to review `Fix parsing of etcd deletion events to return KV.` change.

PR #1075, opened 2019-04-18
Improve health report diags

fasaxc:
## Description

- Log every update at debug.
- Log warnings on every poll for non-live and non-ready reporters.
- Log timeouts.
- Move readiness calculation into a method for simplicity.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/health/health.go
+++ b/lib/health/health.go
@@ -46,26 +46,26 @@ type reporterState struct {
 	timestamp time.Time
 }
 
-func (r *reporterState) RecentlyReady() bool {
+func (r *reporterState) HasReadinessProblem() bool {
 	if !r.reports.Ready {
-		return true
+		return false
 	}
 	if r.TimedOut() {
 		log.WithField("name", r.name).Warn("Report timed out")
-		return false
+		return true
 	}
-	return r.latest.Ready
+	return !r.latest.Ready
 }
 
-func (r *reporterState) RecentlyLive() bool {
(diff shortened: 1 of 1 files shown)


lib/health/health.go
```diff
@@ -136,60 +173,30 @@ func (aggregator *HealthAggregator) Summary() *HealthReport {
...
 	summary := &HealthReport{Live: true, Ready: true}
 
 	// Now for each reporter...
 	for _, reporter := range aggregator.reporters {
-		// Reset Live to false if that reporter is registered to report liveness and hasn't
-		// recently said that it is live.
-		stillLive := reporter.latest.Live && !reporter.TimedOut()
-		if summary.Live && reporter.reports.Live && !stillLive {
+		log.WithField("reporter", reporter).Debug("Checking state of reporter")
+		if !reporter.RecentlyLive() {
```
neiljerram: With this polarity and naming, it feels confusing to me that `reporter.RecentlyLive()` will be `true` if the reporter doesn't report liveness.  I think I would prefer one of these options:

1. Keep the `reporter.reports.Live` condition here, outside of `reporter.RecentlyLive()`.

2. Invert the polarity and naming, to (e.g.) `reporter.HasLivenessProblem()`.

And obviously similarly for readiness.

fasaxc: (2) sounds good to me, reads very nicely

PR #1077, opened 2019-04-22
Reinstate host-local IPAM aggregation

caseydavenport:
We removed host-local IPAM aggregation in v3.6 when we switched KDD mode
over to use Calico IPAM. However, there are some scenarios where we may
still want to use host-local IPAM with BGP.

This PR reinstates the ability to use host-local IPAM with Calico BGP,
and get proper route aggregation.

To enable it, make sure you're using the KDD backend and set `USE_POD_CIDR=true`. 

This PR adds back support for IPIP mode but doesn't attempt to implement VXLAN support. 

- [ ] Tests
- [ ] Documentation
- [ ] Release note

```release-note
Reinstate BGP route aggregation when using host-local IPAM via config option USE_POD_CIDR
```

### DIFF
--- a/lib/backend/syncersv1/felixsyncer/felixsyncerv1.go
+++ b/lib/backend/syncersv1/felixsyncer/felixsyncerv1.go
@@ -15,6 +15,7 @@
 package felixsyncer
 
 import (
+	"github.com/projectcalico/libcalico-go/lib/apiconfig"
 	apiv3 "github.com/projectcalico/libcalico-go/lib/apis/v3"
 	"github.com/projectcalico/libcalico-go/lib/backend/api"
 	"github.com/projectcalico/libcalico-go/lib/backend/model"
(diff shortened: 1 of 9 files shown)


lib/backend/syncersv1/felixsyncer/felixsyncerv1.go
```diff
@@ -72,9 +73,11 @@ func New(client api.Client, callbacks api.SyncerCallbacks) api.Syncer {
 			ListInterface:   model.ResourceListOptions{Kind: apiv3.KindHostEndpoint},
 			UpdateProcessor: updateprocessors.NewHostEndpointUpdateProcessor(),
 		},
-		{
-			ListInterface: model.BlockListOptions{},
-		},
+	}
+
+	// If using Calico IPAM, include IPAM resources the felix cares about.
```
caseydavenport: We could alternatively implement a dummy version of IPAM blocks so that we don't need this extra plumbing. The main thing is that Felix tries to list/watch IPAM blocks and those don't exist when using host-local. 

I don't have a strong preference on how we do this - @fasaxc WDYT?

caseydavenport:
Have tests locally, working on the felix syncer ones before pushing.

fasaxc:
Does this basically get us back to where we were before?  I.e. were we polling (rather than watching) the local node before?

caseydavenport:
> Does this basically get us back to where we were before? I.e. were we polling (rather than watching) the local node before?

Yeah, I hadn't spotted that, and it's a bit surprising, but that file is more-or-less copy/paste from what we had before (with a few naming changes), so I suppose we were just polling before!

Should be easy to add the Watch in though if we want it.

PR #1084, opened 2019-05-14
Setting default to Never for VXLAN and IpIpMode

asincu:
## Description
When applying an ip pool policy with IPV6 CIDR and no VXLAN mode selected via `calicoctl apply -f`
```
---
kind: IPPool
metadata:
  name: ippool-2
spec:
  cidr: 2002::/64
  ipipMode: Never
  natOutgoing: true
```
Calicoctl crashes with `(VXLANMode other than 'Never' is not supported on an IPv6 IP pool)`

## Todos
- [ ] Tests

### DIFF
--- a/lib/apis/v3/ippool.go
+++ b/lib/apis/v3/ippool.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2017 Tigera, Inc. All rights reserved.
+// Copyright (c) 2017,2019 Tigera, Inc. All rights reserved.
 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


lib/apis/v3/ippool.go
```diff
@@ -45,7 +45,7 @@ type IPPoolSpec struct {
 
 	// Contains configuration for VXLAN tunneling for this pool. If not specified,
 	// then this is defaulted to "Never" (i.e. VXLAN tunelling is disabled).
-	VXLANMode VXLANMode `json:"vxlanMode,omitempty" validate:"omitempty,vxlanMode"`
+	VXLANMode VXLANMode `json:"vxlanMode,omitempty" validate:"omitempty,vxlanMode" default:"Never"`
```
caseydavenport: @asincu What is interpreting this tag? I don't see any other fields that have this on them

jpalardy: There's a bit of indirection here; it will get picked up by `go-defaults` ([context](https://github.com/projectcalico/calicoctl/pull/2006/files)).

As per your comment:

> I'm a little bit hesitant to introduce a new library here [...]

it certainly doesn't feel worthwhile if we do this only for this one property.

asincu: @jpalardy @caseydavenport I would propose that we extend the usage of this tag for all fields that require a different default than the ones that go provides.

tomastigera: If there are other use cases, converting them now (in a follow up commit) is an option, but the cool thing is that you do not need to, you can start using it gradually.

There are certainly other cases that might benefit from a reasonable default, e.g. policy order to be set to 0 instead of nil (i just noticed last week). That would also make the calicoctl output nicer, however, is is easily doable using this library?

asincu: @caseydavenport @jpalardy @tomastigera I have added default never to IpIpMode in order for the additional test to pass. Please see https://github.com/projectcalico/calicoctl/pull/2006/commits/84103203b48bb07011cdf44ffe2638b5408eca46

asincu:
Closing this as we decided to go with https://github.com/projectcalico/calicoctl/pull/2015 and https://github.com/projectcalico/libcalico-go/pull/1088

PR #1088, opened 2019-05-28
Update should set defaults before doing validation

asincu:
## Description
IpPool update operation should set defaults before performing validation. The validation procedure is an opinionated one that requires that certain have default values.

This is required to execute commands like update and delete for resources that were created with previous versions.
This relates to: https://github.com/projectcalico/calico/issues/2628

This also impacts operations like `calicoctl apply/delete -f` what will rely on update/create/delete operations after unmarshalling the resources. See https://github.com/projectcalico/calicoctl/pull/2015

### DIFF
--- a/lib/clientv3/ippool.go
+++ b/lib/clientv3/ippool.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2018 Tigera, Inc. All rights reserved.
+// Copyright (c) 2017-2019 Tigera, Inc. All rights reserved.
 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


lib/clientv3/ippool.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2018 Tigera, Inc. All rights reserved.
+// Copyright (c) 2017-2018,2019 Tigera, Inc. All rights reserved.
```
caseydavenport: minor thing, but I think we should just say 2017-2019 here

asincu: @caseydavenport I have updated the copyright.

PR #1089, opened 2019-05-29
Support for supplying NATOutgoingAddress to Felix

maxstr:
## Description
This is a PR required for addressing https://github.com/projectcalico/calico/issues/2222. Addressing the issue also requires a change in Felix which I will link below. We chose to implement this by allowing users to specify the IP address being used rather than using the BGP address by default because it seemed more flexible. For our use case we are using the Kubernetes downward API to supply this address. I've tested this by setting/unsetting `FELIX_NATOUTGOINGADDRESS` in a calico node container using our modified felix. 

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/apis/v3/felixconfig.go
+++ b/lib/apis/v3/felixconfig.go
@@ -220,6 +220,11 @@ type FelixConfigurationSpec struct {
 	// network stack is used.
 	NATPortRange *numorstring.Port `json:"natPortRange,omitempty"`
 
+	// NATOutgoingAddress specifies an address to use when performing source NAT for traffic in a natOutgoing pool that
+	// is leaving the network. By default the address used is an address on the interface the traffic is leaving on
+	// (ie it uses the iptables MASQUERADE target)
+	NATOutgoingAddress string `json:"NATOutgoingAddress,omitempty"`
+
 	// ExternalNodesCIDRList is a list of CIDR's of external-non-calico-nodes which may source tunnel traffic and have
 	// the tunneled traffic be accepted at calico nodes.
 	ExternalNodesCIDRList *[]string `json:"externalNodesList,omitempty"`
--- a/lib/backend/syncersv1/updateprocessors/configurationprocessor_test.go
+++ b/lib/backend/syncersv1/updateprocessors/configurationprocessor_test.go
@@ -70,7 +70,7 @@ var _ = Describe("Test the generic configuration update processor and the concre
 		Kind: apiv3.KindBGPConfiguration,
 		Name: "node.bgpnode1",
 	}
-	numFelixConfigs := 63
+	numFelixConfigs := 64
 	numClusterConfigs := 4
 	numNodeClusterConfigs := 3
 	numBgpConfigs := 3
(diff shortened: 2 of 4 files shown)


lib/apis/v3/felixconfig.go
```diff
@@ -220,6 +220,11 @@ type FelixConfigurationSpec struct {
 	// network stack is used.
 	NATPortRange *numorstring.Port `json:"natPortRange,omitempty"`
 
+	// NATOutgoingAddress specifies an address to use when performing source NAT for traffic in a natOutgoing pool that
+	// is leaving the network. By default the address used is an address on the interface the traffic is leaving on
+	// (ie it uses the iptables MASQUERADE target)
+	NATOutgoingAddress string `json:"NATOutgoingAddress,omitempty"`
```
neiljerram: Did you see the thoughts from @caseydavenport at https://github.com/projectcalico/calico/issues/2222#issuecomment-484292225 ?  If we can meet the need with a global config setting for the interface name or CIDR, that would be nicer than having to set node-specific config for each node in the cluster.

neiljerram:
OK, to follow up on my previous comment...  I see now that:

- You did the Felix change first, using the downward API to set the config in the environment, and that gives you a complete working solution.
- You then added this PR, following @fasaxc's request based on a general desire we have for the data model to be in sync with Felix configuration options.  But you're not relying on the data model for your own solution, hence not picking up that the PR here is incomplete.
- In addition, there are considerations that if we're going to have data model for this, perhaps there are better options.

Let me think through a bit more what the upshot of all that is.  Will write again shortly.

lib/apis/v3/felixconfig.go
maxstr: Will take a look at the missing conversion code momentarily.

Unfortunately I did not see that comment and while implementing was only considering between using a flag to use the BGP address or passing a specific IP. For us, either would work, but being able to pass a specific IP seemed more flexible.

neiljerram:
@maxstr I'm sorry, my comment above about IPPool and missing conversion code was complete rubbish.  Of course the new field is in FelixConfiguration, not IPPool, and there is no per-field conversion code needed for FelixConfiguration.

Also, regarding @caseydavenport 's comment that I pointed to... given that these PRs already Just Work for you, with the downward API, I think we should bank them as they are, and NATOutgoingCIDR or NATOutgoingInterface can be added by further enhancements later if there is sufficient motivation for them.

So, this PR is good to go and I'll now merge it.

neiljerram:
@maxstr We'll also need a docs PR to add the new field at https://github.com/projectcalico/calico/blob/master/master/reference/felix/configuration.md and https://github.com/projectcalico/calico/blob/master/master/reference/resources/felixconfig.md.  Would you like to do that?  I'm happy to do it if you prefer.  It's just a matter of credit where credit's due, I think, but we do need it in the next day or two, in time for the upcoming Calico 3.8 release.

maxstr:
Thanks for the merge and the review! Will try and get you a PR by EOD for the doc updates.

caseydavenport:
Was there ever a docs PR merged for this feature?

caseydavenport:
Nvm, I see that it is present in the reference material for v3.8.

PR #1100, opened 2019-06-13
felixsyncer_e2e_test.go: don't leak Node changes into other tests

neiljerram:
Specifically, the VXLANTunnelMACAddr setting here could cause an
unexpected HostConfig key, in other tests that start up a Felix syncer
and expect to see a specific set of initial updates from it.

### DIFF
--- a/lib/backend/syncersv1/felixsyncer/felixsyncer_e2e_test.go
+++ b/lib/backend/syncersv1/felixsyncer/felixsyncer_e2e_test.go
@@ -19,9 +19,10 @@ import (
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
 	"github.com/projectcalico/libcalico-go/lib/backend/encap"
 	"github.com/projectcalico/libcalico-go/lib/backend/syncersv1/felixsyncer"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/projectcalico/libcalico-go/lib/apiconfig"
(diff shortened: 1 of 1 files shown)


lib/backend/syncersv1/felixsyncer/felixsyncer_e2e_test.go
```diff
@@ -155,7 +183,20 @@ var _ = testutils.E2eDatastoreDescribe("Felix syncer tests", testutils.Datastore
...
+						node, err = c.Nodes().Get(ctx, "127.0.0.1", options.GetOptions{})
+						Expect(err).NotTo(HaveOccurred())
+						node.Spec.BGP = oldBGPSpec
+						node.Spec.VXLANTunnelMACAddr = oldVXLANTunnelMACAddr
+						node, err = c.Nodes().Update(ctx, node, options.SetOptions{})
+						if err == nil {
+							break
+						}
+					}
+					Expect(err).NotTo(HaveOccurred())
```
song-jiang: Maybe I did not spot it last time. So What is the purpose of retry if there is `Expect(err).NotTo(HaveOccurred())`

neiljerram: It checks that it succeeded after 1-5 attempts.

song-jiang: My understanding is once expectation failed, ginkgo will start to complete the test case and run `after Each`, so it probably won't retry 5 times. Am I right?

song-jiang: Oh, I see what you mean. Understood now.

PR #1102, opened 2019-07-03
Add config option for iptables backend

tmjd:
## Release Note

```release-note
None required
```

### DIFF
--- a/lib/apis/v3/felixconfig.go
+++ b/lib/apis/v3/felixconfig.go
@@ -19,9 +19,13 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+type IptablesBackend string
+
 const (
 	KindFelixConfiguration     = "FelixConfiguration"
 	KindFelixConfigurationList = "FelixConfigurationList"
+	IptablesBackendLegacy      = "Legacy"
+	IptablesBackendNFTables    = "NFT"
 )
 
 // +genclient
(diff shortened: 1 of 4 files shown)


lib/validator/v3/validator.go
```diff
@@ -278,6 +279,12 @@ func validateMAC(fl validator.FieldLevel) bool {
 	return true
 }
 
+func validateIptablesBackend(fl validator.FieldLevel) bool {
+	s := fl.Field().String()
+	log.Debugf("Validate Iptables Backend: %s", s)
+	return s == "" || s == "nft" || s == "legacy"
```
caseydavenport: For k8s resources the convention is that named values start with an uppercase.

e.g.,

`iptablesBackend: Legacy` or `iptablesBackend: NFTables` 

We probably want to use a type with some constants for this as well.

```
type IPTablesBackend string
```

lib/apis/v3/felixconfig.go
```diff
@@ -246,6 +246,9 @@ type FelixConfigurationSpec struct {
 	// modes can use XDP. This is not recommended since it doesn't provide better performance than
 	// iptables. [Default: false]
 	GenericXDPEnabled *bool `json:"genericXDPEnabled,omitempty" confignamev1:"GenericXDPEnabled"`
+
+	// IPtablesBackend
+	IptablesBackend string `json:"iptablesBackend,omitempty" validate:"omitempty,iptablesBackend"`
```
caseydavenport: I believe it should be `IPTablesBackend`

tmjd: I'm fine with that but I feel like the command is always `iptables`, no capitalization, so I thought just the 1st I should be capitalized (even though I was inconsistent with that).

caseydavenport: Yeah, I can see that. I guess it doesn't matter much.

lib/apis/v3/felixconfig.go
```diff
@@ -246,6 +250,9 @@ type FelixConfigurationSpec struct {
 	// modes can use XDP. This is not recommended since it doesn't provide better performance than
 	// iptables. [Default: false]
 	GenericXDPEnabled *bool `json:"genericXDPEnabled,omitempty" confignamev1:"GenericXDPEnabled"`
+
+	// IPTablesBackend specifies which backend of iptables will be used. The default is legacy.
+	IPTablesBackend *IPTablesBackend `json:"iptablesBackend,omitempty" validate:"omitempty,iptablesBackend"`
```
fasaxc: There are other iptables config params further up, I think this should be in the same "block" as those and it should follow the same casing.  For better or worse, the existing ones are "IptablesXXX"  (Think I followed the casing used by the iptables team, but I can't say it looks good!)

PR #1107, opened 2019-07-12
Remove NotNets array initialization

rikatz:
Signed-off-by: Ricardo Pchevuzinske Katz <ricardo.katz@serpro.gov.br>

## Description

- Type of fix - This is a bugfix

This PR removes the NotNets initialization into the ``k8sPeerToCalicoFields`` function.

Removing this doesn't make the Policy Controller to always report changed Network Policies and reconcile them, as reported here: https://github.com/projectcalico/kube-controllers/issues/380

- Fixes: #projectcalico/380

## Todos
- [X] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Fix that kube-controllers would reconcile policies unnecessarily.
```


rikatz:
Sure. I'll try to take a look into that later, at least to make a unit test against ``k8sPeerToCalicoFields`` 

There're also some other places in the code that initialize the slice, do you think this might be something we can open a issue to be reviewed later?

rikatz:
@fasaxc Not the best way but...I've inserted some more tests into the converter_test.go to reflect the situation when no Except field is specified (and a double verification of the equality with a nul array, and the not equality with an empty array).

Have tested this against a non patched lib and the tests failed. After I've patched the converter.go code (as this PR does) the test passes successfully, so I think this is enough to test the lib :D

Tks

lib/backend/k8s/conversion/conversion_test.go
```diff
@@ -1645,6 +1645,99 @@ var _ = Describe("Test NetworkPolicy conversion", func() {
...
+
+		// Assert key fields are correct.
+		Expect(pol.Key.(model.ResourceKey).Name).To(Equal("knp.default.test.policy"))
+
+		// Assert value fields are correct.
+		Expect(int(*pol.Value.(*apiv3.NetworkPolicy).Spec.Order)).To(Equal(1000))
+		Expect(pol.Value.(*apiv3.NetworkPolicy).Spec.Selector).To(Equal("projectcalico.org/orchestrator == 'k8s'"))
+		Expect(len(pol.Value.(*apiv3.NetworkPolicy).Spec.Ingress)).To(Equal(1))
+		Expect(pol.Value.(*apiv3.NetworkPolicy).Spec.Ingress[0].Source.Nets[0]).To(Equal("192.168.0.0/16"))
+		Expect(pol.Value.(*apiv3.NetworkPolicy).Spec.Ingress[0].Source.NotNets).NotTo(Equal([]string{}))
```
fasaxc: These two lines can be written as `Expect(pol.Value.(*apiv3.NetworkPolicy).Spec.Ingress[0].Source.Nets[0]).To(BeNil())`

lib/backend/k8s/conversion/conversion_test.go
```diff
@@ -1645,6 +1645,99 @@ var _ = Describe("Test NetworkPolicy conversion", func() {
...
+		Expect(pol.Value.(*apiv3.NetworkPolicy).Spec.Ingress[0].Source.NotNets).To(Equal(nilNotNets))
+
+		// There should be no Egress rules.
+		Expect(len(pol.Value.(*apiv3.NetworkPolicy).Spec.Egress)).To(Equal(0))
```
fasaxc: We get better diags if you use the `HaveLen()` matcher:
```
Expect(pol.Value.(*apiv3.NetworkPolicy).Spec.Egress).To(HaveLen(0))
```

rikatz: Copied/pasted from other tests :D Will improve this here (and maybe in the test I've copied/pasted)

rikatz:
> Thanks; a couple of tips on how to best use the Gomega matchers inline.
> 
> I think the ideal way to test this would be to test the `k8sPeerToCalicoFields` method on its own. Then you could remove all the extra assertions about fields that we're not really interested in in this test. WDYT?
> 
> I.e. something like this:
> 
> ```
> sel, nsSel, nets, notNets := c.k8sPeerToCalicoFields(&networkingv1.NetworkPolicyPeer{
> ...}, "a-namespace")
> Expect(sel).To(Equal(...))
> ...
> Expect(notNets).To(BeNil())
> ```

I'm working on this right now :) I'm thinking about keeping the extra tests (tests are never too much) and make a simple Conversion test suite for this only method, is that ok?

fasaxc:
You're right that having all the tests covers more bases, but there's a cost to having more tests too: 

- more places to update when we change something
- less targeted tests, so more time to diagnose when they fail
- less precise tests, so easy for a maintainer to miss the intent of the test and accidentally "fix" the test when they should fix the code.

rikatz:
@fasaxc So may I remove the bigger duplicate tests and keep the 3 small one directed to k8sPeerToCalicoFields?

rikatz:
I've removed the bigger tests and keeped the unit test of k8sPeerToCalicoFields :)

lib/backend/k8s/conversion/conversion_test.go
```diff
@@ -1639,13 +1639,96 @@ var _ = Describe("Test NetworkPolicy conversion", func() {
...
+		}
+
+		// Parse the policy.
+		podSel, nsSel, nets, notNets := c.k8sPeerToCalicoFields(&np, "default")
+
+		// Assert value fields are correct.
+		Expect(nets[0]).To(Equal("192.168.0.0/16"))
+		Expect(notNets).To(BeNil())
+
+		// As this is a IPBlock rule, podSel and nsSel should be empty (not nil!)
```
fasaxc: strings can't be nil, their "zero value" is the empty string.

lib/backend/k8s/conversion/conversion_test.go
```diff
@@ -1639,13 +1639,96 @@ var _ = Describe("Test NetworkPolicy conversion", func() {
...
+		Expect(nets).To(BeNil())
+		Expect(notNets).To(BeNil())
+
+		// As this is a IPBlock rule, podSel and nsSel should be empty (not nil!)
+		Expect(podSel).To(BeEmpty())
+		Expect(nsSel).To(BeEmpty())
+
+	})
+
+	It("should have empty and Nil NetworkPolicyPeer fields when only Except is declared", func() {
```
fasaxc: I don't think it's valid to have an except with no cidr.  If it is valid then I would have expected it to return a valid notNets (as if cidr was set to 0.0.0.0/0).  Maybe just leave out this case?

rikatz: Yes, removing :)

lib/backend/k8s/conversion/conversion_test.go
rikatz: Yes, but in this case I'm checking against the array and not directly into the string :)

rikatz:
@fasaxc do I have to change anything else? :)

lib/backend/k8s/conversion/conversion_test.go
rikatz: I've rewriten as suggested

PR #1110, opened 2019-07-18
Switch to go modules from glide

rafaelvanoni:


lib/backend/k8s/k8s_test.go
```diff
@@ -2420,7 +2420,6 @@ var _ = testutils.E2eDatastoreDescribe("Test Watch support", testutils.Datastore
 				case e := <-watch.ResultChan():
 					// Got an event. Check it's OK.
 					Expect(e.Error).NotTo(HaveOccurred())
-					Expect(e.Type).To(Equal(api.WatchAdded))
```
fasaxc: Why this change?

go.mod
```diff
@@ -0,0 +1,66 @@
...
+	github.com/pquerna/ffjson v0.0.0-20181028064349-e517b90714f7 // indirect
+	github.com/projectcalico/go-json v0.0.0-20161128004156-6219dc7339ba // indirect
+	github.com/projectcalico/go-yaml v0.0.0-20161201183616-955bc3e451ef // indirect
+	github.com/projectcalico/go-yaml-wrapper v0.0.0-20161127220527-598e54215bee
+	github.com/prometheus/client_golang v0.0.0-20171005112915-5cec1d0429b0
+	github.com/prometheus/client_model v0.0.0-20170216185247-6f3806018612 // indirect
+	github.com/prometheus/common v0.0.0-20171104095907-e3fb1a1acd76 // indirect
+	github.com/prometheus/procfs v0.0.0-20171017214025-a6e9df898b13 // indirect
+	github.com/satori/go.uuid v1.2.0
+	github.com/sirupsen/logrus v1.4.0
```
neiljerram: Out of interest, is there a syntax for saying "<this> is the package name, but <that> (something else) is the repo to get for that package name"?  I guess that's what we need if we need to reference our logrus fork here.  @fasaxc ?

tools.go
```diff
@@ -0,0 +1,34 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// This package contains code generation utilities
+// This package imports things required by build scripts, to force `go mod` to see them as dependencies
```
fasaxc: Do the k8s authors provide this file in a way that we can import it from one of their repos?  Seems like that'd be the best option if possible.

go.mod
rafaelvanoni: Yes, there's a 'replace' directive. For example:
`replace github.com/sirupsen/logrus => github.com/projectcalico/logrus v1.0.4-calico`

tools.go
rafaelvanoni: Good question, I'll see if that can work. The mentions of it that I thought were all towards adding the file instead of importing, tho.

lib/backend/k8s/k8s_test.go
rafaelvanoni: I need to dig a bit more here, but that test was failing with either `WatchAdded` or `WatchModified`

Makefile
```diff
@@ -54,11 +56,8 @@ clean:
 # Building the binary
 ###############################################################################
 # Build the vendor directory.
-vendor: glide.lock
-	# To build without Docker just run "glide install -strip-vendor"
-	# Ensure that the glide cache directory exists.
-	mkdir -p $(HOME)/.glide
-	$(DOCKER_GO_BUILD) glide install --strip-vendor
+vendor: go.mod
```
caseydavenport: Probably want to rely on `go.sum` here, since we populate the vendor directory from the sum file.

lib/backend/k8s/k8s_test.go
rafaelvanoni: Seems like a flake, it's passing now.

caseydavenport:
> What's fakefile for?

I think this is something I accidentally committed a long time ago when I was testing some stuff.... oops

rafaelvanoni:
> > What's fakefile for?
> 
> I think this is something I accidentally committed a long time ago when I was testing some stuff.... oops

Yep, just removing it here.

tools.go
rafaelvanoni: Haven't found one, there are a few different posts where it's suggested to "just add this file".

rafaelvanoni:
> Couple of thoughts:
> 
>     * What's `fakefile` for?
> 
>     * Does the go package cache still work with go modules?  Preserving go's cache of built packages between builds dramatically speeds them up.

Yes, it still works.

Makefile
```diff
@@ -71,7 +71,7 @@ gen-files:
 
 $(BINDIR)/deepcopy-gen: vendor
 	$(DOCKER_GO_BUILD) \
-		sh -c 'go build -o $@ $(LIBCALICO-GO_PKG)/vendor/k8s.io/code-generator/cmd/deepcopy-gen'
+		sh -c 'go build -mod=vendor -o $@ $(LIBCALICO-GO_PKG)/vendor/k8s.io/code-generator/cmd/deepcopy-gen'
```
spikecurtis: When I delete the generated files and then run `make gen-files` it doesn't recreate them for me.  I suspect that this might have something to do with the fact that `PACKAGE_NAME` is `projectcalico/libcalico-go`, but the `--bounding-dirs` is `github.com/projectcalico/libcalico-go`

Makefile
```diff
@@ -106,7 +106,6 @@ check-gen-files: $(GENERATED_FILES)
 check-format: vendor
 	@if $(DOCKER_GO_BUILD) goimports -l lib | grep -v zz_generated | grep .; then \
 	  echo "Some files in ./lib are not goimported"; \
-	  false; \
```
caseydavenport: Why this change? I think this just breaks the check-format target, right?

rafaelvanoni: Yes, the newer version of goimports flags a ton of new errors. I need to go fix those as a follow up task as well as the errors flagged by golangci-lint.

Makefile
rafaelvanoni: Great catch. The problem turned to be our version of deepcopy-gen still needing /go/src/github.com/... Fixed now.

Makefile
caseydavenport: I'm not really comfortable just removing this check - our CI exists for a reason and we don't want to just comment out tests since changes make them fail - can't we just run the new version of `goimports` over the code to fix any new errors?

rafaelvanoni: Sure, there are a ton of new errors tho.

Makefile
```diff
@@ -70,26 +75,23 @@ gen-files:
...
-		sh -c '$(BINDIR)/deepcopy-gen \
-			--v 1 --logtostderr \
-			--go-header-file "./docs/boilerplate.go.txt" \
-			--input-dirs "$(LIBCALICO-GO_PKG)/lib/upgrade/migrator/clients/v1/k8s/custom" \
-			--bounding-dirs "github.com/projectcalico/libcalico-go" \
-			--output-file-base zz_generated.deepcopy'
+	$(DOCKER_GO_BUILD) sh -c '$(BINDIR)/deepcopy-gen \
+		--v 1 --logtostderr \
+		--go-header-file "./docs/boilerplate.go.txt" \
+		--input-dirs "$(LIBCALICO-GO_PKG)/lib/upgrade/migrator/clients/v1/k8s/custom" \
```
spikecurtis: This target is still broken

```suggestion
		--input-dirs "./lib/upgrade/migrator/clients/v1/k8s/custom" \
```

Makefile
```diff
@@ -70,26 +75,23 @@ gen-files:
...
-		sh -c '$(BINDIR)/deepcopy-gen \
-			--v 1 --logtostderr \
-			--go-header-file "./docs/boilerplate.go.txt" \
-			--input-dirs "$(LIBCALICO-GO_PKG)/lib/apis/v3" \
-			--bounding-dirs "github.com/projectcalico/libcalico-go" \
-			--output-file-base zz_generated.deepcopy'
+	$(DOCKER_GO_BUILD) sh -c '$(BINDIR)/deepcopy-gen \
+		--v 1 --logtostderr \
+		--go-header-file "./docs/boilerplate.go.txt" \
+		--input-dirs "$(LIBCALICO-GO_PKG)/lib/apis/v3" \
```
spikecurtis:
```suggestion
		--input-dirs "./lib/apis/v3" \
```

Makefile
rafaelvanoni: Fixed

Makefile
rafaelvanoni: Fixed

PR #1111, opened 2019-07-25
Release IPIP and VXLAN addresses on node deletion

caseydavenport:
I think we were missing logic to do this on node deletion, leading to
blocks hanging around when they need not be.

- [ ] Tests
- [ ] Documentation
- [ ] Release note

```release-note
Fix that tunnel addresses were not released on node deletion when using etcd data store. 
```

### DIFF
--- a/lib/clientv3/node.go
+++ b/lib/clientv3/node.go
@@ -115,6 +115,32 @@ func (r nodes) Delete(ctx context.Context, name string, opts options.DeleteOptio
 			}
 		}
 	}
+
+	// Add in tunnel addresses if they exist for the node.
+	if n, err := r.client.Nodes().Get(ctx, name, options.GetOptions{}); err != nil {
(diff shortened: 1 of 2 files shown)


lib/clientv3/node.go
```diff
@@ -115,6 +115,29 @@ func (r nodes) Delete(ctx context.Context, name string, opts options.DeleteOptio
 			}
 		}
 	}
+
+	// Add in tunnel addresses if they exist for the node.
+	n, err := r.client.Nodes().Get(ctx, name, options.GetOptions{})
+	if err != nil {
+		return nil, err
```
fasaxc: What if the error is a "not found".  In that case, we might still want to clean up the IPs.

caseydavenport:
@fasaxc I think this is ready for another look. Thanks for the review!

PR #1112, opened 2019-07-26
BGP Peer Spec: Add Remote Node ASN Field

tompntn:
## Description
Adds a `nodeAsNumber` field to the BGP Peer spec, which dictates the effective ASN of all remote nodes which have the peering. Currently, if we need all remote nodes in a given peering to have the same ASN we have to manually add the `projectcalico.org/ASNumber` annotation to each node. Adding this field to the spec makes achieving that configuration easy, and can be persisted with yaml in the same place as the other peering information.

This field will take precedence over the remote Node's `spec.bgp.asNumber` and the global default ASN.

If this change to the API is accepted, the relevant changes will be made to the `libcalico` backend and the Calico `confd` fork.

## Todos
- [x] Tests
- [ ] Documentation
- [ ] Release note

## Release Note
```release-note
None required
```

### DIFF
--- a/lib/apis/v3/bgppeer.go
+++ b/lib/apis/v3/bgppeer.go
@@ -51,6 +51,9 @@ type BGPPeerSpec struct {
 	// Selector for the nodes that should have this peering.  When this is set, the Node
 	// field must be empty.
 	NodeSelector string `json:"nodeSelector,omitempty" validate:"omitempty,selector"`
+	// The AS Number that all remote nodes which have this peering will have. Takes precedence
+	// over the remote Node's spec.bgp.asNumber and the global default AS Number.
+	NodeASNumber numorstring.ASNumber `json:"omitempty,nodeAsNumber" validate:"omitempty"`
 	// Selector for the remote nodes to peer with.  When this is set, the PeerIP and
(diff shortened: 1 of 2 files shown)


tmjd:
Thanks for submitting this.
I'm not an expert in the BGP Peering and AS number but what happens if multiple peers select a node that have different AS numbers?

caseydavenport:
I'm not necessarily opposed to this, but my initial thought is that it doesn't feel like the right place for this information, and allowing this to be configured in yet another location makes this more confusing and less clear what the "right" way to do this is for users.

I wonder, since we already have this as an option on the BGPConfiguration object, which has `global` and `node` specific variants. Is the problem here just that BGPConfiguration doesn't have a `nodeSelector` on it?

tompntn:
@caseydavenport @tmjd Adding a `nodeSelector` to the `BGPConfiguration` is a much better idea. Is this an API change you'd be happy with?

caseydavenport:
> Is this an API change you'd be happy with?

I think so. Before we sign off on that we'll need to work through some of the details - for example, what happens when a nodeSelector selects a node that also has a node-specific variant? Do we merge the two, or does one take precedence?

PR #1115, opened 2019-08-06
Send entire node in FelixNodeUpdateProcessor

Brian-McM:
This adds the node under with the NodeKey key in the KV Process method in felixnodeprocessor. To be backwards compatible, this requires changing the return value for the String() function for the HostIPKey to return something different than the NodeKey's String() function, the added NodeKey would be ignored by the watcherCache

## Description
This commit is so we have the necessary node information, specifically the nodes subnet, when implementing the CrossSubnet option for VXLAN 

## Todos
- [x] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/backend/syncersv1/updateprocessors/felixnodeprocessor.go
+++ b/lib/backend/syncersv1/updateprocessors/felixnodeprocessor.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2017 Tigera, Inc. All rights reserved.
+// Copyright (c) 2019 Tigera, Inc. All rights reserved.
 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


lib/backend/syncersv1/updateprocessors/felixnodeprocessor.go
```diff
@@ -145,6 +147,14 @@ func (c *FelixNodeUpdateProcessor) Process(kvp *model.KVPair) ([]*model.KVPair,
 			Value:    vxlanTunlMac,
 			Revision: kvp.Revision,
 		},
+		{
+			Key: model.ResourceKey{
+				Name: name,
+				Kind: "node",
```
caseydavenport: I think we have constants defined for the Kind in the apis/v3 package. See here: https://github.com/projectcalico/libcalico-go/blob/7be0d765781fd762c2b3f0a3f5392b2b132ba3e3/lib/apis/v3/node.go#L24

caseydavenport:
Looks like a couple tests need to be updated still - I think they are just expecting one fewer items than we're sending now, so probably need to bump the number up.

PR #1116, opened 2019-08-08
Allow the vxlan protocol filter to be set in the felix configuration

Brian-McM:
## Description
This allows the user to override the default protocol the vxlan manager uses in felix (protocol 80)

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/apis/v3/felixconfig.go
+++ b/lib/apis/v3/felixconfig.go
@@ -162,6 +162,9 @@ type FelixConfigurationSpec struct {
 	VXLANPort *int `json:"vxlanPort,omitempty"`
 	VXLANVNI  *int `json:"vxlanVNI,omitempty"`
 
+	// RouteProtocol specifies the protocol to use to mark routes managed by felix using the default interface.
+	RouteProtocol int `config:"int;0"`
+
 	// ReportingInterval is the interval at which Felix reports its status into the datastore or 0 to disable.
 	// Must be non-zero in OpenStack deployments. [Default: 30s]
 	ReportingInterval *metav1.Duration `json:"reportingInterval,omitempty" configv1timescale:"seconds" confignamev1:"ReportingIntervalSecs"`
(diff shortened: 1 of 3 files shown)


lib/apis/v3/felixconfig.go
```diff
@@ -162,6 +162,10 @@ type FelixConfigurationSpec struct {
 	VXLANPort *int `json:"vxlanPort,omitempty"`
 	VXLANVNI  *int `json:"vxlanVNI,omitempty"`
 
+	// VXLANProtocolFilter specifies the protocol to mark non vxlan routes that should be managed by the vxlan manager.
+	// If not set, the vxlan manager automatically sets it to 80.
+	VXLANProtocolFilter int `config:"int;0"`
```
caseydavenport: I wonder if we should call this something like `RouteProtocol` (i.e., remove VXLAN). We may add support to Felix to program IPIP routes as well, at which point we'd probably want to use this same protocol.

Also, I think the design has us only setting this protocol on routes which aren't VXLAN routes (unencapp'd).

Brian-McM: The concern that I have about that is that what if we don't want IPIP and VXLAN using the same protocol (if you're switching from IPIP to VXLAN or visa versa and the managers run asynchronously then they would try to manage each others routes), or we want to use protocol filtering for another route table we're managing when vxlan is enabled.

I more wanted to say that routes with this protocol are managed by the vxlan manager (whether they're vxlan routes or not), as opposed to saying only vxlan routes are tagged with these protocols. Maybe I could name it VXLANMgrProtocolFilter to make that clear?

caseydavenport: My assumption to-date was that the "vxlan resolver" and "vxlan manager" components would just become generic "route resolver" and "route manager" components and would be shared infra, and not user-visible (they're an implementation detail rather than a guarantee).

To address wanting multiple protocols, we could just as easily make this a structured field - e.g., a list, so that if we want to add more protocols in the future we can add them. Felix can then pick from the provided list in some deterministic way. 

I don't think the end-user cares which protocol is used for which type of route, right?

Brian-McM: Right, so are you thinking we just have a RouteProtocols config param and have that as a list of protocols that are available for felix to choose from when it needs a protocol for routes?

Brian-McM:
Going to use the changes in this pr https://github.com/projectcalico/libcalico-go/pull/1119 for adding the a route protocol

PR #1117, opened 2019-08-16
Add dry run mode and ability to log when no rules match

jackkleeman:
## Description
New feature: We add a `DryRun` annotation, which means that packets are always
allowed but we log when they would have been dropped. We also add a `LogNoRuleMatch` annotation, which means you can log when packets are dropped without setting WarnMode.

This will be really useful to help test network policies in a risk free way. Simply develop the network policy you like, add the DryRun annotation, and run it for a while, keeping track of logs (eg with https://github.com/box/kube-iptables-tailer/). After you come out of DryRun mode, you can use `LogNoRuleMatch` to keep track of new issues with the policy on an ongoing basis. You can also apply `LogNoRuleMatch` to all policies with an environment variable.

I have tested this manually in my cluster and have added a unit test for each annotation. It will not affect anyone who does not specifically request these features. It affects the calico kube controller only.

I would appreciate advice on where best to document this.

## Todos
- [x] Tests
- [ ] Documentation
- [x] Release note

## Release Note
```release-note
Add `projectcalico.org/DryRun` and `projectcalico.org/LogNoRuleMatch` annotations which allow for easier testing of network policies.
```


jackkleeman:
I think the test failure is a flakey test

caseydavenport:
@jackkleeman could you provide a little bit more detail about your use-case? 

This PR isn't exactly what I would expect for a dry-run capability, so I think some more use-case information would help me in reviewing this.

jackkleeman:
@caseydavenport the way I am using this is that I write a network policy for example to isolate inbound connections to a specific deployment via a label selector. Because I'm not 100% sure that I've correctly enumerated all callers (via labels or ip ranges), I add the `DryRun` annotation. As a result, no packets get dropped; when the policy _would have_ been violated, we instead log, but still allow the packet.

Then I have https://github.com/box/kube-iptables-tailer/ set up to tail my iptables logs on all nodes, so I can immediately tell when there has been some packets that would have been dropped. Then I can whitelist whatever I missed

Once I've decided my networkpolicy is good, I remove the `DryRun` annotation, but keep `LogNoRuleMatch` (or maybe I set this by default). This means that packets will be dropped when a packet is not allowed by the NetworkPolicy, but we always log when this happens, which helps me alert on issues (eg incorrect whitelist, or in fact a vulnerability).

There are some differences between this and 'true' dry run, particularly the fact that if you have two network policies, one that doesn't allow, and one that doesn't allow but is in dry run mode, the packets will be allowed. I can't see a simple way of getting around this but I feel that we can document it. I don't know how common it is to have overlapping policies? Even with this, I think there is a lot of value in this and it is enabling us to roll out a ton of policies generated by static analysis

caseydavenport:
I think it's pretty common to have overlapping policies - one of the most common usage patterns involves a default deny policy which selects all pods, and additional policies which white-list traffic for particular selections.

I think what exists in this PR and the use-case you've described above is actually already implementable using capabilities available natively in Calico network policy resources. I would suggest using those in order to meet this use-case. 

It would be really cool to have a "How to" document on docs.projectcalico.org for how to use Calico LOG rules in conjunction with kube-iptables-tailer, though!

jackkleeman:
> one of the most common usage patterns involves a default deny policy which selects all pods, and additional policies which white-list traffic for particular selections.

I don't think it would make much sense to use dry run mode on the additional policy that only allows new traffic, as dry run is about detecting new packet drops. And yes, dry run would be too blunt of an instrument on the default deny policy. My concern was about situations where two policies select a pod and both don't allow.

> I think what exists in this PR and the use-case you've described above is actually already implementable using capabilities available natively in Calico network policy resources

The above uses capabilities of calico network policy resources - I'm just interested in a way of enabling the dry run behaviour with an annotation on a network policy. If we can come up with a way to do that that avoids this policy overlap issue that would be great!

caseydavenport:
Why not just use a [Calico network policy](https://docs.projectcalico.org/v3.8/reference/resources/networkpolicy) that comes after all of the other ones, and includes rules that do what you want?

e.g.,

```
kind: NetworkPolicy
metadata:
  name: log-unmatched-traffic
  namespace: default
spec:
  selector: all()
  order: 5000
  types:
  - Ingress
  ingress:
  - action: Log
  - action: Allow
```

PR #1123, opened 2019-09-03
BGPConfiguration: ServiceExternalIPs Field

tompntn:
## Description
In order to support advertising service External IP's (https://github.com/projectcalico/calico/issues/2770), we need to whitelist the IP's that we advertise. This is to prevent users from advertising arbitrary IP's into the network.

This PR adds a field to `BGPConfiguration` allowing users to whitelist allowed External IP's:
```
kind: BGPConfiguration
metadata:
  name: default
spec:
  logSeverityScreen: Info
  nodeToNodeMeshEnabled: true
  asNumber: 63400
  serviceExternalIPs:
  - cidr: 104.244.42.129/32
  - cidr: 172.217.3.0/24
```

This list is then serialized and added to the v1 syncer, for consumption by `confd`. Confd support for advertising External IP's is added in https://github.com/projectcalico/confd/pull/266.

Note that this is a global whitelist, so only the value in the `default` bgpconfiguration will be used.

## Todos
- [x] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


caseydavenport:
@tompntn this is looking good, but it looks like there are a couple of test failures.

```
• Failure [0.002 seconds]
BGPConfiguration tests [Datastore] (etcdv3 backend) BGPConfiguration e2e CRUD tests [It] Two fully populated BGPConfigurationSpecs 
/go/src/projectcalico/libcalico-go/vendor/github.com/onsi/ginkgo/extensions/table/table_entry.go:46

  Expected
      <string>: error with field ServiceExternalIPs = '[{104.244.42.129/32} {172.217.3.0/24}]' (Reason: failed to validate Field: ServiceExternalIPs because of Tag: net )
  to contain substring
      <string>: resource does not exist: BGPConfiguration(default) with error:
```

PR #1134, opened 2019-09-28
Expose ClusterIP advertisement through BGP configuration

caseydavenport:
## Description

We already did this for ExternalIPs which is new in v3.10 here: https://github.com/projectcalico/libcalico-go/pull/1123

This PR is to make the ClusterIP advertisement symmetrical. It goes with this corresponding confd PR: https://github.com/projectcalico/confd/pull/272

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Configuration of Kubernetes service Cluster IP advertisement through BGPConfiguration resource.
```


lib/apis/v3/bgpconfig.go
```diff
@@ -41,18 +41,29 @@ type BGPConfiguration struct {
...
+
+// ServiceExternalIPBlock represents a single whitelisted CIDR External IP block.
+type ServiceExternalIPBlock struct {
+	CIDR string `json:"cidr,omitempty" validate:"omitempty,net"`
 }
 
-// SvcExternalIPBlock represents a single whitelisted CIDR External IP block.
-type SvcExternalIPBlock struct {
+// ServiceClusterIPBlock represents a single whitelisted CIDR block for ClusterIPs.
+type ServiceClusterIPBlock struct {
```
neiljerram: I don't want to block this PR, but I'm curious what the benefit of ServiceClusterIPBlock is, instead of the more direct:
```
	ServiceClusterIPs []string
```
With the extra indirection it feels like the construction of these resources will be more awkward in both YAML and golang.

lib/backend/syncersv1/updateprocessors/bgpconfigprocessor.go
```diff
@@ -64,10 +65,28 @@ var nodeMeshToString = func(value interface{}) interface{} {
 	return nodeToNodeMeshDisabled
 }
 
-// We wrap each Service External IP in a SvcExternalIPBlock struct to
+// We wrap each Service Cluster IP in a ServiceExternalIPBlock struct to
```
neiljerram: s/Cluster/External/

lib/clientv3/bgpconfig_e2e_test.go
```diff
@@ -63,8 +63,8 @@ var _ = testutils.E2eDatastoreDescribe("BGPConfiguration tests", testutils.Datas
 		LogSeverityScreen:     "Warning",
 		NodeToNodeMeshEnabled: &ptrFalse,
 		ASNumber:              &nodeASNumber2,
-		ServiceExternalIPs: []apiv3.SvcExternalIPBlock{
-			apiv3.SvcExternalIPBlock{
+		ServiceExternalIPs: []apiv3.ServiceExternalIPBlock{
+			apiv3.ServiceExternalIPBlock{
```
neiljerram: Do you want to cover cluster IP configuration here too?

lib/apis/v3/bgpconfig.go
caseydavenport: @neiljerram good question - the reason we designed it this way is for extensibility down the line. This model allows us to augment blocks with additional attributes if needed, and the yaml difference is minor.

The difference now is this:

```
serviceClusterIPs:
- cidr: 10.0.0.0/8
```

vs 

```
serviceClusterIPs:
- 10.0.0.0/8
```

But, the former allows us to add additional fields if we want to in the future, whereas the second form doesn't allow us to easily do that. For example:

```
serviceClusterIPs:
- cidr: 10.0.0.0/8
  nodeSelector: 'all()'
```

neiljerram: In this case I struggle to imagine what other extensions there could be; but OK.

PR #1135, opened 2019-09-30
Combine together ports of the same protocol

jackkleeman:
## Description

This is a new feature that isn't particularly client facing, intended to reduce the number of iptables rules in cases where we have multiple ports in a k8s network policy. This will hopefully improve application times and reduce errors in cases where people have a lot of ports. This code is already fairly well tested but I have added two new tests to cover particular edge cases that concerned me, where there is no ports provided, and where an empty port struct is provided. This affects the calico kube controllers.

When converting k8s network policies to calico policies, we currently create one calico rule per port x peer combination. Instead, we should probably create one rule per protocol x peer combination, and combine together ports of the same protocol.

As normal, no ports leads to allowing any port and any protocol. The existing behaviour is that any empty port struct, even in a list with other ports, will lead to the creation of a rule that allows all ports. Currently we also create other rules for other ports, but this seems pointless, so I have decided to only produce the allow all rule in this case.

## Todos
- [x] Tests
- [ ] Documentation - not sure this is needed
- [x] Release note

## Release Note

```release-note
Kubernetes NetworkPolicyIngressRules with multiple ports are kept as a single Calico rule per peer where possible, instead of one rule per port/peer combination.
```

### DIFF
--- a/lib/backend/k8s/conversion/conversion.go
+++ b/lib/backend/k8s/conversion/conversion.go
@@ -535,13 +535,43 @@ func (c Converter) k8sRuleToCalico(rPeers []networkingv1.NetworkPolicyPeer, rPor
 		ports = []*networkingv1.NetworkPolicyPort{nil}
 	}
 
-	// Combine destinations with sources to generate rules.
-	// TODO: This currently creates a lot of rules by making every combination of from / ports
-	// into a rule.  We can combine these so that we don't need as many rules!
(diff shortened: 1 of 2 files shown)


caseydavenport:
@jackkleeman thanks for the PR! This is pretty cool. I've given it a quick look and it seems like the right stuff. Planning on doing a more detailed look sometime this week I hope.

PR #1164, opened 2019-11-18
Use client-go@kubernetes-1.15.1 instead of client-go@v12.0.0 for go1.13

rafaelvanoni:
go 1.13 has a bug that prevents using v12.0.0 and the workaround is to use the kubernetes versioned tag. This is causing problems when moving the app-policy repository to 1.13


go.mod
```diff
@@ -47,9 +47,9 @@ require (
 	gopkg.in/go-playground/validator.v9 v9.27.0
 	gopkg.in/tchap/go-patricia.v2 v2.2.6
 	gopkg.in/yaml.v2 v2.2.5 // indirect
-	k8s.io/api v0.0.0-20190620084959-7cf5895f2711
+	k8s.io/api v0.0.0-20190718183219-b59d8169aab5
 	k8s.io/apimachinery v0.0.0-20190612205821-1799e75a0719
-	k8s.io/client-go v12.0.0+incompatible
+	k8s.io/client-go v0.0.0-20190620085101-78d2af792bab
```
tmjd: The hash on this indicates kuberentes-1.15.0, not 1.15.1

rafaelvanoni: Yeah, should be 1.15.0

PR #1166, opened 2019-11-20
Move common Makefile targets to Makefile.common

rafaelvanoni:


Makefile
```diff
@@ -1,50 +1,27 @@
...
+###############################################################################
+MAKE_BRANCH?=$(GO_BUILD_VER)
+MAKE_REPO?=https://raw.githubusercontent.com/projectcalico/go-build/$(MAKE_BRANCH)
+
+Makefile.common: Makefile.common.$(MAKE_BRANCH)
+	cp "$<" "$@"
+Makefile.common.$(MAKE_BRANCH):
+	# Clean up any files downloaded from other branches so they don't accumulate.
+	rm -f Makefile.common.*
+	wget -nv $(MAKE_REPO)/Makefile.common -O "$@"
```
realgaurav: It's debatable but I would imagine `curl` to have higher chances of being available :)
Either way, I suggest adding a target which checks for presence of `wget` or `curl` and then reference it as a dependency here.

Makefile
```diff
@@ -1,50 +1,27 @@
...
-# Define some constants
-#######################
 K8S_VERSION      ?= v1.16.0
-ETCD_VERSION     ?= v3.3.7
-COREDNS_VERSION  ?= 1.5.2
-GO_BUILD_VER     ?= v0.26
-CALICO_BUILD     ?= calico/go-build:$(GO_BUILD_VER)
-PACKAGE_NAME     ?= github.com/projectcalico/libcalico-go
-LOCAL_USER_ID    ?= $(shell id -u $$USER)
 BINDIR           ?= bin
```
realgaurav: Can this be defined in Makefile.common with `bin` as the default value and redefined if needed.

Makefile
```diff
@@ -1,50 +1,27 @@
...
-		--net=host \
-		$(EXTRA_DOCKER_ARGS) \
-		-e LOCAL_USER_ID=$(LOCAL_USER_ID) \
-		-e GOCACHE=/go-cache \
-		-e GOPATH=/go \
-		-v $(CURDIR):/go/src/$(PACKAGE_NAME):rw \
-		-v $(CURDIR)/.go-pkg-cache:/go-cache:rw \
-		-w /go/src/$(PACKAGE_NAME)
-
 DOCKER_GO_BUILD := $(DOCKER_RUN) $(CALICO_BUILD)
```
realgaurav: `DOCKER_GO_BUILD` in Makefile.common?

rafaelvanoni: Fixed

Makefile
rafaelvanoni: That seemed a little too close to the repository specific build rules, so I chose to leave it alone.

Makefile
rafaelvanoni: Fixed

Makefile
```diff
@@ -1,52 +1,28 @@
...
+GO_BUILD_VER=v0.27
 
-## Build binary
-all: test
+###############################################################################
+# Download and include Makefile.common before anything else
+###############################################################################
+MAKE_BRANCH?=$(GO_BUILD_VER)
+MAKE_REPO?=https://raw.githubusercontent.com/projectcalico/go-build/$(MAKE_BRANCH)
+WGET?=/usr/bin/wget
```
realgaurav: I meant something like this--

```
wget: 
    if `which wget`
        $(error "Missing wget, please install wget to proceed")
    endif
```

realgaurav: @rafaelvanoni, apologies for not being clear earlier. I meant to suggest --

```
wget:
ifeq ($(shell which wget),)
    $(error "missing wget, please install wget to proceed")
endif

Makefile.common.$(MAKE_BRANCH): wget
```

Makefile
```diff
@@ -271,26 +237,12 @@ run-coredns: stop-coredns
 stop-coredns:
 	-docker rm -f coredns
 
+st:
```
realgaurav: I missed where will this be used.

Makefile
```diff
@@ -100,8 +76,9 @@ $(BINDIR)/deepcopy-gen: vendor
 ###############################################################################
 # Static checks
 ###############################################################################
-.PHONY: static-checks
-static-checks: check-format check-gen-files
+# TODO: re-enable all linters
+LINT_ARGS += --disable gosimple,unused,structcheck,errcheck,deadcode,varcheck,ineffassign,staticcheck,govet
+LOCAL_CHECKS = goimports check-gen-files
```
realgaurav: How will this be used?

Makefile
rafaelvanoni: Yeah but that's overkill :-)

Makefile
rafaelvanoni: Both of those vars are used in `static-checks` defined in Makefile.common. I'm disabling those linters for now because they're flagging a bunch of errors and fixing them is out of scope for this PR.

Makefile
rafaelvanoni: The `tests` target in Makefile.common requires a `st` target.

Makefile
realgaurav: Not to overemphasize but this could also be a place to lookup `curl` if `wget` is not present.

PR #1170, opened 2019-11-26
Only specify -mod=vendor once in CI

rafaelvanoni:
We already specify the "-mod=vendor" option when running in CI. This PR prevents us from passing it to go twice, which is not accepted.


rafaelvanoni:
This is actually incorrect, needs a bit more work.

PR #1173, opened 2019-12-02
Wrap a mutex around ErrorTreshold

rafaelvanoni:
This fixes one of the race conditions picked up by the `-race` option to `go test` and `ginkgo`, it's just a simple wrapper around the ErrorThreshold variable in the watchersyncer package.

### DIFF
--- a/lib/backend/watchersyncer/watchercache.go
+++ b/lib/backend/watchersyncer/watchercache.go
@@ -16,6 +16,7 @@ package watchersyncer
 
 import (
 	"context"
+	"sync"
 	"time"
 
 	"github.com/sirupsen/logrus"
@@ -50,7 +51,8 @@ type watcherCache struct {
 var (
 	ListRetryInterval = 1000 * time.Millisecond
 	WatchPollInterval = 5000 * time.Millisecond
-	ErrorThreshold    = 15
+	errorThreshold    = 15
+	errorThresholdMtx sync.Mutex
 )
 
 // cacheEntry is an entry in our cache.  It groups the a key with the last known
@@ -134,7 +136,7 @@ mainLoop:
 					wc.onError()
 				}
 
-				if wc.errors > ErrorThreshold {
+				if wc.errors > GetErrorThreshold() {
 					// Trigger a full resync if we're past the error threshold.
 					wc.currentWatchRevision = ""
 					wc.resyncAndCreateWatcher(ctx)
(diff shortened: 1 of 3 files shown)


lib/backend/watchersyncer/watchersyncer_test.go
```diff
@@ -133,7 +133,8 @@ var _ = Describe("Test the backend datastore multi-watch syncer", func() {
 
 		rs.clientWatchResponse(r1, genError)
 		rs.ExpectStatusUnchanged()
-		for i := 0; i < watchersyncer.ErrorThreshold-1; i++ {
+		n := watchersyncer.GetErrorThreshold()
```
robbrockbank: n-1

rafaelvanoni: Fixed !

lib/backend/watchersyncer/watchercache.go
```diff
@@ -428,9 +430,23 @@ func (wc *watcherCache) markAsValid(resourceKey string) {
...
 }
+
+func GetErrorThreshold() int {
+	errorThresholdMtx.Lock()
+	val := errorThreshold
+	errorThresholdMtx.Unlock()
+	return val
+}
+
+func SetErrorThreshold(val int) {
```
ozdanborne: Having `errorThreshold` be a package-level variable makes it harder to test `watcherCache` and requires these mutex functions to prevent race conditions.

How is `watcherCache` used? Is anyone ever updating the `errorThreshold` besides right after initialization? If not, couldn't we just get rid of the mutex & global variable altogether by moving the `errorThreshold` to be a argument passed to `newWatcherCache`?

rafaelvanoni: There are those deferred goroutines in the test package that modify `errorThreshold`, besides when it is initialized. It was already a package-level variable, which seems to be necessary as it is controlling a threshold applicable to all watcherCaches.

lmm: I think that might be better. The only time `errorThreshold` ever changes is in tests.

rafaelvanoni: @lmm Can you clarify what the idea is here.. this was already a global variable, did you mean to change that (for all consumers)?

ozdanborne: Yeah. I'm suggesting (and I think @lmm is agreeing) that we move errorThreshold to be a "readonly" private field on the watcherCache struct, and add it as an argument to newWatcherCache.

ozdanborne: >It was already a package-level variable, which seems to be necessary as it is controlling a threshold applicable to all watcherCaches.

This is the key point. I'm questioning if we really need a setting for "all watcherCaches", and if instead we just need a setting for "each watcherCache".

rafaelvanoni: > This is the key point. I'm questioning if we really need a setting for "all watcherCaches", and if instead we just need a setting for "each watcherCache".

Yup, that's what I'm trying to get at :-)

rafaelvanoni: This check in run() seems to imply it does want to look at all objects

`                                if wc.errors > GetErrorThreshold() {
                                        // Trigger a full resync if we're past the error threshold.
                                        wc.currentWatchRevision = ""
                                        wc.resyncAndCreateWatcher(ctx)
                                }
`

lmm: No, that's just resync on that particular watcherCache. Each of the watcherCache's do their own thing.

rafaelvanoni: Ok, so you do want a per watcherCache threshold.

lmm:
Closing this since we addressed the original issue in https://github.com/projectcalico/libcalico-go/pull/1183

PR #1174, opened 2019-12-03
pin k8s.io/apimachinery to kubernetes-v1.15.0

rafaelvanoni:


lib/backend/k8s/conversion/conversion.go
```diff
@@ -182,12 +182,15 @@ func (c Converter) HasIPAddress(pod *kapiv1.Pod) bool {
 // present, or the calico podIP annotation.
 func (c Converter) getPodIPs(pod *kapiv1.Pod) ([]*cnet.IPNet, error) {
 	var podIPs []string
-	if ips := pod.Status.PodIPs; len(ips) != 0 {
-		log.WithField("ips", ips).Debug("PodIPs field filled in")
-		for _, ip := range ips {
-			podIPs = append(podIPs, ip.IP)
-		}
-	} else if ip := pod.Status.PodIP; ip != "" {
+	// TODO: PodIPs is only present in k8s version > 1.15.0
```
ozdanborne: If podIPs are only present in k8s v1.15, why are we commenting it out in this PR which bumps us to v1.15?

rafaelvanoni: I meant versions `larger` than 1.15.0

ozdanborne: Oh I see. Sort-of tangential though -

It looks like @neiljerram 's PR was pretty deliberate in updating to a version that has those fields: https://github.com/projectcalico/libcalico-go/pull/1168

Are you sure downgrading to a version that doesn't will not impact his work?

rafaelvanoni: Up-grading the /other/ packages instead of downgrading the ones @neiljerram upgraded, per our discussion.

rafaelvanoni:
I'll upgrade the /other/ packages instead of downgrading the ones @neiljerram upgraded, per our discussion.

PR #1176, opened 2019-12-04
move all k8s.io packages to kubernetes v1.16.3

rafaelvanoni:


rafaelvanoni:
> lgtm. I wonder if this will impact downstream components that import libcalico-go because libcalico-go is accessing fields on structs that don't exist in <v1.15 but man_shrugging

Yes.. We need to update typha and felix as well, stand by for PRs on those repos.

PR #1178, opened 2019-12-10
provide a proper Revision when converting k8s to crd NetworkPolicies

breunigs:
## Description

This is a potential fix for https://github.com/projectcalico/typha/issues/335 . The gist of this bug report is that Typha fails to create a watch for a k8s NetworkPolicy because it fails to parse the Revision it finds in a `KVPair`:

```
Failed to create watcher
  ListRoot="/calico/resources/v3/projectcalico.org/networkpolicies"
  error=ResourceVersion is not valid: 346124214
```
The test I added is very much on the unit test side of things. I deployed it to one of our toy clusters, where it appears to work. I.e. it doesn't complain about failing to setup the watch anymore and the `List` from the full resync also disappears. The network policy is honored. I have not run the acceptance/integration testsuite locally because it fails for me even on master. The tests were done with Typha 3.7.5 and this patch applied.

As for the list of components, it's `NetworkPolicy` for sure and all its callers/users. Typha obviously, but so is everything using k8s NetworkPolicy conversion. I'm not familiar enough with the Calico codebase to come up with a list.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

The fix is an optimization in some sense, since the failure to parse the revision caused a full resync. The resync works, so the NetworkPolicies are applied. Not sure if that warrants a release note.

```release-note
fix Kubernetes NetworkPolicy conversion bug (caused unnecessary resyncs in e.g. Typha)
```

### DIFF
--- a/lib/backend/k8s/k8s_test.go
+++ b/lib/backend/k8s/k8s_test.go
@@ -2249,6 +2249,16 @@ var _ = testutils.E2eDatastoreDescribe("Test Watch support", testutils.Datastore
 			defer watch.Stop()
 			ExpectAddedEvent(watch.ResultChan())
 		})
+		It("supports resuming watch from previous revision", func() {
+			watch, err := c.Watch(ctx, model.ResourceListOptions{Kind: apiv3.KindNetworkPolicy}, "")
+			Expect(err).NotTo(HaveOccurred())
+			event := ExpectAddedEvent(watch.ResultChan())
+			watch.Stop()
+
+			watch, err = c.Watch(ctx, model.ResourceListOptions{Kind: apiv3.KindNetworkPolicy}, event.New.Revision)
+			Expect(err).NotTo(HaveOccurred())
+			watch.Stop()
+		})
 	})
 
 	Describe("watching NetworkPolicies (calico)", func() {
--- a/lib/backend/k8s/resources/networkpolicy.go
+++ b/lib/backend/k8s/resources/networkpolicy.go
@@ -340,7 +340,12 @@ func (c *networkPolicyClient) Watch(ctx context.Context, list model.ListInterfac
 		if !ok {
 			return nil, errors.New("NetworkPolicy conversion with incorrect k8s resource type")
 		}
-		return c.K8sNetworkPolicyToCalico(np)
+
+		kvp, err := c.K8sNetworkPolicyToCalico(np)
+		if kvp != nil {
+			kvp.Revision = c.JoinNetworkPolicyRevisions("", kvp.Revision)
+		}
+		return kvp, err
 	}
 	k8sWatch := newK8sWatcherConverter(ctx, "NetworkPolicy (namespaced)", converter, k8sRawWatch)
 


lib/backend/k8s/conversion/conversion.go
```diff
@@ -468,14 +468,15 @@ func (c Converter) K8sNetworkPolicyToCalico(np *networkingv1.NetworkPolicy) (*mo
 	}
 
 	// Build and return the KVPair.
+	rev := c.JoinNetworkPolicyRevisions("", np.ResourceVersion)
```
caseydavenport: I'm not entirely convinced this is the right fix here - `JoinNetworkPolicyRevisions` is meant for resources that are backed by two separate APIs, where we need both revisions available.

In this case, NP is just a single API, and so this function isn't really doing anything. Especially without the corresponding "Split" revisions call. I would guess that it is coincidental that it stops the "Failed to create watcher" log mentioned in the original issue.

I think it was called out that the log is not an error, and expected periodically, so I'm not sure if there is anything to fix here except perhaps lowering the log level to debug?

breunigs: Let's discuss if this is an actual issue first. I'm not sure if I should post here, or in the bug. 

On our production system I see the `watchercache.go 259: Failed to create watcher` error about every minute or two. On our toy cluster, I will see this error message if and only if I update or create a k8s network policy. As you rightly point out, this is not immediate but periodic in some sense, I guess due to the previous watch expiring (I have not understood this bit in the code yet). The production system in question is essentially a build/test tool where namespaces and the associated network policies come and go, so seeing this error so often is not surprising.

The consequence is that it will perform a full resync very often. I know this because of the INFO log `2019-12-10 21:03:35.772 [INFO][8] networkpolicy.go 283: Returning NP KVPs KVPs=&model.KVPairList{KVPairs:[]*model.KVPair{(*model.KVPair)(0xc00ac2b280), … (*model.KVPair)(0xc009097b00)}, Revision:"1191809714/1191809718"}`. This log line contains the memory addresses for all NetworkPolicies in the cluster (about 850 for us right now).

So, as far as my investigation goes typha will perform a full resync on NetworkPolicies everytime any k8s NP is added or changed. Please point out if I misunderstand things here… because with my current level of knowledge it seems surprising if a full resync is desired behaviour here.

> Especially without the corresponding "Split" revisions call

To my understanding, the watchercache passes the "wrong" Revision (e.g. "1234") back to the client here: https://github.com/projectcalico/libcalico-go/blob/2d2a4ff81c856b51ee69705a26f72000252eb531/lib/backend/watchersyncer/watchercache.go#L238
I think the client is a `networkPolicyClient`, which calls into "Split" when setting up a watch: https://github.com/projectcalico/libcalico-go/blob/master/lib/backend/k8s/resources/networkpolicy.go#L322-L327
There's only two places that generate the `ResourceVersion is not valid` message `SplitNetworkPolicyRevision` and `SplitProfileRevision`, so it looks very likely to me that I'm hitting that code path.

> I would guess that it is coincidental that it stops the "Failed to create watcher" log mentioned in the original issue.
I think you're right. I looked through the code again at who calls `K8sNetworkPolicyToCalico`. There's three places in `networkPolicyClient`, one is in `Get`, one in `List` and one in `Watch`. The first two call `kvp.Revision = c.JoinNetworkPolicyRevisions("", kvp.Revision)` more or less immediately after. However, `Watch` doesn't: https://github.com/projectcalico/libcalico-go/blob/2d2a4ff81c856b51ee69705a26f72000252eb531/lib/backend/k8s/resources/networkpolicy.go#L343
Would a better fix be to also call `JoinNetworkPolicyRevisions` here?

breunigs:
Regarding the CLA: It demands some rights which only my boss has been granted by the org. He's willing to sign (in fact he already has), but the CLA Assistant does not offer any way to sign as a company or sign for other accounts. Well, at least we couldn't find it in the CLA's GUI. Please advise.

spikecurtis:
Hi @breunigs, thanks for taking the time to produce this fix.  On the CLA, can you confirm that you and your boss are actual employees of the same entity (e.g. neither of you is legally a contractor)? So, if we get an entity (instead of individual) agreement signed we should be good to go?

breunigs:
@spikecurtis Yes, I am working for "New Work SE" (https://github.com/orgs/new-work/people) and so is my boss @skaes.

skaes:
@spikecurtis I am authorized to sign agreements on behalf of our company, New Work SE. Since today, I'm also co-owner of the new-work organization and we only accept New Work SE employees as members of the org. Can you please let me know how to proceed?

spikecurtis:
@breunigs thanks for this PR.  I've built on top of it to provide a more robust fix. The bug was actually a little deeper in our code and you can see the fix on this PR: #1192 

Closing in favor of that PR, which includes your commit.

PR #1183, opened 2019-12-13
Refactor errorThreshold to fix data race

lmm:
## Description

Based on the discussion in https://github.com/projectcalico/libcalico-go/pull/1173, this PR fixes the data race by giving each watcherCache its own copy of the error threshold.

This isn't as nice as also removing the package-level variables but that would require modifying the syncer API and I'm not sure that's worth it. 

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


lmm:
@rafaelvanoni verified uts and fvs pass with `ginkgo --race`

PR #1192, opened 2020-02-06
Fix revision numbers for NetworkPolicy watches

spikecurtis:
## Description

Fixes revision numbers coming out of watch events, so that we can resume from where we left off if our watch gets cancelled (which happens periodically).

Fixes https://github.com/projectcalico/typha/issues/335

## Todos
- [x] Tests
- [x] Documentation not required
- [x] Release note

## Release Note

```release-note
Fixes an issue where Felix / Typha unnecessarily perform full resyncs of NetworkPolicies
```

### DIFF
--- a/lib/backend/k8s/k8s_test.go
+++ b/lib/backend/k8s/k8s_test.go
@@ -244,15 +244,25 @@ func (c cb) ExpectDeleted(kvps []model.KVPair) {
 // ExpectAddedEvent checks for an api.WatchAdded coming down a specific chan
 // this is used in several tests below
 func ExpectAddedEvent(events <-chan api.WatchEvent) *api.WatchEvent {
+	return expectEventOfType(events, api.WatchAdded)
+}
+
+// ExpectModifiedEvent checks for an api.WatchModified coming down a specific chan
+func ExpectModifiedEvent(events <-chan api.WatchEvent) *api.WatchEvent {
+	return expectEventOfType(events, api.WatchModified)
+}
+
+func expectEventOfType(events <-chan api.WatchEvent, type_ api.WatchEventType) *api.WatchEvent {
 	var receivedEvent *api.WatchEvent
+poll:
 	for i := 0; i < 10; i++ {
 		select {
 		case e := <-events:
 			// Got an event. Check it's OK.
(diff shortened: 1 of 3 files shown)


lib/backend/k8s/k8s_test.go
```diff
@@ -2304,6 +2326,152 @@ var _ = testutils.E2eDatastoreDescribe("Test Watch support", testutils.Datastore
...
+			Expect(err).ToNot(HaveOccurred())
+			Expect(l.KVPairs).To(HaveLen(4))
+
+			// Watch from part way
+			for i := 0; i<4; i++ {
+				revision := l.KVPairs[i].Revision
+				log.WithFields(log.Fields{
+					"revision": revision,
+					"key": l.KVPairs[i].Key.String()}).Info("[Test] starting watch")
+				_, err := c.Watch(ctx, model.ResourceListOptions{Kind: apiv3.KindNetworkPolicy}, revision)
```
caseydavenport: Do we want to stop the watch here as well?

lib/backend/k8s/resources/networkpolicy.go
```diff
@@ -237,6 +237,8 @@ func (c *networkPolicyClient) List(ctx context.Context, list model.ListInterface
 	}
 
 	// List all Namespaced Calico Network Policies.
+	// For consistency, we should split the revision that was input into CRD and K8s revisions,
+	// but the CRD client will fail on any non-empty revision anyway, so don't bother.
```
caseydavenport: > but the CRD client will fail on any non-empty revision anyway

Is this just because the `List` operation doesn't support revisions at all?

lib/backend/k8s/k8s_test.go
spikecurtis: Good catch.

lib/backend/k8s/resources/networkpolicy.go
spikecurtis: Yeah, exactly.  customresource.go line 245 has this to say

```
	if revision != "" {
		return nil, errors.New("Cannot List this resource type specifying a ResourceVersion")
	}
```

PR #1195, opened 2020-02-25
Add configuration values for BPF.

fasaxc:
## Description

Add configuration values for BPF.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/apis/v3/felixconfig.go
+++ b/lib/apis/v3/felixconfig.go
@@ -269,12 +269,14 @@ type FelixConfigurationSpec struct {
 	// BPFEnabled, if enabled Felix will use the BPF dataplane. [Default: false]
 	BPFEnabled *bool `json:"bpfEnabled,omitempty" validate:"omitempty"`
 	// BPFLogLevel controls the log level of the BPF programs when in BPF dataplane mode.  One of "Off", "Info", or
-	// "Debug". [Default: Off].
+	// "Debug".  The logs are emitted to the BPF trace pipe, accessible with the command `tc exec bpf debug`.
+	// [Default: Off].
 	BPFLogLevel string `json:"bpfLogLevel" validate:"omitempty,bpfLogLevel"`
 	// BPFDataIfacePattern is a regular expression that controls which interfaces Felix should attach BPF programs to
-	// in order to catch traffic to/from the network.  This needs to match the interfaces that calico workload traffic
+	// in order to catch traffic to/from the network.  This needs to match the interfaces that Calico workload traffic
 	// flows over as well as any interfaces that handle incoming traffic to nodeports and services from outside the
-	// cluster.  It should not match the workload interfaces (usually named cali...).  [Default: ^(en.*|eth.*|tunl0$)]
+	// cluster.  It should not match the workload interfaces (usually named cali...).
+	// [Default: ^(en.*|eth.*|tunl0$)]
 	BPFDataIfacePattern string `json:"bpfDataIfacePattern,omitempty" validate:"omitempty,regexp"`
 	// BPFConnectTimeLoadBalancingEnabled when in BPF mode, controls whether Felix installs the connection-time load
 	// balancer.  The connect-time load balancer is required for the host to be able to reach Kubernetes services
@@ -287,12 +289,12 @@ type FelixConfigurationSpec struct {
 	// is sent directly from the remote node.  In "DSR" mode, the remote node appears to use the IP of the ingress
 	// node; this requires a permissive L2 network.  [Default: Tunnel]
 	BPFExternalServiceMode string `json:"bpfExternalServiceMode,omitempty" validate:"omitempty,bpfServiceMode"`
-	// BPFKubeProxyIptablesCleanupEnabled, if enabled in BPF mode, Felix will pro-actively clean up kube-proxy's
-	// iptables chains.  Should only be enabled if kube-proxy is not running.  [Default: true]
+	// BPFKubeProxyIptablesCleanupEnabled, if enabled in BPF mode, Felix will proactively clean up the upstream
+	// Kubernetes kube-proxy's iptables chains.  Should only be enabled if kube-proxy is not running.  [Default: true]
 	BPFKubeProxyIptablesCleanupEnabled *bool `json:"bpfKubeProxyIptablesCleanupEnabled,omitempty" validate:"omitempty"`
-	// BPFKubeProxyMinSyncPeriod, controls the minimum time between updates to the dataplane for Kubernetes services.
-	// Lower values give reduced set-up latency.  Higher values reduce Felix CPU usage by batching up more work.
-	// [Default: 1s]
+	// BPFKubeProxyMinSyncPeriod, in BPF mode, controls the minimum time between updates to the dataplane for Felix's
+	// embedded kube-proxy.  Lower values give reduced set-up latency.  Higher values reduce Felix CPU usage by
+	// batching up more work.  [Default: 1s]
 	BPFKubeProxyMinSyncPeriod *metav1.Duration `json:"bpfKubeProxyMinSyncPeriod,omitempty" validate:"omitempty" configv1timescale:"seconds"`
 }
 
--- a/lib/apis/v3/felixconfig.go
+++ b/lib/apis/v3/felixconfig.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2019 Tigera, Inc. All rights reserved.
+// Copyright (c) 2017-2020 Tigera, Inc. All rights reserved.
 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


lib/apis/v3/felixconfig.go
```diff
@@ -265,6 +265,35 @@ type FelixConfigurationSpec struct {
 	// modes can use XDP. This is not recommended since it doesn't provide better performance than
 	// iptables. [Default: false]
 	GenericXDPEnabled *bool `json:"genericXDPEnabled,omitempty" confignamev1:"GenericXDPEnabled"`
+
+	// BPFEnabled, if enabled Felix will use the BPF dataplane. [Default: false]
+	BPFEnabled *bool `json:"bpfEnabled,omitempty" validate:"omitempty,"`
+	// BPFLogLevel controls the log level of the BPF programs when in BPF dataplane mode.  One of "Off", "Info", or
```
neiljerram: If we're exposing this, should we also say where someone can see these logs?

lib/apis/v3/felixconfig.go
```diff
@@ -265,6 +265,35 @@ type FelixConfigurationSpec struct {
...
+	// "Debug". [Default: Off].
+	BPFLogLevel string `json:"bpfLogLevel" validate:"omitempty,bpfLogLevel"`
+	// BPFDataIfacePattern is a regular expression that controls which interfaces Felix should attach BPF programs to
+	// in order to catch traffic to/from the network.  This needs to match the interfaces that calico workload traffic
+	// flows over as well as any interfaces that handle incoming traffic to nodeports and services from outside the
+	// cluster.  It should not match the workload interfaces (usually named cali...).  [Default: ^(en.*|eth.*|tunl0$)]
```
neiljerram: Is this Here To Stay, or just a band-aid for the tech preview?  If the latter, should we tag in some way to be clear that this config field might disappear again without further warning?

Does similar thinking apply to other fields here, or maybe to all the BPF fields?

neiljerram: > It should not match the workload interfaces (usually named cali...).

Feels like we should achieve this in code (using `InterfacePrefix`) and so not document it here.

lib/apis/v3/felixconfig.go
```diff
@@ -265,6 +265,35 @@ type FelixConfigurationSpec struct {
...
+	BPFDataIfacePattern string `json:"bpfDataIfacePattern,omitempty" validate:"omitempty,regexp"`
+	// BPFConnectTimeLoadBalancingEnabled when in BPF mode, controls whether Felix installs the connection-time load
+	// balancer.  The connect-time load balancer is required for the host to be able to reach Kubernetes services
+	// and it improves the performance of pod-to-service connections.  The only reason to disable it is for debugging
+	// purposes.  [Default: true]
```
neiljerram: If debugging-only, do we really need it in the data model here?

lib/apis/v3/felixconfig.go
```diff
@@ -265,6 +265,35 @@ type FelixConfigurationSpec struct {
...
+	BPFConnectTimeLoadBalancingEnabled *bool `json:"bpfConnectTimeLoadBalancingEnabled,omitempty" validate:"omitempty"`
+	// BPFExternalServiceMode in BPF mode, controls how connections from outside the cluster to services (node ports
+	// and cluster IPs) are forwarded to remote workloads.  If set to "Tunnel" then both request and response traffic
+	// is tunneled to the remote node.  If set to "DSR", the request traffic is tunneled but the response traffic
+	// is sent directly from the remote node.  In "DSR" mode, the remote node appears to use the IP of the ingress
+	// node; this requires a permissive L2 network.  [Default: Tunnel]
+	BPFExternalServiceMode string `json:"bpfExternalServiceMode,omitempty" validate:"omitempty,bpfServiceMode"`
+	// BPFKubeProxyIptablesCleanupEnabled, if enabled in BPF mode, Felix will pro-actively clean up kube-proxy's
```
neiljerram: No hyphen in proactively.

lib/apis/v3/felixconfig.go
```diff
@@ -265,6 +265,35 @@ type FelixConfigurationSpec struct {
...
+	// iptables chains.  Should only be enabled if kube-proxy is not running.  [Default: true]
```
neiljerram: Could this be easily autodetected, and the need for a config field removed?

neiljerram: (Or maybe a straightforward script outside our product?)

lib/apis/v3/felixconfig.go
```diff
@@ -265,6 +265,35 @@ type FelixConfigurationSpec struct {
...
+	BPFKubeProxyIptablesCleanupEnabled *bool `json:"bpfKubeProxyIptablesCleanupEnabled,omitempty" validate:"omitempty"`
+	// BPFKubeProxyMinSyncPeriod, controls the minimum time between updates to the dataplane for Kubernetes services.
+	// Lower values give reduced set-up latency.  Higher values reduce Felix CPU usage by batching up more work.
+	// [Default: 1s]
```
neiljerram: Confused about how kube proxy is involved if we also expect it not to be running...?

lib/validator/v3/validator_test.go
```diff
@@ -596,6 +596,18 @@ func init() {
...
+		Entry("should reject an invalid BPFLogLevel value 'badVal'", api.FelixConfigurationSpec{BPFLogLevel: "badVal"}, false),
+		Entry("should accept a valid BPFLogLevel value 'Info'", api.FelixConfigurationSpec{BPFLogLevel: "Info"}, true),
+		Entry("should accept a valid BPFLogLevel value 'Debug'", api.FelixConfigurationSpec{BPFLogLevel: "Debug"}, true),
+		Entry("should accept a valid BPFLogLevel value 'Off'", api.FelixConfigurationSpec{BPFLogLevel: "Off"}, true),
+
+		Entry("should reject a valid BPFExternalServiceMode value 'Foo'", api.FelixConfigurationSpec{BPFExternalServiceMode: "Foo"}, false),
+		Entry("should accept a valid BPFExternalServiceMode value 'Tunnel'", api.FelixConfigurationSpec{BPFExternalServiceMode: "Tunnel"}, true),
+		Entry("should accept a valid BPFExternalServiceMode value 'DSR'", api.FelixConfigurationSpec{BPFExternalServiceMode: "DSR"}, true),
+
+		Entry("should reject an invalid BPFDataIfacePattern value 'eth.*'", api.FelixConfigurationSpec{BPFDataIfacePattern: "*"}, false),
```
neiljerram: description cut'n'paste error

lib/apis/v3/felixconfig.go
```diff
@@ -265,6 +265,35 @@ type FelixConfigurationSpec struct {
...
 	// iptables. [Default: false]
 	GenericXDPEnabled *bool `json:"genericXDPEnabled,omitempty" confignamev1:"GenericXDPEnabled"`
+
+	// BPFEnabled, if enabled Felix will use the BPF dataplane. [Default: false]
+	BPFEnabled *bool `json:"bpfEnabled,omitempty" validate:"omitempty,"`
+	// BPFLogLevel controls the log level of the BPF programs when in BPF dataplane mode.  One of "Off", "Info", or
+	// "Debug". [Default: Off].
+	BPFLogLevel string `json:"bpfLogLevel" validate:"omitempty,bpfLogLevel"`
+	// BPFDataIfacePattern is a regular expression that controls which interfaces Felix should attach BPF programs to
+	// in order to catch traffic to/from the network.  This needs to match the interfaces that calico workload traffic
```
neiljerram: uppercase C in Calico

lib/apis/v3/felixconfig.go
fasaxc: There are use cases where you'd want something other than "ever interface that is not a workload interface"  For example, if you have a bond or something then you may want to run the BPF program on the `bond` instead of the `eth`  .  Also possible you have interfaces that you just want to exclude because you're doing something special with them.

Agree, we could auto-exclude workload interfaces based on the prefix.  That's not a bad idea.

lib/apis/v3/felixconfig.go
fasaxc: Think we may support running without this enabled more thoroughly in future so I put it just on this side of the debug-only line.

lib/apis/v3/felixconfig.go
fasaxc: Added this because it scratched an itch when we were testing.  Agree would be nice to move to auto-detection if possible but I'm not sure it is with things like hyperkube and different PID namespaces :thinking:

lib/apis/v3/felixconfig.go
fasaxc: This one refers to _our_ kube-proxy, inside felix.  Will see if I can make that clearer.

PR #1196, opened 2020-02-26
Return handle with assignment attributes

caseydavenport:
Return the handle, if any, with the assignment attributes. Calling code
may want to check if a handle was used to allocate this address, and if
so what handle was used.

- [ ] Tests
- [ ] Documentation
- [ ] Release note

```release-note
None required
```

### DIFF
--- a/lib/ipam/interface.go
+++ b/lib/ipam/interface.go
@@ -41,8 +41,8 @@ type Interface interface {
 	ReleaseIPs(ctx context.Context, ips []cnet.IP) ([]cnet.IP, error)
 
 	// GetAssignmentAttributes returns the attributes stored with the given IP address
-	// upon assignment.
-	GetAssignmentAttributes(ctx context.Context, addr cnet.IP) (map[string]string, error)
+	// upon assignment, as well as the handle used for assignment (if any).
+	GetAssignmentAttributes(ctx context.Context, addr cnet.IP) (map[string]string, *string, error)
 
 	// IPsByHandle returns a list of all IP addresses that have been
 	// assigned using the provided handle.
--- a/lib/ipam/ipam_test.go
+++ b/lib/ipam/ipam_test.go
@@ -284,9 +284,10 @@ var _ = testutils.E2eDatastoreDescribe("IPAM tests", testutils.DatastoreAll, fun
 		sentinelIP := net.ParseIP("10.0.0.1")
 
 		It("Should return ResourceNotExist on no valid pool", func() {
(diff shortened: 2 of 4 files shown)


lib/ipam/interface.go
```diff
@@ -41,8 +41,8 @@ type Interface interface {
 	ReleaseIPs(ctx context.Context, ips []cnet.IP) ([]cnet.IP, error)
 
 	// GetAssignmentAttributes returns the attributes stored with the given IP address
-	// upon assignment.
-	GetAssignmentAttributes(ctx context.Context, addr cnet.IP) (map[string]string, error)
+	// upon assignment, as well as the handle used for assignment (if any).
+	GetAssignmentAttributes(ctx context.Context, addr cnet.IP) (map[string]string, *string, error)
```
spikecurtis: Why don't we return an `AllocationAttribute` struct?

I think it would be unfortunate to use different terminology on this interface compared with the model.  E.g. "assignment attribute" = `AllocationAttribute.AttrSecondary` and "the handle used for assignment" = `AllocationAttribute.AttrPrimay`

lib/ipam/ipam_test.go
```diff
@@ -321,13 +324,16 @@ var _ = testutils.E2eDatastoreDescribe("IPAM tests", testutils.DatastoreAll, fun
...
+					HandleID: &handle,
 				}
 				err := ic.AssignIP(context.Background(), args)
 				Expect(err).NotTo(HaveOccurred())
 
-				attrs, err := ic.GetAssignmentAttributes(context.Background(), cnet.IP{sentinelIP})
+				attrs, returnedHandle, err := ic.GetAssignmentAttributes(context.Background(), cnet.IP{sentinelIP})
 				Expect(err).NotTo(HaveOccurred())
 				Expect(attrs).To(Equal(ipAttr))
+				Expect(handle).NotTo(BeNil())
```
spikecurtis:
```suggestion
				Expect(returnedHandle).NotTo(BeNil())
```

lib/ipam/interface.go
caseydavenport: Yeah.. the `AllocationAttribute` is an internal representation right now. 

I agree it would be nice to align this API with our models, but I think that is outside the scope of this change. This interface is already defined in different terminology than our internal models - it uses attributes and handles. For example, that's what callers of this interface pass in when allocating an address: https://github.com/projectcalico/libcalico-go/blob/c9b2cded4f4c1fae08477795ea451429a499cdcf/lib/ipam/ipam_types.go#L28-L33

PR #1207, opened 2020-04-03
Add an allow-all profile

lmm:
## Description

This PR adds a synthetic Profile resource with allow-all rules named `projectcalico-default-allow`. This resource is synthetic in that it is not retrieved from the datastore; rather, it is statically defined and returned as needed from the etcdv3 and kubernetes clients. Since it doesn't exist in the datastore, it doesn't get an rv from etcd; we're giving it an rv of `0`.

This profile is meant to be used in host endpoints where a mostly "default-allow" semantic in the absence of policy is desired. In particular, this profile will be attached to auto host endpoints.

The semantics of the allow-all profile are:
- getting/listing profiles always returns the allow-all profile resource
- watching profiles never returns the allow-all resource

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Add a new Profile with allow-all rules named `projectcalico-default-allow`. This profile can be used in host endpoints to provide default-allow in the absence of policy
```


lib/backend/k8s/resources/profile.go
```diff
@@ -128,6 +128,9 @@ func (c *profileClient) Get(ctx context.Context, key model.Key, revision string)
 	rk := key.(model.ResourceKey)
 	if rk.Name == "" {
 		return nil, fmt.Errorf("Profile key missing name: %+v", rk)
+	} else if rk.Name == "allow" {
```
caseydavenport, line 131: I think to do this properly we also need to return the profile in List() and Watch().

lmm: Thanks! List() calls Get() so that looks good but Watch() I've added a conditional to return an error

lib/backend/resources/static.go
```diff
@@ -0,0 +1,53 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	v3 "github.com/projectcalico/libcalico-go/lib/apis/v3"
+	"github.com/projectcalico/libcalico-go/lib/backend/model"
+)
+
+// This file contains helpers for creating hard-coded resources used in the Kubernetes backend.
+const (
+	// AllowProfileName is the name of an "allow-all" profile.
+	AllowProfileName = "projectcalico-allow-all"
```
lmm, line 27: I wonder if there's a good prefix we could use for this. We use `kns.` and `ksa.` already. A prefix could be a nice indicator that this profile was from Calico.

lib/backend/etcdv3/etcdv3.go
```diff
@@ -133,6 +135,14 @@ func (c *etcdV3Client) Create(ctx context.Context, d *model.KVPair) (*model.KVPa
 	logCxt := log.WithFields(log.Fields{"model-etcdKey": d.Key, "value": d.Value, "ttl": d.TTL, "rev": d.Revision})
 	logCxt.Debug("Processing Create request")
 
+	// Can't create the static allow-all profile.
```
caseydavenport, line 138: I think this sort of validation probably belongs up one level in the clientv3 package rather than in the individual backends.

lib/backend/k8s/resources/profile.go
```diff
@@ -240,7 +246,12 @@ func (c *profileClient) Watch(ctx context.Context, list model.ListInterface, rev
 	if len(rlo.Name) != 0 {
 		log.WithField("name", rlo.Name).Debug("Watching a single profile")
 		var err error
-		if strings.HasPrefix(rlo.Name, conversion.NamespaceProfileNamePrefix) {
```
caseydavenport, line 243: This handles validation, but I don't see anywhere in this PR that actually sends the Profile object over the watch.

I think we need adjustments to both List() and Watch() so that they return this new allow Profile() object when called.

lib/backend/resources/static.go
caseydavenport: We usually use the `projectcalico.org/` namespace for labels, but I don't think that `/` is allowed in resource names.

https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names

So, I think what you have is probably as good as we can get.

lib/backend/resources/static.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	// Embed the profile in a KVPair.
+	return &model.KVPair{
+		Key: model.ResourceKey{
+			Name: "allow",
+			Kind: v3.KindProfile,
+		},
+		Value:    profile,
+		Revision: "",
```
caseydavenport, line 51: Hm, for Watch / List this field might be important to populate, in case clients are looking at this.

lib/clientv3/profile_e2e_test.go
```diff
@@ -250,6 +250,44 @@ var _ = testutils.E2eDatastoreDescribe("Profile tests", testutils.DatastoreEtcdV
 			_, outError = c.Profiles().Get(ctx, name2, options.GetOptions{})
 			Expect(outError).To(HaveOccurred())
 			Expect(outError.Error()).To(ContainSubstring("resource does not exist: Profile(" + name2 + ") with error:"))
+			By("Creating Profile (projectcalico-allow-all) and expecting an error")
```
caseydavenport, line 253: Assuming that [this assertion still passes](https://github.com/projectcalico/libcalico-go/pull/1207/files#diff-846336915a67afeb5c93349c4c88e4efR244-R247), then I think there is a bug in your code in the List() implementation (called out above as well).

We should now always expect a single profile on List() calls.

lib/resources/static.go
```diff
@@ -0,0 +1,52 @@
...
+
+const (
+	// AllowProfileName is the name of an "allow-all" profile.
+	AllowProfileName = "projectcalico-allow-all"
+)
+
+// AllowProfile returns a single profile kvp with default allow rules.
+// This profile can be attached to host endpoints to allow traffic in the
+// absence of policy.
+func AllowProfile() *model.KVPair {
```
caseydavenport, line 32: This looks like a duplicate file? Do we need both this and the `lib/backend/resources` one?

lib/backend/k8s/resources/profile.go
caseydavenport: I don't think that's right - List() only calls Get() when an exact Name has been passed in.

For "list all" calls, it does not call Get().

lib/resources/static.go
lmm: That was an error. I removed it.

lib/clientv3/profile_e2e_test.go
lmm: Yep what I had wasn't quite right...

lib/clientv3/profile_e2e_test.go
```diff
@@ -265,10 +360,14 @@ var _ = testutils.E2eDatastoreDescribe("Profile tests", testutils.DatastoreEtcdV
 			Expect(err).NotTo(HaveOccurred())
 			be.Clean()
 
-			By("Listing Profiles with the latest resource version and checking for two results with name1/spec2 and name2/spec2")
+			By("Listing Profiles with the latest resource version and checking for 1 result")
```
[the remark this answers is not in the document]
lmm: Since you last authored your comment, the test has been updated.

lib/clientv3/profile.go
```diff
@@ -41,6 +43,14 @@ type profiles struct {
 // Create takes the representation of a Profile and creates it.  Returns the stored
 // representation of the Profile, and an error, if there is any.
 func (r profiles) Create(ctx context.Context, res *apiv3.Profile, opts options.SetOptions) (*apiv3.Profile, error) {
+	if res.Name == cresources.AllowProfileName {
+		return nil, cerrors.ErrorOperationNotSupported{
+			Operation:  "Create",
+			Identifier: cresources.AllowProfileName,
+			Reason:     "projectcalico-allow-all already exists",
```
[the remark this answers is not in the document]
lmm: 48c5bcb4a3c2462f278ed2390917767326cb2ea6

lib/clientv3/profile.go
```diff
@@ -55,6 +65,14 @@ func (r profiles) Create(ctx context.Context, res *apiv3.Profile, opts options.S
 // Update takes the representation of a Profile and updates it. Returns the stored
 // representation of the Profile, and an error, if there is any.
 func (r profiles) Update(ctx context.Context, res *apiv3.Profile, opts options.SetOptions) (*apiv3.Profile, error) {
+	if res.Name == cresources.AllowProfileName {
+		return nil, cerrors.ErrorOperationNotSupported{
+			Operation:  "Update",
+			Identifier: cresources.AllowProfileName,
+			Reason:     "Cannot modify a built-in profile",
```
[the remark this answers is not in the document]
lmm: That's much better, thanks. 86ee04b34229cfe1e5a22378f3e4508b3cb1914b

lib/backend/etcdv3/etcdv3.go
```diff
@@ -33,12 +33,14 @@ import (
...
 	"github.com/projectcalico/libcalico-go/lib/backend/model"
 	cerrors "github.com/projectcalico/libcalico-go/lib/errors"
+	"github.com/projectcalico/libcalico-go/lib/resources"
 )
 
 var (
 	clientTimeout    = 10 * time.Second
 	keepaliveTime    = 30 * time.Second
 	keepaliveTimeout = 10 * time.Second
+	allowProfileName = fmt.Sprintf("Profile(%s)", resources.AllowProfileName)
```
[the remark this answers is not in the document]
lmm: Thanks 4eb79f4dfa0157b057dd3ba8dde27b79d97719a1

lib/backend/etcdv3/etcdv3.go
```diff
@@ -412,12 +421,38 @@ func (c *etcdV3Client) List(ctx context.Context, l model.ListInterface, revision
...
+// getAllowAllProfile checks whether the statically defined allow-all profile
+// should be returned in the get/list result. If so, its kvpair is returned
+// otherwise it returns nil.
+func getAllowAllProfile(key string, revision string) *model.KVPair {
+	// Either we are getting all profiles or getting the allow-all profile
+	// specifically.
+	if key == "/calico/resources/v3/projectcalico.org/profiles/" || key == "/calico/resources/v3/projectcalico.org/profiles/projectcalico-allow-all" {
+		// If there is no rev at all, we are returning all resources so include the allow-all.
+		// If the rev=0 we also return all resources so include that profile.
+		// If rev=1 we also include the profile; any rev > 1 excludes the profile.
```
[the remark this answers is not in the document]
lmm: Maybe I can improve the comments.

If it helps, I also updated the PR description with a summary of the semantics around this allow-all resource and some of my thoughts around the decisions.

lib/backend/k8s/conversion/conversion.go
```diff
@@ -839,12 +839,14 @@ func (c Converter) SplitProfileRevision(rev string) (nsRev string, saRev string,
...
 	revs := strings.Split(rev, "/")
-	if len(revs) != 2 {
+	if len(revs) > 2 {
 		err = fmt.Errorf("ResourceVersion is not valid: %s", rev)
 		return
 	}
-
+	// ListAndWatch() can send 'ResourceVersion: "0"' as ListOptions.
 	nsRev = revs[0]
-	saRev = revs[1]
```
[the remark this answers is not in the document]
lmm: This is a bug fix that is already in private (`8fde42034188ded5d59fa64da0e18935d884eca9`). It fixes the case where getting/watching Profiles with rv=0 results in `ResourceVersion is not valid: 0`.

lib/backend/etcdv3/etcdv3.go
```diff
@@ -354,6 +358,11 @@ func (c *etcdV3Client) Get(ctx context.Context, k model.Key, revision string) (*
 	}
 	logCxt = logCxt.WithField("etcdv3-etcdKey", key)
 
+	// Handle the static allow-all profile.
+	if kvp := getAllowAllProfile(key, revision); kvp != nil {
```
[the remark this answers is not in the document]
lmm: I can change the name of the function. Maybe `maybeReturnAllowAllProfile` ? It's a mouthful but sort of captures the intent.

I initially had this logic inlined in both the Get and List functions but I thought it was clearer in a separate function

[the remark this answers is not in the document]
lmm: I'm also fine with inlining. Let me circle back to this once we've nailed down the semantics/rv.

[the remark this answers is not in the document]
caseydavenport: For the most part I usually feel that if a function's name needs to be "a mouthful" to describe its purpose then it is a sign that the contract of the function isn't well defined / scoped. Maybe this needs to be two smaller functions that can be composed together, or some other structuring?

lib/clientv3/profile_e2e_test.go
```diff
@@ -377,7 +478,15 @@ var _ = testutils.E2eDatastoreDescribe("Profile tests", testutils.DatastoreEtcdV
 			Expect(err).NotTo(HaveOccurred())
 			testWatcher3 := testutils.NewTestResourceWatch(config.Spec.DatastoreType, w)
 			defer testWatcher3.Stop()
-			testWatcher3.ExpectEvents(apiv3.KindProfile, []watch.Event{
+
+			allowAllRes, err := c.Profiles().Get(ctx, allowAllName, options.GetOptions{})
+			Expect(err).NotTo(HaveOccurred())
+
+			testWatcher3.ExpectEventsAnyOrder(apiv3.KindProfile, []watch.Event{
```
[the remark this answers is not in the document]
lmm: Fixed with ec5fccc3a0ed457c990a4fd14074f1ea12922669

fasaxc:
Surely we want to return this new profile if you list profiles with _any_ resource version and we never want to return it from a watch?

All the watcher/syncer machinery does a list and then a watch from the list's resource version.  I'd expect the list to return the new profile but for the watch to never return it since, from the watch's point of view, the profile has always existed.

lib/backend/k8s/resources/profile.go
lmm: We resolved this over Slack with Shaun. We'll always return the allow-all profile but never in watches.

lib/backend/etcdv3/etcdv3.go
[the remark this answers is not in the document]
lmm: This has been removed now

lmm:
We chatted about the semantics in Slack. We settled on the allow-all profile always being returned in lists but never in watches.

lib/backend/etcdv3/etcdv3.go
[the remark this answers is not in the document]
lmm: This area of code has been refactored based on our decision on semantics.

lib/backend/k8s/conversion/conversion.go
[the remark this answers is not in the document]
lmm: Actually... you're right. I don't think the fix from private is quite right. I've addressed this with: 58cb71234d27c1625d3946abaa0fc1346eaf24a3

In the process I also changed the allow-all rv from 1 to 0.

lib/backend/k8s/conversion/conversion_test.go
```diff
@@ -2586,10 +2586,6 @@ var _ = Describe("Test Namespace conversion", func() {
 		Expect(crdRev).To(Equal("1234"))
 		Expect(k8sRev).To(Equal(""))
 
-		By("failing to convert an invalid combined version")
```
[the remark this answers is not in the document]
lmm: Yep, reverted in 58cb71234d27c1625d3946abaa0fc1346eaf24a3

lib/backend/etcdv3/watcher.go
```diff
@@ -87,20 +87,35 @@ func (wc *watcher) watchLoop() {
 	// When this loop exits, make sure we terminate the watcher resources.
 	defer wc.terminateWatcher()
 
+	// If we are not watching a specific resource then this is a prefix watch.
```
caseydavenport, line 90: I think you copied this comment from below and also left it there.

Probably belongs in its original location rather than here?

lmm: Thanks, fixed in bd21637f122daad89f15c2e049a123bee54e5db5

lib/backend/k8s/conversion/conversion.go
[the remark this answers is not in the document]
caseydavenport: I think this `if` statement is redundant, right? 

We already do `saRev = revs[1]` unconditionally right above.

lib/backend/k8s/k8s_test.go
```diff
@@ -492,6 +506,114 @@ var _ = testutils.E2eDatastoreDescribe("Test Syncer API for Kubernetes backend",
...
+			Expect(found).To(BeFalse())
+		}
+
+		By("existing in our cache", func() {
+			expectedName := "projectcalico-default-allow"
+			Eventually(cb.GetSyncerValuePresentFunc(model.ProfileRulesKey{ProfileKey: model.ProfileKey{expectedName}}), slowCheck...).Should(BeTrue())
+		})
+
+		By("watching all profiles with any rv does not return an event for the default-allow profile", func() {
+			rvs := []string{"", "0"}
```
caseydavenport, line 539: The test says "with any rv" but we only use empty string and 0 here. Should probably throw in a few other RVs as well.

lib/backend/k8s/resources/profile.go
```diff
@@ -240,7 +248,15 @@ func (c *profileClient) Watch(ctx context.Context, list model.ListInterface, rev
...
+
+		// If we're watching the default-allow profile by name, then we exit
+		// with a new profileWatcher now and use fake watches in place
+		// of real ns and sa watchers. This profile watcher will never get any
+		// events because no events will occur for the static default-allow profile.
+		if rlo.Name == resources.DefaultAllowProfileName {
+			log.WithField("rv", revision).Debug("Creating watch on default-allow profile")
```
caseydavenport, line 257: Might want to put the word "fake" in this log, so as not to confuse some unknowing victim in the future. :p

lib/clientv3/profile.go
```diff
@@ -41,6 +44,14 @@ type profiles struct {
 // Create takes the representation of a Profile and creates it.  Returns the stored
 // representation of the Profile, and an error, if there is any.
 func (r profiles) Create(ctx context.Context, res *apiv3.Profile, opts options.SetOptions) (*apiv3.Profile, error) {
+	if res.Name == cresources.DefaultAllowProfileName {
+		return nil, cerrors.ErrorOperationNotSupported{
```
caseydavenport, line 48: Hm, I wonder if the "right" thing to do here is to return an "AlreadyExists" error, rather than a "NotSupoprted". 

I think probably so - this would be the expected behavior from the client's perspective. NotSupported is probably appropriate for Update/Delete, though.

lib/backend/k8s/conversion/conversion.go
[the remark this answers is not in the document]
lmm: Fixed in 3765a79

lib/clientv3/profile.go
lmm: d3c9ca2133f28c52cbaac10d658a252057e0adb1

lib/backend/k8s/resources/profile.go
lmm: Good call :) 98b6175235ed48e801f1aa53c5f4954ab049949b

lib/backend/k8s/k8s_test.go
lmm: 152bb5f3bd7fa2fbec132f2b9ab5acd985112c13

caseydavenport: That commit seems to show you added other values, but this diff still shows only `""` and `"0"`?

lib/backend/k8s/conversion/conversion.go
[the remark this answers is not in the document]
caseydavenport: There are also a bunch of tests below this that also say "any rv" that should be updated to include more than these two values.

lib/backend/k8s/k8s_test.go
caseydavenport: There are also a bunch of tests below this that also say "any rv" that should be updated to include more than these two values.

PR #1208, opened 2020-04-05
Refactor WorkloadEndpoint conversion and clients

Brian-McM:
This is to allow to allow for conversion from a pod to multiple workload endpoints and different implementations of that conversion.

This will help avoid conflicts when merging to the private repo where we have alternative workload endpoint conversion implementations.

### DIFF
--- a/lib/backend/k8s/conversion/workload_endpoint.go
+++ b/lib/backend/k8s/conversion/workload_endpoint.go
@@ -1,0 +1,32 @@
+// Copyright (c) 2016-2020 Tigera, Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// TODO move the WorkloadEndpoint converters to is own package. Some refactoring of the annotation and label constants
+// is necessary to avoid circular imports, which is why this has been deferred.
+package conversion
+
+import (
+	kapiv1 "k8s.io/api/core/v1"
+
+	"github.com/projectcalico/libcalico-go/lib/backend/model"
+)
+
+type WorkloadEndpointConverter interface {
+	VethNameForWorkload(namespace, podName string) string
+	PodToWorkloadEndpoints(pod *kapiv1.Pod) ([]*model.KVPair, error)
+}
+
+func NewWorkloadEndpointConverter() WorkloadEndpointConverter {
+	return &defaultWorkloadEndpointConverter{}
+}
--- a/lib/backend/k8s/resources/watcher.go
+++ b/lib/backend/k8s/resources/watcher.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2016-2018 Tigera, Inc. All rights reserved.
+// Copyright (c) 2016-2020 Tigera, Inc. All rights reserved.
 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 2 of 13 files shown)


lib/backend/k8s/resources/workloadendpoint.go
```diff
@@ -165,59 +170,108 @@ func (c *WorkloadEndpointClient) Get(ctx context.Context, key model.Key, revisio
...
-		}, revision)
+		wepID, err := c.converter.ParseWorkloadEndpointName(l.Name)
 		if err != nil {
-			switch err.(type) {
-			// Return empty slice of KVPair if the object doesn't exist, return the error otherwise.
-			case cerrors.ErrorResourceDoesNotExist:
+			return nil, err
+		}
+
+		// Pods can have multiple WorkloadEndpoints, so check that the "Endpoint" name is specified, otherwise we'll look
```
caseydavenport, line 202: I'm not sure if this is correct - if a `Name` and `Namespace` are both provided, that is an _exact_ identifer for a workload endpoint. We can't do any more specific lookup.

This is apparent by the fact that the `model.ResourceKey` that you pass in to the Get query only has those keys.

lib/backend/k8s/conversion/workload_endpoint.go
```diff
@@ -0,0 +1,50 @@
...
+type PodInterface struct {
+	IsDefaultInterface bool
+	NetworkName        string
+	HostSideIfaceName  string
+	InsidePodIfaceName string
+	InsidePodGW        net.IP
+	IPNets             []*cnet.IPNet
+}
+
+func GetWEPConverter() WorkloadEndpointConverter {
```
caseydavenport, line 48: Should probably call this `NewWorkloadEndpointConverter` to indicate it's creating a new struct, and to be more aligned with typical naming conventions. (we're not actually doing a "get" of any sort, it's just initializing a struct)

lib/backend/k8s/resources/workloadendpoint.go
Brian-McM: The name provided might not be the exact name, it might only be a prefix of the name (which excludes the interface name), and that prefix could uniquely identify a pod, and since the pods and workload used to have a one to one relationship that was good enough to uniquely identify a WorkloadEndpoint.

If you look at how Get worked before you'll see that we don't use the Endpoint. A good example of this scenerio and is in the CNI plugin.

lib/backend/k8s/conversion/workload_endpoint.go
Brian-McM: Good call, I'll change it.

lib/backend/k8s/resources/workloadendpoint.go
Brian-McM: I completely agree with restructuring, I'll do that.

> Otherwise, if the endpoint is not present, and neither is the Pod, we return an empty list - I think this case is not correct. We should return an error, OR return all WEPs on the provided node

 Yes you're right, this case is wrong. I was initially thinking this is what would happen before the these changes, but I see that's not correct, not sure what I was seeing when I did this before.

Brian-McM: Just thinking about this again, the list options could specify this as prefixed when it's really a uniquely identifying name, so I don't think the logic should be split on if it's a prefix or not, we need to know if the Endpoint name was found regardless.

Brian-McM: added "list" and "listUsingName" functions to split up the logic, let me know what you think.

lib/backend/k8s/resources/watcher.go
```diff
@@ -91,43 +102,45 @@ func (crw *k8sWatcherConverter) processK8sEvents() {
...
 						ClosedByRemote: true,
 					},
-				}
+				}}
 			} else {
 				// We have a valid event, so convert it.
-				e = crw.convertEvent(event)
-				if e == nil {
+				events = crw.convertEvent(event)
+				if events == nil {
```
caseydavenport, line 120: Should we do `len(events) == 0` here? Might be safer.

caseydavenport: I believe the main thing this branch covered were cases like this: https://github.com/projectcalico/libcalico-go/blob/master/lib/backend/k8s/resources/workloadendpoint.go#L265

Does it still do that properly? I think probably not because it will be converted into a slice with a single entry here: https://github.com/projectcalico/libcalico-go/pull/1208/files#diff-a77ab7c88a05a1a2822365096c0ed0fdR51

Is that right?

lib/backend/k8s/resources/workloadendpoint.go
```diff
@@ -165,60 +173,112 @@ func (c *WorkloadEndpointClient) Get(ctx context.Context, key model.Key, revisio
...
-			case cerrors.ErrorResourceDoesNotExist:
-				return &model.KVPairList{
-					KVPairs:  []*model.KVPair{},
-					Revision: revision,
-				}, nil
-			default:
-				return nil, err
+			return nil, err
+		}
+		// Find the WorkloadEndpoint that has a name matching the name in the given key
```
caseydavenport, line 244: What happens if we don't find a match here? I think we need to handle that.

e.g., if the wepID.Endpoint is an endpoint that doesn't actually exist.

Brian-McM: returning an empty list now

lib/backend/k8s/resources/watcher.go
Brian-McM: Think I resolved this bug now, added tests for it too

PR #1214, opened 2020-04-09
Add inline kubecfg

tmjd:
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/apiconfig/apiconfig.go
+++ b/lib/apiconfig/apiconfig.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2016-2017 Tigera, Inc. All rights reserved.
+// Copyright (c) 2016-2017,2020 Tigera, Inc. All rights reserved.
 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


lib/apiconfig/apiconfig.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2016-2017 Tigera, Inc. All rights reserved.
+// Copyright (c) 2016-2017,2020 Tigera, Inc. All rights reserved.
```
neiljerram, line 1: Do we have detailed guidance on this?  Personally I would have changed it to `2016-2020`.

tmjd: The pre-commit check, which prompted me to update this, showed a comma separated list was a valid format. I didn't look any deeper into it.

tmjd: Here is what is in the [pre-commit hook ](https://github.com/projectcalico/libcalico-go/blob/master/git-hooks/pre-commit#L67)
```
  echo "Example for updated files (use commas and year ranges):"
  echo "  Copyright (c) 2012,2015-${year} ${copyright_owner}. All rights reserved."
```

PR #1215, opened 2020-04-10
Wireguard configuration in FelixConfig and Node v3 resources.

realgaurav:
## Description

Adding Wireguard configuration to `FelixConfig` and `Node` v3 resources.

## Todos
- [x] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
Added Wireguard configuration.
```


go.sum
```diff
@@ -96,6 +96,9 @@ github.com/google/btree v0.0.0-20180813153112-4030bb1f1f0c/go.mod h1:lNA+9X1NB3Z
 github.com/google/go-cmp v0.2.0/go.mod h1:oXzfMopK8JAjlY9xF4vHSVASa0yLyX7SntLO5aqRK0M=
 github.com/google/go-cmp v0.3.0 h1:crn/baboCvb5fXaQ0IJ1SGTsTVrWpDsCWC8EGETZijY=
 github.com/google/go-cmp v0.3.0/go.mod h1:8QqcDgzrUqlUb/G2PQTWiueGozuR1884gddMywk6iLU=
+github.com/google/go-cmp v0.3.1/go.mod h1:8QqcDgzrUqlUb/G2PQTWiueGozuR1884gddMywk6iLU=
```
robbrockbank, line 99: run `go mod tidy`?

lib/backend/syncersv1/updateprocessors/configurationprocessor_test.go
```diff
@@ -663,6 +664,11 @@ func checkExpectedConfigs(kvps []*model.KVPair, dataType int, expectedNum int, e
 				node := kt.Name
 				Expect(node).To(Equal("mynode"))
 				name = nodeMarker
+			case model.WireguardKey:
+				node := kt.NodeName
+				Expect(node).To(Equal("mynode"))
+				name = nodeNameMarker
+				logrus.Warnf("Wireguard config: %v", kvp.Value)
```
robbrockbank, line 671: i assume this log should not be here.

lib/backend/syncersv1/updateprocessors/configurationprocessor_test.go
```diff
@@ -663,6 +664,11 @@ func checkExpectedConfigs(kvps []*model.KVPair, dataType int, expectedNum int, e
 				node := kt.Name
 				Expect(node).To(Equal("mynode"))
 				name = nodeMarker
+			case model.WireguardKey:
+				node := kt.NodeName
+				Expect(node).To(Equal("mynode"))
+				name = nodeNameMarker
```
robbrockbank, line 670: nodeNameMarker -> wireguardMarker

lib/backend/syncersv1/updateprocessors/felixnodeprocessor_test.go
```diff
@@ -29,7 +29,7 @@ var _ = Describe("Test the (Felix) Node update processor", func() {
 		Kind: apiv3.KindNode,
 		Name: "mynode",
 	}
-	numFelixConfigs := 5
+	numFelixConfigs := 6
```
robbrockbank, line 32: Need to add a test for the wireguard fields.

lib/apis/v3/node.go
```diff
@@ -50,6 +52,15 @@ type NodeSpec struct {
...
 	OrchRefs []OrchRef `json:"orchRefs,omitempty" validate:"omitempty"`
+
+	// Wireguard configuration for this node.
+	Wireguard *NodeWireguardSpec `json:"wireguard,omitempty" validate:"omitempty"`
+}
+
+type NodeStatus struct {
+	// PublicKey is the Wireguard public-key for this node.
+	// wireguardPublicKey validates if the string is a valid base64 encoded key.
+	PublicKey string `json:"publicKey,omitempty" validate:"omitempty,wireguardPublicKey"`
```
robbrockbank, line 63: Now that this is in the Status, the field name has lost the "wireguard" association.  So we probably want to either:
-  Add a wireguard section to the status
-  Rename this to be WireguardPublicKey

Both the spec and status now have only a single wireguard field.  Wondering if we should just remove the wireguard struct and put them as top level fields (such as the vxlan fields).  Given the precedence with vxlan I'm tempted to remove the extra level of hierarchy.  WDYT?  Perhaps ping Casey see what he thinks too.

realgaurav: >Rename this to be WireguardPublicKey

I like this suggestion.

>Given the precedence with vxlan I'm tempted to remove the extra level of hierarchy.
IMO, indirection is helpful if we have a scope of possible new fields in `NodeWireguardSpec` (e.g. interface ipv6 address or per-node MTU, when available). Not that it cannot exist on its own as VXLAN options, I think it keeps configuration clearer.

I will check with @caseydavenport. If he is of the same opinion, I can post an update merging NodeWireguardSpec into Spec.

caseydavenport: I haven't been involved in a lot of the UX design to-date, but I think my rough idea would have been:

- Felix always has wireguard on, with an explicit option to _disable_ it in FelixConfig.
- Felix looks for the `Node.Spec.Wireguard != nil` to indicate that it should do some wg stuff
- The `Node.Spec.Wireguard` section includes whatever configuration we might want to expose, including MTU, etc.


I get that this is different from VXLAN / IPIP in some ways, but I think is a better UX. Namely because it doesn't make the user go back and forth between FelixConfiguration and the Node resource somewhat arbitrarily. WDYT?

caseydavenport: Otherwise, I don't have too strong an opinion. This object is already sorta inconsistent with itself. I think a separate wireguard subsection is probably a good idea, in case we want to add other config options there (although it looks like this PR is putting them all in FelixConfig, hence my above comment)

realgaurav: Thanks Casey. I have kept the wireguard section in Spec and renamed `PublicKey` to `WireguardPublicKey` now.

lib/backend/model/node.go
```diff
@@ -219,3 +226,54 @@ func (options OrchRefListOptions) defaultPathRoot() string {
 func (options OrchRefListOptions) KeyFromDefaultPath(path string) Key {
 	return OrchRefKey{Hostname: options.Hostname}
 }
+
+// The Felix Wireguard Key.
+type WireguardKey struct {
```
caseydavenport, line 231: Understand this might be a late change, but we do send the entire node object to Felix now. https://github.com/projectcalico/libcalico-go/blob/master/lib/backend/syncersv1/updateprocessors/felixnodeprocessor.go#L150-L157

The idea being that we should, for new data, just use the v3 objects and not introduce new v1 config structures.

PR #1217, opened 2020-04-14
Add an OrchRef in KDD node conversion

caseydavenport:
## Description

This matches what we generate in etcd mode, so makes it easier for
common code to work for both data stores. 

e.g., useful for this PR https://github.com/projectcalico/kube-controllers/pull/467

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/backend/k8s/resources/node.go
+++ b/lib/backend/k8s/resources/node.go
@@ -233,6 +233,9 @@ func K8sNodeToCalico(k8sNode *kapiv1.Node, usePodCIDR bool) (*model.KVPair, erro
 	}
 	bgpSpec.IPv4IPIPTunnelAddr = annotations[nodeBgpIpv4IPIPTunnelAddrAnnotation]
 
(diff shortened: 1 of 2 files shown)


lib/backend/k8s/resources/node.go
```diff
@@ -233,6 +233,9 @@ func K8sNodeToCalico(k8sNode *kapiv1.Node, usePodCIDR bool) (*model.KVPair, erro
 	}
 	bgpSpec.IPv4IPIPTunnelAddr = annotations[nodeBgpIpv4IPIPTunnelAddrAnnotation]
 
+	// Add in an orchestrator reference back to the Kubernetes node name.
```
robbrockbank, line 236: nit:  Might be nice to state that this is the same as the node name in the kdd case

robbrockbank:
need to fix UTs

PR #1220, opened 2020-04-15
IPAM StrictAffinity: support to configure it.

gianlucam76:
Those changes are needed for adding a new ipam calicoctl command.

calicoctl ipam configure --strictaffinity=<true|false>

The reason for it is to let customer disable borrowing IP addresses and avoid so
extra routes being advertised when IPAM on a node needs to allocate an
IP address but no IPs are left in the block(s) assigned to it and no new
blocks can be allocated.

Those changes will also be used by another calico ipam show command which
will list all IP addresses currently borrowed (if any).

## Todos
- [ ] Tests
- [ ] Documentation

## Release Note

```release-note
None required
```


lib/ipam/ipam.go
```diff
@@ -1414,24 +1414,62 @@ func (c ipamClient) SetIPAMConfig(ctx context.Context, cfg IPAMConfig) error {
 		return errors.New("Cannot disable 'StrictAffinity' and 'AutoAllocateBlocks' at the same time")
 	}
 
-	allObjs, err := c.client.List(ctx, model.BlockListOptions{}, "")
-	if err != nil {
-		return err
+	// When moving from StrictAffinity true to false, there is a need
```
caseydavenport, line 1417: So, I spoke with @robbrockbank about this and I think we can simplify this a bit.

The reason we added the StrictAffinity field on the block object originally was to prevent extra datastore requests to retrieve the IPAMConfig object, except when allocating new blocks.

However, nowadays, the code is already querying the IPAMConfig object every time we need it, so the information in the block isn't serving that purpose. 

It _maybe_ has some minimal value in that writes to block are atomic, and so you won't get two clients seeing a different value for `StrictAffinity` for a given block (e.g., client1 queries IPAMConfig and it shows "true", client2 queries it and it shows "false"), but I don't think we need to worry about that. if the IPAMConfig is _changing_, then of course the user should expect that maybe some allocations got borrowed. 

So, I think we can do the following:
- Remove this code that updates existing blocks to match the IPAM config
- In the code where we do strict affinity checking, just remove the check for `block.StrictAffinity` altogether.

Then, we'll _always_ just be looking at the value in the IPAMConfig struct. WDYT?

gianlucam76: Makes sense, thanks for the detailed explanation! Will take care of it. Thank you!

lib/ipam/ipam.go
```diff
@@ -1598,3 +1636,62 @@ func (c ipamClient) GetUtilization(ctx context.Context, args GetUtilizationArgs)
...
+	for _, kvp := range blocks.KVPairs {
+		b := kvp.Value.(*model.AllocationBlock)
+		log.Debugf("Got block: %v", b)
+
+		// Find which pool this block belongs to.
+		for _, poolDetails := range details {
+			if b.CIDR.IsNetOverlap(poolDetails.CIDR) {
+				log.Debugf("Block CIDR %v belongs to pool %v", b.CIDR, poolDetails.Name)
+
+				blockDetail := &BlockDetail{CIDR: b.CIDR.IPNet, Node: *b.Affinity}
```
caseydavenport, line 1679: `b.Affinity` might be nil, so we shouldn't do a pointer dereference here without checking that.

lib/ipam/ipam_types.go
```diff
@@ -117,3 +117,39 @@ type PoolUtilization struct {
 	// Utilization for each of this pool's blocks.
 	Blocks []BlockUtilization
 }
+
+// Consumer represent a POD consuming an IP address.
```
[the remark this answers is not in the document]
gianlucam76: Removed any reference to POD. I am not sure though at this point how the table will look in case consumer is not a POD. Is there a way for me to simulate that?

lib/ipam/ipam_test.go
```diff
@@ -600,6 +600,128 @@ var _ = testutils.E2eDatastoreDescribe("IPAM tests", testutils.DatastoreAll, fun
...
+			v4, _, err = ic.AutoAssign(context.Background(), AutoAssignArgs{
+				Num4:     1,
+				Num6:     0,
+				Hostname: node2,
+			})
+			Expect(err).ToNot(HaveOccurred())
+			Expect(len(v4)).To(Equal(1))
+		})
+
+		It("should get no IP when ip stealing is disabled and no IP are available for node", func() {
```
caseydavenport, line 680: "no IP are available for node" -> "no blocks are available for the node"

caseydavenport: It might be a good idea to combine these two tests - i.e., show that you can modify the IPAMConfig and that the changes take effect.

e.g.,

- Allocate an IP
- Borrow an IP, show it works
- Enable StrictAffinity
- Borrow an IP, show it fails
- Disable StrictAffinity
- Borrow an IP, show it works

lib/ipam/ipam_types.go
```diff
@@ -117,3 +117,39 @@ type PoolUtilization struct {
 	// Utilization for each of this pool's blocks.
 	Blocks []BlockUtilization
 }
+
+// Consumer represent an entity consuming an IP address.
```
caseydavenport, line 121: I don't think these structs are used any more, so we can get rid of them.

caseydavenport: Oh, I see - they are still in the calicoctl PR.

I think we can move the definition of these into the calicoctl repo, at least, since that is the only place they are used. No need for them to be publicly defined here.

lib/ipam/ipam.go
```diff
@@ -1414,24 +1414,29 @@ func (c ipamClient) SetIPAMConfig(ctx context.Context, cfg IPAMConfig) error {
...
+			return err
+		}
 	}
 
 	// Write to datastore.
 	obj := model.KVPair{
 		Key:   model.IPAMConfigKey{},
 		Value: c.convertIPAMConfigToBackend(&cfg),
 	}
+	if old != nil {
```
caseydavenport, line 1431: I think we want to enforce that the calling code keeps the revision up-to-date, rather than always update it here.

That way we can get proper compare-and-swap behavior out of this function.

e.g., if two clients call `SetIPAMConfig` at the same time, only one will win and the other will need to retry.

caseydavenport: Hm, I see now, this function is a bit odd. We query it up front and compare to see if there's nothing to do.

I think we _could_ change this to not need to query the old state, and instead it just does an Apply on the provided object, but that is a change from the current behavior. It does, however, make it more aligned with how other create / update functions behave. 

So, I guess you're good to leave this as-is for now.

lib/ipam/ipam.go
```diff
@@ -601,12 +607,6 @@ func (c ipamClient) AssignIP(ctx context.Context, args AssignIPArgs) error {
 			}
 
 			log.Debugf("Block for IP %s does not yet exist, creating", args.IP)
-			cfg, err := c.GetIPAMConfig(ctx)
```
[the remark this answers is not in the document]
gianlucam76: thanks. allocationBlock.assign method used the block.StrictAffinity (which we removed) and I replaced it with IPAM.StrictAffinity so I moved out of the block to have it available in case of no error. I will add it back also inside the loop to address your comment.

lib/ipam/ipam_test.go
```diff
@@ -600,6 +600,128 @@ var _ = testutils.E2eDatastoreDescribe("IPAM tests", testutils.DatastoreAll, fun
 		})
 	})
 
+	Describe("IPAM Pool Details", func() {
+
+		AfterEach(func() {
+			deleteAllPools()
+		})
+
+		It("should get an IP from pool1 when explicitly requesting from that pool", func() {
```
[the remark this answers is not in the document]
gianlucam76: Removed it.

lib/ipam/ipam_test.go
```diff
@@ -600,6 +600,128 @@ var _ = testutils.E2eDatastoreDescribe("IPAM tests", testutils.DatastoreAll, fun
...
+				IPv4Pools: []cnet.IPNet{pool1},
+			}
+
+			v4, _, outErr := ic.AutoAssign(context.Background(), args)
+			Expect(outErr).NotTo(HaveOccurred())
+			Expect(pool1.IPNet.Contains(v4[0].IP)).To(BeTrue())
+		})
+	})
+
+	Describe("IPAM IP stealing", func() {
```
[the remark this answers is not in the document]
gianlucam76: Sorry about that I missed that. Made sure there are no other places where stealing is used.

lib/ipam/ipam_block_reader_writer.go
```diff
@@ -172,7 +172,6 @@ func (rw blockReaderWriter) claimAffineBlock(ctx context.Context, aff *model.KVP
 	affinityKeyStr := "host:" + host
 	block := newBlock(subnet)
 	block.Affinity = &affinityKeyStr
-	block.StrictAffinity = config.StrictAffinity
```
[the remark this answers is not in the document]
gianlucam76: Done

lib/ipam/ipam_types.go
gianlucam76: removed from libcalico.

lib/ipam/ipam_test.go
```diff
@@ -600,6 +600,100 @@ var _ = testutils.E2eDatastoreDescribe("IPAM tests", testutils.DatastoreAll, fun
 		})
 	})
 
+	Describe("IPAM IP borrowing", func() {
+		node1 := "host1"
+		node2 := "host2"
+
+		It("IP borrowing and StrictAffinity", func() {
```
caseydavenport, line 607: nit: generally, It() statements should read like a sentence describing what the test is checking.

e.g., `It("should respect IPAMConfig.StrictAffinity when it is changed")`

PR #1221, opened 2020-04-16
Add BPFDisableUnprivileged config param.

fasaxc:
## Description

Controls whether felix sets the sysctl to disable unprivileged BPF.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


fasaxc:
@neiljerram `nil` means "inherit".  So if you have a node-specific config with nil, that means "use the value from the global config".

PR #1223, opened 2020-04-18
Handle backend watch, if upstream closes channel[ClosedByRemote]

krishgobinath:
## Description
This PR handles Case 1, there are no changes made to Case 2.
1) If upstream Kubernetes API server closes watch the channel with backend libcalico-go, then stop the backend channels connected downstream and don't send any "ErrorTerminated" events back to downstream channels. 
2) If upstream Kubernetes API server, sends Api.WatchError to backend libcalico-go, wrap this status change as an "ErrorTerminated" to downstream. Downstream will resync the channel. 

Local Unit test passed.


lib/backend/k8s/resources/networkpolicy.go
```diff
@@ -437,32 +437,18 @@ func (npw *networkPolicyWatcher) processNPEvents() {
...
-				// so write a warning log and convert to a termination error.
-				log.Warn("Calico NP channel closed")
-				e = api.WatchEvent{
-					Type: api.WatchError,
-					Error: cerrors.ErrorWatchTerminated{
-						ClosedByRemote: true,
-						Err:            errors.New("Calico NP watch channel closed"),
-					},
-				}
+				//Upstream channel is closed by k8s, hence exit from processing watch events
```
caseydavenport, line 440: nit: please add a space after comment markers.

e.g., `// Upstream channel` rather than `//Upstream channel`

lib/backend/k8s/resources/watcher.go
```diff
@@ -104,16 +104,9 @@ func (crw *k8sWatcherConverter) processK8sEvents() {
...
-				// closed by the remote.
-				crw.logCxt.Debug("Watcher terminated by remote")
-				events = []*api.WatchEvent{{
-					Type: api.WatchError,
-					Error: cerrors.ErrorWatchTerminated{
-						Err:            fmt.Errorf("terminating error event from Kubernetes watcher: closed by remote"),
-						ClosedByRemote: true,
-					},
-				}}
+				//Watch channel is closed by remote k8s.hence, return from loop.
```
robbrockbank, line 107: nit:  space after //

lib/backend/k8s/resources/profile.go
```diff
@@ -378,32 +378,18 @@ func (pw *profileWatcher) processProfileEvents() {
...
-				// so write a warning log and convert to a termination error.
-				log.Warn("Profile, namespace watch channel closed.")
-				e = api.WatchEvent{
-					Type: api.WatchError,
-					Error: cerrors.ErrorWatchTerminated{
-						ClosedByRemote: true,
-						Err:            errors.New("Profile namespace watch channel closed."),
-					},
-				}
+				//Watch channel is closed by remote k8s.hence, return from loop.
```
robbrockbank, line 381: nit:  space after //   ... sim elsewhere.

krishgobinath: Taken care. Updated.

lib/backend/k8s/resources/watcher.go
krishgobinath: Taken care. Updated.

lib/backend/k8s/resources/profile.go
```diff
@@ -453,14 +434,10 @@ func (pw *profileWatcher) processProfileEvents() {
 		// Send the processed event.
 		select {
 		case pw.resultChan <- e:
-			// If this is an error event. check to see if it's a terminating one.
-			// If so, terminate this watcher.
+			// If this is an error event. terminate the watcher and bubble error to resenyc.
```
robbrockbank, line 437: Is this necessary now?  Can we not bubble up error events?

If the error is terminating a watcher then we'll see the channel close.  Can't we just bubble up the error without terminating?

lib/backend/k8s/resources/watcher.go
```diff
@@ -128,14 +121,10 @@ func (crw *k8sWatcherConverter) processK8sEvents() {
...
-					// If this is an error event, check to see if it's a terminating one (the
-					// convertEvent method will decide that).  If so, terminate this watcher.
+					// If this is an error event, terminate the watcher and bubble up the watcher error
 					if e.Type == api.WatchError {
-						crw.logCxt.WithError(e.Error).Debug("Watch event was an error event type")
-						if _, ok := e.Error.(cerrors.ErrorWatchTerminated); ok {
-							crw.logCxt.Debug("Watch event indicates a terminated watcher")
-							return
-						}
+						crw.logCxt.Debug("Watch event was an error event type")
```
robbrockbank, line 126: Same here - do we need to terminate.  I think we can just bubble up errors without terminating the watcher.  If the channel gets closed then we'll terminate the watcher through that mechanism.

lib/backend/watchersyncer/watchercache.go
```diff
@@ -117,30 +117,15 @@ mainLoop:
...
-						wc.logger.Debug("Watch was not closed by remote - full resync required")
-						wc.currentWatchRevision = ""
-						wc.onError()
-					}
-					wc.resyncAndCreateWatcher(ctx)
-				} else {
-					wc.onError()
-				}
+				// Handle a WatchError. This error triggered from upstream, all type
+				// of WatchError are treated equally,log the Error and trigger a full resync.
```
robbrockbank, line 121: This is a change of behavior isn't it?  Is it really necessary to resync every time we get an error?

krishgobinath: Yes. I was thinking to keep the error handling part lean. if Upstream reports a watch API error, log the error and act on that. I can bring back previous logic to wait until we reach the threshold limit.

lib/backend/k8s/resources/watcher.go
krishgobinath: Sure, I will update to bubble up the error.

lib/backend/k8s/resources/profile.go
krishgobinath: Sure, I will update to bubble up the error.

lib/backend/k8s/resources/watcher.go
robbrockbank: Still seems to be a problem in the PR :-)

robbrockbank: typo:  k8s.hence

lib/backend/k8s/resources/watcher.go
```diff
@@ -128,14 +121,9 @@ func (crw *k8sWatcherConverter) processK8sEvents() {
 				case crw.resultChan <- *e:
 					crw.logCxt.Debug("Kubernetes event converted and sent to backend watcher")
 
-					// If this is an error event, check to see if it's a terminating one (the
-					// convertEvent method will decide that).  If so, terminate this watcher.
+					// If this is an error event, bubble up the watcher error
```
robbrockbank, line 124: i think this comment is not necessary - all you are doing in the the branch below is logging.  Perhaps just remove this branch completely - it doesn't feel necessary to log here since the calling code can log the event details if it needs to.

lib/backend/watchersyncer/watchercache.go
robbrockbank: I think it's ok.  Just wondering if are simplifying then do we need the threshold error counts elsewhere in this module or whether we can just get rid of that completely?  WDYT?

lib/backend/k8s/resources/watcher.go
krishgobinath: Taken care both cases.

lib/backend/watchersyncer/watchercache.go
krishgobinath: Removed threshold-based error handling in watcher cache.

bcreane:
@krishgobinath please rebase and hit the "resolve" button for any and all issues that @robbrockbank raised that are now closed. Once everything is settled, @robbrockbank please take a final pass through the PR when you're able.

lib/backend/watchersyncer/watchercache.go
```diff
@@ -432,13 +410,10 @@ func (wc *watcherCache) markAsValid(resourceKey string) {
 	}
 }
 
-// onError signals to the syncer that this watcherCache is not in-sync if the number of consecutive errors
-// exceeds the error threshold.  See finishResync() for how the watcherCache goes back to in-sync.
+// onError signals to the syncer that this watcherCache is not in-sync, hence perform resync.
+// See finishResync() for how the watcherCache goes back to in-sync.
```
robbrockbank, line 414: Just spoke with @fasaxc .   Felix does not expect a transition to go from in-sync back to waiting-for-datastore.

I think we can simply remove onError().

lib/backend/watchersyncer/watchercache.go
```diff
@@ -432,13 +410,10 @@ func (wc *watcherCache) markAsValid(resourceKey string) {
...
 func (wc *watcherCache) onError() {
```
robbrockbank, line 415: I think we can either remove, or rename/repurpose this method.

We should be sending WaitForDatastore only if hasSynced is set to "true".  Otherwise we could send multiple WaitForDatastore events.  The owning watchersyncer counts WaitForDatastore/InSync events and they should balance out.

Perhaps:
-  Rename this to be "onWaitForDatastore()" method that sends a WaitForDatastore if not in-sync and then sets the in-sync flag to false.
-  Update the other place where we send WaitForDatastore to call this method instead.

krishgobinath: Update PR as per the above suggestion.

lib/backend/watchersyncer/watchercache.go
```diff
@@ -432,13 +410,12 @@ func (wc *watcherCache) markAsValid(resourceKey string) {
...
-func (wc *watcherCache) onError() {
-	wc.errors++
-	if wc.hasSynced && wc.errors > wc.errorThreshold {
-		wc.logger.WithFields(logrus.Fields{"errors": wc.errors, "threshold": wc.errorThreshold}).Debugf("Exceeded error threshold")
-		wc.hasSynced = false
+// If a syncer isn't in-sync state, the onWaitForDatastore method, sends WaitForDatastore event.
+// See finishResync() for how the watcherCache goes back to in-sync.
+func (wc *watcherCache) onWaitForDatastore() {
+	wc.logger.Debug("Send WaitforDatastore event if sync isn't in-sync")
+	if !wc.hasSynced {
```
robbrockbank, line 417: Should br `if wc.hasSynced`  -  i.e. we only indicate that we are going out-of-sync if we are already synced

lib/backend/watchersyncer/watchercache.go
```diff
@@ -117,30 +114,15 @@ mainLoop:
...
-					}
-					wc.resyncAndCreateWatcher(ctx)
-				} else {
-					wc.onError()
-				}
+				// Handle a WatchError. This error triggered from upstream, all type
+				// of WatchError are treated equally,log the Error and trigger a full resync.
+
+				wc.logger.WithField("EventType", event.Type).Errorf("Watch error received from Upstream")
+				wc.hasSynced = false
```
robbrockbank, line 121: Might be clearer to replace lines 121 and 122 with a call to onWaitForDatastore().

I think it's not strictly necessary because I think hasSynced is always true at this point - but it means you don't have to think about it :-)

lib/backend/watchersyncer/watchercache.go
krishgobinath: Updated PR to reflect as per suggestions.

lib/backend/watchersyncer/watchercache.go
krishgobinath: Thanks for catching this, I missed this one in the previous commit. Updated PR.

robbrockbank:
There is a new PR open for this - so closing this one,

PR #1229, opened 2020-04-27
Add global() selector 

lmm:
## Description

This PR adds support for the `global()` keyword

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Network policy now has the global() namespace selector which selects host endpoints or global network sets
```

### DIFF
--- a/lib/apis/v3/policy.go
+++ b/lib/apis/v3/policy.go
@@ -137,6 +137,9 @@ type EntityRule struct {
 	// For NetworkPolicy, an empty NamespaceSelector implies that the Selector is limited to selecting
 	// only workload endpoints in the same namespace as the NetworkPolicy.
 	//
+	// For NetworkPolicy, `global` NamespaceSelector implies that the Selector is limited to selecting
+	// only GlobalNetworkSet or HostEndpoint.
+	//
 	// For GlobalNetworkPolicy, an empty NamespaceSelector implies the Selector applies to workload
 	// endpoints across all namespaces.
 	NamespaceSelector string `json:"namespaceSelector,omitempty" validate:"omitempty,selector"`
--- a/lib/apis/v3/policy.go
+++ b/lib/apis/v3/policy.go
@@ -137,7 +137,7 @@ type EntityRule struct {
 	// For NetworkPolicy, an empty NamespaceSelector implies that the Selector is limited to selecting
 	// only workload endpoints in the same namespace as the NetworkPolicy.
 	//
-	// For NetworkPolicy, `global` NamespaceSelector implies that the Selector is limited to selecting
+	// For NetworkPolicy, `global()` NamespaceSelector implies that the Selector is limited to selecting
 	// only GlobalNetworkSet or HostEndpoint.
 	//
 	// For GlobalNetworkPolicy, an empty NamespaceSelector implies the Selector applies to workload
--- a/lib/validator/v3/validator.go
+++ b/lib/validator/v3/validator.go
@@ -94,7 +94,7 @@ var (
 	protocolPortsMsg      = "rules that specify ports must set protocol to TCP or UDP or SCTP"
 	protocolIcmpMsg       = "rules that specify ICMP fields must set protocol to ICMP"
 	protocolAndHTTPMsg    = "rules that specify HTTP fields must set protocol to TCP or empty"
-	globalSelectorMsg     = fmt.Sprintf("%v can only be used in an EntityRule namespaceSelector", globalSelector)
+	globalSelectorEntRule = fmt.Sprintf("%v can only be used in an EntityRule namespaceSelector", globalSelector)
 	globalSelectorOnly    = fmt.Sprintf("%v cannot be combined with other selectors", globalSelector)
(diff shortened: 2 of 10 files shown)


lib/apis/v3/policy.go
```diff
@@ -137,6 +137,9 @@ type EntityRule struct {
 	// For NetworkPolicy, an empty NamespaceSelector implies that the Selector is limited to selecting
 	// only workload endpoints in the same namespace as the NetworkPolicy.
 	//
+	// For NetworkPolicy, `global` NamespaceSelector implies that the Selector is limited to selecting
```
caseydavenport, line 140: global -> "Global()"

bcreane:
@lmm please check with @krishgobinath about adding validation - he's planning to add this in the next week or so.

neiljerram:
@lmm Is there already a doc PR for this?  I'd quite like to review that too, preferably before or at the same time as this one.  (Or a design doc?)

Specifically, as well as what @caseydavenport says, I'm wondering if `global()` must be the entire content of any selector field where it is used, or if it is allowed (and makes sense) to combine with other expressions, e.g. `global() && color == 'red'`.

lib/apis/v3/policy.go
```diff
@@ -137,6 +137,9 @@ type EntityRule struct {
 	// For NetworkPolicy, an empty NamespaceSelector implies that the Selector is limited to selecting
 	// only workload endpoints in the same namespace as the NetworkPolicy.
 	//
+	// For NetworkPolicy, `global()` NamespaceSelector implies that the Selector is limited to selecting
+	// only GlobalNetworkSet or HostEndpoint.
```
neiljerram, line 141: Not sure if this is right place to explain every possible selector expression.  Would be OK if it was a special case for just this field, but I don't believe that's the case - right?

lmm:
> Am I right that this is mostly copied from our Enterprise code? Please can you list or identify the changes that are additional to that? (Because I think there's less value in reviewing Enterprise-copied code.)

Thanks @neiljerram . The first 2 commits are cherry-picked from enterprise code. The last commit is newly added. The docs PR (also cherry-picked) is here: https://github.com/projectcalico/calico/pull/3643

> If there isn't already a design or doc that says this, please can you also summarise, for me, where global() is allowed, and why?

`global()` is allowed as a value for an EntityRule's namespaceSelector.

lib/apis/v3/policy.go
lmm: I think you're right, if you feel strongly about this, I can take it out, and port the changes over to the enterprise as well

lmm:
@neiljerram this is ready for another look. One comment re:

> I'm wondering if global() must be the entire content of any selector field where it is used, or if it is allowed (and makes sense) to combine with other expressions, e.g. global() && color == 'red'.

I had a chat with @doublek about this and the intent of the all(), global(), and projectcalico.org/name namespaceSelectors is that they should be used on their own and not in combination. I've added validation for this for global() and projectcalico.org/name.

But I left all() untouched as we already do some canonicalization where all() && some-other-selector results in the selector parser to reduce that to all(), and maybe we want to preserve that behaviour?

lib/validator/v3/validator.go
```diff
@@ -66,6 +67,13 @@ var (
 	// Hostname  have to be valid ipv4, ipv6 or strings up to 64 characters.
 	prometheusHostRegexp = regexp.MustCompile(`^[a-zA-Z0-9:._+-]{1,64}$`)
 
+	// global() cannot be used with other selectors.
+	andOr               = `(&&|\|\|)`
+	globalSelectorRegex = regexp.MustCompile(fmt.Sprintf(`%v global\(\)|global\(\) %v`, andOr, andOr))
```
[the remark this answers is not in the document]
neiljerram: Hmm, I see you already have a test for `"global()||all()"`, so perhaps ` ` in the RE already stands for any amount of whitespace?  (Otherwise, how is that test passing?)

lib/validator/v3/validator.go
```diff
@@ -66,6 +67,13 @@ var (
...
+
+	// projectcalico.org/name cannot be used with other selectors.
+	namespaceSelectorRegex = regexp.MustCompile(fmt.Sprintf(`%v projectcalico.org/name == ".*"|projectcalico.org/name == ".*" %v`, andOr, andOr))
```
neiljerram, line 75: This looks plausible to me, but:
- is it right to add this validation into the current PR?  is there a particular motivation for that?
- what about, for example, `projectcalico/name in {"red", "blue", ...}` - which I think should be allowed on its own, but probably not combined with anything else.

Unless there's a strong reason to include, I think it might be better to split the `projectcalico.org/name` validation out into another PR, so we can discuss its motivation and details more there.

lib/validator/v3/validator.go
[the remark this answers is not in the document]
lmm: We parse the selector string first and use the String() from that result for the validation matching. The parsing will trim out all of the whitespace.

lib/validator/v3/validator.go
lmm: I think you're right - I'll remove the `projectcalico.org/name` bits

PR #1233, opened 2020-05-05
Fix k8sWatcherConverter to return error anytime the event has an error

tmjd:
The change to expand k8s events to multiple api events inadvertently was dropping errors when it should not have. Because the event was a Type Error, which would prevent converting KVPairs,  then the loop did not run since there were no KVPairs, which then would cause no Error event to be returned.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/backend/k8s/resources/watcher.go
+++ b/lib/backend/k8s/resources/watcher.go
@@ -118,7 +118,7 @@ func (crw *k8sWatcherConverter) processK8sEvents() {
 				// We have a valid event, so convert it.
 				events = crw.convertEvent(event)
 				if len(events) == 0 {
-					crw.logCxt.Debug("Event converted to a no-op")
+					crw.logCxt.WithField("event", event).Debug("Event converted to a no-op")
 					continue
 				}
 			}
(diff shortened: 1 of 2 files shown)


lib/backend/k8s/resources/watcher.go
```diff
@@ -155,56 +155,67 @@ func (crw *k8sWatcherConverter) processK8sEvents() {
...
+	if len(kvps) == 0 {
+		return nil
+	}
+
+	return crw.buildEventsFromKVPs(kvps, kevent.Type)
+}
+
+func (crw *k8sWatcherConverter) buildEventsFromKVPs(kvps []*model.KVPair, t kwatch.EventType) []*api.WatchEvent {
+	var getEvent func(*model.KVPair) *api.WatchEvent
+	switch t {
```
Brian-McM, line 197: This switch seems a little redundant, you could set the get even in the above switch (since you already have the same cases). With that in mind, you could just remove this function loop through the kvps at the end of the previous function.

caseydavenport: Yeah, I originally suggested a split like this in my slack chat w/ Erik.

I was imagining it would be called inline the above switch statement, but see now that it's sorta tricky since we don't build the events the same for each type. I'm also OK with re-combining these two functions and just looping over the events like we used to.

caseydavenport: The intention for not doing it in the above switch is to de-couple handling of special events - e.g., Errors and Unexpected types.

That was sorta the bug in the first place - we were handling errors in the same loop as we were handling other update types, and didn't spot the pre-reqs to reach that loop in the first place.

Brian-McM: Makes sense

lib/backend/k8s/resources/watcher_test.go
```diff
@@ -0,0 +1,229 @@
...
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	log "github.com/sirupsen/logrus"
+	kwatch "k8s.io/apimachinery/pkg/watch"
+)
+
+var _ = Describe("Resources watcher ", func() {
+	Describe("convertEvent tests", func() {
+
+		kwc := k8sWatcherConverter{
```
Brian-McM, line 31: I think you should either just create this in each test, or create this in a BeforeEach function, otherwise changes to this watcherConverter in one tests will propagate to the next tests that are run.

PR #1237, opened 2020-05-18
Automatically generate CRD yamls from code

caseydavenport:
## Description

Start generating our CustomResourceDefinition files directly from the golang structures which implement them. 

This PR also:
- Adds schemas to the CRD yamls so that basic validation is performed.
- Switches to v1 instead of v1beta1. 

Note that there are some types used in our APIs which aren't compatible with the generation tools. For example:
- float64 used for ordering in NP / GNP
- `[]*int` used in IPAMBlock
- All of the `numorstring` package types - `Protocol`, `Port`, etc. 

To handle this, I've forked the controller-tools repository to add support to these fields. Ideally, we would eventually remove the need for this fork by switching each of those to supported types, or providing enhancements to the upstream controller-tools repo as is appropriate. 

You can see the changes I needed to make on this branch: https://github.com/caseydavenport/controller-tools/commits/support-floats

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
For NetworkPolicy and GlobalNetworkPolicy, the use of floating point values for the `spec.Order` field is now deprecated, and will be removed entirely in a future release. Please update your policies to use integer values for ordering. 
```

```release-note
Update included CustomResourceDefinitions to use the apiextensions/v1 API group and version, and include schemas for basic validation.
```


caseydavenport:
Tried updating this to use v1.18 k8s and seems we're running into this issue: https://github.com/kubernetes/kubernetes/issues/90904

Hopefully that's fixed before this change makes it into a Calico release.

Makefile
```diff
@@ -171,37 +181,34 @@ run-kubernetes-master: stop-kubernetes-master
 
 	# Wait until the apiserver is accepting requests.
 	while ! docker exec st-apiserver kubectl get nodes; do echo "Waiting for apiserver to come up..."; sleep 2; done
+	sleep 5
```
mgleung, line 184: Why do we need another sleep if we already have the `while` loop to check that the api server is running?

caseydavenport: Yeah, I was meaning to remove this and see. 

It seemed like there was some extra delay in setting up the v1 CRDs compared to the v1beta1 CRDs that was causing some flakiness before. I'll retry without, but seems like some APIs become available faster than others.

PR #1244, opened 2020-06-01
Properly set deleted on block affinities

caseydavenport:
## Description

We use a "deleted" flag on IPAM resources in KDD to help defend against race conditions where the object might be re-created. We do an Update() first to set the bit and invalidate other operations, and then other clients know to ignore it if it has the deleted flag set.

However, we weren't actually setting the deleted flag on block affinities!

This also changes the schema to a bool instead of a string to match blocks, which I think is maybe OK since this field isn't actually getting set today. However, it does mean that older clients (CNI plugins, etc) won't be able to write to this IPAM affinity API onces the new CRDs are applied (i.e., you'll need a new CNI plugin to match the new CRDs) since the schema will have changed. This does leave a small gap during rolling update from <=v3.14 to > v3.14, where old CNI plugins won't be able to allocate new block affinities, but I think that is probably acceptable. 

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Fix race condition during block affinity deletion
```

### DIFF
--- a/config/crd/crd.projectcalico.org_blockaffinities.yaml
+++ b/config/crd/crd.projectcalico.org_blockaffinities.yaml
@@ -39,6 +39,9 @@ spec:
               cidr:
                 type: string
               deleted:
+                description: Deleted indicates that this block affinity is being deleted.
+                  This field is a string for compatibility with older releases that
+                  mistakenly treat this field as a string.
                 type: string
               node:
                 type: string
--- a/lib/backend/k8s/resources/ipam_affinity.go
+++ b/lib/backend/k8s/resources/ipam_affinity.go
@@ -20,6 +20,7 @@ import (
 	"encoding/hex"
 	"fmt"
 	"reflect"
+	"strconv"
 
 	log "github.com/sirupsen/logrus"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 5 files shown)


config/crd/crd.projectcalico.org_blockaffinities.yaml
```diff
@@ -39,7 +39,7 @@ spec:
               cidr:
                 type: string
               deleted:
-                type: string
+                type: boolean
```
caseydavenport, line 42: This change does, unfortunately, mean that older clients (e.g., Calico v3.14 and older) won't be able to write the CRD any longer, because they don't have "omitempty" set so will try to write a string into a boolean field.

The alternative is that we keep it a string, and write some conversion between string / bool in the new code.

fasaxc: What happens if there's data in the API server that is now no longer valid?   Are we allowed to read it back?

I'm inclined to keep it as a string with a comment to say the type is a historic mistake; it is an internal datastructure after all.

caseydavenport: > What happens if there's data in the API server that is now no longer valid? Are we allowed to read it back?

yeah I tested this exact thing yesterday. We're allowed to read it back fine. We just can't write it.

caseydavenport:
@fasaxc thanks for the review, I think this is good to go now.

lib/backend/k8s/resources/ipam_affinity.go
```diff
@@ -140,9 +153,10 @@ func (c blockAffinityClient) toV3(kvpv1 *model.KVPair) *model.KVPair {
 				ResourceVersion: kvpv1.Revision,
 			},
 			Spec: apiv3.BlockAffinitySpec{
-				State: string(state),
-				Node:  host,
-				CIDR:  cidr,
+				State:   string(state),
+				Node:    host,
+				CIDR:    cidr,
+				Deleted: fmt.Sprintf("%t", kvpv1.Value.(*model.BlockAffinity).Deleted),
```
fasaxc, line 159: Do we want to write `""` if the value is `false`?  Probably doesn't matter.

lib/backend/k8s/resources/ipam_affinity_test.go
```diff
@@ -0,0 +1,61 @@
...
+		newKVP, err := be.Get(context.Background(), kvp.Key, "")
+		Expect(err).NotTo(HaveOccurred())
+
+		// Update it to be deleted.
+		newKVP.Value.(*model.BlockAffinity).Deleted = true
+		_, err = be.Update(context.Background(), newKVP)
+
+		// Can no longer see it.
+		_, err = be.Get(context.Background(), kvp.Key, "")
+		Expect(err).To(HaveOccurred())
```
fasaxc, line 59: Worth checking you get the not-exists error here, just in case it's hitting a different error path?

lib/backend/k8s/resources/ipam_affinity.go
fasaxc: Just to mimic omitempty and avoid saving the "default" value?

PR #1249, opened 2020-06-08
Config changes for setting source-destination-check on AWS

realgaurav:
## Description
Config changes for setting source-destination-check on AWS

## Release Note

```release-note
Added configuration for setting source-destination-check value on AWS EC2 instances.
```


lib/apis/v3/felixconfig.go
```diff
@@ -324,6 +324,9 @@ type FelixConfigurationSpec struct {
 	WireguardInterfaceName string `json:"wireguardInterfaceName,omitempty" validate:"omitempty,interface"`
 	// WireguardMTU controls the MTU on the Wireguard interface. See Configuring MTU [Default: 1420]
 	WireguardMTU *int `json:"wireguardMTU,omitempty"`
+
+	// Set source-destination-check on AWS EC2 instances. [Default: DoNothing]
```
caseydavenport, line 328: I think it is worth listing the acceptable values for this as well in the comment.

e.g., "Must be one of: [X, Y, Z]. Default: X" 

Ideally, we'd also include a kubebuilder tag to enumerate accepted values.

e.g.,
```
// +kubebuilder:validation:Enum=DoNothing,Enable,Disable
```

caseydavenport: Also, I think we should define constants for the accepted values.

e.g.,

```
type AWSSrcDstCheckOption string
const (
  DoNothingOption AWSSrcDstCheckOption = "DoNothing"
  . . .
)

...

AWSSrcDstCheck AWSSrcDstCheckOption
```

```

realgaurav: I followed the current pattern but sure I can update the comment.

About `kubebuilder` tag, what's this useful for?

caseydavenport: Current pattern isn't as good as it should be :D 

The kubebuilder tag is interpreted by the CRD generation code, and adds schema and validation information to the CRD itself, so that anyone who is (accidentally, or intentionally) writing data to the CRD directly doesn't put invalid data there.

caseydavenport: > Can you please add why you think this should be defined?:

This is typical for enum fields in our APIs and golang in general. It also means that code that is checking this value can use the constants, and that a discreet set of values are accepted.

PR #1252, opened 2020-06-08
Configuration model for active / passive felix implementation

caseydavenport:
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -221,6 +221,9 @@ run-kubernetes-master: stop-kubernetes-master
 		--server=http://localhost:8080 \
 		apply -f /manifests/test/namespaces.yaml
 
+	# Wait until some resources we expect to exist are present.
+	while ! docker exec st-apiserver kubectl get serviceaccount default; do echo "Waiting for default serviceaccount to be created..."; sleep 2; done
+
 ## Stop the local kubernetes master
 stop-kubernetes-master:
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -221,6 +221,9 @@ run-kubernetes-master: stop-kubernetes-master
 		--server=http://localhost:8080 \
 		apply -f /manifests/test/namespaces.yaml
 
+	# Wait until some resources we expect to exist are present.
```
caseydavenport, line 224: Spotted this when trying to run some tests with a focus - it takes 10-15 secs for the SA to show up.

We get away with it if we run all the tests at once, because we get lucky with ordering. but if you run a test that expects the SA to exist first, then it fails!

config/crd/crd.projectcalico.org_felixconfigurations.yaml
```diff
@@ -73,6 +73,10 @@ spec:
                   node appears to use the IP of the ingress node; this requires a
                   permissive L2 network.  [Default: Tunnel]'
                 type: string
+              bpfKubeProxyEndpointSlicesEnabled:
```
caseydavenport, line 76: This is cruft from another PR that forgot to run `make gen-crds`

PR #1262, opened 2020-06-18
BGP communities and configurable BGP port 

Suraiya-Hameed:
## Description
This PR has resource changes needed to support to BGP communities and configurable BGP port via APIs. 

As part of this PR, `BGPConfiguration` will accept new fields as follows,
```yaml
spec:
  listenPort: 179
  communities:
  - name: community-name-1
    value: 100:5632
  - name: community-name-2
    value: 30:100:5684
  prefixAdvertisements:
  - cidr: 192.168.1.0/16
    communities:
    - community-name-1
    - community-name-2
    - 54200:963
```
|Field|Description|
|-|-|
|spec.listenPort|Port where BGP protocol should listen. Defaults to 179|
|spec.communities|List of community name/identifier and their community name|
|spec.prefixAdvertisements|IPv4 or IPv6 CIRDs and their communities. If a route belongs to a prefix, all its corresponding communities will be applied. |

**spec.communities**
|Field|Description|
|-|-|
|name|Name/Identifier for the community|
|value|Community value either in `aa:nn` format or `aa:nn:mm`.</br>If `aa:nn` format is used, the community is sent as a standard community, where both `aa` and `nn` are 16 bit numbers.</br>If `aa:nn:mm` format is used, it will be sent as a BGP large community, where `aa`, `nn` and `mm` are all 32 bit numbers.`aa` is an AS Number, `nn` and `mm` are per-AS identifier.|

**spec.prefixAdvertisements**
|Field|Description|
|-|-|
|cidr|If a route belongs to this IPv4 or IPv6 CIRD, communities will be applied|
|communities|List of communities names whose values are defined(in spec.communities) or community value either in `aa:nn` or `aa:nn:mm` format.|

`PeerIP` field in `BGPPeer` will accept optional port number. Calico node will peer with BGP speaker over this port.
```yaml
spec:
  peerIP: 192.10.2.1:179
```
|Field|Description|
|-|-|
|peerIP|Format <IP> or <IPv4>:<port> or [<IPv6>]:<port> </br>IPv4 or IPv6 address of this peer followed by an optional port on this peer to which instances of BGP will be talking to. Default port value 179. |

Related issues:
https://github.com/projectcalico/calico/issues/3049
https://github.com/projectcalico/confd/pull/326
## Todos
- [x] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
Add support for BGP communities and configurable BGP ports
```


Suraiya-Hameed:
When working on confd side of these changes, I'll submit a separate PR to libcalico-go if bgpsyncer needs changes.

lib/apis/v3/bgppeer.go
```diff
@@ -44,8 +44,10 @@ type BGPPeerSpec struct {
 	// If this is not set, this represents a global peer, i.e. a peer that peers with
 	// every node in the deployment.
 	Node string `json:"node,omitempty" validate:"omitempty,name"`
-	// The IP address of the peer.
-	PeerIP string `json:"peerIP" validate:"omitempty,ip"`
+	// The IP address of the peer followed by an optional port number to peer with.
+	// If port number is given, format should be `[<IPv6>]:port` or `<IPv4>:<port>` for IPv4.
+	// Default port number is 179.
+	PeerIP string `json:"peerIP" validate:"omitempty"`
```
caseydavenport, line 50: It might be nice to define a new validator type for the new "IP:port" type so it could be re-used elsewhere if / when needed.

I think this PR just includes this in the validator code for the BGPPeerSpec type?

lib/backend/syncersv1/updateprocessors/bgpconfigprocessor.go
```diff
@@ -100,3 +105,39 @@ var svcClusterIpsToString = func(value interface{}) interface{} {
 
 	return strings.Join(ipCidrs, ",")
 }
+
+// return JSON encoded string of CommunityKVPair
+var communitiesToString = func(value interface{}) interface{} {
```
caseydavenport, line 110: I don't think there's actually any reason we need to convert these to a string. We should be able to send the structured data over as the Value of a KVPair.

caseydavenport: Same for prefixAdvertisements and listenPort

lib/clientv3/bgpconfig.go
```diff
@@ -153,3 +167,37 @@ func (r bgpConfigurations) ValidateDefaultOnlyFields(res *apiv3.BGPConfiguration
...
+// If a value in Spec.PrefixAdvertisements[x].Communities does not match standard or large BGP community format,
+// check if it is a community name with value already defined in Spec.Communities
+func (r bgpConfigurations) ValidatePrefixAdvertisements(res *apiv3.BGPConfiguration) error {
+	pa := res.Spec.PrefixAdvertisements
+	communityKVPairs := res.Spec.Communities
+
+	for _, v := range pa {
+		for _, community := range v.Communities {
+			if !standardCommunity.MatchString(community) && !largeCommunity.MatchString(community) {
+				isValidCommunity := isCommunityNameInKVParis(community, communityKVPairs)
```
caseydavenport, line 180: typo: "InKVParis"

However, a KVPair is a specific type used in libcalico, so let's not use KVPair here in the API (see earlier comments).

Maybe just `isCommunityDefined` or something like that.

lib/validator/v3/validator.go
```diff
@@ -1342,6 +1366,64 @@ func validateRouteTableRange(structLevel validator.StructLevel) {
...
+	}
+}
+
+func isValidCommunity(communityValue string, fieldName string, structLevel validator.StructLevel) bool {
+	if standardCommunity.MatchString(communityValue) {
+		validateCommunityValue(communityValue, fieldName, structLevel, false)
+	} else if largeCommunity.MatchString(communityValue) {
+		validateCommunityValue(communityValue, fieldName, structLevel, true)
+	} else {
+		return false
```
caseydavenport, line 1403: Do we need to `structLevel.ReportError` in this branch?

lib/backend/syncersv1/updateprocessors/bgpconfigprocessor.go
```diff
@@ -100,3 +105,39 @@ var svcClusterIpsToString = func(value interface{}) interface{} {
 
 	return strings.Join(ipCidrs, ",")
 }
+
+// return JSON encoded string of CommunityKVPair
```
Suraiya-Hameed, line 109: I'm not quite sure if I'm doing it the right way by marshalling the struct. Let me know whats the best option her, I'll revisit this again while doing confd changes, I'll have more understanding on it by then.

lib/apis/v3/bgppeer.go
```diff
@@ -44,8 +44,10 @@ type BGPPeerSpec struct {
 	// If this is not set, this represents a global peer, i.e. a peer that peers with
 	// every node in the deployment.
 	Node string `json:"node,omitempty" validate:"omitempty,name"`
-	// The IP address of the peer.
-	PeerIP string `json:"peerIP" validate:"omitempty,ip"`
+	// The IP address of the peer followed by an optional port number to peer with.
+	// If port number is given, format should be `[<IPv6>]:port` or `<IPv4>:<port>` for IPv4.
```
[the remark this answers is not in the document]
Suraiya-Hameed: yes, we would be using the `listenPort` of `BGPConfiguration` in that case.

lib/validator/v3/validator.go
Suraiya-Hameed: Not for the else condition, `isValidCommunity` is called in 2 places, one while validating `specs.community` and other while validating `specs.prefixAdvertisements.community`, in the later case, if it is not a valid community it could be a community name that is already defined (we do this validation in lib/clientv3/bgpconfig), and calling function takes care of `structLevel.ReportError` in former case.

lib/apis/v3/bgppeer.go
Suraiya-Hameed: right, it was in validator for BGPPeerSpec, I've refactored it now.

lib/backend/syncersv1/updateprocessors/bgpconfigprocessor.go
Suraiya-Hameed: I've updated the communities and prefixAdvertisements to send data as Value of a KVPair. 
I see string is return as it is in many cases, doing the same for listenPort which is just an integer.

lib/clientv3/bgpconfig.go
```diff
@@ -51,6 +57,10 @@ func (r bgpConfigurations) Create(ctx context.Context, res *apiv3.BGPConfigurati
 		return nil, err
 	}
 
+	if err := r.ValidatePrefixAdvertisements(res); err != nil {
```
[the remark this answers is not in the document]
Suraiya-Hameed: The reason I added this in clientv3 instead of validator is because we need access to both `sepc.communities` and `spec.prefixAdvertisement.communities` to make sure all the community names passed are pre-defined. 

The validator I added in lib/validator/v3/validator.go are specific to `sepc.communities` and `spec.prefixAdvertisement.communities`, so they don't have access to other fields. 

I see your point, I'll update validator to cover the entire `sepc` of BGPConfig, that way we can have all the validation in one place.

lib/apis/v3/bgppeer.go
```diff
@@ -44,8 +44,11 @@ type BGPPeerSpec struct {
...
+	// Default port number is 179.
+	// Target BGP port for this peer will be the `listenPort` set in `BGPConfiguration`
```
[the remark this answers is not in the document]
Suraiya-Hameed: It was to address https://github.com/projectcalico/libcalico-go/pull/1262#discussion_r443688754
hmm, it needs to be rephrased then.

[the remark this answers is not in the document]
Suraiya-Hameed:
```
// The IP address of the peer followed by an optional port number to peer with.
// If port number is given, format should be `[<IPv6>]:port` or `<IPv4>:<port>` for IPv4.
// Default port number is 179.
// Port number used by this peer if it is targeting other calico/node will be the `listenPort` set in `BGPConfiguration`
```
Is this more clear?

lib/backend/syncersv1/updateprocessors/bgpconfigprocessor.go
Suraiya-Hameed: Looks like we need to serialize it, https://github.com/projectcalico/confd/blob/master/pkg/backends/calico/client.go#L919.

caseydavenport: That serialization isn't relevant to this code, though, and should work just fine even without this logic. Note the generic json Marshaling at the end: https://github.com/projectcalico/libcalico-go/blob/dfe3bf9034f6399f882ed7c02f98030d472234c8/lib/backend/model/keys.go#L370

lib/backend/syncersv1/updateprocessors/bgpconfigprocessor.go
```diff
@@ -100,3 +103,42 @@ var svcClusterIpsToString = func(value interface{}) interface{} {
...
+	communities := value.([]apiv3.Community)
+	if len(communities) == 0 {
+		return nil
+	}
+	return &model.KVPair{
+		Key: model.ResourceKey{
+			Name: "communities",
+		},
+		Value: &communities,
+	}
```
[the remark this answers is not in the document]
caseydavenport: Yeah, I agree - these should not be necessary (I commented to the same effect as well).

lib/backend/syncersv1/updateprocessors/bgpconfigprocessor.go
Suraiya-Hameed: Returning struct as value of KVPair throws error in confd while serializing (https://github.com/projectcalico/confd/blob/master/pkg/backends/calico/client.go#L919).

Reason seems to be the valueType for GlobalBGPConfigKey is string, so we don't reach the line you pointed above.
https://github.com/projectcalico/libcalico-go/blob/dfe3bf9034f6399f882ed7c02f98030d472234c8/lib/backend/model/keys.go#L354

lib/apis/v3/bgppeer.go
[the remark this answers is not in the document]
neiljerram: To check my understanding, I think you are proposing this here:
`If the PeerIP address matches another Calico node, the PeerIP port is ignored.`

If that's right, I think it's wrong, because we don't have similar logic for the `ASNumber` field: 
https://github.com/projectcalico/confd/blob/master/pkg/backends/calico/client.go#L429:L432

Note that we specifically allow for a different `ASNumber` when computing any reverse peerings:
https://github.com/projectcalico/confd/blob/master/pkg/backends/calico/client.go#L462

Also your current proposal would rule out use cases that could be useful in future, where there is another BGP daemon (than calico-node) listening on a Calico node with a different port.

Net, I suggest removing this proposal behaviour, and the "Port number used by this peer if it is targeting other calico/node will be the `listenPort` set in `BGPConfiguration`" line.

lib/clientv3/bgpconfig.go
```diff
@@ -153,3 +170,19 @@ func (r bgpConfigurations) ValidateDefaultOnlyFields(res *apiv3.BGPConfiguration
 
 	return nil
 }
+
+// Currently the name of a configuration resource has a strict format.  It is either "default"
+// for the global default values, or "node.<nodename>" for the node specific values.  Returns an
+// error if the name is in neither format.
```
neiljerram, line 176: Why is this being added here and now?  If we think there's a requirement for this, would it be better to do in a separate PR that also applies the same validation to FelixConfiguration?

lib/clientv3/bgpconfig.go
```diff
@@ -39,6 +41,13 @@ type bgpConfigurations struct {
 	client client
 }
 
+var (
+	standardCommunity = regexp.MustCompile(`^(\d+):(\d+)$`)
+	largeCommunity    = regexp.MustCompile(`^(\d+):(\d+):(\d+)$`)
```
neiljerram, line 46: Are `standardCommunity` and `largeCommunity` still needed here?

lib/clientv3/bgpconfig.go
Suraiya-Hameed: This was an existing check, I copied it from updateprocessor as we will not go through that flow for BGPConfig anymore. https://github.com/projectcalico/libcalico-go/blob/master/lib/backend/syncersv1/updateprocessors/configurationprocessor.go#L332
It doesn't fit in anywhere in libcalico-go apart from clientv3/bgpconfig or we can move it to confd, log an error if the resource name doesn't match this check and skip processing that resource like we do now. I'll move this check to confd so existing behaviour doesn't change.

neiljerram: Thanks for explaining; I think I vote for confd.

lib/apis/v3/bgppeer.go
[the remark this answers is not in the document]
Suraiya-Hameed: >If the PeerIP address matches another Calico node, the PeerIP port is ignored.

I think that not the case.

My understanding of @caseydavenport comment was, say if calico/node has non-default listen_port, then the BGP peer during reverse peering will use this calico/node's ip and listen_port for peering (aka auto-detect calico/node's port number to peer with).

PeerIP port will be used only during forward peering where calico/node targets the peer, during revere peering where peers target the calico/node it will be based on the listen_port of the target node.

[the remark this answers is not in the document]
neiljerram: Thanks @Suraiya-Hameed, that specification sounds perfect to me, so I guess I just misunderstood this sentence of the comment.  Can you think of a way to clarify it?  In particular I think "if it is targeting other calico/node" is misleading.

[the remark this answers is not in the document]
Suraiya-Hameed: I've updated one of the confd test to cover this case https://github.com/projectcalico/confd/blob/946cbe7428a6e09476fa90b00de33ce67a715187/tests/mock_data/calicoctl/explicit_peering/global/input.yaml

In the `input.yaml`, when `kube-master` is peering with `bgppeer-2`,  `bgppeer-2` has `PeerIP` without port, plus it is also a calico/node(`kube-node-2`). If `kube-node-2` has a `listenPort` set, we will use that during peering.

This will be the outcome, 
```
# For peer /global/peer_v4/10.192.0.4
protocol bgp Global_10_192_0_4 from bgp_template {
  neighbor 10.192.0.4 port 150 as 64567; 
  #`bgppeer-2` dint have any port set, but `kube-node-2` had a listenPort, so we are using that port when `kube-master` adds `bgppeer-2` as neighbour.
}
```

neiljerram:
@caseydavenport Are there any points from you that still need to be addressed on this PR?  If not, could you update your review?

@Suraiya-Hameed Subject to Casey's approval, I recommend rebasing and merging this.  Then you can also merge a pin bump for Typha, and that will simplify the remaining iterations on the confd code.

lib/apis/v3/bgpconfig.go
```diff
@@ -55,6 +55,15 @@ type BGPConfigurationSpec struct {
 	// ServiceClusterIPs are the CIDR blocks from which service cluster IPs are allocated.
 	// If specified, Calico will advertise these blocks, as well as any cluster IPs within them.
 	ServiceClusterIPs []ServiceClusterIPBlock `json:"serviceClusterIPs,omitempty" validate:"omitempty,dive" confignamev1:"svc_cluster_ips"`
+
+	// Communities contain list of BGP community values and their arbitrary names for tagging routes.
+	Communities []Community `json:"communities,omitempty" validate:"omitempty,dive" confignamev1:"communities"`
+
+	// PrefixAdvertisement lists the communities to be advertised per prefix.
```
caseydavenport, line 62: This comment (while technically correct for now) isn't how I think we should document this field.

It should be something like this:

```
PrefixAdvertisements contains per-prefix advertisement configuration.
```

caseydavenport: The distinction being that while Communities are the only thing you can currently set here, the PrefixAdvertisements sub-section is intended to be more general.

Also, this comment has a typo: "PrefixAdvertisement" -> "PrefixAdvertisements"

lib/apis/v3/bgpconfig.go
```diff
@@ -67,6 +76,26 @@ type ServiceClusterIPBlock struct {
 	CIDR string `json:"cidr,omitempty" validate:"omitempty,net"`
 }
 
+// Community contains community value and its name.
+// BGP community value can be of format `aa:nn` or `aa:nn:mm`.
+// If `aa:nn` format is used, standard community will be used, where `aa` and `nn` are 16 bit number.
+// If `aa:nn:mm` format is used, large community will be used, where `aa`, `nn` and `mm` are 32 bit number.
+// Where, `aa` is an AS Number, `nn` and `mm` are per-AS identifier.
+type Community struct {
+	Name  string `json:"name,omitempty" validate:"required"`
```
caseydavenport, line 85: These fields should have user-facing comments on them (they can be brief).

lib/apis/v3/bgpconfig.go
```diff
@@ -67,6 +76,26 @@ type ServiceClusterIPBlock struct {
...
+}
+
+// PrefixAdvertisement contains communities that should be applied to a route if it belong to a prefix.
+// `Communities` can be list of either community names already defined in `Specs.Communities` or community value of format `aa:nn` or `aa:nn:mm`.
+// If `aa:nn` format is used, standard community will be used, where `aa` and `nn` are 16 bit number.
+// If `aa:nn:mm` format is used, large community will be used, where `aa`, `nn` and `mm` are 32 bit number.
+// Where,`aa` is an AS Number, `nn` and `mm` are per-AS identifier.
+type PrefixAdvertisement struct {
+	CIDR        string   `json:"cidr,omitempty" validate:"required,net"`
+	Communities []string `json:"communities,omitempty" validate:"required"`
```
caseydavenport, line 96: Same here - should have comments on CIDR and Communities.

lib/apis/v3/bgpconfig.go
```diff
@@ -67,6 +76,26 @@ type ServiceClusterIPBlock struct {
 	CIDR string `json:"cidr,omitempty" validate:"omitempty,net"`
 }
 
+// Community contains community value and its name.
+// BGP community value can be of format `aa:nn` or `aa:nn:mm`.
```
caseydavenport, line 80: This is misplaced - comment describing the syntax for a community should be on the `Value` section, not on the community struct.

lib/apis/v3/bgpconfig.go
```diff
@@ -67,6 +76,26 @@ type ServiceClusterIPBlock struct {
...
+// If `aa:nn:mm` format is used, large community will be used, where `aa`, `nn` and `mm` are 32 bit number.
+// Where, `aa` is an AS Number, `nn` and `mm` are per-AS identifier.
+type Community struct {
+	Name  string `json:"name,omitempty" validate:"required"`
+	Value string `json:"value,omitempty" validate:"required"`
+}
+
+// PrefixAdvertisement contains communities that should be applied to a route if it belong to a prefix.
+// `Communities` can be list of either community names already defined in `Specs.Communities` or community value of format `aa:nn` or `aa:nn:mm`.
+// If `aa:nn` format is used, standard community will be used, where `aa` and `nn` are 16 bit number.
```
caseydavenport, line 91: Same here - struct level comments should describe the struct, and field-level comments should describe the individual fields. 

This is important because these comments will get auto-generated into API documentation (eventually) and custom resource definitions.

caseydavenport: My suggestion for the comment here would be:

```
PrefixAdvertisement configures advertisement properties for the specified CIDR. 
```

lib/apis/v3/bgpconfig.go
caseydavenport, line 85: We should have standard name field validation performed on this name as well.

lib/apis/v3/bgpconfig.go
```diff
@@ -55,6 +55,17 @@ type BGPConfigurationSpec struct {
 	// ServiceClusterIPs are the CIDR blocks from which service cluster IPs are allocated.
 	// If specified, Calico will advertise these blocks, as well as any cluster IPs within them.
 	ServiceClusterIPs []ServiceClusterIPBlock `json:"serviceClusterIPs,omitempty" validate:"omitempty,dive" confignamev1:"svc_cluster_ips"`
+
+	// Communities contain list of BGP community values and their arbitrary names for tagging routes.
```
caseydavenport, line 59: "Communities is a list" rather than "Communities contain list"

lib/apis/v3/bgpconfig.go
```diff
@@ -67,6 +78,29 @@ type ServiceClusterIPBlock struct {
...
+	// Where, `aa` is an AS Number, `nn` and `mm` are per-AS identifier.
+	// +kubebuilder:validation:Pattern=`^(\d+):(\d+)$|^(\d+):(\d+):(\d+)$`
+	Value string `json:"value,omitempty" validate:"required"`
+}
+
+// PrefixAdvertisement configures advertisement properties for the specified CIDR.
+type PrefixAdvertisement struct {
+	// CIDR for which properties should be advertised.
+	CIDR string `json:"cidr,omitempty" validate:"required,net"`
+	// `Communities` can be list of either community names already defined in `Specs.Communities` or community value of format `aa:nn` or `aa:nn:mm`.
```
caseydavenport, line 97: Drop the backticks around Communities here.

PR #1264, opened 2020-06-26
Added felix config param; FeatureDetectOverride

uablrek:
## Description

New feature.

This complements the PR https://github.com/projectcalico/felix/pull/2310. It makes it possible to set the parameter with an environment variable in the Calico manifest for K8s;
```
            - name: FELIX_FEATUREDETECTOVERRIDE
              value: "MASQFullyRandom=false"
```

It is part of the fix for issue; https://github.com/projectcalico/felix/issues/2089

Tested manually with the https://github.com/projectcalico/felix/pull/2310 applied.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/apis/v3/felixconfig.go
+++ b/lib/apis/v3/felixconfig.go
@@ -87,10 +87,10 @@ type FelixConfigurationSpec struct {
 	// attempts to acquire the iptables lock if it is not available. Lower values make Felix more
 	// responsive when the lock is contended, but use more CPU. [Default: 50ms]
 	IptablesLockProbeInterval *metav1.Duration `json:"iptablesLockProbeInterval,omitempty" configv1timescale:"milliseconds" confignamev1:"IptablesLockProbeIntervalMillis"`
-	// IptablesFeatureDetectOverride is used to override the feature detection. Values are specified in
+	// FeatureDetectOverride is used to override the feature detection. Values are specified in
 	// a comma separated list. A value of "true" or "false" will force the feature, empty or omitted values
 	// are auto-detected. Features; SNATFullyRandom, MASQFullyRandom, RestoreSupportsLock
-	IptablesFeatureDetectOverride string `json:iptablesFeatureDetectOverride",omitempty"`
+	FeatureDetectOverride string `json:iptablesFeatureDetectOverride",omitempty"`
 	// IpsetsRefreshInterval is the period at which Felix re-checks all iptables
 	// state to ensure that no other process has accidentally broken Calico’s rules. Set to 0 to
 	// disable iptables refresh. [Default: 90s]
--- a/lib/apis/v3/felixconfig.go
+++ b/lib/apis/v3/felixconfig.go
@@ -90,7 +90,7 @@ type FelixConfigurationSpec struct {
 	// FeatureDetectOverride is used to override the feature detection. Values are specified in
 	// a comma separated list. A value of "true" or "false" will force the feature, empty or omitted values
 	// are auto-detected. Features; SNATFullyRandom, MASQFullyRandom, RestoreSupportsLock
-	FeatureDetectOverride string `json:iptablesFeatureDetectOverride",omitempty"`
+	FeatureDetectOverride string `json:"featureDetectOverride",omitempty"`
 	// IpsetsRefreshInterval is the period at which Felix re-checks all iptables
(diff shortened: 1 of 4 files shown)


lib/apis/v3/felixconfig.go
```diff
@@ -87,6 +87,10 @@ type FelixConfigurationSpec struct {
 	// attempts to acquire the iptables lock if it is not available. Lower values make Felix more
 	// responsive when the lock is contended, but use more CPU. [Default: 50ms]
 	IptablesLockProbeInterval *metav1.Duration `json:"iptablesLockProbeInterval,omitempty" configv1timescale:"milliseconds" confignamev1:"IptablesLockProbeIntervalMillis"`
+	// IptablesFeatureDetectOverride is used to override the feature detection. Values are specified in
+	// a comma separated list. A value of "true" or "false" will force the feature, empty or omitted values
+	// are auto-detected. Features; SNATFullyRandom, MASQFullyRandom, RestoreSupportsLock
+	IptablesFeatureDetectOverride string `json:iptablesFeatureDetectOverride",omitempty"`
```
fasaxc, line 93: Suggest making this just featureDetectOverride here too.

I think we need validation here to stop a user from setting a garbage value that would confuse felix.  Unfortunately, that means duplicating the function from felix.  Here's a sample commit that did some validation + tests: https://github.com/projectcalico/libcalico-go/commit/7f5b8c620b5caf2cf79e757d905e03e58db317c8#diff-5b3418236d048b3157da11cc90e65ad8

lib/apis/v3/felixconfig.go
```diff
@@ -87,6 +87,10 @@ type FelixConfigurationSpec struct {
 	// attempts to acquire the iptables lock if it is not available. Lower values make Felix more
 	// responsive when the lock is contended, but use more CPU. [Default: 50ms]
 	IptablesLockProbeInterval *metav1.Duration `json:"iptablesLockProbeInterval,omitempty" configv1timescale:"milliseconds" confignamev1:"IptablesLockProbeIntervalMillis"`
+	// FeatureDetectOverride is used to override the feature detection. Values are specified in
+	// a comma separated list. A value of "true" or "false" will force the feature, empty or omitted values
+	// are auto-detected. Features; SNATFullyRandom, MASQFullyRandom, RestoreSupportsLock
+	FeatureDetectOverride string `json:"featureDetectOverride",omitempty"`
```
fasaxc, line 93: `"featureDetectOverride",omitempty"` should be be `"featureDetectOverride,omitempty"` (the omitempty is an instruction to the json parser so it needs to be inside the `json:` tag.

lib/apis/v3/felixconfig.go
```diff
@@ -87,6 +87,10 @@ type FelixConfigurationSpec struct {
 	// attempts to acquire the iptables lock if it is not available. Lower values make Felix more
 	// responsive when the lock is contended, but use more CPU. [Default: 50ms]
 	IptablesLockProbeInterval *metav1.Duration `json:"iptablesLockProbeInterval,omitempty" configv1timescale:"milliseconds" confignamev1:"IptablesLockProbeIntervalMillis"`
+	// FeatureDetectOverride is used to override the feature detection. Values are specified in
+	// a comma separated list. A value of "true" or "false" will force the feature, empty or omitted values
```
fasaxc, line 91: Think you need to say what the format of the items in the list is.

uablrek: I added an example. That's what I like to have when reading doc myself.

lib/apis/v3/felixconfig.go
uablrek: Ooops, typo. Thanks for spotting it.

uablrek:
Updated. I will work on validations later.

uablrek:
Test fails with;
```
FelixConfiguration.crd.projectcalico.org "myfelixconfig" is invalid: spec.featureDetectOverride: Required value
```
but the value shall be optional. I will try to find out what's wrong.

uablrek:
@fasaxc Shouldn't the "Required value" fail in tests (comment above) go away now when the "omitempty" bug is fixed?

fasaxc:
Did you do `make gen-files`?  I think that's how you update the generated schema.

PR #1266, opened 2020-07-04
Maintaining original next hop on specific bgppeer

gunboe:
## Description
An option to keep the original nexthop field when routes are sent to a BGP Peer. Setting "true" configures the selected BGP Peers node to use the "next hop keep;" instead of "next hop self;"(default) in the specific branch of the Node on "bird.cfg". Thus routes will be send to the external Router maintaining the BGP Original Next Hop. So traffic incoming from external Router can be forwarded directly to the Nodes, without through the Route Reflector.

This is a **New Feature** that target:
 - (https://github.com/projectcalico/calico/issues/3666)
 - (https://github.com/projectcalico/confd/issues/308)
 - (https://github.com/projectcalico/node/issues/279)
 - (https://github.com/projectcalico/confd/pull/257#issuecomment-567058423)
and others discussions...

This PR affects basically **projectcalico/confd** adding a new field in BGPpeer: `KeepOriginalNextHop` and generate a new Boolean Key in `bird.cfg.template` defined by `next_hop_key` per BGPpeer Node, with default value equal to `False`.

Also, I have extended kubernetes CRD (that is include in libcalico-go) and installed it in my Cluster to test. The last step was recompile `calicoctl` with the new `libcalico-go` to help to create new objects with the new field.

A simple use case is create a new BGPpeer using the Route Reflectors to communicate with external BGPpeer ("my-global-peer"):
```bash
## Disable the default BGP node-to-node mesh 
calicoctl patch bgpconfiguration default -p '{"spec": {"nodeToNodeMeshEnabled": “false”}}'
```
Make BGPpeers (Route Reflector and BGPpeer):
```yaml
kind: BGPPeerList
items:
- apiVersion: projectcalico.org/v3
  kind: BGPPeer
  metadata:
    name: bgppeer-rr
  spec:
    nodeSelector: all()
    peerSelector: route-reflector == 'true'
- apiVersion: projectcalico.org/v3
  kind: BGPPeer
  metadata:
    name: my-global-peer
  spec:
    asNumber: 65000
    nodeSelector: route-reflector == "true"
    peerIP: 10.79.63.42
    keepOriginalNextHop: true
```
So, is expect the Nodes send theirs routes to RR ( "bgppeer-rr") and this one sends all of routes keeping original Next Hopes to external BGP peer ("my-global-peer").
```
        [-----------]                        [---------------]
        [  Node01   ]   w/ next hop keep     [External Router]
        [bgppeer-rr ]  ------------------>   [my-global-peer ]
        [-----------]                        [---------------]
          ^   ^   ^                                 |
    ------|   |   |-------                          |
    |         |          |          Directly Fwd    |
[Node02]   [Node03]   [NodeXX]  <--------------------
```

IMPORTANT: I did very few tests with and without the reflecting router.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

### DIFF
--- a/config/crd/crd.projectcalico.org_bgppeers.yaml
+++ b/config/crd/crd.projectcalico.org_bgppeers.yaml
@@ -64,6 +64,12 @@ spec:
                   remote AS number comes from the remote node’s NodeBGPSpec.ASNumber,
                   or the global default if that is not set.
                 type: string
+              keepOriginalNextHop:
+                description: 'Option to keep the original nexthop field when routes are sent to
+                  a BGP Peer. Setting "true" configures the selected BGP Peers node to use the
+                  "next hop keep;" instead of "next hop self;"(default) in the specific branch
+                  of the Node on "bird.cfg".'
+                type: boolean
             required:
             - asNumber
             - peerIP
--- a/lib/apis/v3/bgppeer.go
+++ b/lib/apis/v3/bgppeer.go
@@ -61,6 +61,10 @@ type BGPPeerSpec struct {
 	// NodeBGPSpec.IPv6Address specified.  The remote AS number comes from the remote
 	// node’s NodeBGPSpec.ASNumber, or the global default if that is not set.
 	PeerSelector string `json:"peerSelector,omitempty" validate:"omitempty,selector"`
+	// Option to keep the original nexthop field when routes are sent to a BGP Peer.
+	// Setting "true" configures the selected BGP Peers node to use the "next hop keep;"
+	// instead of "next hop self;"(default) in the specific branch of the Node on "bird.cfg".
+	KeepOriginalNextHop bool `json:"keepOriginalNextHop,omitempty"`
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object


lib/backend/model/bgppeer.go
```diff
@@ -34,8 +35,9 @@ var (
 )
 
 type NodeBGPPeerKey struct {
-	Nodename string `json:"-" validate:"omitempty"`
-	PeerIP   net.IP `json:"-" validate:"required"`
+	Nodename            string `json:"-" validate:"omitempty"`
+	PeerIP              net.IP `json:"-" validate:"required"`
+	KeepOriginalNextHop bool   `json:"-" validate:"omitempty"`
```
fasaxc, line 40: Unless we're doing something very odd downstream, I don't think this belongs in the key for this resource

lib/backend/model/bgppeer.go
```diff
@@ -63,12 +65,13 @@ func (key NodeBGPPeerKey) valueType() (reflect.Type, error) {
...
-	return fmt.Sprintf("BGPPeer(node=%s, ip=%s)", key.Nodename, key.PeerIP)
+	return fmt.Sprintf("BGPPeer(node=%s, ip=%s, keep_next_hop=%s)", key.Nodename, key.PeerIP, strconv.FormatBool(key.KeepOriginalNextHop))
 }
 
 type NodeBGPPeerListOptions struct {
-	Nodename string
-	PeerIP   net.IP
+	Nodename            string
+	PeerIP              net.IP
+	KeepOriginalNextHop bool
```
fasaxc, line 74: Same for the list options, the new field should only go in the value type.

lib/backend/model/bgppeer.go
```diff
@@ -183,5 +187,6 @@ type BGPPeer struct {
 	// value as a string in to the backend, because confd templating
 	// converts large uints to float e notation which breaks the BIRD
 	// configuration.
-	ASNum numorstring.ASNumber `json:"as_num,string"`
+	ASNum               numorstring.ASNumber `json:"as_num,string"`
+	KeepOriginalNextHop bool                 `json:"keep_next_hop,omitempty"`
```
fasaxc, line 191: Probably need conversion logic to fill this field in when converting from the "v3" type to the "model" type.  Confd uses the old "model" types, I believe so we have to convert from the v3 to the model types.

gunboe: Thank you @fasaxc for your feedk back! I am a very nubie here, it was hard to follow the "bread crumb trail", mainly try to understand different api kinds and its connections. I suppose that is there an updating process in course. So ts possible that I did unnecessary things. But I could not see where to convert types and at the end confd assumed a boolean value in the Bird Template. It is possible some strange values (not a boolean) inadvertently be used?

lib/backend/model/bgppeer.go
gunboe: In deed! It is unnecessary, I'll remake up

gunboe:
I did new tests but now without kubernetes datastore, using etcd directly.Thats Ok.

fasaxc:
LGTM, will leave this open until the confd changes are approved

gunboe:
@fasaxc I've made a last minute commit correcting a typo. Also, do you want me to squash this in one commit?

gunboe:
Hi @fasaxc  and @neiljerram, I did a Rebase with master branch and an Squash with commit typos. I hope this help to facilitate merge process to the next release, soonly, I hope. Thanks.

PR #1269, opened 2020-07-16
Support peer with same IP but different port

Suraiya-Hameed:
## Description
Extend GlobalBGPPeerKey and NodeBGPPeerKey to differentiate between peers with same IP but different ports.

https://github.com/projectcalico/confd/pull/341#discussion_r452835494

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/backend/model/bgppeer.go
+++ b/lib/backend/model/bgppeer.go
@@ -36,6 +36,7 @@ var (
 type NodeBGPPeerKey struct {
 	Nodename string `json:"-" validate:"omitempty"`
 	PeerIP   net.IP `json:"-" validate:"required"`
+	Port     string `json:"-" validate:"omitempty"`
 }
 
 func (key NodeBGPPeerKey) defaultPath() (string, error) {
@@ -45,8 +46,12 @@ func (key NodeBGPPeerKey) defaultPath() (string, error) {
 	if key.Nodename == "" {
 		return "", errors.ErrorInsufficientIdentifiers{Name: "node"}
 	}
+	ipPort := key.PeerIP.String()
+	if key.Port != "" {
+		ipPort = ipPort + "-" + key.Port
+	}
 	e := fmt.Sprintf("/calico/bgp/v1/host/%s/peer_v%d/%s",
-		key.Nodename, key.PeerIP.Version(), key.PeerIP)
+		key.Nodename, key.PeerIP.Version(), ipPort)
 	return e, nil
 }
 
@@ -63,7 +68,7 @@ func (key NodeBGPPeerKey) valueType() (reflect.Type, error) {
 }
 
 func (key NodeBGPPeerKey) String() string {
-	return fmt.Sprintf("BGPPeer(node=%s, ip=%s)", key.Nodename, key.PeerIP)
+	return fmt.Sprintf("BGPPeer(node=%s, ip=%s, port=%s)", key.Nodename, key.PeerIP, key.Port)
 }
 
 type NodeBGPPeerListOptions struct {
@@ -113,14 +118,19 @@ func (options NodeBGPPeerListOptions) KeyFromDefaultPath(path string) Key {
 
 type GlobalBGPPeerKey struct {
 	PeerIP net.IP `json:"-" validate:"required"`
+	Port   string `json:"-" validate:"omitempty"`
 }
 
 func (key GlobalBGPPeerKey) defaultPath() (string, error) {
 	if key.PeerIP.IP == nil {
 		return "", errors.ErrorInsufficientIdentifiers{Name: "peerIP"}
 	}
+	ipPort := key.PeerIP.String()
+	if key.Port != "" {
+		ipPort = ipPort + "-" + key.Port
+	}
 	e := fmt.Sprintf("/calico/bgp/v1/global/peer_v%d/%s",
-		key.PeerIP.Version(), key.PeerIP)
+		key.PeerIP.Version(), ipPort)
 	return e, nil
 }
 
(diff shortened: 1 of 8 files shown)


Suraiya-Hameed:
yes, that's right, there are no calls to `KeyFromDefaultPath` in confd. I've updated inverse logic to extract port from key path for consistency.

>Also can you add some UTs? I would suggest in keys_test.go.

I've added tests to bpgpeer_test.go, `KeyFromDefaultPath` in [keys.go](https://github.com/projectcalico/libcalico-go/blob/master/lib/backend/model/keys.go#L195) doesn't have a case for *BGPPeerKey, as it handles BGPPeer as v3 key (`/calico/resources/v3/projectcalico`).

lib/backend/model/bgppeer_test.go
```diff
@@ -0,0 +1,70 @@
...
+	func(strKey string, expected Key, isGlobalPeer bool) {
+		var key Key
+		if isGlobalPeer {
+			key = (GlobalBGPPeerListOptions{}).KeyFromDefaultPath(strKey)
+		} else {
+			key = (NodeBGPPeerListOptions{}).KeyFromDefaultPath(strKey)
+		}
+
+		Expect(key).To(Equal(expected))
+		serialized, err := KeyToDefaultPath(expected)
```
Suraiya-Hameed, line 36: These tests check for path to key conversion (`KeyFromDefaultPath`), I dint add tests for the other direction (key to path), as we are already making use of `KeyToDefaultPath` for validation.

fasaxc:
> there are no calls to KeyFromDefaultPath in confd

@Suraiya-Hameed FYI, it's used in typha so it's needed indirectly.  Typha uses it to serialise the key for sending to its clients (including confd).

lib/backend/model/bgppeer.go
```diff
@@ -31,11 +31,13 @@ var (
 	matchGlobalBGPPeer = regexp.MustCompile("^/?calico/bgp/v1/global/peer_v./([^/]+)$")
 	matchHostBGPPeer   = regexp.MustCompile("^/?calico/bgp/v1/host/([^/]+)/peer_v./([^/]+)$")
 	typeBGPPeer        = reflect.TypeOf(BGPPeer{})
+	ipPortSeparator    = "-"
 )
 
 type NodeBGPPeerKey struct {
 	Nodename string `json:"-" validate:"omitempty"`
 	PeerIP   net.IP `json:"-" validate:"required"`
+	Port     string `json:"-" validate:"omitempty"`
```
fasaxc, line 40: I don't think we should use string for the type of Port. I'd be inclined to use `uint16` since it naturally validates.

lib/backend/model/bgppeer.go
```diff
@@ -185,3 +195,18 @@ type BGPPeer struct {
...
+
+func extractIPAndPort(ipPort string) ([]byte, string) {
+	arr := strings.Split(ipPort, ipPortSeparator)
+	if len(arr) == 2 {
+		return []byte(arr[0]), arr[1]
+	}
+	return []byte(ipPort), ""
+}
+
+func combineIPAndPort(ip string, port string) string {
```
fasaxc, line 207: I think you should map port 179 (the default BGP port) to the old style of key.  That will ensure that you can't have an old-style peer and a new-style peer both with the same port.  It'll also mean that old typha clients and new typhas will interop more cleanly.

lib/backend/model/bgppeer_test.go
fasaxc: Good use of table driven test

lib/backend/model/bgppeer.go
Suraiya-Hameed: The reason for using string here was, we get port in `<ip>:<port>` format https://github.com/projectcalico/libcalico-go/blob/master/lib/apis/v3/bgppeer.go#L51 and later use net.SplitHostPort to split them, which returns port as string. Port used here is just to generate key, to avoid converting between string to uint16 and the back to string, I went for this. But looks like we do the same for IP, I'll update it to uint16

lib/backend/model/bgppeer.go
Suraiya-Hameed: We are not setting default port to 179 explicitly anywhere, so I think if user explicitly sets 179 in new-style there will be a conflict.

Suraiya-Hameed: so we will always get the old-style peer key if port is not explicit set.

lib/backend/model/bgppeer.go
fasaxc: The key is part of the public API so it's worth the extra conversion work to make the definition of key more "correct"

lib/backend/model/bgppeer.go
fasaxc: How about doing these translations:
```
BGPPeerKey{..., Port: 179} -> "/calico/bgp/v1/.../peer_v4/10.0.0.5"
"/calico/bgp/v1/.../peer_v4/10.0.0.5" -> BGPPeerKey{..., Port: 179} 
```
I.e. never use 179 in the string version of the key but fill it in on read so that it's explicit for new clients.

I think we should avoid allowing an old and new-style version of hte 179 peer being created; presumably that would cause BIRD to fail and effectively break the cluster?

Suraiya-Hameed: Thanks for clarifying @fasaxc, you're right, having an old peer and a new peer with port 179 will cause issue. Not using 179 in key will help with that :+1:

PR #1272, opened 2020-07-27
Additional validation for BGP communities

Suraiya-Hameed:
## Description

This is mainly to address the misunderstanding where if `communities` are set, user should use them in `prefixAdvertisements` or else they will not be advertised. During user testing we figured it wasn't obvious that `prefixAdvertisements` and `communities` should be used in tandem. 

This PR simply makes sure that if `spec.communities` are set `spec.prefixAdvertisements` should not be empty. It would help avoid human error where user adds entries in communities field and think it will get propagated, but they in fact missed a step in the configuration. 

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/validator/v3/validator.go
+++ b/lib/validator/v3/validator.go
@@ -1412,9 +1412,13 @@ func validateBGPConfigurationSpec(structLevel validator.StructLevel) {
 		}
 	}
 
+	if (len(spec.PrefixAdvertisements) == 0) && (len(communities) != 0) {
+		structLevel.ReportError(reflect.ValueOf(communities), "Spec.Communities[]", "",
+			reason("communities are defined but not used in Spec.PrefixAdvertisement[]."), "")
+	}
+
 	// check if Spec.PrefixAdvertisement.Communities are valid
-	pas := spec.PrefixAdvertisements
-	for _, pa := range pas {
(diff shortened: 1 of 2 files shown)


lib/validator/v3/validator.go
```diff
@@ -1412,9 +1412,13 @@ func validateBGPConfigurationSpec(structLevel validator.StructLevel) {
 		}
 	}
 
+	if (spec.PrefixAdvertisements == nil || len(spec.PrefixAdvertisements) == 0) && (communities != nil || len(communities) != 0) {
```
neiljerram, line 1415: The `== nil` conditions are superfluous, because https://play.golang.org/p/eswiL5pZVMc

lib/validator/v3/validator.go
```diff
@@ -1412,9 +1412,13 @@ func validateBGPConfigurationSpec(structLevel validator.StructLevel) {
...
+		structLevel.ReportError(reflect.ValueOf(communities), "Spec.Communities[]", "",
+			reason("community is defined but never used in Spec.PrefixAdvertisement[]."), "")
```
neiljerram, line 1417: I would suggest slight tweak: "communities are defined but not used ..."

PR #1276, opened 2020-08-05
Added Windows support and new API EnsureBlock to IPAM

song-jiang:
## Description

Added Windows support and new API EnsureBlock to IPAM
- Added ability to reserve ips for a new ipam block.
- Added EnsureBlock API 
- Refactored `autoAssign`

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Calico IPAM support for Windows nodes
```

### DIFF
--- a/lib/ipam/ipam.go
+++ b/lib/ipam/ipam.go
@@ -18,6 +18,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math/bits"
+	"runtime"
 
 	log "github.com/sirupsen/logrus"
 
@@ -49,7 +51,8 @@ const (
 )
 
 var (
-	ErrBlockLimit = errors.New("cannot allocate new block due to per host block limit")
+	ErrBlockLimit      = errors.New("cannot allocate new block due to per host block limit")
+	ErrNoQualifiedPool = errors.New("cannot find a qualified ippool")
 )
 
 // NewIPAMClient returns a new ipamClient, which implements Interface.
@@ -97,7 +100,7 @@ func (c ipamClient) AutoAssign(ctx context.Context, args AutoAssignArgs) ([]net.
 				return nil, nil, fmt.Errorf("provided IPv4 IPPools list contains one or more IPv6 IPPools")
 			}
 		}
-		v4list, err = c.autoAssign(ctx, args.Num4, args.HandleID, args.Attrs, args.IPv4Pools, 4, hostname, args.MaxBlocksPerHost)
+		v4list, err = c.autoAssign(ctx, args.Num4, args.HandleID, args.Attrs, args.IPv4Pools, 4, hostname, args.MaxBlocksPerHost, args.HostReservedAttrIPv4s)
 		if err != nil {
 			log.Errorf("Error assigning IPV4 addresses: %v", err)
 			return v4list, nil, err
@@ -112,7 +115,7 @@ func (c ipamClient) AutoAssign(ctx context.Context, args AutoAssignArgs) ([]net.
 				return nil, nil, fmt.Errorf("provided IPv6 IPPools list contains one or more IPv4 IPPools")
 			}
 		}
-		v6list, err = c.autoAssign(ctx, args.Num6, args.HandleID, args.Attrs, args.IPv6Pools, 6, hostname, args.MaxBlocksPerHost)
+		v6list, err = c.autoAssign(ctx, args.Num6, args.HandleID, args.Attrs, args.IPv6Pools, 6, hostname, args.MaxBlocksPerHost, args.HostReservedAttrIPv6s)
 		if err != nil {
 			log.Errorf("Error assigning IPV6 addresses: %v", err)
 			return v4list, v6list, err
@@ -125,7 +128,7 @@ func (c ipamClient) AutoAssign(ctx context.Context, args AutoAssignArgs) ([]net.
 // getBlockFromAffinity returns the block referenced by the given affinity, attempting to create it if
 // it does not exist. getBlockFromAffinity will delete the provided affinity if it does not match the actual
 // affinity of the block.
-func (c ipamClient) getBlockFromAffinity(ctx context.Context, aff *model.KVPair) (*model.KVPair, error) {
+func (c ipamClient) getBlockFromAffinity(ctx context.Context, aff *model.KVPair, rsvdAttr *HostReservedAttr) (*model.KVPair, error) {
 	// Parse out affinity data.
 	cidr := aff.Key.(model.BlockAffinityKey).CIDR
 	host := aff.Key.(model.BlockAffinityKey).Host
@@ -155,7 +158,7 @@ func (c ipamClient) getBlockFromAffinity(ctx context.Context, aff *model.KVPair)
 
 			// Claim the block, which will also confirm the affinity.
 			logCtx.Info("Attempting to claim the block")
-			b, err := c.blockReaderWriter.claimAffineBlock(ctx, aff, *cfg)
+			b, err := c.blockReaderWriter.claimAffineBlock(ctx, aff, *cfg, rsvdAttr)
 			if err != nil {
 				logCtx.WithError(err).Warn("Error claiming block")
 				return nil, err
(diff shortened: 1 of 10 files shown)


lib/ipam/ipam.go
```diff
@@ -478,30 +516,117 @@ func (c ipamClient) autoAssign(ctx context.Context, num int, handleID *string, a
...
+		client:                c,
+		version:               version,
+		host:                  host,
+		pools:                 pools,
+		remainingAffineBlocks: affBlocks,
+		hostReservedAttr:      rsvdAttr,
+		allowNewClaim:         true,
+	}
+
+assignLoop:
```
caseydavenport, line 576: Can we structure this to remove the need for these loop label markers? In general we've tried to avoid these in Calico since there's usually a better way to represent the same logic. WDYT?

lib/ipam/ipam.go
```diff
@@ -478,30 +516,117 @@ func (c ipamClient) autoAssign(ctx context.Context, num int, handleID *string, a
...
+				if _, ok := err.(cerrors.ErrorResourceUpdateConflict); ok {
+					log.WithError(err).Debug("CAS Error assigning from new block - retry")
+
+					// At this point, block b's Unallocated field has been reduced already.
+					// We should get the original block from datastore again.
+					blockCIDR := b.Key.(model.BlockKey).CIDR
+					b, err = c.blockReaderWriter.queryBlock(ctx, blockCIDR, "")
+					if err != nil {
+						logCtx.WithError(err).Warn("Failed to get block again after update conflict")
+						break retryLoop
```
caseydavenport, line 617: Actually, unless I'm reading this wrong, I think we could just remove the `retryLoop` tag entirely and the code would still work? We're never calling `break assignLoop` from within the inner loop, so `break retryLoop` is the same as just `break`

lib/ipam/ipam.go
```diff
@@ -1391,18 +1518,31 @@ func (c ipamClient) GetAssignmentAttributes(ctx context.Context, addr net.IP) (m
...
-		return nil, err
+	} else {
+		config = c.convertBackendToIPAMConfig(obj.Value.(*model.IPAMConfig))
+	}
+	if detectOS(ctx) == "windows" {
+		// When a Windows node owns a block, it creates a local /26 subnet object and as far as we know, it can't
+		// do a longest-prefix-match between the subnet CIDR and a remote /32.  This means that we can't allow
+		// remote hosts to borrow IPs from a Windows-owned block; and Windows hosts can't borrow IPs either.
+		// Force Windows nodes to always behave as if StrictAffinity is enabled.  This means that they won't borrow
+		// and they'll mark any blocks that they allocate as Strict so others can't borrow from them.
```
caseydavenport, line 1541: Hm, this is a problem - the Linux code doesn't look at the block.StrictAffinity field any more, and instead only looks at the global configuration. We changed this in a recent release to enable the strict affinity code to behave more naturally.

I think that means that, instead of forcing this to be "true" here, we need to return an error if the global strict affinity configuration is set to "false" and we detect that we're running on a Windows node. That will prevent Windows nodes from launching pods at all so long as strict affinity is not enabled.

WDYT?

lib/ipam/ipam_block.go
```diff
@@ -32,15 +32,15 @@ import (
 
 // windwowsReservedHandle is the handle used to reserve addresses required for Windows
 // networking so that workloads do not get assigned these addresses.
-const windowsReservedHandle = "windows-reserved-IPAM-handle"
+const WindowsReservedHandle = "windows-reserved-IPAM-handle"
```
caseydavenport, line 35: might have spoke about this before but, I think this will be converted to lowercase when we write the Handle object to the k8s API since k8s resource names must be lowercase.

I think we should have it lowercase here as well so that all references to the handle name are identical.

lib/ipam/ipam_block.go
```diff
@@ -53,6 +53,33 @@ func newBlock(cidr cnet.IPNet) allocationBlock {
...
 	}
 
+	if rsvdAttr != nil {
+		// Reserve IPs based on host reserved attributes.
+		// For example, with windows OS, the following IP addresses of the block are
+		// reserved. This is done by pre-allocating them during initialization
+		// time only.
+		// IPs : x.0, x.1, x.2 and x.bcastAddr (e.g. x.255 for /24 subnet)
+
+		log.Infof("Block %s reserving IPs", b.CIDR.String())
```
caseydavenport, line 63: I think we use `WithFields` elsewhere to add the block to the log. Would be good to do that here as well in a consistent way so its easy to search through all the logs pertaining to a block. 

e.g., https://github.com/projectcalico/libcalico-go/blob/master/lib/ipam/ipam.go#L807

lib/ipam/ipam_block.go
```diff
@@ -53,6 +53,33 @@ func newBlock(cidr cnet.IPNet) allocationBlock {
...
+		// nil attributes
+		attrs := make(map[string]string)
+		attrs["note"] = rsvdAttr.Note
```
caseydavenport, line 66: We've defined common attributes here, probably worth doing "note" there as well so any external controller can reliably get it: https://github.com/projectcalico/libcalico-go/blob/master/lib/ipam/ipam.go#L42-L48

lib/ipam/ipam_block.go
```diff
@@ -146,14 +173,15 @@ func getHostAffinity(block *model.AllocationBlock) string {
...
-func (b allocationBlock) numFreeAddresses() int {
+func (b allocationBlock) NumFreeAddresses() int {
 	return len(b.Unallocated)
 }
 
-// empty returns true if the block has released all of its assignable addresses,
-// and returns false if any assignable addresses are in use.
-func (b allocationBlock) empty() bool {
-	return b.containsOnlyReservedIPs()
+func (b allocationBlock) empty(windowsHost bool) bool {
```
caseydavenport, line 180: Do we need this change? I thought the old approach worked for both windows and linux hosts.

lib/ipam/ipam_block_reader_writer.go
```diff
@@ -98,19 +98,6 @@ func (rw blockReaderWriter) findUnclaimedBlock(ctx context.Context, host string,
 		return nil, fmt.Errorf("no configured Calico pools for node %s", host)
 	}
 
-	// List blocks up front to reduce number of queries.
```
caseydavenport, line 101: This is a bad merge, I think - we want to keep this code.

lib/ipam/ipam_block_reader_writer.go
```diff
@@ -120,9 +107,14 @@ func (rw blockReaderWriter) findUnclaimedBlock(ctx context.Context, host string,
 		for subnet := blocks(); subnet != nil; subnet = blocks() {
 			// Check if a block already exists for this subnet.
 			log.Debugf("Getting block: %s", subnet.String())
-			if _, ok := exists[subnet.String()]; !ok {
-				log.Infof("Found free block: %+v", *subnet)
-				return subnet, nil
+			_, err := rw.queryBlock(ctx, *subnet, "")
```
caseydavenport, line 110: Related to merge conflict above, we don't want to query the block on every iteration here.

lib/ipam/ipam_block_reader_writer.go
```diff
@@ -370,10 +363,6 @@ func (rw blockReaderWriter) queryBlock(ctx context.Context, blockCIDR cnet.IPNet
 	return rw.client.Get(ctx, model.BlockKey{CIDR: blockCIDR}, revision)
 }
 
-func (rw blockReaderWriter) listBlocks(ctx context.Context, revision string) (*model.KVPairList, error) {
```
caseydavenport, line 373: part of the bad merge mentioned above - should keep this function.

lib/ipam/ipam_block_reader_writer_test.go
```diff
@@ -130,6 +130,11 @@ func (c *fakeClient) Clean() error {
 	return nil
 }
 
+func (c *fakeClient) Close() error {
```
caseydavenport, line 133: Do we actually need this? I don't think this is required for the client interface.

lib/ipam/ipam.go
song-jiang: The problem is some user would like to join windows nodes to an exiting cluster which has global strict affinity set to `false`. This would not going to work for them.

lib/ipam/ipam_block.go
song-jiang: Good point. I did not see any issue of upper case but it is a good idea to make it lower case.

lib/ipam/ipam.go
song-jiang: Yeah. Code still works the same without tags. I'll get it removed.

lib/ipam/ipam_block.go
song-jiang: The code is same as private master and v3.12. I think we need this change.

lib/ipam/ipam_block.go
song-jiang: Do we rely on `note` for anything? I thought it is just for description purpose.

lib/ipam/ipam_block_reader_writer_test.go
song-jiang: No. But seems it is there in private.

lib/ipam/ipam.go
caseydavenport: But what can we do about that?

I think the only thing we can reasonably do is to reject creation of pods on Windows nodes if the global strict affiinity is false. 

The problem is, there is no longer any concept of a per-node strictaffinity option. It was removed.

lib/ipam/ipam_block.go
caseydavenport: Nah, I'm pretty sure those branches are wrong. Lets talk about it.

song-jiang: Reverted to OS version

PR #1279, opened 2020-08-10
Change default datastoretype

frozenprocess:
## Description
This PR changes datastore type to `kubernetes` if `etcd` is not specified by user.[successor to this PR](https://github.com/projectcalico/calicoctl/pull/2159)

## Todos
- [ ] Tests
- [x] Documentation
- [ ] Release note

## Release Note

```release-note
Default to Kubernetes datastore
```
@robbrockbank @caseydavenport

### DIFF
--- a/lib/apiconfig/apiconfig.go
+++ b/lib/apiconfig/apiconfig.go
@@ -41,7 +41,7 @@ type CalicoAPIConfig struct {
 
 // CalicoAPIConfigSpec contains the specification for a Calico CalicoAPIConfig resource.
 type CalicoAPIConfigSpec struct {
-	DatastoreType DatastoreType `json:"datastoreType" envconfig:"DATASTORE_TYPE" default:"etcdv3"`
+	DatastoreType DatastoreType `json:"datastoreType" envconfig:"DATASTORE_TYPE"`
 	// Inline the ectd config fields
 	EtcdConfig
 	// Inline the k8s config fields.
--- a/lib/apiconfig/load.go
+++ b/lib/apiconfig/load.go
@@ -43,9 +43,8 @@ func LoadClientConfig(filename string) (*CalicoAPIConfig, error) {
 	} else if c.Spec.DatastoreType == "" {
 		c.Spec.DatastoreType = Kubernetes
 	}
-	// Changing Kubeconfig prevents a default error attitude
-	// when user has set no configuration.
-	if c.Spec.DatastoreType == Kubernetes && c.Spec.Kubeconfig == "" {
+
+	if os.Getenv("KUBERNETES_SERVICE_HOST") == "" && c.Spec.DatastoreType == Kubernetes && c.Spec.Kubeconfig == "" {
 		c.Spec.Kubeconfig = filepath.Join(os.Getenv("HOME"),".kube","config")
 	}
 
--- a/lib/apiconfig/load.go
+++ b/lib/apiconfig/load.go
@@ -4,6 +4,8 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"os"
+	"path/filepath"
 
 	"github.com/kelseyhightower/envconfig"
 	log "github.com/sirupsen/logrus"
@@ -15,6 +17,8 @@ import (
 // LoadClientConfig loads the ClientConfig from the specified file (if specified)
 // or from environment variables (if the file is not specified).
 func LoadClientConfig(filename string) (*CalicoAPIConfig, error) {
+	var c *CalicoAPIConfig
+	var err error
 
 	// Override / merge with values loaded from the specified file.
 	if filename != "" {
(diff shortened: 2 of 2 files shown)


lib/apiconfig/load.go
```diff
@@ -75,3 +80,23 @@ func LoadClientConfigFromEnvironment() (*CalicoAPIConfig, error) {
 
 	return c, nil
 }
+
+// ClientConfigOverride changes datastore type to Kubernetes when EtcdV3 endpoint is not
+// explicitly defined by user.
+func ClientConfigOverride(cfg *CalicoAPIConfig) (*CalicoAPIConfig, error) {
+	if cfg.Spec.EtcdEndpoints == "" && cfg.Spec.DatastoreType == EtcdV3 {
```
frozenprocess, line 87: I feel like checking `EtcdEndpoints` is enough, it can prevent an error in case user set `Kubernetes` as `datastoretype` but forgets to set `K8sAPIEndpoint` or `Kubeconfig`.

@robbrockbank @caseydavenport

lib/apiconfig/load.go
```diff
@@ -29,7 +31,10 @@ func LoadClientConfig(filename string) (*CalicoAPIConfig, error) {
 		}
 		return c, nil
```
frozenprocess, line 32: Just to double check, if user is using a calicoctl config file we will not override anything?

frozenprocess:
Things to consider :
--context ,--kubeconfig arguments will cause an error.
example:
` kubectk calico --kubeconfig=~/.kube/myshinyconfig get nodes

lib/apiconfig/load.go
```diff
@@ -75,3 +80,23 @@ func LoadClientConfigFromEnvironment() (*CalicoAPIConfig, error) {
 
 	return c, nil
 }
+
+// ClientConfigOverride changes datastore type to Kubernetes when EtcdV3 endpoint is not
+// explicitly defined by user.
+func ClientConfigOverride(cfg *CalicoAPIConfig) (*CalicoAPIConfig, error) {
```
frozenprocess, line 86: I've created a separate function just to be more clear about whats happening. Using `LoadClientConfigFromEnvironment` for everything was a bit confusing for future troubleshooting imo.

caseydavenport: I would like to see the existing LoadClientConfigFromEnvironment function enhanced to match the current desired behavior, rather than split that logic into two places. While it might be simpler to think about them separately within the context of this change, longer term it will be a lot easier to understand if client configuration loading is done in only one place.

frozenprocess:
Instead of giving `DATASTORE_TYPE` a default value, we create an empty one and only override if user did not define one explicitly.
I moved default Kubernetes logic to `LoadClientConfig` function to cover both ways of creating configuration object.

@caseydavenport WDYT?
@robbrockbank

lib/apiconfig/apiconfig.go
```diff
@@ -91,6 +93,10 @@ func NewCalicoAPIConfig() *CalicoAPIConfig {
 		TypeMeta: metav1.TypeMeta{
 			Kind:       KindCalicoAPIConfig,
 			APIVersion: apiv3.GroupVersionCurrent,
+		},Spec: CalicoAPIConfigSpec{
```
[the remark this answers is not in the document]
frozenprocess: This is the final version. Setting `kubeconfig` is necessary imo because it can help when no configuration can be found and instead of returning an error we use Kubernetes config file for connection.

The only scenario that I could've think of for default value of `kubeconfig` as being problematic is when user sets `K8sAPIEndpoint` manually. However, since `K8sAPIEndpoint` takes precedence over `kubeconfig` it should be safe to use it like this.

> [Additionally, we probably want to change the defaults listed in the struct tags.](https://github.com/projectcalico/libcalico-go/pull/1279#pullrequestreview-467187632)

Since this is a default value I thought it is safe to set in struct tags to address @caseydavenport  request.
It can be moved to [load.go line 43 too](https://github.com/projectcalico/libcalico-go/pull/1279/files#diff-098f827fb283aa9a2a88906bc2e691e6R43).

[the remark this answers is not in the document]
robbrockbank: Yup move to load.go as you suggested (i.e. only initialize it if we've determined we are using kubernetes datastore). Then I think we are good to go. Thanks!

lib/apiconfig/load.go
```diff
@@ -23,28 +27,37 @@ func LoadClientConfig(filename string) (*CalicoAPIConfig, error) {
...
+	// otherwise set default Datastoretype to Kubernetes
+	if c.Spec.DatastoreType == "" && c.Spec.EtcdEndpoints != "" {
+		c.Spec.DatastoreType = EtcdV3
+	} else if c.Spec.DatastoreType == "" {
+		c.Spec.DatastoreType = Kubernetes
 	}
-	return LoadClientConfigFromEnvironment()
+	// Changing Kubeconfig prevents a default error attitude
+	// when user has set no configuration.
+	if c.Spec.DatastoreType == Kubernetes && c.Spec.Kubeconfig == "" {
```
frozenprocess: Using this logic makes breaks running calicoctl inside a pod. `thanks to @caseydavenport for point it out`

caseydavenport: hehe, sorry I didn't get around to actually posting a comment on here!

I think we also need to check for the presence of the in cluster config env vars, and only set kubeconfig if those aren't set. WDYT?

frozenprocess: No worries, Alex told me about it and I got excited :D

Are we talking about `CALICO_CTL_CONTAINER` or something like `KUBERNETES_PORT`?

We could also check for secrets folder. WDYT?
`/var/run/secrets`

caseydavenport: I was thinking KUBERNETES_SERVICE_HOST / KUBERNETES_SERVICE_PORT

caseydavenport (approved):

fasaxc:
This PR broke Typha, which uses `LoadClientConfigFromEnvironment` and expected the config to be defaulted.  @frozenprocess why did you remove the `default:` tag in the struct (rather than changing it to "kubernetes")?

fasaxc:
I put up a Typha fix to handle this case: https://github.com/projectcalico/typha/pull/400

frozenprocess:
> This PR broke Typha, which uses `LoadClientConfigFromEnvironment` and expected the config to be defaulted.  @frozenprocess why did you remove the `default:` tag in the struct (rather than changing it to "kubernetes")?

Oh boy 😈! 
There was an overwrite case if I remember correctly. I'll have to get back to you on exactly why I changed it to an empty default.

PR #1284, opened 2020-08-15
Handle deletion of weps in KDD

fasaxc:
## Description

This fixes handling of pods that get stuck in Terminating state _after_ the CNI plugin has been called:

- Pod delete sets the deletion timestamp on the resource but the resource is not actually deleted until it has been torn down.
- Pod gets its grace period to shut down.
- After the grace period, `kubelet` issues a CNI DEL, which succeeds, removes the interface and returns the IP to the pool.
- One of the following phases of teardown fails; the pod get stuck in Terminating state, with a deletion timestamp but `kubelet` is unable to clean it up.
- To libcalico-go, the pod looks to still be alive with its old IP so we include it in policy and workload endpoint calculations.
- The IP is free and it ends up getting allocated to another pod.  This confuses policy and makes felix think it has duplicate endpoints.

Fix:
* Zero out the annotations used for signaling in the pod IP and use deletion timestamp + zero annotation to signal that the IP is gone.
* Leave the podIP alone since it's the Kubernetes norm to preserve that.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
In Kubernetes API Datastore mode, record when a pod is deleted from the network; this prevents pods that are stuck in Terminating state from being treated as active pods, resulting in duplicate IP errors and incorrect IP set calculation.
```

### DIFF
--- a/lib/backend/k8s/resources/workloadendpoint.go
+++ b/lib/backend/k8s/resources/workloadendpoint.go
@@ -60,7 +60,7 @@ func (c *WorkloadEndpointClient) Create(ctx context.Context, kvp *model.KVPair) 
 	// Note: it's a bit odd to do this in the Create, but the CNI plugin uses CreateOrUpdate().  Doing it
 	// here makes sure that, if the update fails: we retry here, and, we don't report success without
 	// making the patch.
-	return c.patchPodIP(ctx, kvp)
+	return c.patchInPodIPs(ctx, kvp)
 }
 
 func (c *WorkloadEndpointClient) Update(ctx context.Context, kvp *model.KVPair) (*model.KVPair, error) {
@@ -68,7 +68,7 @@ func (c *WorkloadEndpointClient) Update(ctx context.Context, kvp *model.KVPair) 
 	// As a special case for the CNI plugin, try to patch the Pod with the IP that we've calculated.
 	// This works around a bug in kubelet that causes it to delay writing the Pod IP for a long time:
 	// https://github.com/kubernetes/kubernetes/issues/39113.
-	return c.patchPodIP(ctx, kvp)
+	return c.patchInPodIPs(ctx, kvp)
 }
 
 func (c *WorkloadEndpointClient) DeleteKVP(ctx context.Context, kvp *model.KVPair) (*model.KVPair, error) {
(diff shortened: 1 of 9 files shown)


lib/backend/k8s/resources/workloadendpoint.go
```diff
@@ -60,64 +60,78 @@ func (c *WorkloadEndpointClient) Create(ctx context.Context, kvp *model.KVPair)
...
 	log.Debugf("PATCHing pod with IPs: %v", ips)
-	wepID, err := c.converter.ParseWorkloadEndpointName(kvp.Key.(model.ResourceKey).Name)
+	key := kvp.Key
+	return c.patchPodIPAnnotations(key, kvp.Revision, kvp.UID, ips)
+}
+
+// patchOutPodIPs sets our pod IP annotations to empty strings; this is used to signal that the IP has been removed
+// from the pod at teardown.
+func (c *WorkloadEndpointClient) patchOutPodIPs(ctx context.Context, key model.Key, revision string, uid *types.UID) (*model.KVPair, error) {
+	return c.patchPodIPAnnotations(key, revision, uid, nil)
```
caseydavenport, line 102: Might be worth a follow on PR to remove this section of the CNI plugin, since I think it's dead code with this change? https://github.com/projectcalico/cni-plugin/blob/0c2676672f697b1ce3ac1ad5f27b57135fb5395c/pkg/k8s/k8s.go#L509

Not necessarily a problem, but could be confusing in the future when we don't remember why it's there :D

lib/backend/k8s/resources/workloadendpoint.go
```diff
@@ -127,25 +141,69 @@ func (c *WorkloadEndpointClient) patchPodIP(ctx context.Context, kvp *model.KVPa
 	return kvps[0], nil
 }
 
-const annotationNameValueTemplate = `%s: %s`
-const annotationPatchTemplate = `{"metadata": {"annotations": {%s}}}`
+// convertPatchErrorToCalico is a specialised version of K8sErrorToCalico that understands patch-specific errors.
+func convertPatchErrorToCalico(err error, id interface{}) error {
+	if ke, ok := err.(kerrors.APIStatus); ok {
```
caseydavenport, line 146: Is it worth just enhancing the main convert error logic to include this? 

I wouldn't be surprised if we have bugs elsewhere in the code (or introduce them later) because we don't realize there's a patch-specific implementation of errors, and it seems like this could be safely added there.

PR #1289, opened 2020-08-25
Change Unicode apostrophe in API comments to ASCII apostrophe

neiljerram:
These API comments are automatically propagated to CRD YAMLs, which
don't support Unicode apostrophes.

## Release Note

```release-note
None required
```

### DIFF
--- a/config/crd/crd.projectcalico.org_felixconfigurations.yaml
+++ b/config/crd/crd.projectcalico.org_felixconfigurations.yaml
@@ -105,10 +105,10 @@ spec:
                   `tc exec bpf debug`. [Default: Off].'
                 type: string
               chainInsertMode:
-                description: 'ChainInsertMode controls whether Felix hooks the kernel’s
+                description: 'ChainInsertMode controls whether Felix hooks the kernel''s
                   top-level iptables chains by inserting a rule at the top of the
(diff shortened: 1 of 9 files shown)


config/crd/crd.projectcalico.org_felixconfigurations.yaml
```diff
@@ -105,10 +105,10 @@ spec:
                   `tc exec bpf debug`. [Default: Off].'
                 type: string
               chainInsertMode:
-                description: 'ChainInsertMode controls whether Felix hooks the kernel’s
+                description: 'ChainInsertMode controls whether Felix hooks the kernel''s
                   top-level iptables chains by inserting a rule at the top of the
                   chain or by appending a rule at the bottom. insert is the safe default
-                  since it prevents Calico’s rules from being bypassed. If you switch
+                  since it prevents Calico''s rules from being bypassed. If you switch
```
fasaxc, line 111: Doubled `'` in this file

neiljerram: Yes, I believe that's correct for quoting a single quote within a single quoted string ! http://blogs.perl.org/users/tinita/2018/03/strings-in-yaml---to-quote-or-not-to-quote.html

fasaxc (approved):

PR #1296, opened 2020-09-01
Add new FelixConfiguration knobs for disabling encap drop rules

doublek:
## Description

Relates to 
- projectcalico/felix#2484
- docs: projectcalico/calico#3961

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Add FelixConfiguration parameters to explicitly allow encapsulated packets from workloads.
```

### DIFF
--- a/lib/apis/v3/felixconfig.go
+++ b/lib/apis/v3/felixconfig.go
@@ -180,6 +180,13 @@ type FelixConfigurationSpec struct {
 	VXLANPort *int `json:"vxlanPort,omitempty"`
 	VXLANVNI  *int `json:"vxlanVNI,omitempty"`
 
+	// AllowVXLANPacketsFromWorkloads controls whether Felix will add a rule to drop VXLAN encapsulated traffic
+	// from workloads [Default: false]
+	AllowVXLANPacketsFromWorkloads *bool `json:"allowVXLANPacketsFromWorkloads,omitempty"`
+	// AllowIPIPPacketsFromWorkloads controls whether Felix will add a rule to drop IPIP encapsulated traffic
+	// from workloads [Default: false]
+	AllowIPIPPacketsFromWorkloads *bool `json:"allowIPIPPacketsFromWorkloads,omitempty"`
+
 	// ReportingInterval is the interval at which Felix reports its status into the datastore or 0 to disable.
 	// Must be non-zero in OpenStack deployments. [Default: 30s]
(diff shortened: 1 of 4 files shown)


lib/apis/v3/felixconfig.go
```diff
@@ -180,6 +180,13 @@ type FelixConfigurationSpec struct {
 	VXLANPort *int `json:"vxlanPort,omitempty"`
 	VXLANVNI  *int `json:"vxlanVNI,omitempty"`
 
+	// AllowVXLANPacketsFromWorkloads controls whether Felix will add a rule to drop VXLAN encapsulated traffic
+	// from workloads [Default: false]
+	AllowVXLANPacketsFromWorkloads *bool `json:"allowVXLANPacketsFromWorkloads,omitempty"`
+	// AllowIPIPPacketsFromWorkloads controls whether Felix will add a rule to drop IPIP encapsulated traffic
+	// from workloads [Default: false]
+	AllowIPIPPacketsFromWorkloads *bool `json:"allowIPIPPacketsFromWorkloads,omitempty"`
```
fasaxc, line 188: Does it need to be a `*bool` if `false` is the default?  (Or maybe that's just our convention?)

fasaxc (approved):
One question but LGTM

lib/apis/v3/felixconfig.go
fasaxc: Oh, guess it's for inheritance, never mind.

lib/apis/v3/felixconfig.go
```diff
@@ -180,6 +180,13 @@ type FelixConfigurationSpec struct {
 	VXLANPort *int `json:"vxlanPort,omitempty"`
 	VXLANVNI  *int `json:"vxlanVNI,omitempty"`
 
+	// AllowVXLANPacketsFromWorkloads controls whether Felix will add a rule to drop VXLAN encapsulated traffic
+	// from workloads [Default: false]
+	AllowVXLANPacketsFromWorkloads *bool `json:"allowVXLANPacketsFromWorkloads,omitempty"`
```
caseydavenport, line 185: Now that we're using Kubernetes CRD generation with schemas, I think we need to explicit about defining optional fields and defaults.

So, this field should have comments like this on it in order to generate the correct CRDs. 

```
// +optional
// kubebuilder:default:=false
```

Source: https://book.kubebuilder.io/reference/markers/crd-validation.html

PR #1297, opened 2020-09-01
Configure max blocks per-host in IPAMConfig

caseydavenport:
## Description

Allows setting a global option for max blocks per-host. This is a nifty
tunable that, in conjunction with strict affinity, can help mitigate scenarios where one or more hosts end up
attempting to allocate more blocks than they should.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Support global configuration for max blocks per node
```

### DIFF
--- a/config/crd/crd.projectcalico.org_ipamconfigs.yaml
+++ b/config/crd/crd.projectcalico.org_ipamconfigs.yaml
@@ -38,6 +38,10 @@ spec:
             properties:
               autoAllocateBlocks:
                 type: boolean
+              maxBlocksPerHost:
+                description: MaxBlocksPerHost, if non-zero, is the max number of blocks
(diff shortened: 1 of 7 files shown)


caseydavenport:
I noticed that `AssignIP` doesn't currently have a check against the max blocks per host param. 

@song-jiang is this intentional? I think `AssignIP` is used to assign specific IP addresses in some scenarios.

song-jiang:
I don't think it is intentional. I think it should add max blocks per host check similar to `AutoAssign`. Will you add this on your PR?

caseydavenport:
@song-jiang yep, will do!

song-jiang (approved):
LGTM

PR #1299, opened 2020-09-02
Attempt to release other blocks even if one isn't empty

caseydavenport:
## Description

When we try to release affinities for a host, we don't want to do it if
there are borrowed addresses. However, we also don't want to prevent
cleaning up of empty blocks if only one or two have borrowed addresses.

Previously, we'd return on the first error. With this change, we'll
store the error and continue trying other blocks, returning the stored
error after we try all the remaining blocks.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Release affinities for all empty blocks on deleted nodes, even if some still have allocations. 
```

### DIFF
--- a/lib/ipam/ipam.go
+++ b/lib/ipam/ipam.go
@@ -1111,6 +1111,7 @@ func (c ipamClient) ReleaseHostAffinities(ctx context.Context, host string, must
 		return err
 	}
 
+	var storedError error
 	versions := []int{4, 6}
 	for _, version := range versions {
 		blockCIDRs, _, err := c.blockReaderWriter.getAffineBlocks(ctx, hostname, version, nil)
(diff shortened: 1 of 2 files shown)


lib/ipam/ipam.go
```diff
@@ -1124,12 +1125,15 @@ func (c ipamClient) ReleaseHostAffinities(ctx context.Context, host string, must
 				if _, ok := err.(errBlockClaimConflict); ok {
 					// Claimed by a different host.
 				} else {
-					return err
+					// Store the error for later so we can return it.
+					// We don't want to return just yet so we can do a best-effort
+					// attempt at releasing the other CIDRs for this host.
+					storedError = err
```
Brian-McM, line 1131: It may not be important, but there's a slight change in what error is returned. If multiple blocks would have failed, the previous code would return the error for the first (and only) block that failed. This code returns the error for the last block that failed.

Also, do you think it would be worth it to either aggregate the errors into one error, or return multiple list of errors?

Brian-McM (approved):
Looks good, just had one minor comment.

lib/ipam/ipam.go
caseydavenport: Yeah, I considered this. Overall I figured it was unimportant because I don't think any of the calling code cares about _which_ block actually failed, and actually can't reliably rely on that anyway.

IMO logging out the intermediate errors is enough (they'll show up in the log) and we'll just return the final error to indicate the operation failed.

PR #1300, opened 2020-09-02
Allow host affinity release even if pool doesn't exist

caseydavenport:
## Description

We had an unnecessary limitation when freeing a host's blocks that they
belonged to an IP pool. This PR fixes that.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Enable releasing block affinities even if pool has been deleted.
```


Brian-McM:
If requiring the block belonging to an IP pool was an unnecessary requirement, would it make sense to instead remove that requirement from ReleaseAffinity so this change can be propagated to other uses of this function? I understand if it's a breaking change though.

Brian-McM (approved):

PR #1303, opened 2020-09-02
Add ownerref / finalizer to IPAM handles

caseydavenport:
## Description

This PR is intended to enable the use of Kubernetes GC mechanisms to track leaked IP
allocations in the cluster, enabling a controller to monitor and remove
the allocations at a later point in time.

Idea is this:
- Attach an owner reference back to the pod or node that triggered the allocation, so when it is deleted, so is the IPAM handle.
- Use a finalizer to prevent removal of the IPAM handle, and force an external controller or user to explicitly approve the deletion.

In this way, the controller can make sure the allocation is properly released and not orphaned. 

This PR on its own does not implement the GC, and tries to maintain the existing behavior for IPAM resources. However, it enables calling code to add finalizers and owner references to IPAM allocations by specifying attributes. Namely, the resource kind, name, and uid. 

Follow on work is required to fully enable this. Namely:
- calico/node should pass the kind, name and UID, for tunnel IP allocations.
- calico/cni should pass the kind, name, and UID, for pod IP allocations.
- kube-controllers should be enhanced to perform GC of finalizing handles. 

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/backend/k8s/resources/ipam_handle.go
+++ b/lib/backend/k8s/resources/ipam_handle.go
@@ -35,7 +35,7 @@ import (
 const (
 	IPAMHandleResourceName = "IPAMHandles"
 	IPAMHandleCRDName      = "ipamhandles.crd.projectcalico.org"
-	DeletedFinalizer       = "deleted.projectcalico.org"
+	IPAMFinalizer          = "ipam.projectcalico.org"
 )
 
 func NewIPAMHandleClient(c *kubernetes.Clientset, r *rest.RESTClient) K8sResourceClient {
@@ -94,14 +94,11 @@ func (c ipamHandleClient) toV3(kvpv1 *model.KVPair) *model.KVPair {
 	block := kvpv1.Value.(*model.IPAMHandle).Block
 	ownerRef := kvpv1.Value.(*model.IPAMHandle).OwnerReferences
 
-	// Check if we have enough information to add a reference to the owner
-	// of this IPAM allocation. Currently, we only do this for IPs belong to pods
-	// and IPs belonging to nodes.
 	var finalizers []string
 	if len(ownerRef) != 0 {
 		// Add a finalizer so that when the parent is deleted, we have an
 		// opportunity to take action on related resources (e.g. blocks).
-		finalizers = []string{DeletedFinalizer}
+		finalizers = []string{IPAMFinalizer}
 	}
 
 	return &model.KVPair{
--- a/lib/backend/k8s/resources/ipam_handle.go
+++ b/lib/backend/k8s/resources/ipam_handle.go
@@ -70,15 +70,15 @@ type ipamHandleClient struct {
 func (c ipamHandleClient) toV1(kvpv3 *model.KVPair) *model.KVPair {
 	handle := kvpv3.Value.(*apiv3.IPAMHandle).Spec.HandleID
 	block := kvpv3.Value.(*apiv3.IPAMHandle).Spec.Block
-	attrs := kvpv3.Value.(*apiv3.IPAMHandle).Spec.Attrs
+	refs := kvpv3.Value.(*apiv3.IPAMHandle).OwnerReferences
 	return &model.KVPair{
 		Key: model.IPAMHandleKey{
 			HandleID: handle,
 		},
 		Value: &model.IPAMHandle{
-			HandleID: handle,
-			Block:    block,
-			Attrs:    attrs,
+			HandleID:        handle,
+			Block:           block,
+			OwnerReferences: refs,
 		},
 		Revision: kvpv3.Revision,
 	}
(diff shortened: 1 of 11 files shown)


bcreane:
@caseydavenport - any tests (e.g. UTs) planned for this PR?

caseydavenport:
@bcreane yes, absolutely! That's what I am working on now. But would like some initial review on the code and approach before I lock it down too much.

lib/backend/k8s/resources/ipam_handle.go
```diff
@@ -89,6 +92,43 @@ func (c ipamHandleClient) toV3(kvpv1 *model.KVPair) *model.KVPair {
 	name := c.parseKey(kvpv1.Key)
 	handle := kvpv1.Key.(model.IPAMHandleKey).HandleID
 	block := kvpv1.Value.(*model.IPAMHandle).Block
+	attrs := kvpv1.Value.(*model.IPAMHandle).Attrs
+
+	// Determine who owns this handle, if possible.
```
caseydavenport, line 97: I need to add support for multiple owners, since a handle can be used for multiple IP assignments.

lib/backend/k8s/resources/ipam_handle.go
```diff
@@ -35,6 +35,7 @@ import (
 const (
 	IPAMHandleResourceName = "IPAMHandles"
 	IPAMHandleCRDName      = "ipamhandles.crd.projectcalico.org"
+	DeletedFinalizer       = "deleted.projectcalico.org"
```
fasaxc, line 38: I think if I saw this finalizer in `kubectl get -o yaml` output I'd think that it meant that the handle had been deleted already.  How about `ipam.projectcalico.org`?

fasaxc (changes requested):
Think this is a good idea; couple of thoughts for things that could go wrong:
* User deletes our CRD as part of a botched upgrade or in an effort to clean out the datastore to start over. Not sure what would happen?
* User manually deletes a handle, do we do anything?

lib/backend/k8s/resources/ipam_handle.go
caseydavenport: Yes, very much agree

caseydavenport:
> User manually deletes a handle, do we do anything?

If a user manually deletes a handle, then it will follow the same procedure as if the pod was deleted. It will go into "finalizing" state and then will get deleted when the yet-to-be-introduced controller sees the finalizing handle and actions it. 

> User deletes our CRD as part of a botched upgrade or in an effort to clean out the datastore to start over. Not sure what would happen?

Hm, not sure. I would guess the same behavior as today (handles just disappear) but I will test this.

caseydavenport:
> If a user manually deletes a handle

Today, the IP would remain allocated but leaked because the handle would be gone when we go to free the allocation. 

With this change, the pod would remain with the IP but the IP would be freed up in IPAM and not leaked. I suppose we could add an additional check that the pod is actually no longer running before we do the garbage collection. Then, we would keep the IP around until it was no longer needed, but also no lose track of the allocation via the handle (unless a user manually removed the fnializer themselves, in which case all bets are off)

caseydavenport:
> Hm, not sure. I would guess the same behavior as today (handles just disappear) but I will test this.

I tested this out today. Looks like the CRD itself blocks on the finalization of all the individual handles in the cluster, so it would wait for our controller to come in and release all the IP addresses.

lib/ipam/ipam.go
```diff
@@ -1325,12 +1330,23 @@ func (c ipamClient) ReleaseByHandle(ctx context.Context, handleID string) error
...
 	handle := allocationHandle{obj.Value.(*model.IPAMHandle)}
 
-	for blockStr, _ := range handle.Block {
+	for blockStr := range handle.Block {
 		_, blockCIDR, _ := net.ParseCIDR(blockStr)
 		if err := c.releaseByHandle(ctx, handleID, *blockCIDR); err != nil {
 			return err
 		}
 	}
+
```
caseydavenport, line 1339: Found this while testing out various weird scenarios (e.g., deleting some resources but not others).

In this case, if IP addresses are manually free'd without actually deleting the handle, then subsequent calls to `ReleaseByHandle` will never delete the handle itself.

This makes sure that we delete the handle even if `ReleaseByHandle` didn't find any IPs to free.

robbrockbank:
No particular comments, but had a couple of questions:
-  IIUC we'll end up with an owner ref per pod IP.  Does this scale ok in terms of (a) potentially having a large number of owner references on the same resource, and (b) having a large number of garbage collection refs. 
-  I don't recall the issues that prompted this PR, but was it a purely KDD issue?

caseydavenport:
> IIUC we'll end up with an owner ref per pod IP. Does this scale ok in terms of (a) potentially having a large number of owner references on the same resource, and (b) having a large number of garbage collection refs.

We shouldn't have a large number of owner references on the same IPAM handle resource, but we might have a large number referring to the same pod resource if we assign multiple addresses to a pod (due to a bug). And yeah, we could potentially have a lot of GC refs. 

Honestly, I'm not sure how this scales. I'll see if I can find any documentation or prior work in this area. But in the absence of that, will need to do some testing myself. 

> I don't recall the issues that prompted this PR, but was it a purely KDD issue?

It's not really one specific issue, rather a trend of IP leak issues over time. And not purely KDD. This obviously isn't going to improve anything for etcd clusters, but should improve reliability of GC for kdd clusters by a lot.

caseydavenport:
> Honestly, I'm not sure how this scales. I'll see if I can find any documentation or prior work in this area. But in the absence of that, will need to do some testing myself.

For what it's worth, I'm not too worried about this. Every pod launched via a Deployment / DaemonSet / ReplicaSet has an OwnerReference back to the object that created it, so this seems to be a pretty mainline type of use for this.

fasaxc (approved):
I think we should give this a go but, since it's slightly risky in those corner cases I'd be inclined to add a flag to disable it when you come to add the function to the CNI plugin.

PR #1304, opened 2020-09-08
remove spurious annotations and creationtimestamps from crds

bcreane:
## Description
Remove incorrect `annotations` and `creationTimestamp` from crds, added in https://github.com/projectcalico/calico/pull/3579. Requires cherry-pick back to `release-v3.15` and corresponding manifest changes in https://github.com/projectcalico/calico/.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/config.patch
+++ b/config.patch
@@ -11,3 +11,183 @@ diff -Naur config.orig/crd/crd.projectcalico.org_ipamblocks.yaml config/crd/crd.
                  type: array
                attributes:
                  items:
+--- config.orig/crd/crd.projectcalico.org_bgpconfigurations.yaml	2020-09-15 17:41:44.686361905 +0000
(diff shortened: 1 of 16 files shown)


caseydavenport:
@bcreane this looks good to me, but I think these will get added back in whenever someone runs the "make gen-crds" target. 

Do you think there's some logic we could add to that target to sneakily remove this cruft as part of generating the files? Guessing there is some series of shell magics we could use? https://github.com/projectcalico/libcalico-go/blob/a370cde17823112605e717a200bcc3a42d0f75a7/Makefile#L62-L65

bcreane:
@caseydavenport - back to you for review.

caseydavenport (approved):
@bcreane this is beautiful, thank you!

We'll just have to remember to update config.patch if we add any new resources.

bcreane:
@caseydavenport - any need to cherry-pick into release branches?

PR #1307, opened 2020-09-10
node addresses

tomastigera:
## Description

    add NodeAddressKubeInternalIP and NodeAddressKubeExternalIP
    
    Include the Internal/External IP of a node k8s node in the list of
    addresses in a calico node.
    
    If an address is already included from BPG as NodeAddressBGPIP, it will
    appear twice on the list. It is the users task to filter out the address
    of the type it wants and and watch for possible duplicates.

    Node.Spec.Addresses contains all known node's addresses
    
    Modeled after k8s Node.Status.Addresses, it is a list of <type:address>
    pairs that aggregate known addresses of a node. This makes it easy to
    add new address types and/or multiple addresses of the same type.
    
    Currently, the list aggregates addresses from all calico node fields.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


lib/apis/v3/node.go
```diff
@@ -98,6 +103,27 @@ type NodeWireguardSpec struct {
 	InterfaceIPv4Address string `json:"interfaceIPv4Address,omitempty" validate:"omitempty,ipv4"`
 }
 
+// NodeAddressType represents a node address type as a string
+type NodeAddressType string
+
+const (
+	NodeAddressIP                NodeAddressType = "IP"
+	NodeAddressVxlanTunnelIP     NodeAddressType = "VxlanTunnelIP"
```
fasaxc, line 111: Nit: Go style convention is to put acronyms in upper case.  So, should be `NodeAddressVXLANTunnelIP`; k8s style convention is that all the characters in the acronym should be int he same case so should be `VXLANTunnelIP` (and we try to follow their guidelines for our v3 API).

fasaxc: Reference: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#naming-conventions

lib/apis/v3/node.go
```diff
@@ -98,6 +103,27 @@ type NodeWireguardSpec struct {
...
+	NodeAddressVxlanMAC          NodeAddressType = "VxlanMAC"
+	NodeAddressIPIPTunnelIP      NodeAddressType = "IPIPTunnelIP"
+	NodeAddressWireguardTunnelIP NodeAddressType = "WireguardTunnelIP"
+	NodeAddressK8sInternalIP     NodeAddressType = "K8sInternalIP"
```
fasaxc, line 115: I _think_ we have other fields that reference k8s that use `KubeFoo` rather than `K8sFoo` so probably best to use that

lib/backend/k8s/resources/node.go
```diff
@@ -300,6 +303,76 @@ func K8sNodeToCalico(k8sNode *kapiv1.Node, usePodCIDR bool) (*model.KVPair, erro
...
+				Type:    apiv3.NodeAddressK8sExternalIP,
+				Address: kaddr.Address,
+			}
+		default:
+			continue
+		}
+
+		// k8s InternalIP and possibly ExternalIP are likely already in the list
+		// from the BPG spec IPv4 or IPv6 fields. Find them and upgrade them to
+		// the k8s type.
```
fasaxc, line 356: Think I'd leave our this step; the client can dedupe if it matches their use case but, in general, I prefer to have the raw data in the datamodel and helper functions to massage it on read if that's useful.  (It's difficult to rejig the datamodel later if we decide that we've lost some information.)

fasaxc (changes requested):
Few nits/suggestions but was broadly what I was expecting from our discussion :-)

lib/backend/k8s/resources/node.go
tomastigera: I was considering that, and than I opted for deduping here since each user, instead of just looping over the list would need to check if it is already seen an address, not to use it 2x - which may make using this more difficult. My reasoning was that "upgrading" `NodeAddressIP` to `NodeAddressK8sXXXIP` only adds information. It is still an IP, but also known to K8s.

If we pass the an address 2x, should we also make the `NodeAddressIP` to be more precise like `NodeAddressBGPIP?

fasaxc (changes requested):

lib/apis/v3/node.go
```diff
@@ -98,6 +103,27 @@ type NodeWireguardSpec struct {
...
+	NodeAddressWireguardTunnelIP NodeAddressType = "WireguardTunnelIP"
+	NodeAddressKubeInternalIP    NodeAddressType = "KubeInternalIP"
+	NodeAddressKubeExternalIP    NodeAddressType = "KubeExternalIP"
+)
+
+// NodeAddress represents an address assigned to the node annotated with its type.
+type NodeAddress struct {
+	// Type denotes what sort of address it is.
+	Type NodeAddressType `json:"type"`
+	// Address is a string rpresentation of the actual address.
```
caseydavenport, line 123: typo: rpresentation

lib/apis/v3/node.go
```diff
@@ -98,6 +103,27 @@ type NodeWireguardSpec struct {
...
+// NodeAddressType represents a node address type as a string
+type NodeAddressType string
+
+const (
+	NodeAddressBGPIP             NodeAddressType = "BGPIP"
+	NodeAddressVXLANTunnelIP     NodeAddressType = "VXLANTunnelIP"
+	NodeAddressVXLANMAC          NodeAddressType = "VXLANMAC"
+	NodeAddressIPIPTunnelIP      NodeAddressType = "IPIPTunnelIP"
+	NodeAddressWireguardTunnelIP NodeAddressType = "WireguardTunnelIP"
+	NodeAddressKubeInternalIP    NodeAddressType = "KubeInternalIP"
```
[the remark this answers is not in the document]
caseydavenport: If we do actually need to track the source of the address, my suggestion would be to extend the NodeAddress struct rather than try to put two bits of information (type and source) into a single field. 

e.g.,

```yaml
- address: 1.2.3.4/32
  type: InternalIP
  source: Kubernetes

```

lib/apis/v3/node.go
```diff
@@ -98,6 +103,27 @@ type NodeWireguardSpec struct {
...
+	NodeAddressKubeExternalIP    NodeAddressType = "KubeExternalIP"
+)
+
+// NodeAddress represents an address assigned to the node annotated with its type.
+type NodeAddress struct {
+	// Type denotes what sort of address it is.
+	Type NodeAddressType `json:"type"`
+	// Address is a string rpresentation of the actual address.
+	Address string `json:"address"`
```
caseydavenport, line 124: We probably want to implement validation on this struct and its fields, right?

caseydavenport (changes requested):

lib/apis/v3/node.go
```diff
@@ -98,6 +103,27 @@ type NodeWireguardSpec struct {
 	InterfaceIPv4Address string `json:"interfaceIPv4Address,omitempty" validate:"omitempty,ipv4"`
 }
 
+// NodeAddressType represents a node address type as a string
+type NodeAddressType string
+
+const (
+	NodeAddressBGPIP             NodeAddressType = "BGPIP"
+	NodeAddressVXLANTunnelIP     NodeAddressType = "VXLANTunnelIP"
+	NodeAddressVXLANMAC          NodeAddressType = "VXLANMAC"
```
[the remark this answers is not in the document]
tomastigera: Sure, i just followed the naming in the structs, makes sense

lib/apis/v3/node.go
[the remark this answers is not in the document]
tomastigera: Not sure about the source. It is more about what is the role of the address. What is a meaning of an "internal IP"? Perhaps in the module that uses it, which is always executed in kubernetes, the meaning is well defined and the const should perhaps defined there.

If we extend the struct we loose the nice compatibility properties.

lib/apis/v3/node.go
tomastigera: Since we never use a value written by a user, it feels redundant.

lib/apis/v3/node.go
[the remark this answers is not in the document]
caseydavenport: > If we extend the struct we loose the nice compatibility properties.

Hm, say more about this. What compatibility properties are you referring to? 

>  It is more about what is the role of the address. What is a meaning of an "internal IP"?

Yes, I agree. And IIUC, the role of the address is not "An internal IP that we learned from Kubenrnetes" but "An internal IP", right?

An internal IP is one that can be used to access the node from within the cluster, I think.

lib/apis/v3/node.go
caseydavenport: I thought the plan was to make this field writable for etcd mode to work. Right?

tomastigera: Right, at some point, yes. I was pondering validation. For the type, if it is unknown, the consumer skips it, as it does not understand the address. If the consumer is older, there is a chance that it does not know a new but validated type. For the address itself, we would need to validate the combinations address-type. If we leave taking care of duplicities to the user, why not to let it check that an IP is an IP (unless we provide types for 4/6 it needs to figure that already) etc.

But I am happy to add a validator for the tupple.

lib/apis/v3/node.go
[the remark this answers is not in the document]
tomastigera: > > If we extend the struct we loose the nice compatibility properties.
>
> Hm, say more about this. What compatibility properties are you referring to?

If we wanted to add an address that does not have a source, but some other property (well, for instance a version like IPv4/6), then some addresses use one field, others other or all fields etc. A union would be nice. Imo an `int` would serve us well, but the string also carries the human readable version, which is nice :) Feels like structuring got us here.

> > It is more about what is the role of the address. What is a meaning of an "internal IP"?
> 
> Yes, I agree. And IIUC, the role of the address is not "An internal IP that we learned from Kubenrnetes" but "An internal IP", right?

It is an IP assigned by kubernetes and it is internal to a kubernetes cluster. Technically the BPG address (which we learn by a range of methods) is usually assigned by the same source, has a different role it seems to me. But KubeInternalIPv4 feels a bit too much. I guess the cleanest / future proof might be:

```
type NodeAddress struct {
    Type NodeAddressType // IP, MAC, DNS, HostName etc.
    Annotations map[string]string
}

k8sInternalIPv4 := {
    Address: "1.2.3.4",
    Type: NodeAddressTypeIP,
    Annotations: map[string]string{"version":"v4", "source":"kubernetes-internal"},
}

eth0IP := {
    Address: "1.2.3.4",
    Type: NodeAddressTypeIP,
    Annotations: map[string]string{"version":"v4", "source":"netlink", "iface":"eth0"},
}
```

It is a bit more heavyweight and uncessary.

If we did the deduplication of address here, we could merge the annotations (what about source?) but that would not work well with etcd, so better to leave it to the user...

tomastigera (commented):
Happy to add the validator once we agree on the type format.

caseydavenport:
>     Annotations: map[string]string{"version":"v4", "source":"netlink", "iface":"eth0"},

I don't think we would want to add an annotations map to this sub-structure. The whole reason this is a slice of structures rather than a slice of strings is so that we can extend with structured data. So adding an arbitrary key/value map would defeat that purpose, as does overloading a single field to have multiple meanings.

tomastigera:
Afte some discussion with @caseydavenport , we decided to go simpler and to fix what we currently need rather then trying to predict future usecases and merge to many address types together. However, we will leave room for future extensions.

PR #1312, opened 2020-09-14
Restore defaulting to LoadConfigFromXXX methods.

fasaxc:
## Description

#1279 change the default datastore type to kubernetes but, as a side effect, it also stopped LoadConfigFromEnvironment from defaulting the datastore type.  This broke Typha, calicoctl and Node, which use that method on the assumption that it does defaulting.

This PR refactors the code to make sure that defaulting happens in all the public LoadConfigXXX functions.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/apiconfig/load.go
+++ b/lib/apiconfig/load.go
@@ -1,3 +1,17 @@
+// Copyright (c) 2017-2020 Tigera, Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package apiconfig
 
 import (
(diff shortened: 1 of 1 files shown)


lib/apiconfig/load.go
```diff
@@ -17,60 +31,38 @@ import (
...
+		return nil, fmt.Errorf("syntax error in %s: %v", filename, err)
 	}
 
-	return c, err
+	return c, nil
 }
 
-// LoadClientConfig loads the ClientConfig from the supplied bytes containing
-// YAML or JSON format data.
+// LoadClientConfig loads the ClientConfig from the supplied bytes containing YAML or JSON format data.
```
neiljerram, line 58: nit: LoadClientConfig -> LoadClientConfigFromBytes

lib/apiconfig/load.go
```diff
@@ -81,20 +73,49 @@ func LoadClientConfigFromBytes(b []byte) (*CalicoAPIConfig, error) {
...
+	if c.Spec.DatastoreType == Kubernetes {
+		// Default to using $(HOME)/.kube/config, unless another means has been configured.
+		switch {
+		case c.Spec.Kubeconfig != "":
+			log.WithField("kubeconfig", c.Spec.Kubeconfig).Debug("kubeconfig provided.")
+		case c.Spec.K8sAPIEndpoint != "":
+			log.WithField("apiEndpoint", c.Spec.K8sAPIEndpoint).Debug("API endpoint provided.")
+		case os.Getenv("HOME") == "":
+			// No home directory, can't build a default config path.
+			log.Debug("No home directory, default path doesn't apply.")
```
neiljerram, line 114: Warning?  Bearing in mind that this means there won't be any effective k8s config.

neiljerram (approved):
Two nits, but I'm happy to leave those to your judgement.

Will last week's Typha change still be OK?  I believe, after this PR, Typha will no longer see `""` - is that OK?

lib/apiconfig/load.go
fasaxc: I _think_ some legitimate paths go through here (Typha maybe?) because they fill in some values after this step.

PR #1323, opened 2020-09-28
Update go.mod to golang 1.15 and update ginkgo / gomega libraries

Brian-McM:
## Description

We're using the build image that has 1.15 so we should bump the go.mod version so we're actually building a golang 1.15 binary. The test packages are old and the new ones have some Panic testing abilities I'd like to use.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


realgaurav (approved):
It doesn't have to be in this PR OR please ignore me if you already have some other process in workflow to address ballooning go.sum (i.e. `go mod tody`)

Brian-McM:
@realgaurav it's already been run, there's a mod tidy make target.

PR #1324, opened 2020-09-29
Node.Addresses list all address at which the node is reachable.

tomastigera:
This allows to pass multiple address, e.g. the external and internal
address of a node in kubernetes cluster.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


lib/backend/k8s/resources/node.go
```diff
@@ -300,6 +303,26 @@ func K8sNodeToCalico(k8sNode *kapiv1.Node, usePodCIDR bool) (*model.KVPair, erro
 	}, nil
 }
 
+func listAllAddresses(calicoNode *apiv3.Node, k8sNode *kapiv1.Node) {
```
fasaxc, line 306: I'd change list to `fill` to emphasise that it mutates the calicoNode

fasaxc (approved):
Just one nit but looks good.

PR #1325, opened 2020-10-01
Concurrent operations on handles via UID / revision

caseydavenport:
## Description

Found some race conditions that exist when using ReleaseByHandle. This code does a few things:

- Correctly returns UID on resources backed by CRDs. 
- Updates ReleaseByHandle to take revision and UID information.

I originally raised this PR to address a scenario I found while working on the IPAM handle GC controller using finalizers (this approach was abandoned, but the race condition still exists in the API). In my hacking, I encountered a race condition where:

1. CNI plugin goes to allocate an IP address, creates a handle
1. CNI plugin attempts to write the block, but gets a CAS error
1. CNI plugin deletes handle in preparation for a retry.
1. CNI plugin retries the above steps 1-3, successfully allocates the IP. Handle gets a new UID.

In this scenario, the IPAM handle would enter finalizing state in step 3 above, kube-controllers would see that finalizing handle, and call `ReleaseByHandle`. This would race with the retry attempt, thus releasing the IP after the CNI plugin successfully allocates it on the second attempt. TL;DR - ReleaseByHandle releases IPs that might have been allocated after the calling code decided it was safe to release the IPs. 

Despite ultimately taking a different approach on the IPAM GC controller, I think this PR is probably still useful. Here is another related PR that showcases an existing race condition with handles which results in us accidentally releasing an IP address that is still in use: https://github.com/projectcalico/kube-controllers/pull/489

The changes in this PR prevent the race condition by allowing users of the API to provide revision and UID information on the ReleaseByHandle API.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/ipam/ipam.go
+++ b/lib/ipam/ipam.go
@@ -1073,7 +1073,15 @@ func (c ipamClient) releaseIPsFromBlock(ctx context.Context, handleMap map[strin
 		// Success - decrement handles.
 		logCtx.Debugf("Decrementing handles: %v", handles)
 		for handleID, amount := range handles {
-			if _, err := c.decrementHandle(ctx, handleMap[handleID], blockCIDR, amount); err != nil {
+			// Check if we have a cached handle for this ID.
+			h, ok := handleMap[handleID]
+			if !ok {
+				h, err = c.blockReaderWriter.queryHandle(ctx, handleID, "")
+				if err != nil {
+					return nil, err
+				}
+			}
+			if _, err := c.decrementHandle(ctx, h, blockCIDR, amount); err != nil {
 				logCtx.WithError(err).Warn("Failed to decrement handle")
 			}
 		}
--- a/lib/ipam/ipam.go
+++ b/lib/ipam/ipam.go
@@ -25,7 +25,6 @@ import (
 	"golang.org/x/sync/semaphore"
 
 	log "github.com/sirupsen/logrus"
-	"k8s.io/apimachinery/pkg/types"
 
 	v3 "github.com/projectcalico/api/pkg/apis/projectcalico/v3"
 	libapiv3 "github.com/projectcalico/libcalico-go/lib/apis/v3"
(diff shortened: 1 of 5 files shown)


lib/apis/v3/ipam_handle.go
```diff
@@ -39,6 +39,9 @@ type IPAMHandle struct {
 type IPAMHandleSpec struct {
 	HandleID string         `json:"handleID"`
 	Block    map[string]int `json:"block"`
+
+	// +optional
+	Deleted bool `json:"deleted"`
```
caseydavenport, line 44: Turns out we just weren't writing or using this field ever :cry:

lib/backend/k8s/resources/ipam_handle.go
```diff
@@ -93,6 +96,12 @@ func (c ipamHandleClient) toV3(kvpv1 *model.KVPair) *model.KVPair {
 	handle := kvpv1.Key.(model.IPAMHandleKey).HandleID
 	block := kvpv1.Value.(*model.IPAMHandle).Block
 	ownerRef := kvpv1.Value.(*model.IPAMHandle).OwnerReferences
+	del := kvpv1.Value.(*model.IPAMHandle).Deleted
+
+	var uid types.UID
+	if kvpv1.UID != nil {
+		uid = *kvpv1.UID
+	}
```
caseydavenport, line 104: Make sure we include UID in the request if one is given

lib/ipam/ipam.go
```diff
@@ -531,7 +531,6 @@ func (s *blockAssignState) findOrClaimBlock(ctx context.Context, minFreeIps int)
 					logCtx.Errorf(errString)
 					return nil, false, errors.New(errString)
 				}
-				s.datastoreRetryCount++
```
caseydavenport, line 534: This was dead code.

lib/ipam/ipam.go
```diff
@@ -785,26 +784,36 @@ func (c ipamClient) AssignIP(ctx context.Context, args AssignIPArgs) error {
...
+				// consistent with allocation state.
+				if _, decErr := c.decrementHandle(ctx, h, blockCIDR, 1); decErr != nil {
+					log.WithError(decErr).Warn("Failed to decrement handle")
+				}
+			}
 			if _, ok := err.(cerrors.ErrorResourceUpdateConflict); ok {
 				log.WithError(err).Debug("CAS error assigning IP - retry")
 				continue
 			}
-
```
caseydavenport, line 801: Moved this up a few lines so that we decrement the handle even if we hit a CAS on the block (since we'll increment it on retry so we need to also decrement it here).

lib/ipam/ipam.go
```diff
@@ -1458,27 +1472,50 @@ func (c ipamClient) IPsByHandle(ctx context.Context, handleID string) ([]net.IP,
...
+		}
+	}
 	return nil
 }
 
-func (c ipamClient) releaseByHandle(ctx context.Context, handleID string, blockCIDR net.IPNet) error {
+// releaseByHandle releases all addresses with the given handle object, and returns an updated handle object or nil if
+// the handle was deleted, or nil and an error if an error was encountered.
+func (c ipamClient) releaseByHandle(ctx context.Context, handle *model.KVPair, blockCIDR net.IPNet) (*model.KVPair, error) {
+	handleID := handle.Key.(model.IPAMHandleKey).HandleID
```
robbrockbank, line 1518: I think this retry processing is not valid anymore if passing in revision and UUID.

Who would now handle this retry?  The calling code?   Wondering if passing in the revision is correct.
Is the IPAM data going to contain a UUID and handle?

caseydavenport:

PR #1326, opened 2020-10-02
Properly respect Deleted flag

caseydavenport:
## Description

We were never actually writing the deleted flag to the datastore. This flag is used to get "Compare and Delete" behavior, when the k8s API has none. 

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


fasaxc (commented):
Worth an FV test so we don't regress it again?

PR #1334, opened 2020-10-20
updated default in doc for BPFDataIfacePattern

tomastigera:
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/config/crd/crd.projectcalico.org_felixconfigurations.yaml
+++ b/config/crd/crd.projectcalico.org_felixconfigurations.yaml
@@ -67,7 +67,7 @@ spec:
                   that Calico workload traffic flows over as well as any interfaces
                   that handle incoming traffic to nodeports and services from outside
                   the cluster.  It should not match the workload interfaces (usually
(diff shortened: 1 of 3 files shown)


neiljerram (changes requested):
This change is good, but the PR should also include the same change in the CRD file.

If you run `./install-git-hooks` from the root of the libcalico-go repo, I think it will do that automatically for you when you try to commit.

neiljerram (approved):
LGTM, thanks!

caseydavenport:
I was lamenting this yesterday. It's a bit sad that our defaulting logic is implemented in Felix and our the defaulting documentation is implemented in libcalico-go. 

I think there are a few other places where we do this as well (I'm looking at you, confd). But for the most part we're moving towards putting defaulting in libcalico-go.

I think it's a nicer user-experience if fields that actually have defaults get populated when the resource is created, so that on a `Get` call they can see all of the choices the system made for them. It's less magic. 

To be clear, in no way asking you to do anything about it!

PR #1247, opened 2020-06-05
Handle backend watch, if upstream closes channel[ClosedByRemote] 

krishgobinath:
This PR handles following things.

If upstream Kubernetes API server closes watch the channel with backend libcalico-go, then stop the backend channels connected downstream and don't send any "ErrorTerminated" events back to downstream channels.

If upstream Kubernetes API server, sends Api.WatchError to backend libcalico-go, wrap this status change as an "ApiError" to downstream. Downstream will resync the channel.

Local Unit test passed.

PR #1337, opened 2020-10-29
etcd backend not passing watcher errors up the stack, and errors not …

robbrockbank:
…triggering full resync in watcher cache

## Description
This contains a couple of fixes

-  In etcd backend, make sure errors due to watches are passed as an event - this is to ensure real errors result in a re-sync.  In particular compacted revisions manifest as an error on the watch channel - we were not sending an error and just terminating the channel causing us to loop indefinitely trying to use a compacted revision.

-  Once in-sync, do not go out of sync in the watcher syncer.  There is no point in changing status, we already cache the revisions so that once we are connected again we send the correct updates.  Felix was not using the status change, and the status change was actually causing issues with other components that were not expecting the status change.

-  Remove the ErrorWatchTerminated definition so that we don't try to use it.  This forces us to fix up applications that were relying on this error type.

Should fix: https://github.com/projectcalico/calico/issues/4109

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/backend/etcdv3/watcher.go
+++ b/lib/backend/etcdv3/watcher.go
@@ -98,6 +98,7 @@ func (wc *watcher) watchLoop() {
 		if kvps, err = wc.listCurrent(); err != nil {
 			log.Errorf("failed to list current with latest state: %v", err)
 			// Error considered as terminating error, hence terminate watcher.
+			wc.sendError(err)
 			return
 		}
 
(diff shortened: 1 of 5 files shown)


mgleung (approved):
LGTM

doublek (approved):
LGTM - the etcd watcher fixes are straightforward and looks correct. I looked through the scenarios for the InSync fix and I think it should work as expected.

neiljerram:
@robbrockbank Thanks for handling this.

At the risk of being annoying - since I haven't understood the detail here - can I just ask if the combination of #1247 + this still makes sense?  I'm asking because it appears to me that this PR reverts about half of the changes in #1247 .

(Also cc @caseydavenport , as the author of #1247 )

PR #1343, opened 2020-11-13
Ensure that in dual stack environments, IPv4 addresses are not leaked…

mgleung:
… when IPv6 addresses are not available

## Description

Fixes cases where if both IPv4 and IPv6 addresses are attempted to be auto assigned and one of the assignments fails, the other address is cleaned up appropriately.

Fixes https://github.com/projectcalico/cni-plugin/issues/908

## Todos
- [x] Tests
- [x] Documentation
- [x] Release note

## Release Note

```release-note
Fix an IPAM auto assignment bug that caused IP addresses to be leaked in dual stack environments when assignment partially succeeded.
```

### DIFF
--- a/lib/ipam/ipam.go
+++ b/lib/ipam/ipam.go
@@ -127,7 +127,7 @@ func (c ipamClient) AutoAssign(ctx context.Context, args AutoAssignArgs) ([]net.
 			log.Errorf("Error assigning IPV6 addresses: %v", err)
 		}
 
-		if len(v6list) == 0 {
+		if len(v6list) != args.Num6 {
 			if len(v4list) != 0 {
 				// Release any v4 IPs that were assigned when we were expecting to assign
 				// both v4 and v6 IPs and v6 IP assignment did not complete.
--- a/lib/ipam/ipam.go
+++ b/lib/ipam/ipam.go
@@ -108,6 +108,12 @@ func (c ipamClient) AutoAssign(ctx context.Context, args AutoAssignArgs) ([]net.
 		}
 	}
 
+	if len(v4list) != args.Num4 && args.Num6 != 0 {
+		// No need to assign any V6 if V4 has already failed.
+		log.Debugf("Skipping assignment of IPv6 addresses since IPv4 assignment has failed")
+		return v4list, nil, nil
+	}
+
 	if args.Num6 != 0 {
 		// If no err assigning V4, try to assign any V6.
 		log.Debugf("Assigning IPv6 addresses")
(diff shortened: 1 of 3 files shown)


lib/ipam/ipam.go
```diff
@@ -119,7 +125,24 @@ func (c ipamClient) AutoAssign(ctx context.Context, args AutoAssignArgs) ([]net.
...
+				// both v4 and v6 IPs and v6 IP assignment did not complete.
+				// This should only occur in dual stack environments.
+				// This prevents v4 IPs from being assigned when v6 assignment fails.
+				v4IPs := []net.IP{}
+				for _, v4 := range v4list {
+					v4IPs = append(v4IPs, *net.ParseIP(v4.IP.String()))
+				}
+				_, releaseErr := c.ReleaseIPs(ctx, v4IPs)
+				if releaseErr != nil {
+					log.Errorf("Error releasing IPv4 addresses %+v on IPv6 address assignment failure: %s", v4IPs, err)
```
caseydavenport, line 142: I think if we hit this line, we're going to leak the v4 addresses - we'll tell the caller that no v4 addresses were allocated, but actually some were.

lib/ipam/ipam.go
```diff
@@ -119,7 +125,24 @@ func (c ipamClient) AutoAssign(ctx context.Context, args AutoAssignArgs) ([]net.
 		v6list, err = c.autoAssign(ctx, args.Num6, args.HandleID, args.Attrs, args.IPv6Pools, 6, hostname, args.MaxBlocksPerHost, args.HostReservedAttrIPv6s)
 		if err != nil {
 			log.Errorf("Error assigning IPV6 addresses: %v", err)
-			return v4list, v6list, err
+		}
+
+		if len(v6list) != args.Num6 {
```
caseydavenport, line 130: I think this fundamentally changes the way this function was intended to behave. Admittedly, we're not really using it in that way ourselves, but the intention of the function was always that it would allocate as many addresses as it could in order to satisfy the request and return them to the caller.

Imagine this scenario:
- Call auto-assign asking for 3 IPv6 addresses and 1 IPv4 address.
- We only allocate 2 IPv6 addresses, and 1 IPv4 address.

Currently, the function will return 2 IPv6 addresses and 1 IPv4 address.

With this change, it will return 2 IPv6 addresses and 0 IPv4 addresses.

caseydavenport: Same goes for the reverse scenario (where we hit the v6 case above). We'll no longer be doing a best-effort allocation of as many addresses as we can to satisfy the request.

caseydavenport (changes requested):
So, I'm not yet 100% sold on putting this logic into AutoAssign.

Based on the current API, I believe it's the caller's responsibility to determine if their request was adequately satisfied and either:
- Free the addresses 
- Ask for more addresses
- Something else.

I might be missing something, but my current sense is that this fix should be in the IPAM plugin rather than here.

mgleung:
Closing in favor of https://github.com/projectcalico/cni-plugin/pull/1010

PR #1345, opened 2020-11-17
Handle malformed handle IDs

caseydavenport:
## Description

In https://github.com/projectcalico/cni-plugin/issues/821 we introduced
a bug that caused some handles to be written to the datastore using the
wrong format. Namely, `k8s-pod-network.<id>\r\neth0`. 

This resulted in an inability to properly release these handles, because
the IPAM handle name doesn't match the name for the handle in the IPAM
block. 

This PR introduces santization logic which is intended to remove the
bugged suffix from handles before release, so that we properly match.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Fix releasing of improperly formatted handles as a result of host-local IPAM migration bug. 
```

### DIFF
--- a/lib/ipam/ipam_block.go
+++ b/lib/ipam/ipam_block.go
@@ -337,11 +337,18 @@ func (b allocationBlock) attributeRefCounts() map[int]int {
 func (b allocationBlock) attributeIndexesByHandle(handleID string) []int {
 	indexes := []int{}
 	for i, attr := range b.Attributes {
-		if attr.AttrPrimary != nil && *attr.AttrPrimary == handleID {
+		if attr.AttrPrimary != nil && sanitizeHandle(*attr.AttrPrimary) == handleID {
 			indexes = append(indexes, i)
 		}
 	}
 	return indexes
+}
+
+// sanitizeHandle fixes any improperly formatted handles that we might come across.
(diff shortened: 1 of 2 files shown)


caseydavenport:
@fasaxc this seems to fix the test I added here: https://github.com/projectcalico/kube-controllers/pull/557

Which is broken on without this fix.

lib/ipam/ipam_block.go
```diff
@@ -337,13 +337,19 @@ func (b allocationBlock) attributeRefCounts() map[int]int {
...
 			indexes = append(indexes, i)
 		}
 	}
 	return indexes
 }
 
+// sanitizeHandle fixes any improperly formatted handles that we might come across (i.e., due to other bugs
+// causing improperly written data.)
+func sanitizeHandle(handleID string) string {
+	return strings.Split(handleID, "\r")[0]
```
fasaxc, line 350: Since workarounds like this are hard-won, I'd err on the side of a detailed comment:
```
// Work around malformed handles written by host-local to Calico IPAM migration after
// host-local IPAM changed its file format: https://github.com/projectcalico/cni-plugin/issues/821.
```

fasaxc (approved):

PR #1350, opened 2020-12-10
replaces unicode quotation marks in code comments with ascii double quotes

marshallford:
## Description
See title. Also fixes some unicode apostrophes as well. bug fix. A step towards resolving https://github.com/projectcalico/calico/issues/3846 and related to https://github.com/projectcalico/libcalico-go/pull/1289.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/config/crd/crd.projectcalico.org_felixconfigurations.yaml
+++ b/config/crd/crd.projectcalico.org_felixconfigurations.yaml
@@ -135,11 +135,11 @@ spec:
                   traffic that goes from a workload endpoint to the host itself (after
                   the traffic hits the endpoint egress policy). By default Calico
                   blocks traffic from workload endpoints to the host itself with an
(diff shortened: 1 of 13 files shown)


caseydavenport:
@marshallford thank you!

What do you think about some sort of check or test to make sure these don't accidentally get added back? They're pretty hard to spot in a code-review, and I'm assuming they are being added accidentally.

caseydavenport (approved):
LGTM! A follow on with a test would be great, but this is useful on its own so merging now.

marshallford:
> What do you think about some sort of check or test to make sure these don't accidentally get added back? They're pretty hard to spot in a code-review, and I'm assuming they are being added accidentally.

That sounds like a great follow on PR!

PR #1357, opened 2021-01-07
Add support for Kubernetes EndPort field

rikatz:
Signed-off-by: Ricardo Pchevuzinske Katz <ricardo.katz@gmail.com>

## Description

This is a new feature that is being developed at [kubernetes #97058](https://github.com/kubernetes/kubernetes/pull/97058)

With this feature, setting endPort will create a range of Ports in Network Policy instead of a single port, so libcalico-go needs to be aware of what to do with that new field.

This field is going to be Feature Gated at Kubernetes API Server for the alpha release, but this is not a problem for the lib and the kubernetes controller to be aware of this field.

Note for reviewer: The PR at k/k wasn't merged yet, so it's not possible to test/change go.mod. Once this is merged at k/k and released, this PR can go forward

## Todos
- [x] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Kubernetes Network Policy EndPort is now supported
```


caseydavenport:
> lib/upgrade/migrator/felixconfig_test.go:26:2: could not import github.com/projectcalico/libcalico-go/lib/backend/syncersv1/updateprocessors (lib/backend/syncersv1/updateprocessors/globalnetworkpolicyprocessor.go:22:2: could not import github.com/projectcalico/libcalico-go/lib/backend/k8s/conversion (lib/backend/k8s/conversion/conversion.go:405:8: p.EndPort undefined (type "k8s.io/api/networking/v1".NetworkPolicyPort has no field or method EndPort))) (typecheck)

Looks like we might need to bump the API version that we're using in order to do this?

rikatz:
Yeap this is why I've put in the description we should wait the PR in k/k to get merged :(

Is there an alternative way Casey?

Thanks!

caseydavenport:
Nope sorry, I missed that in the description! I was too hasty in my reading :)

rikatz:
No worries. Once this gets approved we can move forward here.

Also I still need to finish my KEP review :)

rikatz:
Good news, endPort PR got merged :)

will check how to bump the client-go to a newer version, or maybe we should wait for v1.21 release and bump it to v1.21 directly?

caseydavenport (approved):
This LGTM - @rikatz it looks like you need to rebase, though!

rikatz:
@caseydavenport I will, but I wanted to wait the stable v1.21 lib release :D 

Otherwise we will use the v1.21.0-beta.0 

Do you want to follow with the beta release? Otherwise I'm just waiting the final release to happen, and then I'll adjust here and poke you or Shaun on Slack :)

caseydavenport:
@rikatz I think I'd rather move forward merging with the beta, and then update to the GA release once it comes out. We can get a head start on testing, etc. with the beta version.

The only complication is that once this is merged, we'll also need to update downstream components (i.e., felix, cni, etc) to use the matching `k8s.io/X` library versions so there isn't version skew, and ensure that those build properly.

rikatz:
@caseydavenport rebased, and updated to rc-0 (Release candidate at least is better than beta :P )

Do you want me to open a PR updating the library also on felix, controller, etc? Can do this quickly here

rikatz:
Actually, I guess it's better to do this once the new libcalico is released, so update libcalico + k8s libs, right? :)

caseydavenport:
Once this is merged we'll need to update the pin in the other repositories ASAP. We don't wait until a release, and generally keep the pins as up to date as possible to make sure our testing is running the latest code.

rikatz:
argh, didn't know I needed to generate files :/

Re-generating here :)

If you want to ping me once this gets merged on Slack, I can take care of this today :D

PR #1359, opened 2021-01-12
set workload endpoint veth name with a legal size

fzu-huang:
network interfece name's max-size is 15, with self-specificed prefix, we should make sure our veth name size <= 15

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/backend/k8s/conversion/workload_endpoint_default.go
+++ b/lib/backend/k8s/conversion/workload_endpoint_default.go
@@ -33,6 +33,8 @@ import (
 	"github.com/projectcalico/libcalico-go/lib/numorstring"
 )
 
+const linuxInterfaceMaxSize = 15
+
 type defaultWorkloadEndpointConverter struct{}
 
 // VethNameForWorkload returns a deterministic veth name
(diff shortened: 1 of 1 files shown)


caseydavenport:
Hm, I think I would rather enforce that the interface prefix is less than or equal to 4 characters. I don't want to start trimming entropy from the hash when the interface prefix starts getting too large. 

What are you trying to use for the interface prefix?

fzu-huang:
> Hm, I think I would rather enforce that the interface prefix is less than or equal to 4 characters. I don't want to start trimming entropy from the hash when the interface prefix starts getting too large.
> 
> What are you trying to use for the interface prefix?

`FELIX_INTERFACEPREFIX` is a configurable parameter, which is clearly stated in the official documentation, so in theory I can use ‘veth-’ as the parameter value. But in fact this will cause iptables rule write failure, because the prefix 'veth-' plus the pod hash has exceeded 15 bytes

caseydavenport:
> Maybe we should limit FELIX_INTERFACEPREFIX value's size.

Yes, this is exactly what I am thinking. I think that's the correct fix here.

fzu-huang:
the source of this pr (163yun/libcalico-go) will be maintained with a specific workloadendpoint-naming-implementation.  

This pr now move to https://github.com/projectcalico/libcalico-go/pull/1376

PR #1361, opened 2021-01-27
Support multiple kubeconfig files via env var 'KUBECONFIG'

electricjesus:
Addresses projectcalico/calico#4312

## Description
Prior to this fix, calicoctl does not recognise multiple kubeconfig files specified via KUBECONFIG env var.

This changeset provides more or less copied behaviour of kubectl where it uses `filepath.Split()` and `ClientConfigLoadingRules.Precedence` if multiple kubeconfig targets do exist. 

Falls back to current `ExplicitPath` behaviour setting if config typically set as a single file specified only.

## Todos
- [x] Tests
- [x] Documentation
- [x] Release note

## Release Note

```release-note
`calicoctl` now supports multiple kubeconfig files via KUBECONFIG env var, just as `kubectl` would:

Example  usage:

    export KUBECONFIG=~/.kube/config:~/.kube/kubeconfig_2
    
    calicoctl get node
    calicoctl --context config2 get node
```

### DIFF
--- a/lib/backend/k8s/k8s.go
+++ b/lib/backend/k8s/k8s.go
@@ -17,6 +17,7 @@ package k8s
 import (
 	"context"
 	"fmt"
+	"path/filepath"
 	"reflect"
 	"sync"
 
@@ -210,6 +211,35 @@ func NewKubeClient(ca *apiconfig.CalicoAPIConfigSpec) (api.Client, error) {
 	return kubeClient, nil
 }
 
+// deduplicate removes any duplicated values and returns a new slice, keeping the order unchanged
+// 	based on deduplicate([]string) []string found in k8s.io/client-go/tools/clientcmd/loader.go#634
+// 	Copyright 2014 The Kubernetes Authors.
+func deduplicate(s []string) []string {
+	encountered := map[string]struct{}{}
+	ret := make([]string, 0)
+	for i := range s {
+		if _, ok := encountered[s[i]]; ok {
+			continue
+		}
+		encountered[s[i]] = struct{}{}
+		ret = append(ret, s[i])
+	}
+	return ret
+}
+
+// fill out loading rules based on filename(s) encountered in specified kubeconfig
+func fillLoadingRulesFromKubeConfigSpec(loadingRules *clientcmd.ClientConfigLoadingRules, kubeConfig string) {
+	fileList := filepath.SplitList(kubeConfig)
+
+	if len(fileList) > 1 {
+		loadingRules.Precedence = deduplicate(fileList)
+		loadingRules.WarnIfAllMissing = true
+		return
+	}
+
+	loadingRules.ExplicitPath = kubeConfig
+}
+
 func CreateKubernetesClientset(ca *apiconfig.CalicoAPIConfigSpec) (*rest.Config, *kubernetes.Clientset, error) {
 	// Use the kubernetes client code to load the kubeconfig file and combine it with the overrides.
 	configOverrides := &clientcmd.ConfigOverrides{}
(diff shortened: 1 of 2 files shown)


lib/backend/k8s/k8s.go
```diff
@@ -210,6 +211,33 @@ func NewKubeClient(ca *apiconfig.CalicoAPIConfigSpec) (api.Client, error) {
...
+			continue
+		}
+		encountered[s[i]] = true
+		ret = append(ret, s[i])
+	}
+	return ret
+}
+
+// fill out loading rules based on filename(s) encountered in specified kubeconfig
+func fillLoadingRulesFromKubeConfigSpec(loadingRules *clientcmd.ClientConfigLoadingRules, kubeConfig string) {
```
matthewdupre, line 229: Looks sensible.  I guess the main reason to copy this from client-go's NewDefaultClientConfigLoadingRules() is to maintain our slightly different environment variable setup?

electricjesus: Indeed. Almost all of our env vars are setup under calicoctl and passed down as `*apiconfig.CalicoAPIConfigSpec` that would have all the env vars and command line arguments already pre-combined

lib/backend/k8s/k8s.go
```diff
@@ -210,6 +211,33 @@ func NewKubeClient(ca *apiconfig.CalicoAPIConfigSpec) (api.Client, error) {
 	return kubeClient, nil
 }
 
+// deduplicate removes any duplicated values and returns a new slice, keeping the order unchanged
+func deduplicate(s []string) []string {
+	encountered := map[string]bool{}
+	ret := make([]string, 0)
+	for i := range s {
```
fasaxc, line 218: You can do
```
for _, v := range s {
   if encountered[v] {...
```
to directly iterate over the values instead of having to index each time.

lib/backend/k8s/k8s_client_test.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+var _ = Describe("CreateKubernetesClientset fillLoadingRulesFromKubeConfigSpec", func() {
+
+	When("There are multiple Kubeconfig files specified", func() {
```
fasaxc, line 26: No need to change this because you've only got one test in each block and it would just add a lot of boilerplate(!) but once you've got more than one test in a When/Describe/Context block, it should look like this:
```
When/Describe/Context("there are multiple foos", func() {
  var foos []foo
  var theThing thing
  BeforeEach(func() {
    foos = makeTheFoos()
    theThing = thing.New(thing.WithFoos(foos))
  })
  It("should do bar", func() {
    Expect(theThing).To(DoBar())
  })
  It("should do biff", func() {
    Expect(theThing).To(DoBiff())
  })
})
```
i.e. where you have `When("xyz", func()...)`, "xyz" should be done by the `BeforeEach` block nested inside the `When` so that hte set-up is shared between all the `It`s

lib/backend/k8s/k8s.go
electricjesus: I think `v` would be a copy now instead of s[i], which would be a reference.. yeah it's weird for such a small func :smile: 

I just pirated this function from https://github.com/kubernetes/client-go/blob/master/tools/clientcmd/loader.go#L635 but we can improve on it certainly

lib/backend/k8s/k8s_client_test.go
electricjesus: Right, forgot about BeforeEach! I'll just change it, seems to be structurally better even if it has a lot of boilerplate

lib/backend/k8s/k8s.go
fasaxc: `for _, v` is semantically a copy, sure, but in practice I'd expect zero difference here; string is a pointer-like type and the compiler is likely to spot that it doesn't need to actually copy.  Also not like this is hot-path critical code!

If you took that code from the k8s repo, please stick a copyright header on it as required by the Apache license.

electricjesus: Added the copyright, I just left the function mostly intact

fasaxc (approved):

marvin-tigera:
OK, I will merge the pull request when it's ready, leave the commits as is when I merge it, and leave the branch after I've merged it.

PR #1363, opened 2021-02-11
Fix libcalico returning a generic kubernetes error

tommasopozzetti:
## Description
In case of a missing WorkloadEndpoint object, libcalico returns a generic Kubernetes error rather than a specific Calico one, which can result in consumers of the library not being able to identify it as a "Not Found" error, and properly ignore it where they need to do so.
An example of such bug is shown in projectcalico/calico#4235 and this change has fixed the issue that is described there.

```release-note
Properly report not found when WorkloadEndpoint doesn't exist. Fixes https://github.com/projectcalico/calico/issues/4235
```


caseydavenport:
/sem-appove

Sorry for missing this!

caseydavenport (approved):

PR #1364, opened 2021-02-11
Speed up releasing of many IPs at once by doing each block in parallel

caseydavenport:
## Description

Do each block in its own goroutine. In the majority of cases the
goroutines will be acting on separate blocks / handles / etc and so can
be run in parallel without any conflicts.

Also, query IP pools once instead of once for each IP. 

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


caseydavenport:
The other thing I'd like to do in this area is pre-fetch handles rather than do a query for each, but I think that should be a separate PR.

lib/ipam/ipam.go
```diff
@@ -854,17 +876,39 @@ func (c ipamClient) ReleaseIPs(ctx context.Context, ips []net.IP) ([]net.IP, err
...
+		}(*cidr, ips)
+	}
+
+	// Read the response from each goroutine.
+	err = nil
+	for i := 0; i < len(ipsByBlock); i++ {
+		r := <-resultChan
+		log.Debugf("Received response #%d from release goroutine", i)
+		if r.Error != nil && err == nil {
+			err = r.Error
```
mgleung, line 907: Should we have a way to join all the possible errors together? Seems like we would want to report all of the IPs we attempted to release.

caseydavenport: Maybe. I figured that we do log out every error as we hit them, and this is just to report back that _some_ error occurred.

We might be able to return a []error? But I'm not sure what clients would do with that vs. just a single error.

caseydavenport:
> The other thing I'd like to do in this area is pre-fetch handles rather than do a query for each, but I think that should be a separate PR.

Well... I decided to add that here as well for fun. The second commit includes a change that lists handles rather than queries them each individually, for this particular ReleaseIPs() workflow.

lib/ipam/ipam.go
mgleung: My read of this was we could hit a bunch of errors but we would only be reporting the last one that we receive? Maybe we just get around this by just logging a warning and then return back the last error that we get as per usual?

caseydavenport: Yep, we log an error up on line 892 already

lib/ipam/ipam.go
```diff
@@ -922,10 +965,22 @@ func (c ipamClient) releaseIPsFromBlock(ctx context.Context, ips []net.IP, block
 			}
 		}
 
+		// List all handles, so we don't need to query them individually.
+		allHandles, err := c.blockReaderWriter.listHandles(ctx, "")
```
fasaxc, line 969: This feels like an expensive operation to do per-block?  Could you pass in the map and then refresh it after a failure in the loop here?

I suppose you could also have a heuristic: if there's only on IP to free, query only that handle.

lib/ipam/ipam.go
```diff
@@ -854,17 +875,39 @@ func (c ipamClient) ReleaseIPs(ctx context.Context, ips []net.IP) ([]net.IP, err
...
+	}
+	resultChan := make(chan retVal, len(ipsByBlock))
+	log.Infof("Starting %d routines to release IP addresses", len(ipsByBlock))
 	for cidrStr, ips := range ipsByBlock {
 		_, cidr, _ := net.ParseCIDR(cidrStr)
-		unalloc, err := c.releaseIPsFromBlock(ctx, ips, *cidr)
-		if err != nil {
-			log.Errorf("Error releasing IPs: %v", err)
-			return nil, err
+		go func(cidr net.IPNet, ips []net.IP) {
```
fasaxc, line 888: Should probably set a limit on parallelism here; generally better to have GO_MAX_PROC goroutines working relatively fast than 1000 goroutines all competing for API server resource, I reckon.

I've used this pattern in the past in felix:
```
	sem := semaphore.NewWeighted(int64(maxWorkers))
	for ... {
		if err := sem.Acquire(context.Background(), 1); err != nil {
			// Should only happen if the context finishes.
			log.WithError(err).Panic("Failed to acquire semaphore")
		}
		
		go func() {
			defer sem.Release(1)
			...
		}()
	})
```

fasaxc (changes requested):
Couple of thoughts from me

caseydavenport:
@fasaxc @mgleung I think this is ready for another look

lib/ipam/ipam.go
```diff
@@ -925,7 +995,7 @@ func (c ipamClient) releaseIPsFromBlock(ctx context.Context, ips []net.IP, block
 		// Success - decrement handles.
 		logCtx.Debugf("Decrementing handles: %v", handles)
 		for handleID, amount := range handles {
-			if err := c.decrementHandle(ctx, handleID, blockCIDR, amount); err != nil {
+			if err := c.decrementHandle(ctx, handleID, blockCIDR, amount, handleMap[handleID]); err != nil {
```
mgleung, line 998: Is it possible for `handleMap` to be `nil` here if there are 2 or less IPs being released?

mgleung (approved):
LGTM

fasaxc (approved):
LGTM

PR #1368, opened 2021-02-16
Add a default max blocks per host

caseydavenport:
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
By default, limit each node to 20 IP address blocks. This value can be overridden through IPAM configuration. 
```


mgleung (approved):
LGTM

PR #1370, opened 2021-02-26
Calico-4373: default missing protocol to TCP; don't let single port overwrite all ports

mattfenwick:
## Description

Bug fix: fixes https://github.com/projectcalico/calico/issues/4373.  There were a couple problems with port parsing that this addresses:

 - missing protocols were not treated as TCP, contradicting the network policy spec
 - allowing a specific port for a protocol could cause an 'allow all ports' for a protocol to be lost

Test plan:

```
libcalico-go/lib/backend/k8s/conversion$ go test .

libcalico-go$ sudo make test
```

## Todos
- [x] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
When interpreting Kubernetes NetworkPolicy ports, Calico now interprets an empty port struct as "all TCP" as per the NetworkPolicy spec.  Previously, empty structs were ignored. 
```

### DIFF
--- a/lib/backend/k8s/conversion/conversion.go
+++ b/lib/backend/k8s/conversion/conversion.go
@@ -391,18 +391,15 @@ func (c converter) k8sRuleToCalico(rPeers []networkingv1.NetworkPolicyPeer, rPor
 		if p.Port != nil {
 			portval := intstr.FromString(p.Port.String())
 			port.Port = &portval
-
-			// TCP is the implicit default (as per the definition of NetworkPolicyPort).
-			// Make the default explicit here because our data-model always requires
-			// the protocol to be specified if we're doing a port match.
-			port.Protocol = &protoTCP
 		}
 		if p.Protocol != nil {
 			protval := kapiv1.Protocol(fmt.Sprintf("%s", *p.Protocol))
 			port.Protocol = &protval
 		} else {
-			protval := kapiv1.ProtocolTCP
-			port.Protocol = &protval
+			// TCP is the implicit default (as per the definition of NetworkPolicyPort).
+			// Make the default explicit here because our data-model always requires
+			// the protocol to be specified if we're doing a port match.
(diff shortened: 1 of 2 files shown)


lib/backend/k8s/conversion/conversion.go
```diff
@@ -400,6 +400,9 @@ func (c converter) k8sRuleToCalico(rPeers []networkingv1.NetworkPolicyPeer, rPor
 		if p.Protocol != nil {
 			protval := kapiv1.Protocol(fmt.Sprintf("%s", *p.Protocol))
 			port.Protocol = &protval
+		} else {
```
fasaxc, line 403: I think this makes line 398 redundant so you can remove that line and its comment

lib/backend/k8s/conversion/conversion.go
```diff
@@ -430,7 +433,14 @@ func (c converter) k8sRuleToCalico(rPeers []networkingv1.NetworkPolicyPeer, rPor
 		}
 
 		pStr := protocol.String()
-		protocolPorts[pStr] = append(protocolPorts[pStr], calicoPorts...)
+		// treat nil as 'all ports'
+		if calicoPorts == nil {
```
fasaxc, line 437: Comment above on line 426 is now out of date; `calicoPorts` can be nil, now.

fasaxc (changes requested):
Looking good, just a couple of comment-level nits.

lib/backend/k8s/conversion/conversion.go
mattfenwick: 👍

lib/backend/k8s/conversion/conversion.go
mattfenwick: 👍

fasaxc (approved):

PR #1376, opened 2021-03-11
select a prefix to build interface name in legal size

fzu-huang:
## Description
network interfece name's max-size is 15, with self-specificed prefix, we should make sure our veth name size <= 15
ref: https://github.com/projectcalico/libcalico-go/pull/1359 (which will be close)

## Todos
- [x] Tests
- [x] Documentation
- [x] Release note

## Release Note

Will choose a suitable prefix from environment: `FELIX_INTERFACEPREFIX` which is not empty and has a length< 5.

### DIFF
--- a/lib/backend/k8s/conversion/workload_endpoint_default.go
+++ b/lib/backend/k8s/conversion/workload_endpoint_default.go
@@ -33,6 +33,14 @@ import (
 	"github.com/projectcalico/libcalico-go/lib/numorstring"
 )
 
+var (
+	defaultPrefix = "cali"
+
+	maxInterfaceNameSize      = 15
+	interfaceNameHashPartSize = 11
+	maxInterfacePrefixSize    = maxInterfaceNameSize - interfaceNameHashPartSize
+)
+
 type defaultWorkloadEndpointConverter struct{}
 
 // VethNameForWorkload returns a deterministic veth name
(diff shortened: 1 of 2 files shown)


fzu-huang:
@caseydavenport   This pr came from https://github.com/projectcalico/libcalico-go/pull/1359 . Can you review it, please ?

lib/backend/k8s/conversion/workload_endpoint_default.go
```diff
@@ -42,17 +49,21 @@ func (wc defaultWorkloadEndpointConverter) VethNameForWorkload(namespace, podnam
 	// veth name and mac addr.
 	h := sha1.New()
 	h.Write([]byte(fmt.Sprintf("%s.%s", namespace, podname)))
-	prefix := os.Getenv("FELIX_INTERFACEPREFIX")
+	var prefix string
+	splits := strings.Split(os.Getenv("FELIX_INTERFACEPREFIX"), ",")
```
caseydavenport, line 53: Why do we want to support a list of interface prefixes now? This seems like a controversial change. Could you explain a bit more about your use-case?

caseydavenport (changes requested):
This isn't quite what I was imagining - I don't think I understand your use-case. Could you provide a bit more information?

I was imagining a simple change to validate the prefix:

e.g.

```go
if len(prefix) > maxInterfacePrefixSize {
  log.Warn("Interface prefix too large - must be at most %d characters", maxInterfacePrefixSize)
}
```

lib/backend/k8s/conversion/workload_endpoint_default.go
fzu-huang: libcalico-go has already support a list of interface prefixes before this commit,  although I don't think its implementation is complete.

You can see that in [document](https://docs.projectcalico.org/archive/v3.18/reference/felix/configuration#bare-metal-specific-configuration)

Also, you can see the code arround here in master branch:
```
prefix := os.Getenv("FELIX_INTERFACEPREFIX")
	if prefix == "" {
		// Prefix is not set. Default to "cali"
		prefix = "cali"
	} else {
		// Prefix is set - use the first value in the list.
		splits := strings.Split(prefix, ",")
		prefix = splits[0]
	}
```

I think we should check each prefix's length.

caseydavenport: The Felix InterfacePrefix does accept more than a single option, but the handling of the env var in libcalico does not _really_. It only uses the first option. The intention of multiple values is so that Felix can run on systems with multiple types of workloads - e.g., OpenStack VMs and Kubernetes pods. The code that we are looking at here is scoped to Kubernetes-only, and so only supports a single prefix (the first in the list). If it suddenly started picking a different value, things would break badly.

I don't think it's appropriate to change this behavior without doing it "properly", and properly would mean a re-design of how the interface prefix is handled in libcalico. 

>  I don't think I understand your use-case. Could you provide a bit more information?

I'd still like an answer to this question - what exactly are you trying to do that's not working today?

fzu-huang: Thank you very much for your correction and patience! I think I already know the meaning and purpose of `FELIX_INTERFACEPREFIX`.
The original code is for the k8s scenario, so the first option of the environment variable is directly extracted, so I think，
I only need to check the length of this option, right?

caseydavenport: @fzu-huang I'm still confused about why you want this change - could you please tell me more about what you're trying to achieve? 

I don't think it's a good behavior to default to "cali" if the provided prefix is wrong.

fzu-huang: emm, you're right. Maybe I should log out and return an error here (if first option of FELIX_INTERFACEPREFIX has a size over than 4 )  .  The function ` VethNameForWorkload` will only be called by `podToDefaultWorkloadEndpoint`  and it will judge error .

caseydavenport: > could you please tell me more about what you're trying to achieve?

I can't really comment on what the correct implementation is without an understanding of what outcome you're looking for. What doesn't work today, that you would like to work after this PR is merged?

PR #1387, opened 2021-03-22
Fill internal or external IP for HostIPKey if BGP is disabled

sridhartigera:
HostIPKey is filled with bgp address. If BGP is disabled, the hostIP we pass to the BPF programs is nil and the host end point programs fail to attach. Hence as a fallback, fill in the internal ip or external IP.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/apis/v3/node.go
+++ b/lib/apis/v3/node.go
@@ -64,6 +64,7 @@ type NodeSpec struct {
 type NodeAddress struct {
 	// Address is a string representation of the actual address.
 	Address string `json:"address" validate:"net"`
+	Type string `json:"type" validate:"omitempty"`
 }
 
 type NodeStatus struct {
--- a/lib/apis/v3/node.go
+++ b/lib/apis/v3/node.go
@@ -23,6 +23,9 @@ import (
 const (
 	KindNode     = "Node"
 	KindNodeList = "NodeList"
+	CalicoNodeIP = "CalicoNodeIP"
+	InternalIP   = "InternalIP"
+	ExternalIP   = "ExternalIP"
 )
 
 // +genclient
@@ -64,7 +67,9 @@ type NodeSpec struct {
 type NodeAddress struct {
 	// Address is a string representation of the actual address.
 	Address string `json:"address" validate:"net"`
-	Type string `json:"type" validate:"omitempty"`
+
+	// Type is the node IP type
+	Type string `json:"type,omitempty" validate:"omitempty,ipType"`
 }
 
 type NodeStatus struct {
--- a/lib/backend/k8s/resources/node.go
+++ b/lib/backend/k8s/resources/node.go
@@ -306,17 +306,19 @@ func K8sNodeToCalico(k8sNode *kapiv1.Node, usePodCIDR bool) (*model.KVPair, erro
 func fillAllAddresses(calicoNode *apiv3.Node, k8sNode *kapiv1.Node) {
 	if bgp := calicoNode.Spec.BGP; bgp != nil {
 		if addr := bgp.IPv4Address; addr != "" {
-			calicoNode.Spec.Addresses = append(calicoNode.Spec.Addresses, apiv3.NodeAddress{Address: addr})
+			calicoNode.Spec.Addresses = append(calicoNode.Spec.Addresses, apiv3.NodeAddress{Address: addr, Type: "bgp"})
 		}
 		if addr := bgp.IPv6Address; addr != "" {
-			calicoNode.Spec.Addresses = append(calicoNode.Spec.Addresses, apiv3.NodeAddress{Address: addr})
+			calicoNode.Spec.Addresses = append(calicoNode.Spec.Addresses, apiv3.NodeAddress{Address: addr, Type: "bgp"})
 		}
 	}
 
 	for _, kaddr := range k8sNode.Status.Addresses {
 		switch kaddr.Type {
-		case kapiv1.NodeInternalIP, kapiv1.NodeExternalIP:
-			calicoNode.Spec.Addresses = append(calicoNode.Spec.Addresses, apiv3.NodeAddress{Address: kaddr.Address})
+		case kapiv1.NodeInternalIP:
+			calicoNode.Spec.Addresses = append(calicoNode.Spec.Addresses, apiv3.NodeAddress{Address: kaddr.Address, Type: "int"})
+		case kapiv1.NodeExternalIP:
+			calicoNode.Spec.Addresses = append(calicoNode.Spec.Addresses, apiv3.NodeAddress{Address: kaddr.Address, Type: "ext"})
 		default:
 			continue
 		}
--- a/lib/backend/k8s/resources/node.go
+++ b/lib/backend/k8s/resources/node.go
@@ -306,19 +306,19 @@ func K8sNodeToCalico(k8sNode *kapiv1.Node, usePodCIDR bool) (*model.KVPair, erro
 func fillAllAddresses(calicoNode *apiv3.Node, k8sNode *kapiv1.Node) {
 	if bgp := calicoNode.Spec.BGP; bgp != nil {
 		if addr := bgp.IPv4Address; addr != "" {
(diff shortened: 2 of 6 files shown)


lib/backend/k8s/resources/node.go
```diff
@@ -306,17 +306,19 @@ func K8sNodeToCalico(k8sNode *kapiv1.Node, usePodCIDR bool) (*model.KVPair, erro
 func fillAllAddresses(calicoNode *apiv3.Node, k8sNode *kapiv1.Node) {
 	if bgp := calicoNode.Spec.BGP; bgp != nil {
 		if addr := bgp.IPv4Address; addr != "" {
-			calicoNode.Spec.Addresses = append(calicoNode.Spec.Addresses, apiv3.NodeAddress{Address: addr})
+			calicoNode.Spec.Addresses = append(calicoNode.Spec.Addresses, apiv3.NodeAddress{Address: addr, Type: "bgp"})
```
caseydavenport, line 309: Types should follow Kubernetes API conventions for enums.

So basically casing.

```
- BGP
- ExternalIP
- InternalIP
```

IMO there's no reason to reinvent the types. Let's just use the same values k8s already does.

caseydavenport: The "BGP" value is probably not what we want to call it, given it's used for more than just BGP nowadays anyway, and isn't orthogonal with InternalIP (most cases they are probably the same thing). For example, this IP will be set in VXLAN mode even with BGP/BIRD disabled entirely. 

I would suggest one of the following:
- Treat this address as an `InternalIP` and have the felix code use the first InternalIP it finds in the list.
- Call this type something more generic - "AutodetectedIP" or something.

lib/apis/v3/node.go
```diff
@@ -64,6 +64,7 @@ type NodeSpec struct {
 type NodeAddress struct {
 	// Address is a string representation of the actual address.
 	Address string `json:"address" validate:"net"`
+	Type string `json:"type" validate:"omitempty"`
```
fasaxc, line 67: Type should be optional and validated

fasaxc: The types should follow k8s naming conventions

lib/backend/k8s/resources/node.go
fasaxc: `CalicoNodeIP` ?

lib/backend/syncersv1/updateprocessors/felixnodeprocessor.go
```diff
@@ -93,6 +93,37 @@ func (c *FelixNodeUpdateProcessor) Process(kvp *model.KVPair) ([]*model.KVPair,
 				}
 			}
 		}
+		// Look for internal node address, if BGP is not running
+		if ipv4 == nil {
+			var ip *cnet.IP
+			var cidr *cnet.IPNet
+			for _,addr := range node.Spec.Addresses {
```
fasaxc, line 100: This loop and the one below are super similar; best to make a utility function:
```
if ipv4 == nil {
  ipv4 = findNodeAddress(node, "CalicoNodeIP" /* should actually be a constant */)
}
if ipv4 == nil {
  ipv4 = findNodeAddress(node, "InternalIP" /* should actually be a constant */)
}
if ipv4 == nil {
  ipv4 = findNodeAddress(node, "ExternalIP" /* should actually be a constant */)
}
```

lib/backend/syncersv1/updateprocessors/felixnodeprocessor.go
```diff
@@ -93,6 +93,37 @@ func (c *FelixNodeUpdateProcessor) Process(kvp *model.KVPair) ([]*model.KVPair,
...
+				if addr.Type == "int" {
+					ip, cidr, err = cnet.ParseCIDROrIP(addr.Address)
```
fasaxc, line 102: Need to check that it is an IPv4 address

fasaxc (dismissed):

lib/backend/k8s/resources/node.go
caseydavenport: CalicoNodeIP is a bit ambiguous as well isn't it? Because calico/node might be using an IP that isn't a CalicoNodeIP after this PR is merged :grin: 

I'd rather not have APIs tied to a particular software component (it's contribued, but what if we decide in the future that some component other than calico/node should do IP auto-detection?)

fasaxc: I was thinking of `CalicoNodeIP` as the "Node IP that has been configured in the Calico datamodel" rather than the "IP determined by calico/node" but `AutodetectedIP` or `CalicoAutodetectedIP` would be fine by me.  @sridhartigera suggest you and @caseydavenport discuss this and come to an agreement.

lib/backend/syncersv1/updateprocessors/felixnodeprocessor.go
```diff
@@ -93,6 +93,19 @@ func (c *FelixNodeUpdateProcessor) Process(kvp *model.KVPair) ([]*model.KVPair,
 				}
 			}
 		}
+		// Look for internal node address, if BGP is not running
```
caseydavenport, line 96: @fasaxc @sridhartigera I want to check my understanding here.

IIRC, newer version of Felix and Typha ignore the HostIPKey and instead use the full v3 Node object.

e.g. 
- https://github.com/projectcalico/felix/blob/c46999eec127ecec7dec03f5f23201e1d6eff987/calc/vxlan_resolver.go#L82-L83
- https://github.com/projectcalico/felix/blob/bbd5c2b04230b1f23006ad44fab1ed6ef79e1270/calc/l3_route_resolver.go#L160-L166

So I think in practice we could do without these lines? The only reason these would be needed is if we're running a new version of Typha against a vey old version of Felix, since the new Typha would need to send data model updates in a way that the old felix understands.

caseydavenport: I think the main thing I want to make sure is that any new code we're writing on the other side of this in Felix is looking at the v3 object and not at the HostIPKey, since that would be a step in the wrong direction.

lib/backend/k8s/resources/node.go
caseydavenport: Works for me!

lib/backend/syncersv1/updateprocessors/felixnodeprocessor.go
sridhartigera: We still use the HostIPKey 
https://github.com/projectcalico/felix/blob/master/calc/dataplane_passthru.go#L43. and the BPF end point manager receives HostMetaDataUpdate message from eventSequencer.

caseydavenport: sad.  I guess we're stuck with this for now.

It would be nice to clean up that code to use the full v3 resource. At that point we could stop mucking around with HostIPKeys in this code so we don't need to send the same information in multiple formats.

caseydavenport: So I think that means the CalicoNodeIP type is essentially unused until we switch the Felix code over to use the v3 resource.

Once Felix is switched, felix can determine which IP to use based on the list of addresses rather than the HostIPKey.

lib/backend/k8s/resources/node.go
sridhartigera: Can we go with AutoDetectedIP?

lib/backend/syncersv1/updateprocessors/felixnodeprocessor.go
sridhartigera: Right, CalicoNodeIP is unused for now.

lib/backend/k8s/resources/node.go
caseydavenport: I'm fine with CalicoNodeIP for now

caseydavenport (approved):

PR #1393, opened 2021-03-30
Add IPv6 network support for failsafe rules

mgleung:
## Description

Adds IPv6 network support for the failsafe rules to uptake https://github.com/projectcalico/felix/pull/2742

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


lib/apis/v3/felixconfig.go
```diff
@@ -224,17 +224,19 @@ type FelixConfigurationSpec struct {
...
+	// [Default: tcp:22, udp:68, tcp:179, tcp:2379, tcp:2380, tcp:6443, tcp:6666, tcp:6667]
 	FailsafeInboundHostPorts *[]ProtoPort `json:"failsafeInboundHostPorts,omitempty"`
 	// FailsafeOutboundHostPorts is a comma-delimited list of UDP/TCP ports and CIDRs that Felix will allow outgoing traffic from host endpoints to
 	// irrespective of the security policy. This is useful to avoid accidentally cutting off a host with incorrect configuration. Each port
-	// should be specified as tcp:<port-number> or udp:<port-number>. For back-compatibility, if the protocol is not specified, it defaults
-	// to "tcp". If a CIDR is not specified, it will default to `0.0.0.0/0`. To disable all outbound host ports, use the value none.
-	// The default value opens etcd's standard ports to ensure that Felix does not get cut off from etcd as well as allowing DHCP and DNS.
-	// [Default: tcp:0.0.0.0/0:179, tcp:0.0.0.0/0:2379, tcp:0.0.0.0/0:2380, tcp:0.0.0.0/0:6443, tcp:0.0.0.0/0:6666, tcp:0.0.0.0/0:6667, udp:0.0.0.0/0:53, udp:0.0.0.0/0:67]
+	// should be specified as tcp:<port-number> or udp:<port-number>. For IPv6 networks, the CIDR must also be specified between square brackets
```
caseydavenport, line 235: For the FelixConfiguration API model, I don't think we should require the IPv6 address be within brackets, right?

Because the syntax is unambiguous and has its own field:

```
- protocol: TCP
  port: 443
  net: fe80::beef
```

caseydavenport: It looks like the code expects it to be without brackets here as well, so the user-facing comment shouldn't suggest that it needs them.

caseydavenport: Do we need validation that if brackets are specified we handle them correctly?

mgleung: The issue with colons is more about how the separation works in the env configuration case where we have to specify everything as parts separated by colons. You're right in that it is not required if we are to use YAML representation, but in the string representation we need to somehow separate the port and protocol from the address. I can clarify that the brackets are only required for the env configuration case, but I'm not sure how we would like to specify differences between YAML and env usage.

caseydavenport: This comment only applies to the yaml / datastore case, though.

caseydavenport (approved):

PR #1397, opened 2021-04-01
Fix handling of empty podIPs annotation

fasaxc:
## Description

* Fix that `IsFinished` wasn't checking whether the `podIPs` annotation was actually present.  Since the zero value for a string is `""`, we need to check for presence too or we'll be confused for non-Calico CNIs that don't set the annotation.
* Fix up various other inconsistencies:
  * A "finished" pod shouldn't be "ready"
  * `HasIPAddress` and `getPodIPs` didn't honour the annotation.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Fix that Felix would incorrectly treat any deleted pod as immediately finished.  This meant that pods networked with non-Calico CNIs would not have connectivity in the termination grace period.
```

### DIFF
--- a/lib/backend/k8s/conversion/conversion.go
+++ b/lib/backend/k8s/conversion/conversion.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2016-2020 Tigera, Inc. All rights reserved.
+// Copyright (c) 2016-2021 Tigera, Inc. All rights reserved.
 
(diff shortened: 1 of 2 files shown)


caseydavenport (approved):

wy1079385180:
why rollout restart sts can still see pod  ip in annotation？

PR #1398, opened 2021-04-05
Split calico / k8s NP

caseydavenport:
## Description

To-date, in KDD mode we have treated k8s NP and Calico NP as the same
resource on the frontend API. This is a holdover from etcd mode, where
k8s NPs are translated and written into etcd, and so they appear to be
the same resource.

In KDD mode, it would be better to handle them separately, so that
`calicoctl get networkpolicies` doesn't return k8s NP, etc., etc.

Part of https://github.com/projectcalico/calico/issues/2923

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
calicoctl will no longer display Kubernetes network polices when operating in KDD mode
```


lib/backend/model/resource.go
```diff
@@ -98,6 +98,11 @@ func init() {
 		"networkpolicies",
 		reflect.TypeOf(apiv3.NetworkPolicy{}),
 	)
+	registerResourceInfo(
+		KindKubernetesNetworkPolicy,
+		"networkpolicies",
```
mgleung, line 103: Won't this cause the plural resource mapping to be overwritten to only reference K8s network policies?

caseydavenport: Yes!

Although, I think that map might be dead code now? Still a really good catch. I'll fix it.

mgleung: Yeah, I guess it's really only used for etcd mode. Maybe it's wrong to even have  an entry for K8s network policy in that case since K8s network policies are treated through the same keywords in etcd mode?

mgleung: Actually we might need to use it for its string representation: https://github.com/projectcalico/libcalico-go/blob/79bef43e1a47f8b63f4944550872e12a3bda7e75/lib/backend/model/resource.go#L163 so it should be fine. etcd I guess will just know everything as `networkpolicies` and won't have anything written for `kubenetworkpolicies`

lib/backend/k8s/k8s_test.go
```diff
@@ -2639,35 +2639,40 @@ var _ = testutils.E2eDatastoreDescribe("Test Watch support", testutils.Datastore
...
-
-			// // There should be no more events
-			// Expect(watch.ResultChan()).ToNot(Receive())
-			//watch.Stop()
+			Expect(watch.ResultChan()).ToNot(Receive())
+			watch.Stop()
+
+			// Make a second change to one of the NPs
+			for _, kvp := range l.KVPairs {
+				name := strings.TrimPrefix(kvp.Value.(*apiv3.NetworkPolicy).Name, "knp.default.")
```
mgleung, line 2656: I'm guessing we shouldn't need this prefix anymore now that everything is separated right?

mgleung (approved):
LGTM

PR #1400, opened 2021-04-07
Move FindNodeAddresses API to node object

sridhartigera:
FindNodeAddresses is used to fetch the CalicoNodeIP/InternalIP/ExternalIP given the IP type. Moving this API to Node object so that this can be used by other modules. 

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/resources/node.go
+++ b/lib/resources/node.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2017 Tigera, Inc. All rights reserved.
+// Copyright (c) 2021 Tigera, Inc. All rights reserved.
 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/lib/resources/node.go
+++ b/lib/resources/node.go
@@ -1,0 +1,43 @@
+// Copyright (c) 2017 Tigera, Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package resources
+
+import (
+	log "github.com/sirupsen/logrus"
+
+        apiv3 "github.com/projectcalico/libcalico-go/lib/apis/v3"
+	cnet "github.com/projectcalico/libcalico-go/lib/net"
+
+)
+// FindNodeAddress returns node address of the specified type. Type can be one of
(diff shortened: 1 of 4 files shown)


caseydavenport:
@sridhartigera we don't normally attach helper functions to the v3 API objects, instead we use package-level functions that accept a v3 API object as an input.

Could you adjust this PR to match that approach? This will be important in the near future, as I am planning on moving the v3 structure definitions to a new repository (and I don't want to bring function logic with them).

caseydavenport:
> Moving this API to Node object so that this can be used by other modules.

@sridhartigera which other modules are you referring to? Who else needs this logic?

sridhartigera:
> > Moving this API to Node object so that this can be used by other modules.
> 
> @sridhartigera which other modules are you referring to? Who else needs this logic?

I am planning to use this logic in felix, specifically in calc/l3_route_resolver.

sridhartigera:
> @sridhartigera we don't normally attach helper functions to the v3 API objects, instead we use package-level functions that accept a v3 API object as an input.
> 
> Could you adjust this PR to match that approach? This will be important in the near future, as I am planning on moving the v3 structure definitions to a new repository (and I don't want to bring function logic with them).

you mean something like this.
func FindNodeAddress(node *v3.Node, ipType String)?

sridhartigera:
> > @sridhartigera we don't normally attach helper functions to the v3 API objects, instead we use package-level functions that accept a v3 API object as an input.
> > Could you adjust this PR to match that approach? This will be important in the near future, as I am planning on moving the v3 structure definitions to a new repository (and I don't want to bring function logic with them).
> 
> you mean something like this.
> func FindNodeAddress(node *v3.Node, ipType String)?
I was thinking of replicating the similar logic in felix. Is that preferred? If we are ok to have this in libcalico-go, can I have the same FindNodeAddresses func in node.go ?

lib/resources/node.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	log "github.com/sirupsen/logrus"
+
+	apiv3 "github.com/projectcalico/libcalico-go/lib/apis/v3"
+	cnet "github.com/projectcalico/libcalico-go/lib/net"
+)
+
+// FindNodeAddress returns node address of the specified type. Type can be one of
+// CalicoNodeIP, InternalIP or ExternalIP
+func FindNodeAddress(node *apiv3.Node, ipType string) (*cnet.IP, *cnet.IPNet) {
```
caseydavenport, line 26: Rather than pass a string here, we should probably use the apiv3 type instead.

caseydavenport: Oh, right. I see that they are just defined as strings: https://github.com/projectcalico/libcalico-go/blob/master/lib/apis/v3/node.go#L26-L28

Normally we define a type for enums like this so that we can pass around typed values, like we do here: https://github.com/projectcalico/libcalico-go/blob/master/lib/apis/v3/policy.go#L21-L27

I suppose in this case the Kubernetes API defines the equivalent field as a string, so maybe this is OK.

caseydavenport (approved):

PR #1422, opened 2021-05-11
Use a spec field on WEPs for ServiceAccount name

caseydavenport:
## Description

A third option for fixing https://github.com/projectcalico/calico/issues/4529

Fixes https://github.com/projectcalico/calico/issues/4529

Idea is:
- Stop setting serviceaccount as a label on the WEP 
- Instead include it as a spec field. Convert to a label on Syncer API
to Felix so selection still works the same.

In kdd mode, this alone will be enough. The CNI plugin will stop sending
the label through, and since Felix reads directly from the Pod API will
keep receiving the same info over the Syncer.

For etcd mode, we need to make an additional change so that
kube-controllers properly updates old WEPs in the data store to include
the new spec field.

Alternatively, for etcd mode we could have kube-controllers add the
projectcalico.org/serviceaccount label itself. Any existing clusters
will not have any labels > 63 characters, and any new workloads that do
have such a long serviceaccount will be handled by the new CNI plugin
populating the spec field. 

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Fix issue with serviceaccount names larger than 63 characters.
```

### DIFF
--- a/lib/backend/k8s/conversion/conversion_test.go
+++ b/lib/backend/k8s/conversion/conversion_test.go
@@ -788,6 +788,80 @@ var _ = Describe("Test Pod conversion", func() {
 			apiv3.LabelServiceAccount:        "sa-test",
 		}
 		Expect(wep.Value.(*apiv3.WorkloadEndpoint).ObjectMeta.Labels).To(Equal(expectedLabels))
+		Expect(wep.Value.(*apiv3.WorkloadEndpoint).Spec.ServiceAccountName).To(Equal("sa-test"))
+
+		// Assert ResourceVersion is present.
+		Expect(wep.Revision).To(Equal("1234"))
+	})
+
+	It("should parse a Pod with long serviceaccount", func() {
+		longName := "serviceaccount-name-that-is-too-long-to-be-used-as-a-kubernetes-label-because-it-exceeds-the-character-limit"
+		pod := kapiv1.Pod{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "podA",
+				Namespace: "default",
+				Annotations: map[string]string{
+					"arbitrary": "annotation",
+				},
+				Labels: map[string]string{
+					"labelA": "valueA",
+					"labelB": "valueB",
+				},
+				ResourceVersion: "1234",
+			},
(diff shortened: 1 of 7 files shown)


lib/backend/k8s/conversion/conversion_test.go
```diff
@@ -788,6 +788,7 @@ var _ = Describe("Test Pod conversion", func() {
 			apiv3.LabelServiceAccount:        "sa-test",
 		}
 		Expect(wep.Value.(*apiv3.WorkloadEndpoint).ObjectMeta.Labels).To(Equal(expectedLabels))
+		Expect(wep.Value.(*apiv3.WorkloadEndpoint).Spec.ServiceAccountName).To(Equal("sa-test"))
```
robbrockbank, line 791: maybe worth a quick copy and paste test with long SA name?

lib/apis/v1/openapi_generated.go
```diff
@@ -83,21 +83,18 @@ func schema_libcalico_go_lib_apis_v1_BGPPeer(ref common.ReferenceCallback) commo
 				Properties: map[string]spec.Schema{
 					"TypeMetadata": {
 						SchemaProps: spec.SchemaProps{
-							Default: map[string]interface{}{},
-							Ref:     ref("github.com/projectcalico/libcalico-go/lib/apis/v1/unversioned.TypeMetadata"),
+							Ref: ref("github.com/projectcalico/libcalico-go/lib/apis/v1/unversioned.TypeMetadata"),
```
robbrockbank, line 86: What's going on here then?

caseydavenport: Was because of an outdated generation tool cached locally - should be fixed now :)

robbrockbank (approved):
LGTM

cmukai-cisco:
I deployed the patch release with this in it and I am getting the following message:

`Message:      OCI runtime create failed: container_linux.go:367: starting container process caused: exec: "/install-cni.sh": stat /install-cni.sh: no such file or directory: unknown`

I was using v3.15.1 prior to this so this may be due to the large version change.

PR #1436, opened 2021-06-24
Improve IPAM logging info returned by AutoAssign()

coutinhop:
## Description
Improve logging output from AutoAssign() for more clarity, especially when not all requested IPs are successfully assigned, stating the reasons for it.

Add `type IPAMAssignmentInfo struct` to ipam.go and return v4 and v6 versions of it from autoAssign() and propagate it to AutoAssign().

Change signature AutoAssign() in interface.go, and change every call of both functions to accommodate the new return values.

IPAMAssignmentInfo has as String() method which returns a formatted log message of the success/failure/reasons of the assignment. I would appreciate feedback on the message/output, thanks.

Add IPAMAssignmentInfo verification to AutoAssign fv tests in ipam_test.go and ipam_win_test.go and add a IPAMAssignmentInfo.String() ut test case to ipam_test.go.

Also: fix a typo in a log msg in findOrClaimBlock() and make AutoAssign tests in ipam_test.go check the error obtained against the expected (instead of just expecting any error to occur).

## Todos
- [x] Tests
- [ ] Documentation
- [x] Release note

## Release Note
```release-note
Added enhanced error logging for IPAM failures
```

### DIFF
--- a/lib/ipam/interface.go
+++ b/lib/ipam/interface.go
@@ -36,7 +36,7 @@ type Interface interface {
 	// which is useful for dataplanes that need to know the subnet (such as Windows).
 	//
 	// In case of error, returns the IPs allocated so far along with the error.
-	AutoAssign(ctx context.Context, args AutoAssignArgs) ([]cnet.IPNet, []cnet.IPNet, error)
+	AutoAssign(ctx context.Context, args AutoAssignArgs) (*IPAMAssignments, *IPAMAssignments, error)
 
 	// ReleaseIPs releases any of the given IP addresses that are currently assigned,
 	// so that they are available to be used in another assignment.
--- a/lib/ipam/ipam.go
+++ b/lib/ipam/ipam.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"math/bits"
 	"runtime"
+	"strings"
 
 	"golang.org/x/sync/semaphore"
 
@@ -86,7 +87,7 @@ type ipamClient struct {
 // and the list of the assigned IPv6 addresses.
 //
 // In case of error, returns the IPs allocated so far along with the error.
-func (c ipamClient) AutoAssign(ctx context.Context, args AutoAssignArgs) ([]net.IPNet, []net.IPNet, error) {
+func (c ipamClient) AutoAssign(ctx context.Context, args AutoAssignArgs) (*IPAMAssignments, *IPAMAssignments, error) {
 	// Determine the hostname to use - prefer the provided hostname if
 	// non-nil, otherwise use the hostname reported by os.
 	hostname, err := decideHostname(args.Hostname)
@@ -95,7 +96,7 @@ func (c ipamClient) AutoAssign(ctx context.Context, args AutoAssignArgs) ([]net.
 	}
 	log.Infof("Auto-assign %d ipv4, %d ipv6 addrs for host '%s'", args.Num4, args.Num6, hostname)
 
-	var v4list, v6list []net.IPNet
+	var v4ia, v6ia *IPAMAssignments
 
 	if args.Num4 != 0 {
 		// Assign IPv4 addresses.
@@ -105,10 +106,10 @@ func (c ipamClient) AutoAssign(ctx context.Context, args AutoAssignArgs) ([]net.
 				return nil, nil, fmt.Errorf("provided IPv4 IPPools list contains one or more IPv6 IPPools")
 			}
 		}
-		v4list, err = c.autoAssign(ctx, args.Num4, args.HandleID, args.Attrs, args.IPv4Pools, 4, hostname, args.MaxBlocksPerHost, args.HostReservedAttrIPv4s)
+		v4ia, err = c.autoAssign(ctx, args.Num4, args.HandleID, args.Attrs, args.IPv4Pools, 4, hostname, args.MaxBlocksPerHost, args.HostReservedAttrIPv4s)
 		if err != nil {
 			log.Errorf("Error assigning IPV4 addresses: %v", err)
-			return v4list, nil, err
+			return v4ia, nil, err
 		}
 	}
 
@@ -120,14 +121,14 @@ func (c ipamClient) AutoAssign(ctx context.Context, args AutoAssignArgs) ([]net.
 				return nil, nil, fmt.Errorf("provided IPv6 IPPools list contains one or more IPv4 IPPools")
 			}
 		}
-		v6list, err = c.autoAssign(ctx, args.Num6, args.HandleID, args.Attrs, args.IPv6Pools, 6, hostname, args.MaxBlocksPerHost, args.HostReservedAttrIPv6s)
+		v6ia, err = c.autoAssign(ctx, args.Num6, args.HandleID, args.Attrs, args.IPv6Pools, 6, hostname, args.MaxBlocksPerHost, args.HostReservedAttrIPv6s)
 		if err != nil {
 			log.Errorf("Error assigning IPV6 addresses: %v", err)
-			return v4list, v6list, err
+			return v4ia, v6ia, err
 		}
 	}
 
-	return v4list, v6list, nil
+	return v4ia, v6ia, nil
 }
 
 // getBlockFromAffinity returns the block referenced by the given affinity, attempting to create it if
@@ -483,7 +484,7 @@ func (s *blockAssignState) findOrClaimBlock(ctx context.Context, minFreeIps int)
 	if config.AutoAllocateBlocks {
 		for i := 0; i < datastoreRetries; i++ {
 			// Claim a new block.
-			logCtx.Infof("No more affine blocks, but need to claim more block -- allocate another block")
+			logCtx.Infof("No more affine blocks, but need to claim more blocks -- allocate another block")
 
 			// First, try to find an unclaimed block.
 			logCtx.Info("Looking for an unclaimed block")
@@ -548,7 +549,44 @@ func (s *blockAssignState) findOrClaimBlock(ctx context.Context, minFreeIps int)
 	return nil, false, errors.New("failed to find or claim a block")
 }
 
-func (c ipamClient) autoAssign(ctx context.Context, num int, handleID *string, attrs map[string]string, requestedPools []net.IPNet, version int, host string, maxNumBlocks int, rsvdAttr *HostReservedAttr) ([]net.IPNet, error) {
+type IPAMAssignments struct {
+	IPs              []net.IPNet       // assigned IP addresses
+	IPVersion        int               // IP version (4 or 6)
+	NumRequested     int               // number of requested IP addresses (not all may be assigned)
+	HostReservedAttr *HostReservedAttr // reserved addresses at start and/or end of blocks
+	Msgs             []string          // warning/error messages to be rendered in case there are any issues with the assignment
+}
+
+func (i *IPAMAssignments) AddMsg(msg string) {
+	for _, m := range i.Msgs {
+		if msg == m {
+			// Don't add duplicate msgs
+			return
+		}
+	}
+	i.Msgs = append(i.Msgs, msg)
+}
+
+func (i *IPAMAssignments) PartialFulfillmentError() error {
+	if len(i.IPs) < i.NumRequested {
+		var b strings.Builder
+
+		fmt.Fprintf(&b, "Assigned %d out of %d requested IPv%d addresses", len(i.IPs), i.NumRequested, i.IPVersion)
+
+		for _, m := range i.Msgs {
+			fmt.Fprintf(&b, "; %v", m)
+		}
+
+		if i.HostReservedAttr != nil {
+			fmt.Fprintf(&b, "; HostReservedAttr: %v", i.HostReservedAttr.Handle)
+		}
+
+		return errors.New(b.String())
+	}
+	return nil
+}
+
+func (c ipamClient) autoAssign(ctx context.Context, num int, handleID *string, attrs map[string]string, requestedPools []net.IPNet, version int, host string, maxNumBlocks int, rsvdAttr *HostReservedAttr) (*IPAMAssignments, error) {
 
 	// First, get the existing host-affine blocks.
 	logCtx := log.WithFields(log.Fields{"host": host})
@@ -562,7 +600,6 @@ func (c ipamClient) autoAssign(ctx context.Context, num int, handleID *string, a
 	}
 
 	logCtx.Debugf("Found %d affine IPv%d blocks for host: %v", len(affBlocks), version, affBlocks)
-	ips := []net.IPNet{}
 
 	// Record how many blocks we own so we can check against the limit later.
 	numBlocksOwned := len(affBlocks)
@@ -592,6 +629,12 @@ func (c ipamClient) autoAssign(ctx context.Context, num int, handleID *string, a
 	}
 	logCtx.Debugf("Host must not use more than %d blocks", maxNumBlocks)
 
+	ia := &IPAMAssignments{
+		IPVersion:        version,
+		NumRequested:     num,
+		HostReservedAttr: rsvdAttr,
+	}
+
 	s := &blockAssignState{
 		client:                c,
 		version:               version,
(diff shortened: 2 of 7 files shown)


lib/ipam/interface.go
```diff
@@ -36,7 +36,7 @@ type Interface interface {
 	// which is useful for dataplanes that need to know the subnet (such as Windows).
 	//
 	// In case of error, returns the IPs allocated so far along with the error.
-	AutoAssign(ctx context.Context, args AutoAssignArgs) ([]cnet.IPNet, []cnet.IPNet, error)
+	AutoAssign(ctx context.Context, args AutoAssignArgs) ([]cnet.IPNet, []cnet.IPNet, *IPAMAssignmentInfo, *IPAMAssignmentInfo, error)
```
mgleung, line 39: I think it might be a good idea to include the assigned IP addresses as well into this. So your return values would be just `IPAMAssignmentInfo` (or whatever this will be called after it's merged with the return info) and `error`

coutinhop: Hm, I see, it makes sense. Thanks, Matt! I guess we could call it `IPAMAssignment` and have it contain the []cnet.IPNet slice inside it. Is that name good?

lib/ipam/ipam.go
```diff
@@ -548,7 +549,44 @@ func (s *blockAssignState) findOrClaimBlock(ctx context.Context, minFreeIps int)
...
+func (i *IPAMAssignmentInfo) String() string {
+	if i == nil {
+		return ""
+	}
+
+	ret := fmt.Sprintf("Assigned %v out of %v requested IPv%v addresses", i.NumAssigned, i.NumRequested, i.IPVersion)
+
+	switch {
+	case i.NoFreeAffineBlocks && i.StrictAffinity:
+		ret += "; No more free affine blocks and strict affinity enabled"
```
mgleung, line 573: The nice golang way to do this is with `strings.Builder` https://golang.org/pkg/strings/#Builder

coutinhop: Oh, TIL! Will use it, thanks!

lib/ipam/ipam.go
```diff
@@ -548,7 +549,44 @@ func (s *blockAssignState) findOrClaimBlock(ctx context.Context, minFreeIps int)
...
+	NumBlocksOwned     int
+	MaxNumBlocks       int
+	ExhaustedPools     []string
+	StrictAffinity     bool
+	NoFreeAffineBlocks bool
+	HostReservedAttr   *HostReservedAttr
+}
+
+func (i *IPAMAssignmentInfo) String() string {
+	if i == nil {
```
caseydavenport, line 565: You don't actually need this check, even though it's a pointer receiver.

Here's an example: https://play.golang.org/p/YkUMSPgpFZN

coutinhop: TIL as well, thanks!

lib/ipam/ipam.go
```diff
@@ -592,6 +630,15 @@ func (c ipamClient) autoAssign(ctx context.Context, num int, handleID *string, a
 	}
 	logCtx.Debugf("Host must not use more than %d blocks", maxNumBlocks)
 
+	iPAMAssignmentInfo := &IPAMAssignmentInfo{
```
caseydavenport, line 633: nit: should name this `ipamAssignmentInfo` (keep consistent case for acronyms)

lib/ipam/interface.go
caseydavenport, line 39: We should update the interface comment to explain the new return values

caseydavenport: Another nice simplification might be to remove the `[]cnet.IPNet` return values altogether and embed them in the returned struct instead. This reduces the number of return values, as well as the number of places we're tracking the same information (e.g., num assigned IPs).

Maybe could call the struct `IPAMResult`

lib/ipam/ipam.go
```diff
@@ -548,7 +549,44 @@ func (s *blockAssignState) findOrClaimBlock(ctx context.Context, minFreeIps int)
 	return nil, false, errors.New("failed to find or claim a block")
 }
 
-func (c ipamClient) autoAssign(ctx context.Context, num int, handleID *string, attrs map[string]string, requestedPools []net.IPNet, version int, host string, maxNumBlocks int, rsvdAttr *HostReservedAttr) ([]net.IPNet, error) {
+type IPAMAssignmentInfo struct {
+	NumRequested       int
```
caseydavenport, line 553: I think it would be nice to see (brief) comments on each of these fields indicating how they are used / are intended to be populated.

lib/ipam/ipam.go
```diff
@@ -712,6 +762,7 @@ func (c ipamClient) autoAssign(ctx context.Context, num int, handleID *string, a
 						break
 					}
 					logCtx.Infof("Successfully assigned IPs from non-affine block %s", blockCIDR.String())
+					iPAMAssignmentInfo.NumAssigned += len(newIPs)
```
caseydavenport, line 765: Rather than try to catch all the places we might allocate IPs, could we set NumAssigned at the end of the function?

Or going a step further, see my suggestion about just embedding the allocated IPs in the struct rather than having two separate return values. WDYT?

lib/ipam/ipam.go
```diff
@@ -548,7 +549,44 @@ func (s *blockAssignState) findOrClaimBlock(ctx context.Context, minFreeIps int)
...
+	switch {
+	case i.NoFreeAffineBlocks && i.StrictAffinity:
+		ret += "; No more free affine blocks and strict affinity enabled"
+	case i.NumBlocksOwned >= i.MaxNumBlocks:
+		ret += "; IPAM block limit reached"
+	case len(i.ExhaustedPools) > 0:
+		ret += "; No IPs available in pools"
+	}
+	if i.NumAssigned != i.NumRequested {
+		ret += fmt.Sprintf("; no free affine blocks: %v, strict affinity: %v, assigned blocks: %v, block limit: %v, exhausted IP pools: %v", i.NoFreeAffineBlocks, i.StrictAffinity, i.NumBlocksOwned, i.MaxNumBlocks, i.ExhaustedPools)
```
caseydavenport, line 580: I wonder if this is too much information to display on every failed request. For example, if we hit the block limit for a node, do we need to say anything about exhausted IP pools? Or can we limit the message to just the issue that was actually encountered?

caseydavenport (changes requested):
Looking good! I added a bunch of thoughts / suggestions. Take a look at let me know if you agree / disagree / have questions at all. Happy to discuss.

I also plan to take a second look later today / tomorrow as well to check for other scenarios we might want to cover.

As a note for myself, here are some possible failure modes to check in my next go through:

- [ ] No IP pools select the node.
- [ ] Need a new block, but there are none available.
- [ ] Need a new block, but auto-assign blocks is disabled.
- [ ] Need to borrow an IP, but strict affinity is enabled.
- [ ] No IP addresses remaining in any valid IP pools.
- [ ] Exact IP requested via AssignIP, but ... something?

lib/ipam/ipam.go
```diff
@@ -686,6 +735,7 @@ func (c ipamClient) autoAssign(ctx context.Context, num int, handleID *string, a
 				// Grab a new random block.
 				blockCIDR := newBlockCIDR()
 				if blockCIDR == nil {
+					iPAMAssignmentInfo.ExhaustedPools = append(iPAMAssignmentInfo.ExhaustedPools, p.Spec.CIDR)
```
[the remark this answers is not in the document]
caseydavenport: Maybe nothing useful can be done about this, and it's a very rare and obscure case, so no big deal. Best thing we could do would perhaps be to include a disclaimer - "IP pool is exhaused (but btw we hit an error just before this)"

lib/ipam/interface.go
coutinhop: Will do!

lib/ipam/ipam.go
```diff
@@ -548,7 +549,44 @@ func (s *blockAssignState) findOrClaimBlock(ctx context.Context, minFreeIps int)
...
+		return ""
+	}
+
+	ret := fmt.Sprintf("Assigned %v out of %v requested IPv%v addresses", i.NumAssigned, i.NumRequested, i.IPVersion)
+
+	switch {
+	case i.NoFreeAffineBlocks && i.StrictAffinity:
+		ret += "; No more free affine blocks and strict affinity enabled"
+	case i.NumBlocksOwned >= i.MaxNumBlocks:
+		ret += "; IPAM block limit reached"
```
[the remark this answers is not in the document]
coutinhop: I agree that's better, thanks!

lib/ipam/ipam.go
coutinhop: Yes, that was @mgleung's suggestion as well (embedding the `[]cnet.IPNet` slice inside the struct), and again I agree it's better! Will do that.

lib/ipam/ipam.go
coutinhop: I had initially thought this way, but after some discussion with @mgleung I ended up including everything. I can see advantages and disadvantages to both, I don't know... Maybe we could have a `DumpValues()` method which does this, and defer the choice of calling either `String()` or `DumpValues()` (not married to this name, though), or both?

lib/ipam/ipam.go
```diff
@@ -548,7 +549,44 @@ func (s *blockAssignState) findOrClaimBlock(ctx context.Context, minFreeIps int)
...
+	IPVersion          int
+	NumBlocksOwned     int
+	MaxNumBlocks       int
+	ExhaustedPools     []string
+	StrictAffinity     bool
+	NoFreeAffineBlocks bool
+	HostReservedAttr   *HostReservedAttr
+}
+
+func (i *IPAMAssignmentInfo) String() string {
```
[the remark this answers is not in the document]
coutinhop: Yes, will do this (using `strings.Builder` https://golang.org/pkg/strings/#Builder as @mgleung suggested)

lib/ipam/ipam.go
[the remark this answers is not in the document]
coutinhop: To be completely honest, I was led to this by the debug message on line 739 ("All addresses exhausted in pool %s"). But I'm not sure I follow, does `randomBlockGenerator()` keep track of all the blocks it returned (so that if we skip it due to data store query failure, it will not ever be generated again)?

lib/ipam/ipam.go
```diff
@@ -548,7 +549,44 @@ func (s *blockAssignState) findOrClaimBlock(ctx context.Context, minFreeIps int)
...
+type IPAMAssignments struct {
+	IPs              []net.IPNet       // assigned IP addresses
+	IPVersion        int               // IP version (4 or 6)
+	NumRequested     int               // number of requested IP addresses (not all may be assigned)
+	HostReservedAttr *HostReservedAttr // reserved addresses at start and/or end of blocks
+	Msgs             []string          // warning/error messages to be rendered in case there are any issues with the assignment
+}
+
+func (i *IPAMAssignments) AddMsg(msg string) {
+	for _, m := range i.Msgs {
```
mgleung, line 561: nit: might be more efficient to put this in a map so we don't need to do a loop through all of them to check. It might not matter too much either way since I don't expect this list to really get super long usually.

mgleung: I'm also wondering if we're only using `Msgs` to just keep a list of additional info if we should even store it as separate strings or if we should just keep it as a string builder and then just `WriteString` the new messages to it. As far as I can tell we only want to keep them separate so we can do the comparison to see if we've already seen the error message before right?

lib/ipam/ipam.go
```diff
@@ -548,7 +549,44 @@ func (s *blockAssignState) findOrClaimBlock(ctx context.Context, minFreeIps int)
...
+}
+
+func (i *IPAMAssignments) PartialFulfillmentError() error {
+	if len(i.IPs) < i.NumRequested {
+		var b strings.Builder
+
+		fmt.Fprintf(&b, "Assigned %d out of %d requested IPv%d addresses", len(i.IPs), i.NumRequested, i.IPVersion)
+
+		if len(i.Msgs) > 0 {
+			fmt.Fprintf(&b, "; %v", strings.Join(i.Msgs, ";"))
```
mgleung, line 577: I'm also wondering if we're only using `Msgs` to just keep a list of additional info if we should even store it as separate strings or if we should just keep it as a string builder and then just `WriteString` the new messages to it. As far as I can tell we only want to keep them separate so we can do the comparison to see if we've already seen the error message before right?

lib/ipam/ipam.go
coutinhop: @caseydavenport 's reasoning for keeping the string list is to have the "breadcrumb" information of what error/warning was hit until we get into the current state. So a plain map would lose that ordering, I had considered a map with some kind of helper structure to keep the ordering, but I didn't think we would have too great a number of msgs that would make the `for` loop too bad on performance... Yes, a string builder should work, but I'm not sure about filtering duplicates with it, would there be an efficient way of doing this?

lib/ipam/ipam.go
mgleung: do we need to do the join with `"; "` instead of `";"`?

lib/ipam/ipam.go
```diff
@@ -603,26 +646,31 @@ func (c ipamClient) autoAssign(ctx context.Context, num int, handleID *string, a
...
+		rem := num - len(ia.IPs)
 		if maxNumBlocks > 0 && numBlocksOwned >= maxNumBlocks {
 			s.allowNewClaim = false
 		}
 
 		b, newlyClaimed, err := s.findOrClaimBlock(ctx, 1)
 		if err != nil {
 			if _, ok := err.(noFreeBlocksError); ok {
+				if config.StrictAffinity {
+					ia.AddMsg("No more free affine blocks and strict affinity enabled")
```
mgleung, line 661: Should we also add a message for the case where it's not enforcing `StrictAffinity` which should mean we couldn't find any unclaimed blocks?

coutinhop: We would break out of this loop in either case, but we would check non-affine blocks further below (if strict affinity is not enabled). Do you think we should change this to "No more free affine blocks, strict affinity: %v" and add this message independent of strict affinity?

lib/ipam/ipam.go
```diff
@@ -672,20 +721,22 @@ func (c ipamClient) autoAssign(ctx context.Context, num int, handleID *string, a
 	// If we need to support non-strict affinity and no auto-allocation of
 	// blocks, then we should query the actual allocation blocks and assign
 	// from those.
-	rem := num - len(ips)
+	rem := num - len(ia.IPs)
 	if config.StrictAffinity != true && rem != 0 {
```
mgleung, line 725: We could check the opposite of this condition and add a log stating that strict affinity is enabled but we were unable to find addresses in any of the affine blocks

lib/ipam/ipam.go
```diff
@@ -649,11 +697,12 @@ func (c ipamClient) autoAssign(ctx context.Context, num int, handleID *string, a
 					continue
 				}
 				logCtx.WithError(err).Warningf("Failed to assign IPs in newly allocated block")
+				ia.AddMsg("Failed to assign IPs in newly allocated block")
```
mgleung, line 700: Should we say which block we failed to assign IPs to?

lib/ipam/ipam.go
coutinhop: Good catch, thanks! Will fix

lib/ipam/ipam.go
coutinhop: Basically move the "no free affine blocks" addMsg() call to here? I like that, think this is more clear. Will do, thanks!

lib/ipam/ipam.go
coutinhop: We talked offline and ended up deciding blocks are internal information to calico and there is not much gain in specifying them. We do specify which IP pools in the logging, though.

lib/ipam/ipam.go
mgleung: oh i think I misread `findOrClaimBlock` yeah I don't think my original comment matters here.

mgleung (approved):
LGTM

coutinhop (commented):
Synced with @caseydavenport and the comments were addressed. This code seems to have been updated already.

PR #1437, opened 2021-06-24
Add go routine stack trace dumping tooling

Brian-McM:
This commit adds tooling that can be used to dump go routine stack trace information on running go programs. This tooling is useful because it's sometimes difficult to reproduce issues that occur in production, but by enabling this in a production cluster you can get a stack trace of what the program is doing, which will aid in diagnosing possible issues.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/debugtools/stacktrace/go_routine_stacktrace_server.go
+++ b/lib/debugtools/stacktrace/go_routine_stacktrace_server.go
@@ -1,0 +1,73 @@
+// Copyright (c) 2021 Tigera, Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package grstacktrace contains utilities for getting go routine stack traces from running go programs. A go program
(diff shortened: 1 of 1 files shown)


tmjd:
I think instead of using an HTTP server for this we should use a signal handler. Then I think we should be able to enable dumping stack traces much more freely.

Brian-McM:
> I think instead of using an HTTP server for this we should use a signal handler. Then I think we should be able to enable dumping stack traces much more freely.

Enabling it would be the same either way, you need to set an env variable regardless (I don't think we want this enabled all the time).

What signal are you hoping to use? One of the kill signals? I'm not opposed to doing that, but I'm not sure that makes dumping this more _freely_. With the server you don't need to kill the pod, so you can let the issue continue to run and diagnose it. You may need to dump a stack trace multiple times because when you dump it it may not be running the problem. We may also want to run other tools after to get more info, like mem / cpu profiling.

Like I said, I'm not opposed to using a signal but I disagree that it allows us to dump stack traces more freely.

tmjd:
My thought is that if it is a signal handler then we can enable it all the time. If it is an HTTP endpoint then I think the default would need to be disabled and we would need a special flag to enable it.
I would like to have this capability all the time, though granted if it is a signal I think the only way to fire it would to be able to actually get on a node and use docker or podman utilities but that is also the reason I'd be comfortable having it enabled all the time.

As for which Signals, probably SIGUSR1 or SIGUSR2. IIRC Felix already sets up one of those to dump information.

Brian-McM:
OK that makes sense and sounds good. I had thought that the only signals you could send were kill signals, I don't really know a lot about this but I'll read up on it.

fasaxc:
Felix currently uses both SIGUSR1 and SIGUSR2.  One to dump a CPU profile and one to dump a memory profile.

We could just overload those so that we dump a CPU profile and a stack trace, for example.  If we do add a debug server, how about using a unix domain socket instead of a TCP socket; that'd avoid security issues with opening extra ports, which some customers are very worried about.

caseydavenport:

PR #1443, opened 2021-06-30
Add EndpointSlice and new policy fields to syncer

caseydavenport:
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Support for Services in NetworkPolicy egress rules
```

### DIFF
--- a/lib/validator/v3/validator.go
+++ b/lib/validator/v3/validator.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2016-2020 Tigera, Inc. All rights reserved.
+// Copyright (c) 2016-2021 Tigera, Inc. All rights reserved.
 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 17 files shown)


lib/validator/v3/validator.go
```diff
@@ -1203,6 +1226,22 @@ func validateNetworkPolicy(structLevel validator.StructLevel) {
 		}
 	}
 
+	// Services are only allowed on egress rules.
```
fasaxc, line 1229: Think we should also prevent services in the egress Destination struct

lib/validator/v3/validator_test.go
```diff
@@ -1947,6 +1947,64 @@ func init() {
...
+									Name:      "service1",
+									Namespace: "default",
+								},
+							},
+						},
+					},
+				},
+			}, false,
+		),
+
```
fasaxc, line 2007: Worth shooting for full coverage in the validation logic; think there are a few more cases to cover

fasaxc (changes requested):
Generally looking good to me

caseydavenport:
@fasaxc I think this is ready for another look.

Also @mgleung this is the next part of the PR you reviewed today

lib/validator/v3/validator.go
```diff
@@ -1333,6 +1387,33 @@ func validateGlobalNetworkPolicy(structLevel validator.StructLevel) {
...
```
mgleung, line 1390: Do we not also need to check that services are not set as the source on the egress rules for the GNP?

mgleung (approved):
LGTM

PR #1458, opened 2021-07-17
Simplify processing of v3 resources

robbrockbank:
## Description

Use a couple of regexes in the v3 resource -> internal key processing rather than handling for each specific resource type.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/backend/model/keys.go
+++ b/lib/backend/model/keys.go
@@ -23,12 +23,10 @@ import (
 	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


bcreane (approved):
This is very elegant. I don't see any issues with the new `kindLower` search w/global and NS'ed bug catchery. Feel free to wait for an expert if you want a more critical 👁️ .

fasaxc (approved):
LGTM

PR #1471, opened 2021-07-26
Add HostPort support

AloysAugustin:
## Description
This PR adds parsing of the HostPort / HostIP field for the EndpointPorts, in order to enable HostPort support.
A side effect of this is that not all ports in a WorkloadEndpoint's Ports list have a name. This didn't seem to create problems in our testing, but we may not have covered all cases.

This PR doesn't include tests for now, we'll add them if you agree on the approach 🙂 . One possible discussion point is on the filtering of the Ports list, we chose to parse all the Pod's ports, but we could parse only those that are named or that have a hostPort.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
WorkloadEndpoints now support hostPorts
```


caseydavenport (approved):
This looks good to me - the thing that I'm not 100% sure about is if any downstream projects expect that `port.Name` is always set on ports. That seems to be the one major assumption that has changed here, since previously ports were only included if they were named.

We'll need to be careful of that when looking at consumers of the ports field.

caseydavenport:
> we chose to parse all the Pod's ports, but we could parse only those that are named or that have a hostPort.

It's probably beneficial to filter out ports that we know won't be used, since it can potentially save some memory.

AloysAugustin:
Oops! I missed a lot of required updates, this should be fixed now. I also updated the parsing to omit the ports that have neither a name nor a host port.

caseydavenport (approved):

PR #1480, opened 2021-07-28
Update syncer test framework to have better kvp comparison and defaulted filtering of updates

robbrockbank:
Few changes:
- Remove some of the less than ideal sync tester methods and replace with something I hope is more robust
- Add a new fv-fast makefile target for testing fvs when you write them
- Add endpoint slices to felix FVs
- When dumping the syncer cache include details of whether an entry has been explicitly checked - this should make it a lot easier to spot unexpected cache entries and make debugging a lot faster.
- Add a utils for namespace deletion which waits for completion. Update tests to use that

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/backend/syncersv1/felixsyncer/felixsyncer_e2e_test.go
+++ b/lib/backend/syncersv1/felixsyncer/felixsyncer_e2e_test.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2020 Tigera, Inc. All rights reserved.
+// Copyright (c) 2017-2021 Tigera, Inc. All rights reserved.
 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


lib/backend/syncersv1/felixsyncer/felixsyncer_e2e_test.go
```diff
@@ -565,10 +614,6 @@ var _ = testutils.E2eDatastoreDescribe("Felix syncer tests", testutils.Datastore
 			syncTester.ExpectStatusUpdate(api.ResyncInProgress)
 			syncTester.ExpectCacheSize(len(current))
 			for _, e := range current {
-				if config.Spec.DatastoreType == apiconfig.Kubernetes {
-					// Don't check revisions for K8s since the node data gets updated constantly.
-					e.Revision = ""
-				}
```
robbrockbank, line 571: I've built this into the default sanitizer in the syncTester.

song-jiang (approved):
LGTM. I'm more of learning than actually reviewing the code.

PR #1484, opened 2021-08-05
List / Watch optimizations

caseydavenport:
## Description

- [x] Start from resourceVersion=0 to serve watches and Lists from the API server's in-memory cache instead of going all the way
to etcd.
- [ ] Use watch bookmarks
- [x] Set `resourceVersionMatch=NotOlderThan` to make sure we get recent data.
- [x] Use paginated List() calls for all resources.

May result in minor consistency changes, but hopefully with the benefit
of substantially reduced API server load. 

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Optimize lists and watches made against the Kubernetes API
```

### DIFF
--- a/lib/backend/k8s/k8s.go
+++ b/lib/backend/k8s/k8s.go
@@ -509,6 +509,7 @@ func buildCRDClientV1(cfg rest.Config) (*rest.RESTClient, error) {
 		if err != nil {
 			log.WithError(err).Fatal("failed to add calico resources to scheme")
 		}
+		metav1.AddToGroupVersion(scheme.Scheme, schema.GroupVersion{Group: "crd.projectcalico.org", Version: "v1"})
 	})
 	return cli, nil
 }
--- a/lib/backend/k8s/resources/customresource.go
+++ b/lib/backend/k8s/resources/customresource.go
@@ -21,7 +21,6 @@ import (
 
 	log "github.com/sirupsen/logrus"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -30,7 +29,6 @@ import (
 	scheme "k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
-	"k8s.io/client-go/tools/pager"
 
 	"github.com/projectcalico/libcalico-go/lib/backend/api"
 	"github.com/projectcalico/libcalico-go/lib/backend/model"
@@ -237,15 +235,16 @@ func (c *customK8sResourceClient) List(ctx context.Context, list model.ListInter
 	logContext := log.WithFields(log.Fields{
 		"ListInterface": list,
 		"Resource":      c.resource,
+		"Type":          "CustomResource",
 	})
-	logContext.Debug("List Custom K8s Resource - paged if necessary")
-	kvps := []*model.KVPair{}
+	logContext.Debug("Received List request")
 
 	// Attempt to convert the ListInterface to a Key.  If possible, the parameters
 	// indicate a fully qualified resource, and we'll need to use Get instead of
 	// List.
 	if key := c.listInterfaceToKey(list); key != nil {
 		logContext.Debug("Performing List using Get")
+		kvps := []*model.KVPair{}
 		if kvp, err := c.Get(ctx, key, revision); err != nil {
 			// The error will already be a Calico error type.  Ignore
 			// error that it doesn't exist - we'll return an empty
@@ -289,42 +288,14 @@ func (c *customK8sResourceClient) List(ctx context.Context, list model.ListInter
 		}
 		return out, nil
 	}
-
-	lp := pager.New(listFunc)
-	opts := metav1.ListOptions{ResourceVersion: revision}
-	if revision != "" {
-		opts.ResourceVersionMatch = metav1.ResourceVersionMatchNotOlderThan
-	}
-	result, _, err := lp.List(ctx, opts)
-	if err != nil {
-		return nil, K8sErrorToCalico(err, list)
-	}
-
-	// For each item in the response, convert it to a KVPair and add it to the list.
-	forEach := func(obj runtime.Object) error {
-		res := obj.(Resource)
-		kvp, err := c.convertResourceToKVPair(res)
+	convertFunc := func(r Resource) ([]*model.KVPair, error) {
+		kvp, err := c.convertResourceToKVPair(r)
 		if err != nil {
-			logContext.WithError(err).WithField("Item", res).Warning("unable to process resource, skipping")
-			return nil
+			return nil, err
 		}
-		kvps = append(kvps, kvp)
-		return nil
+		return []*model.KVPair{kvp}, nil
 	}
-	err = meta.EachListItem(result, forEach)
-	if err != nil {
-		return nil, K8sErrorToCalico(err, list)
-	}
-
-	// Extract list revision information.
-	m, err := meta.ListAccessor(result)
-	if err != nil {
-		return nil, err
-	}
-	return &model.KVPairList{
-		KVPairs:  kvps,
-		Revision: m.GetResourceVersion(),
-	}, nil
+	return pagedList(ctx, logContext, revision, list, convertFunc, listFunc)
 }
 
 func (c *customK8sResourceClient) Watch(ctx context.Context, list model.ListInterface, revision string) (api.WatchInterface, error) {
--- a/lib/backend/k8s/resources/customresource.go
+++ b/lib/backend/k8s/resources/customresource.go
@@ -232,7 +232,7 @@ func (c *customK8sResourceClient) Get(ctx context.Context, key model.Key, revisi
 }
 
 // List lists configured Custom K8s Resource instances in the k8s API matching the
-// supplied ListInterface. It will use list pagining if necessary to reduce the load on the Kubernetes API server.
+// supplied ListInterface. It will use list paging if necessary to reduce the load on the Kubernetes API server.
 func (c *customK8sResourceClient) List(ctx context.Context, list model.ListInterface, revision string) (*model.KVPairList, error) {
 	logContext := log.WithFields(log.Fields{
 		"ListInterface": list,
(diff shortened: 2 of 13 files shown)


lib/backend/watchersyncer/watchercache.go
```diff
@@ -180,7 +181,7 @@ func (wc *watcherCache) resyncAndCreateWatcher(ctx context.Context) {
 			}
 
 			// Start the sync by Listing the current resources.
-			l, err := wc.client.List(ctx, wc.resourceType.ListInterface, "")
+			l, err := wc.client.List(ctx, wc.resourceType.ListInterface, "0")
```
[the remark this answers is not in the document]
caseydavenport: Added `resourceVersionMatch=NotOlderThan` whenever revision is set. 

Bookmarks will be the next step, but haven't added them yet - they require a bit more plumbing and test modifications than I originally hoped.

lib/backend/k8s/resources/node.go
```diff
@@ -153,24 +157,48 @@ func (c *nodeClient) List(ctx context.Context, list model.ListInterface, revisio
...
 	}
 
-	// Listing all nodes.
-	nodes, err := c.clientSet.CoreV1().Nodes().List(ctx, metav1.ListOptions{ResourceVersion: revision})
-	if err != nil {
-		return nil, K8sErrorToCalico(err, list)
+	listFunc := func(ctx context.Context, opts metav1.ListOptions) (runtime.Object, error) {
+		nodes, err := c.clientSet.CoreV1().Nodes().List(ctx, opts)
+		if err != nil {
+			return nil, fmt.Errorf("failed to list nodes: %s", err)
```
caseydavenport, line 163: Need to return the actual error type here, not wrapped.

lib/backend/k8s/resources/workloadendpoint.go
```diff
@@ -291,29 +295,50 @@ func (c *WorkloadEndpointClient) listUsingName(ctx context.Context, listOptions
...
-			continue
+		if !c.converter.IsValidCalicoWorkloadEndpoint(pod) {
+			return nil
 		}
 
-		kvps, err := c.converter.PodToWorkloadEndpoints(&pod)
+		kvps, err := c.converter.PodToWorkloadEndpoints(pod)
 		if err != nil {
-			return nil, err
+			return K8sErrorToCalico(err, listOptions)
```
caseydavenport, line 310: We convert below - can return the unwrapped error here.

lib/backend/k8s/resources/customresource.go
```diff
@@ -263,62 +267,77 @@ func (c *customK8sResourceClient) List(ctx context.Context, list model.ListInter
...
-	elem := reflect.ValueOf(reslOut).Elem()
-	items := reflect.ValueOf(elem.FieldByName("Items").Interface())
-	for idx := 0; idx < items.Len(); idx++ {
-		res := items.Index(idx).Addr().Interface().(Resource)
-		if kvp, err := c.convertResourceToKVPair(res); err == nil {
-			kvps = append(kvps, kvp)
-		} else {
+	lp := pager.New(listFunc)
+	opts := metav1.ListOptions{ResourceVersion: revision}
+	if revision != "" {
```
caseydavenport, line 295: I think this is the right way to do this. There are three scenarios:

- revision = "" -> resourceVersionMatch not supported.
- revision = "0" -> resourceVersionMatch=NotOlderThan -> Any (from API server cache)
- revision = <non-zero> -> resourceVersionMatch=NotOlderThan -> Not older than the given value.

lib/backend/k8s/resources/customresource.go
```diff
@@ -263,62 +267,77 @@ func (c *customK8sResourceClient) List(ctx context.Context, list model.ListInter
...
 		KVPairs:  kvps,
-		Revision: reslOut.GetListMeta().GetResourceVersion(),
+		Revision: m.GetResourceVersion(),
 	}, nil
 }
 
 func (c *customK8sResourceClient) Watch(ctx context.Context, list model.ListInterface, revision string) (api.WatchInterface, error) {
 	// Build watch options to pass to k8s.
-	opts := metav1.ListOptions{ResourceVersion: revision, Watch: true}
+	opts := metav1.ListOptions{ResourceVersion: revision, Watch: true, AllowWatchBookmarks: false}
```
caseydavenport, line 332: We can allow watch bookmarks in a future increment - I'd like to get this one in first since bookmarks are slightly more substantial.

lib/backend/k8s/resources/customresource.go
```diff
@@ -263,62 +267,77 @@ func (c *customK8sResourceClient) List(ctx context.Context, list model.ListInter
...
+	}
+	err = meta.EachListItem(result, forEach)
+	if err != nil {
+		return nil, K8sErrorToCalico(err, list)
+	}
+
+	// Extract list revision information.
+	m, err := meta.ListAccessor(result)
+	if err != nil {
+		return nil, err
```
lmm, line 322: Should this be a Calico error too?

lib/backend/k8s/resources/kubeendpointslice.go
```diff
@@ -91,29 +94,44 @@ func (c *endpointSliceClient) Get(ctx context.Context, key model.Key, revision s
...
 	}
 
-	// Add in the Revision information.
-	kvps.Revision = endpointSlices.ResourceVersion
+	// Extract the list revision information.
+	m, err := meta.ListAccessor(result)
+	if err != nil {
+		return nil, err
+	}
+	kvps.Revision = m.GetResourceVersion()
```
lmm, line 134: This section of code starting from where a new pager is created down to here is almost the same in a few files - do you think it's worth extracting this into a helper function?

lib/backend/k8s/resources/kubenetworkpolicy.go
```diff
@@ -91,39 +94,52 @@ func (c *networkPolicyClient) Get(ctx context.Context, key model.Key, revision s
...
 	log.Debug("Received List request on Kubernetes NetworkPolicy type")
-	// List all of the k8s NetworkPolicy objects.
-	networkPolicies := networkingv1.NetworkPolicyList{}
-	req := c.clientSet.NetworkingV1().RESTClient().
-		Get().
-		Resource("networkpolicies")
-	err := req.Do(ctx).Into(&networkPolicies)
-	if err != nil {
-		log.WithError(err).Info("Unable to list K8s Network Policy resources")
-		return nil, K8sErrorToCalico(err, list)
```
neiljerram, line 102: Does it make any difference going from this code with `RESTClient` to the `NetworkPolicies()` interface?  If not, was it just oversight that this old code was here in this form for so long?

lib/backend/k8s/resources/profile.go
```diff
@@ -188,42 +191,81 @@ func (c *profileClient) List(ctx context.Context, list model.ListInterface, revi
...
+	err = meta.EachListItem(result, forEach)
+	if err != nil {
+		return nil, K8sErrorToCalico(err, list)
+	}
+
+	// Extract the list revision information.
+	m, err = meta.ListAccessor(result)
+	if err != nil {
+		return nil, err
+	}
```
neiljerram, line 263: Yes, it definitely feels now like we need a common utility for this passage of code.

lib/backend/k8s/k8s.go
```diff
@@ -509,6 +509,7 @@ func buildCRDClientV1(cfg rest.Config) (*rest.RESTClient, error) {
 		if err != nil {
 			log.WithError(err).Fatal("failed to add calico resources to scheme")
 		}
+		metav1.AddToGroupVersion(scheme.Scheme, schema.GroupVersion{Group: "crd.projectcalico.org", Version: "v1"})
```
neiljerram, line 512: What is this added line for?  It doesn't seem obviously connected with the rest of the change.

lib/backend/k8s/resources/kubeendpointslice.go
```diff
@@ -91,29 +94,44 @@ func (c *endpointSliceClient) Get(ctx context.Context, key model.Key, revision s
...
+	listFunc := func(ctx context.Context, opts metav1.ListOptions) (runtime.Object, error) {
+		return c.clientSet.DiscoveryV1beta1().EndpointSlices("").List(ctx, opts)
 	}
 
+	// For each object returned from the API server, add it to a KVPairList.
 	kvps := model.KVPairList{KVPairs: []*model.KVPair{}}
-	for _, es := range endpointSlices.Items {
-		kvp, err := c.EndpointSliceToKVP(&es)
+	kvps.Revision = revision
+	forEach := func(obj runtime.Object) error {
```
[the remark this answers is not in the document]
neiljerram: (And so on in other similar cases)

neiljerram (commented):
Coding looks OK, modulo some queries and nits.  Not sure I'm confident enough on the overall concept though; would be good for @robbrockbank or @fasaxc to check that when they are back.

lib/backend/k8s/resources/customresource.go
caseydavenport: Hm, I don't think it needs to be since it's not an error returned from the server, but probably safe to do anyway.

lib/backend/k8s/resources/customresource.go
```diff
@@ -263,62 +267,77 @@ func (c *customK8sResourceClient) List(ctx context.Context, list model.ListInter
...
+	if revision != "" {
+		opts.ResourceVersionMatch = metav1.ResourceVersionMatchNotOlderThan
+	}
+	result, _, err := lp.List(ctx, opts)
+	if err != nil {
+		return nil, K8sErrorToCalico(err, list)
+	}
+
+	// For each item in the response, convert it to a KVPair and add it to the list.
+	forEach := func(obj runtime.Object) error {
```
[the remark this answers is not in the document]
caseydavenport: Yeah, I suppose I could. I found it clearer to define it separately than to try to inline into the function call which felt like too much nesting to me, but happy to change.

lib/backend/k8s/resources/kubenetworkpolicy.go
caseydavenport: Yeah - the code used to use a RESTClient because we wrote this well before client-go had native support for the NetworkPolicy API (e.g., when networkpolicies were in alpha stage).

I think we have just never updated it since.

lib/backend/k8s/k8s.go
caseydavenport: I  was getting errors from the newly introduced list pager about improper registration that led me down some github issues which suggested this as the fix.

Errors looked like this:

> v1.ListOptions is not suitable for converting to "crd.projectcalico.org/v1" in scheme "pkg/runtime/scheme.go:100"

lib/backend/k8s/resources/customresource.go
caseydavenport: Actually, I feel that it should NOT be converted to a datastore error, since the error is not due to interacting with the datastore. Rather, it's just returning the given object as a List interface type.

lib/backend/k8s/resources/profile.go
caseydavenport: I can do a common utility - but I think it conflicts with the idea of inlining the `forEach` and `listFunc` functions - those will vary for each client and will need to be passed in to the common utility.

caseydavenport:
@lmm @neiljerram thanks for the review - I've made a common function for the list paging and it looks a lot nicer now, thanks!

I think this is ready for a final review.

lmm (approved):
LGTM - I left a minor comment. Not sure why CI is failing - semaphore isn't loading that job for me for some reason.

PR #1496, opened 2021-09-01
Use Info log level on allocation performance measurement tests

coutinhop:
Use Info log level (as opposed to Debug) on allocation performance measurement tests in ipam_test.go

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/ipam/ipam_test.go
+++ b/lib/ipam/ipam_test.go
@@ -147,6 +147,7 @@ var _ = testutils.E2eDatastoreDescribe("IPAM tests", testutils.DatastoreAll, fun
 		var hostname string
 		var err error
 		var pool20, pool32, pool26 []cnet.IPNet
+		var origLogLevel log.Level
 
 		BeforeEach(func() {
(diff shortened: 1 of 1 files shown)


lib/ipam/ipam_test.go
```diff
@@ -186,6 +186,9 @@ var _ = testutils.E2eDatastoreDescribe("IPAM tests", testutils.DatastoreAll, fun
 		})
 
 		Measure("It should be able to allocate a single address quickly - blocksize 32", func(b Benchmarker) {
+			origLevel := log.GetLevel()
```
caseydavenport, line 189: Rather than do this within each Measure() block, you can set the log level in the `BeforeEach` call above, and then revert it to the original value in `AfterEach()`

coutinhop: Right, I missed that, will do! Thanks, @caseydavenport !

mgleung (approved):
LGTM

PR #1508, opened 2021-09-27
[allowed-use] Introduce support for allowed uses of IPAM pools.

fasaxc:
## Description

When a client auto-assigns an IP it must set the intended use field on its request.  Only pools that allow that use will be considered.

Upstream some refactoring done in the Enterprise codebase.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


caseydavenport (approved):

marvin-tigera:
OK, I will merge the pull request when it's ready, leave the commits as is when I merge it, and leave the branch after I've merged it.

PR #1509, opened 2021-09-27
[ip-reservations] Implement IP reservations

fasaxc:
## Description

Add support for reserving particular IPs so that they won't be used by Calico IPAM.

Depends on #1508 

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


lib/ipam/ipam_test.go
```diff
@@ -446,6 +456,64 @@ var _ = testutils.E2eDatastoreDescribe("IPAM tests", testutils.DatastoreAll, fun
 		})
 	})
 
+	Describe("Reservation tests", func() {
```
song-jiang, line 459: We need to add a test case where an IP is already been reserved in a block? For example, windows nodes reserves 4 IPs for a block?

song-jiang (commented):
Looking good with several suggestions.

PR #1510, opened 2021-09-28
Add release Makefile targets

tmjd:
## Description

Adding Makefile targets for creating git tags using the same targets we use in other repos.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -262,3 +262,45 @@ ci: clean static-checks test
 ###############################################################################
 update-pins: update-api-pin
 
+###############################################################################
+# Release
+###############################################################################
+
+## Tags and builds a release from start to finish.
+release: release-prereqs
+	$(MAKE) VERSION=$(VERSION) release-tag
+
+	@echo ""
+	@echo "Release tagged. Next, push the tag."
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -262,3 +262,44 @@ ci: clean static-checks test
...
+	@echo ""
+
+## Produces a git tag for the release.
+release-tag: release-prereqs release-notes
+	git tag $(VERSION) -F release-notes-$(VERSION)
+
+## Generates release notes.
+release-notes: release-prereqs
+	echo "This release is for build purposes only and this should not be referenced externally." > release-notes-$(VERSION)
+	echo "libcalico is not a stable interface." >> release-notes-$(VERSION)
```
caseydavenport, line 286: Probably worth including a link to `projectcalico/api` since that is the actual public API repo.

caseydavenport (approved):
one comment but don't need to look again

caseydavenport (approved):
one comment but don't need to look again

PR #1511, opened 2021-10-01
Add simple first and interval logger

robbrockbank:
## Description
Add a simple first and interval logger to log utils.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/logutils/ratelimitedlogger.go
+++ b/lib/logutils/ratelimitedlogger.go
@@ -1,0 +1,306 @@
+// Copyright (c) 2021 Tigera, Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package logutils
+
+import (
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	fieldLogSkipped = "logs-skipped"
+	fieldLogNextLog = "next-log"
+	defaultInterval = 5 * time.Minute
+)
+
+// NewRateLimitedLogger returns a RateLimitedLogger which can be used for rate limited logging.
+//
+// Methods are essentially the same as the logrus logging methods, but there is no Panic or Fatal log since these don't
+// make much sense for rate limited logging.
+//
+// Log requests are only processed if allowed by the logging level. The first processed log will always be written.
+// Subsequent processed logs will not be written within the configured time interval. Once the time interval has passed
+// the next log will be written. The logs include additional fields specifying the number of skipped logs and the
+// minimum time for the next expected log.  The Force() method can be used to ensure the log is written - this resets
+// the time for the next log.
+//
+// Typical use might be as follows:
+//
+//   logger := NewRateLimitedLogger().WithField("key": "my-key")
+//   for {
+//     logger.Infof("Checking some stuff: %s", myStuff)
+//     complete = doSomeStuff()
+//     if complete {
+//       break
+//     }
+//   }
+//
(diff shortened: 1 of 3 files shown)


lib/logutils/firstandinterval.go
```diff
@@ -0,0 +1,296 @@
...
+	if logger.level() >= logrus.ErrorLevel {
+		if entry := logger.logEntry(); entry != nil {
+			entry.Errorln(args...)
+		}
+	}
+}
+
+// level returns the log level associated with the logger.  (copied from logrus since this is not a public method)
+func (logger *FirstAndIntervalLogger) level() logrus.Level {
+	return logrus.Level(atomic.LoadUint32((*uint32)(&logger.entry.Logger.Level)))
```
fasaxc, line 295: I normally check against `logrus.GetLevel()` is this doing something different?

lib/logutils/firstandinterval.go
```diff
@@ -0,0 +1,296 @@
...
+	entry    *logrus.Entry
+
+	// The number skipped since the last processed log.
+	skipped  int
+
+	// Whether to force the next log to be processed.
+	force    bool
+
+	// Lock used to access to this data. This lock is never held while writing a log.
+	lock     sync.RWMutex
```
fasaxc, line 81: `sync.Mutex` is a bit faster if you're not using the `RLock()` function

lib/logutils/firstandinterval.go
```diff
@@ -0,0 +1,296 @@
...
+// The first log is always processed.  Subsequent logs are only processed if at least the interval time has passed since
+// the last log request. Processed logs contain information about the number of skipped logs and the expected earliest
+// time of the next log. Force() may be used to force the next log to be processed.  Note that sufficient log level is
+// still required for a processed log to actually be written.
+//
+// Typical use might be as follows:
+//
+//   logger := NewFirstAndIntervalLogger(5 * time.Minute).WithField("key": "my-key")
+//   for {
+//     logger.Infof("Checking some stuff: %s", myStuff)
```
fasaxc, line 44: I think the user will need to be careful not to do `WithField()` inside a loop.  If they do (and IIUC) then that makes a child logger with its own timestamps so the parent logger's next log time won't get updated.

fasaxc (commented):
Definitely a good idea, couple of thoughts inline

lib/logutils/firstandinterval.go
robbrockbank: Oh goddamit ... I don't know what I was thinking!  Yup - indeed this does not work very well unless there is a shared timestamp.  So perhaps this should be an atomic timestamp instead.

lib/logutils/firstandinterval.go
robbrockbank: Because I allowed the Logger to be passed in on the constructor I think we could therefore not use the logrus.GetLevel() since that always pulled from the standard logger.

robbrockbank: That said, I only added that to the constructor for testing ... so other suggestions welcome.

lib/logutils/firstandinterval.go
robbrockbank: Changed.

lib/logutils/ratelimitedlogger.go
```diff
@@ -0,0 +1,308 @@
...
+//       break
+//     }
+//   }
+//
+//   // Use force to ensure our final log is printed and it contains the summary info about the number of skipped logs.
+//   Logger.Force().Info("Finished checking stuff")
+//
+// The config is an optional parameter. If not specified, default values are used (see RateLimitedLoggerConfig for
+// details about the default values).
+func NewRateLimitedLogger(config *RateLimitedLoggerConfig) *RateLimitedLogger {
```
fasaxc, line 67: Suggest using the options pattern to avoid needing to pass nil:
```
type RateLimitedLoggerOpt func(*RateLimitedLogger)

func OptInterval(d time.Duration) RateLimitedLoggerOpt {
  return func(r *RateLimitedLogger) {
    r.data.interval = d
  }
}

func NewRateLimitedLogger(opts ...RateLimitedLoggerOpt) *RateLimitedLogger 
...

r := &RateLimitedLogger{
		data: &intervalData{
			nextLog:  time.Now(),
			interval: defaultInterval,
		},
		entry: logrus.NewEntry(logger),
	}
	for _, opt :=range opts {
		opt(r)
	}	
	return r
}
```

PR #1519, opened 2021-10-08
(Staged)Policies should not have AOF=true; only global policies should

rene-dekker:
## Description
Network policies were previously created with the ApplyOnForward flag, which should only be available for global network policies. I do not know exactly what the impact of this bug is to end users, but if you have code that depends on applyOnForward only being available for GlobalNetworkPolicies, this could have resulted into bugs.

## Todos
- [x] Release note

## Release Note

```release-note
- Fixed a bug that set `applyOnForward=true` for all network policies and staged network policies. This flag is meant only for global network policies and should be set to` false` for non-global policies.
```


caseydavenport:

PR #1539, opened 2021-10-28
Update IP Pool CRD to contain disableBGPExport

coutinhop:
## Todos
- [x] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -15,7 +15,7 @@ require (
 	github.com/leodido/go-urn v0.0.0-20181204092800-a67a23e1c1af // indirect
 	github.com/onsi/ginkgo v1.14.1
 	github.com/onsi/gomega v1.10.1
-	github.com/projectcalico/api v0.0.0-20211027233830-6b11a0f525e1
+	github.com/projectcalico/api v0.0.0-20211102181812-edfaf495a5c8
 	github.com/projectcalico/go-json v0.0.0-20161128004156-6219dc7339ba // indirect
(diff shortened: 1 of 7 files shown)


go.mod
```diff
@@ -39,6 +39,7 @@ require (
 )
 
 replace (
+	github.com/projectcalico/api => github.com/coutinhop/api v0.0.0-20211029224736-72ed23299ba3
```
mgleung, line 42: We should point this to our API once https://github.com/projectcalico/api/pull/40 is merged

mgleung (approved):
LGTM should just need an update once the dependencies are merged

caseydavenport (changes requested):
We need a UT in the clientv3 suite to confirm that modifying this field is respected (e.g., when written, can be queried and the value is correct)

Probably goes in this suite: https://github.com/projectcalico/libcalico-go/blob/master/lib/clientv3/ippool_e2e_test.go

coutinhop:
> We need a UT in the clientv3 suite to confirm that modifying this field is respected (e.g., when written, can be queried and the value is correct)
> 
> Probably goes in this suite: https://github.com/projectcalico/libcalico-go/blob/master/lib/clientv3/ippool_e2e_test.go

Got it, will do! Thanks @caseydavenport

go.mod
caseydavenport: This still needs updating.

caseydavenport (approved):

PR #1549, opened 2021-11-19
Add IPAM capability to release block affinity given a block object

caseydavenport:
## Description

Using `ReleaseAffinity` asserts that the given CIDR is within an IP pool
- it needs to do this becuase its arguments allow for CIDRs that span
several blocks / pools. 

this newer function is smaller in scope, so can avoid that limitation,
and is more appropriate for functions like the IPAM GC controller which
has blocks in-hand when attempting to release their affinity.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/ipam/interface.go
+++ b/lib/ipam/interface.go
@@ -80,7 +80,7 @@ type Interface interface {
 	ReleasePoolAffinities(ctx context.Context, pool cnet.IPNet) error
 
 	// ReleaseBlockAffinity releases the affinity of the exact block provided.
-	ReleaseBlockAffinity(ctx context.Context, block *model.AllocationBlock, mustBeEmpty bool)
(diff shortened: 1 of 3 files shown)


caseydavenport:
Spotted this because of this error showing up in IPAM GC:

```
2021-11-19 22:28:26.277 [WARNING][1] ipam.go 569: unable or unwilling to release affinity for block blockCIDR="172.20.56.224/27" error=The requested CIDR (172.20.56.224/27) is not within any configured pools.
```

The pool was disabled because it was being emptied, so it should be valid to release its affinity even when disabled. 

However, ReleaseAffinities asserts that it is within a non-disabled IP pool. This is a nicer way to do it IMO.

mgleung (approved):

PR #1551, opened 2021-11-26
fix type assertion

Icarus9913:
fix type assertion to avoid panic.


fasaxc:
Have you observed this panic?  I would expect all the objects returned to have the correct type so the check should be unnecessary.

PR #1555, opened 2021-12-06
Add support for filtering debug logs on filename.

fasaxc:
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/lib/logutils/logutils.go
+++ b/lib/logutils/logutils.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2016-2019 Tigera, Inc. All rights reserved.
+// Copyright (c) 2016-2019,2021 Tigera, Inc. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -20,6 +20,7 @@ import (
 	"io"
 	"os"
 	"path"
+	"regexp"
 	"runtime"
 	"sort"
 	"strings"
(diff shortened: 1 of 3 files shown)


lib/logutils/logutils.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2016-2019 Tigera, Inc. All rights reserved.
+// Copyright (c) 2016-2019,2021 Tigera, Inc. All rights reserved.
```
neiljerram, line 1: I would simplify that to `2016-2021`.  Not saying that you necessarily should too, but curious to know if what I would do would be wrong.

lib/logutils/logutils_test.go
```diff
@@ -168,15 +175,45 @@ var _ = Describe("BackgroundHook log flushing tests", func() {
...
+		BeforeEach(func() {
+			hookOpts = append(hookOpts, WithDebugFileRegexp(regexp.MustCompile("another_file_for_test")))
+		})
+
+		It("should filter debug logs", func() {
+			logger.Debug("Hello")
+			debugFromAnotherFile(logger, "What?")
+			var ql QueuedLog
+			Eventually(c).Should(Receive(&ql))
+			Expect(string(ql.Message)).To(ContainSubstring(`level=debug msg="What?"`))
```
neiljerram, line 206: Should there be a check here for not getting a "Hello" log?

lib/logutils/logutils_test.go
```diff
@@ -501,3 +538,35 @@ func (s *mockSyslogWriter) Crit(m string) error {
 	_, err := fmt.Fprintf((*io.PipeWriter)(s), "CRITICAL %s", m)
 	return err
 }
+
+var benchOut bool
+var BenchIn bool
```
neiljerram, line 543: Not sure what these guys are doing?  Is benchOut just to avoid complaints about the MatchString result being unused?  BenchIn?  Or are they part of the benchmark infrastructure magic?

neiljerram (changes requested):
One substantive point: I think we're missing a check in the test code.  Then just a couple of ignorable nits.

lib/logutils/logutils_test.go
fasaxc: It's a general benchmark-related trick.  Assigning/reading from a global variable prevents the compiler from optimising away the whole loop.

lib/logutils/logutils.go
fasaxc: I _think_ you're only supposed to include the years when changes were made but I'm not sure it truly matters when we've got git history to precisely track every change.

lib/logutils/logutils_test.go
fasaxc: I was relying on implicitly receiving the next log to signal that the first one was filtered.  But, no harm in being explicit so I added an extra line.

neiljerram (approved):

